PR #241, opened 2020-02-04
Add new Code Snippet extension

### LINKED ISSUE #275: Extend Metadata service to support storing code-snippet
Enable metadata to store code snippets and integrate with the UI to enable consumption of these items into Python Scripts or Notebook cells

lresende:
The Code Snippet extension enables users to easily bookmark
 reusable code to be inserted in their notebook cells or
 Python scripts.

 The current implementation provides the backend services,
 and the initial direction of the user interface for the extension
 while additional UI is still being developed

Current UI:
image

In order to display code snippets:
1. In `/[Jupyter_path]/metadata/` create the directory `code-snippets/`
2. Add a couple json files as below
```
{
	"schema_name": "code-snippet",
	"display_name": "bloh",
	"description": "bloh description here",
	"metadata": {
		"language": "python",
		"code": [
			"def create_project_temp_dir():",
			"   temp_dir = tempfile.gettempdir()",
			"   project_temp_dir = os.path.join(temp_dir, 'elyra')",
			"   if not os.path.exists(project_temp_dir):",
			"     os.mkdir(project_temp_dir)",
			"   return project_temp_dir"
		]
	}
}
```
3. [To access code snippets metadata from browser](http://localhost:8888/api/metadata/code-snippets)

Fixes #275 #276 and #390


elyra/metadata/code_snippet.py
```diff
@@ -0,0 +1,146 @@
...
+from jupyter_core.application import base_flags
+from traitlets import Instance, Dict, Unicode, Bool
+from elyra._version import __version__
+
+
+class CodeSnippet(Metadata):
+    namespace = 'code-snippets'
+
+
+class AppUtilMixin(object):
```
kevin-bates: PR #389 moves this to the `mixins.py` file to eliminate redundancy since this existed in the runtime application as well.

elyra/metadata/code_snippet.py
```diff
@@ -0,0 +1,146 @@
...
+            print()
+            self.print_help()
+        self.exit(exit_status)
+
+    def _confirm_required(self, name, value):
+        if value is None or len(value) == 0:
+            self._log_and_exit("'{}' is a required parameter.".format(name), display_help=True)
+
+
+class ListCodeSnippets(Application, AppUtilMixin):
```
kevin-bates: Mixins should be listed prior to the superclass.

elyra/metadata/code_snippet.py
```diff
@@ -0,0 +1,146 @@
...
+            for runtime in sorted_runtimes:
+                invalid = ""
+                if runtime.reason and len(runtime.reason) > 0:
+                    invalid = "**INVALID** ({})".format(runtime.reason)
+                print("  %s  %s  %s" % (runtime.name.ljust(max_name_len),
+                                        runtime.resource.ljust(max_resource_len),
+                                        invalid))
+
+
+class RemoveCodeSnippets(Application, AppUtilMixin):
```
kevin-bates: Ditto on ordering.

karlaspuldaro:
@kevin-bates 
For now, we are only displaying code snippets manually added to `Library/Jupyter/metadata/code-snippets/` folder as json files. In the future we want to be able to create them through our application as well.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  // TODO: Use code mirror to display code
+  // TODO: implement copy to clipboard command
+  // TODO: implement insert code to file editor command (first check for code language matches file editor kernel language)
+  renderCodeSnippetsDisplay(): Array<JSX.Element> {
+    const codeSnippetDisplayList: Array<JSX.Element> = this.state.codeSnippets.map(
+      (codeSnippet: any, index: number) => {
+        const displayName =
+          '[' + codeSnippet.language + '] ' + codeSnippet.displayName;
```
kevin-bates, line 61: Will these be ordered by language? (Or can they be?)  This doesn't need to happen now, but when there are lots of snippets across multiple languages, this ordering would be helpful.  (Then order by displayName within each language.)

packages/code-snippet/src/CodeSnippet.ts
```diff
@@ -0,0 +1,72 @@
...
+    return codeSnippets;
+  }
+
+  // TODO: Test this function
+  async findByLanguage(language: string): Promise<ICodeSnippet[]> {
+    const allCodeSnippets: ICodeSnippet[] = await this.findAll();
+    const codeSnippetsByLanguage: ICodeSnippet[] = [];
+
+    for (const codeSnippet of allCodeSnippets) {
+      if (codeSnippet.language === language) {
```
kevin-bates, line 65: We'll probably want a way to fetch snippets of a given language from the server at some point.  I doubt specifically named snippets would be fetched more than, say, Python snippets, so perhaps the GET request by name on metadata could say... if the namespace is 'code-snippets' use the 'resource' to filter the language.

The request would resemble `/api/metadata/code-snippets/python` where `python` is the resource and the handler knows to get all snippets, but return only those whose language is 'python'.  The `findAll()` method could then take an optional language parameter that, when specified, extends the endpoint with its value `/<language>`.

elyra/metadata/schemas/code-snippet.json
```diff
@@ -0,0 +1,32 @@
...
+    "metadata": {
+      "description": "Additional data specific to this metadata",
+      "type": "object",
+      "properties": {
+        "language": {
+          "description": "Code snippet implementation language",
+          "type": "string"
+        },
+        "code": {
+          "type": "array"
```
kevin-bates, line 25: Is there a reason for storing the code as an array? I understand the notebook stores data in its `"source":` elements as an array - one item per code line - but it seems more natural to store the code in a free-form string.

Is the widget simply responsible for placing the code into the clipboard, then the user pastes the clipboard into the appropriate cell of the notebook?  If so, I would think it would be better to treat the entry as a string rather an array.  A string is also easier to edit, etc. in case someone wants to edit a code-snippet outside the application.

Probably not a big deal, just curious.

akchinSTC:
[Binder](https://mybinder.org/v2/gh/lresende/elyra/lresende-code-snippet?urlpath=lab/)

packages/code-snippet/src/CodeSnippetWidget.tsx
lresende: We were thinking about ordering, but also a way to select only one language.

lresende:
@karlaspuldaro The tests are failing because the test environment has no `code-snippets metadata` available. We probably should handle it better on the UI, maybe by only requesting the metadata when showing for the first time? we should also refresh it periodically to accommodate new?

packages/code-snippet/src/CodeSnippetWidget.tsx
kevin-bates: :smile: https://github.com/elyra-ai/elyra/pull/241#discussion_r401272154

kevin-bates:
The screenshot in the description looks like it strips formatting, making it difficult to understand what the snippet might be doing.  Is the plan to display snippets in their proper format?

Should a code-snippet also include an optional `description` field?  This could be presented as a tooltip hover thing or more explicitly in the display.  Some snippets will be too complicated to easily understand their intent and a simple description would be helpful, especially when collapsed in the display.

karlaspuldaro:
@kevin-bates 
My last commit fixed the code formatting, it looks like that now
image
When clicking on the code area, you can also see a scroll bar if there's more to show. We plan to also show code highlight in the preview.

And I agree adding a description would be helpful, we already provide that field in the original json file but we're not using it yet
image

Also, we're looking at failing tests, should have a fix soon.

kevin-bates:
@karlaspuldaro - the reformatted snippet looks great!  Thanks.

I don't see the description field defined in the `code-snippet.json` schema file.  That's what I was basing my question from.

ajbozarth:
I will try to check this out tomorrow to give UI feedback, but from looking at the most recent screenshot I think there needs to be some sort of border or separator around/below the code snippet "preview" especially when it would have a scroll. The lack of separation between the preview lines and the snippet "listings" looks "messy"

As I said I will do a more detailed review tomorrow once I'm done with my 2.0 branch work.

lresende:
@ajbozarth @kevin-bates I believe we are done with the last couple of suggestions and the build is passing (thanks @karlaspuldaro). Please review, we are looking for blocks, and we will continue on UI and other enhancements with small and interactive PRs.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,155 @@
...
+const BUTTON_CLASS = 'elyra-button';
+const COPY_ICON_CLASS = 'elyra-copy-icon';
+const INSERT_ICON_CLASS = 'elyra-add-icon';
+const CODE_SNIPPET_NAME_CLASS = 'elyra-codeSnippet-name';
+const CODE_SNIPPET_BUTTONS_WRAPPER_CLASS = 'elyra-codeSnippet-buttons';
+
+/**
+ * A widget for code-snippet.
+ */
+
```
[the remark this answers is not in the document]
lresende: Shouldn't these be handled by lint?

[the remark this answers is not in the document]
ajbozarth: I thought it was, but that must only be the python linter

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,155 @@
...
+              className={CODE_SNIPPET_NAME_CLASS}
+            >
+              <ExpandableComponent displayName={displayName}>
+                <pre>{codeSnippet.code.join('\n')}</pre>
+              </ExpandableComponent>
+            </div>
+            <div className={CODE_SNIPPET_BUTTONS_WRAPPER_CLASS}>
+              <div key="copyButton">
+                <button
+                  className={BUTTON_CLASS + ' ' + COPY_ICON_CLASS}
```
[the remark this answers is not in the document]
karlaspuldaro: That's a good idea, should we do this in a separate issue?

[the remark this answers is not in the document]
ajbozarth: This would be ok in a separate issue, just make sure to open it if not included here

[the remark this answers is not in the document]
karlaspuldaro: Since it will affect other packages and not only code snippets, we can resolve this in #399

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,155 @@
...
+ */
+
+class CodeSnippetTable extends React.Component<
+  { codeSnippets: ICodeSnippet[] },
+  {}
+> {
+  // TODO: Use code mirror to display code
+  // TODO: implement copy to clipboard command
+  // TODO: implement insert code to file editor command (first check for code language matches file editor kernel language)
+  renderCodeSnippetsDisplay(): Array<JSX.Element> {
```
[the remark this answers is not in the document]
karlaspuldaro: Good catch. This was implemented back when we were dealing with table element instead, I can see the code can still be further simplified.

ajbozarth:
I opened #400 to handle my comments about the error handling. @karlaspuldaro once you finish addressing my previous comments I'll take another pass through the code. Also once you provide an example code-snippet file as request in scrum I will complete my UI review and share that here as well.

ajbozarth:
Here's my UI review, overall I believe this can be merged as is and these UI comments can be addressed in a follow up issue:

- All the buttons need some css work, similar to my changes to how the arrows in the error dialog look in the jupyterlab-2.0 branch we should remove the colored background from the button icons and instead add a mouse-over background color
- When "open" the code snippet itself needs some sort of board or separator, this would make it clear that the section scrolls as well as generally looking cleaner.
- As part of the above UI addition the code section should probably include a resize ability to show more of the code snippet (I believe this is already on the roadmap)
- A bug(?) I noticed was that the copy and add button disappear when the code snippet is "opened" I believe those buttons should still be available in that case.

All these issues can be addressed in future work if desired, if they wont be addressed here you should open a new issue (or comment on a related open issue) referencing this comment.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+
+export class CodeSnippetWidget extends ReactWidget {
+  // A signal to tell CodeSnippetTable component to render the list of code snippets
+  private renderCodeSnippetsSignal = new Signal<this, ICodeSnippet[]>(this);
+
+  // Request code snippets from server
+  async fetchData(): Promise<ICodeSnippet[]> {
+    const codeSnippetManager = new CodeSnippetManager();
```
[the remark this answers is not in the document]
karlaspuldaro: good point, gonna do the same with our render signal instance

packages/code-snippet/src/ExpandableComponent.tsx
```diff
@@ -0,0 +1,86 @@
...
+const UP_ICON_CLASS = 'elyra-upArrow-icon';
+
+/**
+ * A React component for expandable containers.
+ */
+export interface IExpandableComponentProps {
+  displayName: string;
+}
+
+export class ExpandableComponent extends React.Component<
```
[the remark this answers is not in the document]
karlaspuldaro: That's right, and also to be able to reuse this anywhere else we might need some kind of item with a display name that expands into some detailed info. We can open an issue for that.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,155 @@
...
+    });
+  }
+
+  render(): React.ReactElement {
+    return (
+      <div className={CODE_SNIPPETS_CLASS}>
+        <header className={CODE_SNIPPETS_HEADER_CLASS}>
+          {'</> Code Snippets'}
+        </header>
+        <UseSignal signal={this.renderCodeSnippetsSignal} initialArgs={[]}>
```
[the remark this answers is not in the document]
karlaspuldaro: Yeah, for now we are just allowing data fetch once the extension is in use. This behavior can change in the future once we add support for creating snippets directly in the UI.
This event listener is actually handy if we want to accomplish anything similar, where widgets notify react components of a state update, so they can render the DOM appropriately.

karlaspuldaro:
Thx @ajbozarth, I agree with all your UI suggestions :)
And I'm ok with addressing them in a next step since they're not blockers but enhancements.
The copy and add buttons should still be there if you increase the extension window width, looks like it's a simple CSS fix that can be grouped with the 'resize issue' on your 3rd bullet point.

packages/code-snippet/src/CodeSnippetWidget.tsx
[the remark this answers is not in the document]
ajbozarth: Fair enough, we can also reevaluate this after the updates to the submission util in #400

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,154 @@
...
+                className={BUTTON_CLASS + ' ' + INSERT_ICON_CLASS}
+                onClick={(): void => {
+                  console.log('INSERT CODE BUTTON CLICKED');
+                }}
+              ></button>
+            </div>
+          </div>
+        </div>
+      );
+    };
```
ajbozarth, line 87: Add a empty line after this line for better readability

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -0,0 +1,154 @@
...
+interface ICodeSnippetProps {
+  codeSnippets: ICodeSnippet[];
+}
+class CodeSnippetTable extends React.Component<ICodeSnippetProps> {
+  // TODO: Use code mirror to display code
+  // TODO: implement copy to clipboard command
+  // TODO: implement insert code to file editor command (first check for code language matches file editor kernel language)
+
+  render(): React.ReactElement {
+    const renderCodeSnippet = (codeSnippet: ICodeSnippet): JSX.Element => {
```
ajbozarth, line 53: This can be declared as private outside the render function for better readability

packages/code-snippet/src/ExpandableComponent.tsx
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as React from 'react';
+import '../style/index.css';
```
ajbozarth, line 18: like above this should probably be reordered
```suggestion
import '../style/index.css';

import * as React from 'react';
```

packages/code-snippet/src/ExpandableComponent.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * The CSS class added to code snippet widget.
+ */
+
```
ajbozarth, line 23: remove space between comment and relevant code

PR #369, opened 2020-03-16
Make exception information more user friendly

### LINKED ISSUE #307: Make exception information more user friendly
From #279.  Author: @lresende

Looks like we are receiving more details on the client-side, but I would like to make this a little more user-friendly. Maybe we could get inspired on the old Windows error dialog boxes:

image

And have something like:

```
Error title
Error Message ...
Error details in a collapsable section (closed by default)?
```

In order to accomplish this, we might need to send more specific error details from the backend. In the past, I have accomplished something similar by distilling some of this info into specific response HTTP headers.

karlaspuldaro:
These changes introduce a server-side mixin for returning errors in a JSON-formatted response.  The JSON will contain `reason` and `message` fields, along with an optional `traceback` field.  Clients should tolerate `traceback` not being found.

It also contains client-side changes for displaying such payloads.  Screenshots related to these changes can be found [here](https://github.com/elyra-ai/elyra/issues/307#issuecomment-599579080).

Fixes #307

           in the file; or

           it.

### DIFF
--- a/elyra/metadata/metadata.py
+++ b/elyra/metadata/metadata.py
@@ -153,11 +153,11 @@ class MetadataStore(ABC):
             validate(instance=metadata, schema=schema, format_checker=draft7_format_checker)
         except ValidationError as ve:
             # Because validation errors are so verbose, only provide the first line.
-            msg = "Schema validation failed for metadata '{}' in namespace '{}' with error: {}"
             first_line = str(ve).partition('\n')[0]
-            # Log full message, raise with only first line of message.
-            self.log.error(msg.format(name, self.namespace, str(ve)))
-            raise ValidationError(msg.format(name, self.namespace, first_line))
+            msg = "Schema validation failed for metadata '{}' in namespace '{}' with error: {}.".\
+                format(name, self.namespace, first_line)
+            self.log.error(msg)
+            raise ValidationError(msg)
 
 
 class FileMetadataStore(MetadataStore):
--- a/elyra/metadata/metadata.py
+++ b/elyra/metadata/metadata.py
@@ -292,7 +292,7 @@ class FileMetadataStore(MetadataStore):
                         if metadata is not None:
                             resources.append(metadata)
         else:  # namespace doesn't exist, treat as KeyError
-            raise KeyError("Metadata Namespace '{}' was not found!".format(self.namespace))
+            raise KeyError("Metadata namespace '{}' was not found!".format(self.namespace))
 
         if name:  # If we're looking for a single metadata and we're here, then its not found
             raise KeyError("Metadata '{}' in namespace '{}' was not found!".format(name, self.namespace))
--- a/elyra/metadata/metadata.py
+++ b/elyra/metadata/metadata.py
@@ -292,7 +292,7 @@ class FileMetadataStore(MetadataStore):
                         if metadata is not None:
                             resources.append(metadata)
         else:  # namespace doesn't exist, treat as KeyError
-            raise KeyError("Namespace '{}' was not found!".format(self.namespace))
+            raise KeyError("Metadata Namespace '{}' was not found!".format(self.namespace))
 
         if name:  # If we're looking for a single metadata and we're here, then its not found
             raise KeyError("Metadata '{}' in namespace '{}' was not found!".format(name, self.namespace))
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -145,7 +145,11 @@ class KfpPipelineProcessor(PipelineProcessor):
             try:
                 kfp.compiler.Compiler().compile(cc_pipeline, pipeline_path)
             except Exception as ex:
-                raise RuntimeError('Error compiling pipeline {} at {}'.format(pipeline_name, pipeline_path), repr(ex))
+                if "urllib3.connection.HTTPConnection" in str(ex.reason):
+                    raise RuntimeError('Error connecting to pipeline server {}'.format(api_endpoint))
+                else:
+                    raise RuntimeError('Error compiling pipeline {} at {}'.
+                                       format(pipeline_name, pipeline_path), str(ex))
 
             self.log.info("Kubeflow Pipeline successfully compiled.")
             self.log.debug("Kubeflow Pipeline available at %s", pipeline_path)
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -27,6 +27,7 @@ from elyra.util.archive import create_temp_archive
 from elyra.util.cos import CosClient
 from kubernetes.client.models import V1EnvVar
 from notebook.pipeline import NotebookOp
+from urllib3.exceptions import MaxRetryError
 
 
 class KfpPipelineProcessor(PipelineProcessor):
@@ -144,12 +145,11 @@ class KfpPipelineProcessor(PipelineProcessor):
             # Compile the new pipeline
             try:
                 kfp.compiler.Compiler().compile(cc_pipeline, pipeline_path)
+            except MaxRetryError as connex:
+                raise RuntimeError('Error connecting to pipeline server {}'.format(api_endpoint))
             except Exception as ex:
-                if "urllib3.connection.HTTPConnection" in str(ex.reason):
-                    raise RuntimeError('Error connecting to pipeline server {}'.format(api_endpoint))
-                else:
-                    raise RuntimeError('Error compiling pipeline {} at {}'.
-                                       format(pipeline_name, pipeline_path), str(ex))
+                raise RuntimeError('Error compiling pipeline {} at {}'.
+                                   format(pipeline_name, pipeline_path), str(ex))
 
             self.log.info("Kubeflow Pipeline successfully compiled.")
             self.log.debug("Kubeflow Pipeline available at %s", pipeline_path)
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -145,7 +145,7 @@ class KfpPipelineProcessor(PipelineProcessor):
             # Compile the new pipeline
             try:
                 kfp.compiler.Compiler().compile(cc_pipeline, pipeline_path)
-            except MaxRetryError as connex:
+            except MaxRetryError:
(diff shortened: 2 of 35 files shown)


packages/application/src/submission.tsx
```diff
@@ -14,35 +14,44 @@
...
-      body: res_body ? (
-        <p>
-          {res_body}
-          <br />
-          {default_body}
-        </p>
-      ) : (
-        <p>{default_body}</p>
-      ),
+      body:
```
kevin-bates: General question.  Is the `SubmissionHandler` used for all requests to the server?   I believe there are interactions when retrieving the metadata and submitting a notebook.  Will those go through the `SubmissionHandler` or be able to use `ErrorDialogContent`?

karlaspuldaro: Yes, it also applies to errors in submitting a notebook. The dialog box title would be 'Error submitting notebook', then we'd get the same ErrorDialogContent rendered below.

lresende:
Looks like we have some places where we are using `raise web.HTTPError(500, repr(ex))` in the handlers and in this case, it looks like exceptions are not very well displayed.

image

I really don't think we should show ANY non-user-friendly message to the user directly, and these should be all in the details page.

kevin-bates:
Thanks Luciano - this is good information.  Just FYI, you're still using the previous approach.  That same issue with the changes will look at follows:
image

That said, what you're suggesting requires that each location within the server handle exceptions and we need to define **exactly** what we want displayed.

> I really don't think we should show ANY non-user-friendly message to the user directly, and these should be all in the details page.

How does the server decide, given an exception from some third-party, whether the exception is "non-user-friendly"?

Exceptions are essentially composed of three elements (I know folks are aware of these items, this is purely for clarity):
1. A reason.  This is typically the name of the `Exception` class.  E.g., `ValueError`, `KeyError`.  This element is returned in the error JSON as `reason`.
2. A message.  This is typically some helpful text that indicates the _context_ in which error occurred.  For a `KeyError`, this would simply be the key that isn't found. `'my-missing-key'`.  If that part of the code properly handles `KeyError`, then the `KeyError` might be raised with a better message: `Runtime 'my-missing-key' does not exist`.  Depending on the third-party library, this portion of the exception can be quite large and detailed.  In fact, it appears that KFP returns a _list_ of messages! The compilation error we're using as examples is a list of 2 items and these changes return the first of the two in the `message`. The message element is returned in the error JSON as `message`.
3. A traceback.  This is the set of calls leading to where the exception was raised.  The traceback element, if available, is returned in the error JSON as `traceback`.

What we need to decide is what is considered "Error details"?  Currently, these changes consider `traceback` as the "Error details".  However, it sounds like you want portions of the message somehow separated into the `message` and the `details`.  What does that separation?  It has to be at the source of the exception and, if using a third-party library, it must know what that library throws.

Do you want all tracebacks to remain on the server?  I.e., should the server log this?  I think that's probably a good way to go.  However, unless each location of each exception is caught and handled with a well-defined structure **at the source**, there's nothing that a general handler can do.  If we want to strip off the traceback (and leave it on the server) that's fine, but the bigger question is how to separate the "details" from the "message".

Please advise.  Do we want to go down the road of defining one-off, extension-specific, error handling?  Or just do a better job of handling errors at the source, returning to the user only `reason` and `message` and logging the traceback on the server at the source and forgoing the `Details` dropdown altogether?

lresende:
I think our implementation is going in the right direction, but we only tested one scenario that is easy to reproduce (e.g. cutting access to the runtime server). I think we need to validate cases where we are wrapping errors into an HTTPError versus when we are just passing exceptions that are thrown from the services/apis and make sure it is working as desired. 

In my comment above, I look into how handlers were handling errors and when we exercise the case where the handlers are raising HTTPError, the error messages are still now very user-friendly on the UI as shown in the image below:

image

I just think we should look into our handlers and standardized the way we handle errors in the backend.

lresende:
We should also check some of the messages we are using for errors and check if they make sense, like in the error above, should we actually say "error compiling...." or actually "connection error"?

kevin-bates:
Here's the code that produces the "error compiling" message.  Given this is entirely third-party, I'm not sure what else we can do other than tweak the textual portion we add.

https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_kfp.py#L144-L148

Are you suggesting that this error handler just log the result from the third party then raise what is essentially "error compiling pipeline" - which then gets sent through the handlers to the user?

packages/application/src/submission.tsx
```diff
@@ -157,3 +166,62 @@ export class SubmissionHandler {
     );
   }
 }
+
+interface IProps {
```
ajbozarth: This seems like too common a name, maybe change to `IErrorDialogProps`

packages/application/src/submission.tsx
```diff
@@ -157,3 +166,62 @@ export class SubmissionHandler {
...
+    ) : null;
+
+    return (
+      <div className={MESSAGE_DISPLAY}>
+        {this.props.reason + ': ' + this.props.message}
+        <br />
+        <br />
+        {details}
+        <br />
+        <div>{this.props.default_msg}</div>
```
ajbozarth: Do we still need to include the default message if we're already including the traceback? IIUC we would now be showing everything the user would expect to find in the logs

packages/application/src/submission.tsx
```diff
@@ -157,3 +166,62 @@ export class SubmissionHandler {
...
+          {this.props.traceback}
+        </div>
+      </div>
+    ) : null;
+
+    return (
+      <div className={MESSAGE_DISPLAY}>
+        {this.props.reason + ': ' + this.props.message}
+        <br />
+        <br />
```
ajbozarth: does including all these breaks make sense if `details == null`? If not maybe they should be part of `details`

karlaspuldaro: Good catch. The first adds a new line to message, the second adds a blank line. 
We could move one of them to `details`, so if we get an empty traceback it looks like 
image

packages/application/src/submission.tsx
karlaspuldaro: I had that conversation with @kevin-bates and we decided to leave it just because we would very likely still get extra info from the logs, such as prior to the stack trace for instance

ajbozarth:
@karlaspuldaro thanks for addressing my inline comments, do you have any answers as to my questions in the review comment?

karlaspuldaro:
@ajbozarth Sorry I missed your extra comments.
Oops that empty css file shouldn't be there. 
And you're right, I will look into reusing jupyterlab arrows instead.

karlaspuldaro:
So regarding the icons, after some research and discussing it with @ajbozarth , since I could only find a jlab 1.2.x reference to a down arrow icon only - which we could reference but the line is slightly thicker than our up arrow icon - we thought it'd best to keep both our svg icons here for consistency.
Once we move to 2.0 all icons will be cleaned up anyway due to jlab new icon usage.

akchinSTC:
Are the new svg icons we are using picked up from someplace or did we/design team make them?
Wondering if we need to add license headers to them e.g. 
https://raw.githubusercontent.com/eclipse/che/master/dashboard/src/assets/svg/administration.svg

lresende:
When I force a server not found, I am getting the new exception:
image

When I force metadata not available (e.g. mv resources resources-ok) looks like it's not going through the new dialog? 

image

Is it the case where there is no extra info available ?

karlaspuldaro:
@lresende That's right, in that case there is no traceback to display, so we don't see that option with the expand button.

elyra/pipeline/processor_kfp.py
```diff
@@ -145,7 +145,11 @@ def cc_pipeline():
             try:
                 kfp.compiler.Compiler().compile(cc_pipeline, pipeline_path)
             except Exception as ex:
-                raise RuntimeError('Error compiling pipeline {} at {}'.format(pipeline_name, pipeline_path), repr(ex))
+                if "urllib3.connection.HTTPConnection" in str(ex.reason):
```
kevin-bates: What class is `ex` here?  Looking for a string in the reason doesn't feel right. I think it might be better if we could tie this to one or two of [these exception classes](https://urllib3.readthedocs.io/en/latest/reference/#module-urllib3.exceptions).

elyra/metadata/metadata.py
```diff
@@ -292,7 +292,7 @@ def _load_metadata_resources(self, name=None, validate_metadata=True, include_in
                         if metadata is not None:
                             resources.append(metadata)
         else:  # namespace doesn't exist, treat as KeyError
-            raise KeyError("Namespace '{}' was not found!".format(self.namespace))
+            raise KeyError("Metadata Namespace '{}' was not found!".format(self.namespace))
```
kevin-bates: `Namespace` should probably be lower case.

lresende:
I have made a few changes, and now the dialogs will show like this:

When there is no stacktrace available:
image


When stacktrace is sent back:
image

and when details is shown:
image

kevin-bates:
I don't mind you removing the reason, although I could see it being useful sometimes.  

I do not agree with the notion that because there's a traceback, there's no need to check the log for more details.  I think there could very well be important information in the log file that is not contained in the messages or traceback information.  I recommend the 'check log' reminder be unconditional.

Could you please fix the test that broke when you changed the error message?

Other items:
I wish we had a consistent look for the message.  Not sure why sometimes its quoted and sometimes it's not.  I suspect the time that is isn't is when a list of messages is provided and we take the first.  Not sure if it's worth "trimming" surrounding quotes to achieve consistency.

I also think we should discuss the idea of returning a "key" (probably a timestamp) that is displayed so that the appropriate location in the log file(s) can be more easily discovered.  For example:

`Check the JupyterLab log for more details at 2020-03-19 11:31:02`

Where the timestamp value is pulled from a `timestamp` field in the response. This can be a different issue, but I think it would be super helpful to administrators or power users (that have access to the logs).

lresende:
@kevin-bates Thank you for the very detailed feedback as usual. I believe I have addressed all your concerns and suggestion and the error dialog now is looking like:

image

Any other suggestions?

lresende:
Thanks @kevin-bates , there were no new changes to the master since the last merge a couple days ago.

PR #370, opened 2020-03-16
Initial commit to support binder

### LINKED ISSUE #328: Add support to run Elyra on binder
Similar to JupyterLab, we should have a binder buttom on our readme as well as in our PRs.

akchinSTC:
Fixes #328


akchinSTC:
Try it out
[Binder](https://mybinder.org/v2/gh/akchinSTC/elyra-1/binder-328-v2?urlpath=lab)

binder/start
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+import sys
+import shutil
+import os
+
+argv = sys.argv[1:] + ['--config', 'binder/jupyter_notebook_config.py']
+print(argv)
+
+os.execv(shutil.which(argv[0]), argv)
```
kevin-bates: nit: We should probably add a newline here.

Makefile
```diff
@@ -63,8 +63,8 @@ npm-packages: lerna-build
 	$(call PACKAGE_LAB_EXTENSION,notebook-scheduler)
 	$(call PACKAGE_LAB_EXTENSION,pipeline-editor)
 	$(call PACKAGE_LAB_EXTENSION,python-runner)
-	cd dist && curl -O $$(npm view @jupyterlab/git@0.9.0 dist.tarball --userconfig=./npm_config) && cd -
-	cd dist && curl -O $$(npm view @jupyterlab/toc@2.0.0 dist.tarball --userconfig=./npm_config) && cd -
+	cd dist && curl -o jupyterlab-git-0.9.0.tgz $$(npm view @jupyterlab/git@0.9.0 dist.tarball) && cd -
+	cd dist && curl -o jupyterlab-toc-2.0.0.tgz $$(npm view @jupyterlab/toc@2.0.0 dist.tarball) && cd -
```
lresende: Why are we changing these?

binder/jupyter_notebook_config.py
```diff
@@ -0,0 +1,21 @@
+# Configuration file for jupyter-notebook.
```
lresende: License header

binder/jupyter_notebook_config.py
```diff
@@ -0,0 +1,21 @@
...
+#
+#  Local IP addresses (such as 127.0.0.1 and ::1) are allowed as local, along
+#  with hostnames configured in local_hostnames.
+c.NotebookApp.allow_remote_access = True
+
+## The default URL to redirect to from `/`
+c.NotebookApp.default_url = '/lab'
+
+## The directory to use for notebooks and kernels.
+c.NotebookApp.notebook_dir = '/tmp/workspace'
```
lresende: Why we are using /tmp/workspace here? can't we use Jovyan/workdir which is the default?

binder/postBuild
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
lresende: License header

binder/postBuild
```diff
@@ -0,0 +1,21 @@
...
+make bdist
+
+# Install Elyra Wheel
+pip install dist/*.whl
+
+# Rebuild Jupyter Lab to enable extensions
+jupyter lab build
+
+# Create starting working directory
+mkdir -p /tmp/workspace
```
lresende: Same comment as above regarding using jovyan/workdir

binder/start
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
```
lresende: License header

binder/start
```diff
@@ -0,0 +1,9 @@
...
+import sys
+import shutil
+import os
+
+argv = sys.argv[1:] + ['--config', 'binder/jupyter_notebook_config.py']
```
lresende: Minor questions: Do we need to use the config.py here? can't we just pass things like `--ip=0.0.0.0`? the rest is mostly defaulted configuration?

Makefile
akchinSTC: after the binder build, after starting up lab, an error pops up requesting another rebuild stating the only reason being the git extension tar name is different than the one it currently had (`git-0.9.0.tgz` instead of `jupyterlab-git-0.9.0.tgz`). Slightly altering name was quickest fix and seemed trivial. Same version, just different filename.

binder/jupyter_notebook_config.py
akchinSTC: I can remove this. /tmp/workspace was supposed to be the default starting location and would contain some sample notebooks and pipelines to quickly test functionality.

binder/postBuild
akchinSTC: same as above

binder/start
akchinSTC: removed it completely

PR #373, opened 2020-03-18
Moved handling of docker image list to utils

### LINKED ISSUE #360: Synchronize the list of docker images between Submit Notebook and Submit Pipeline
Submit Notebook:
image

Submit pipeline:
image

ajbozarth:
Created a util in application for getting the list of supported docker images
 
Fixes #360 

           in the file; or

           it.

### DIFF
--- a/packages/application/src/services.ts
+++ b/packages/application/src/services.ts
@@ -29,7 +29,7 @@ const dockerImages: { [key: string]: string } = {
 /**
  * A utilities class for various elyra services.
  */
-export class ElyraServices {
+export class FrontendServices {
   static getDockerImages(): { [key: string]: string } {
     return dockerImages;
   }
--- a/packages/application/src/services.ts
+++ b/packages/application/src/services.ts
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2018-2020 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


ajbozarth:
Opening this as is for feedback and review. I've confirmed this works from the UI side of things but I was having trouble with submissions and don't believe they're caused by my changes. Lots of feedback and review appreciated.

ajbozarth:
For selected I just recreated what was already happening. I figured it would be addressed when we make the notebook submission form persist values (which is an open issue). And as for the first image being the default on first load for both notebooks and pipelines, I also figured that when we switched to a metadata service we would make sure to address that.

packages/application/src/services.ts
```diff
@@ -0,0 +1,36 @@
...
+  'tensorflow/tensorflow:2.0.0-gpu-py3': 'Tensorflow 2.0 w/ GPU',
+  'pytorch/pytorch:1.2-cuda10.0-cudnn7-runtime': 'Pytorch 1.2 w/ CUDA-runtime',
+  'pytorch/pytorch:1.2-cuda10.0-cudnn7-devel': 'Pytorch 1.2 w/ CUDA-devel',
+  'amancevice/pandas:0.25.0-jupyter': 'Pandas'
+};
+
+/**
+ * A utilities class for various elyra services.
+ */
+export class ElyraServices {
```
lresende: I am over the fence if we should call this Elyraxxxx, maybe FrontendServices or ApplicationServices

ajbozarth: I have no feeling about the name, I picked it on a whim just to get something down. I think I had though I'd eventually put the service classes here, but later realized they would be better in their own classes/files.

We can pick a name real quick in scrum and I'll update this

PR #374, opened 2020-03-19
Add auto-scrolling to follow output in python runner

marthacryan:
Uses the ScrollingWidget class from jupyterlab/logconsole package
to enable automatic scrolling when new output appears in python
runner. 

Fixes #82. 
Fixes #355.

 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #377, opened 2020-03-20
Add styling to the up/down buttons

marthacryan:
Followup to comments made in #374. Fixed up the CSS so that the text wasn't under the arrows, and added svg's for up/down arrows. Now the output area looks like this:
image
On hover, the arrows are highlighted in gray:
image

 
           in the file; or

           it.

### DIFF
--- a/packages/python-runner/style/index.css
+++ b/packages/python-runner/style/index.css
@@ -34,7 +34,6 @@
 .elyra-PythonEditor-OutputArea-output {
   padding: var(--jp-code-padding);
   border: var(--jp-border-width) solid transparent;
-  /*margin-right: 64px;*/
 }
 
 .elyra-PythonEditor-scrollTop {
--- a/packages/python-runner/style/index.css
+++ b/packages/python-runner/style/index.css
@@ -41,7 +41,7 @@
   top: 33px;
   right: 21px;
   z-index: 1;
-  background-image: url('./up-arrow-icon.svg');
+  background-image: var(--up-arrow);
   background-repeat: no-repeat;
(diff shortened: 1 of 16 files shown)


packages/python-runner/style/index.css
```diff
@@ -21,6 +21,7 @@
 .elyra-PythonEditor-OutputArea-child {
   border-top: 1px solid var(--jp-border-color2);
   border-bottom: 1px solid var(--jp-border-color2);
+  margin-right: 64px;
```
ajbozarth: this should be moved to `.elyra-PythonEditor-OutputArea-output` otherwise the bottom board can look odd
Screen Shot 2020-03-20 at 3 07 27 PM

ajbozarth: it also looks a bit odd with the change, but I would personally say better than without
Screen Shot 2020-03-20 at 3 09 10 PM
Screen Shot 2020-03-20 at 3 09 18 PM

marthacryan: Yeah I did that deliberately but I wasn't sure which looked weirder. I could also make the background white instead of transparent?

marthacryan: How does this look? 
image

ajbozarth: I'd say it's a decent middle ground between the two not so great choices. Lets go with that

marthacryan:
> We will need to update the license file with these icons, see the #369 for example.

Changed the style sheet in python runner so that it references the icon from the application package instead, so this is no longer an issue.

PR #386, opened 2020-03-30
Throw exception when multipipeline or supernode found when parsing pipeline blueprint

### LINKED ISSUE #359: remove supernode in pipeline editor?
On the Elyra pipeline dashboard, users can create supernode to cluster their notebook nodes. However, it seems like the kfp processor doesn't have any logic for supernode and it will fail during submission. Is there any documentation on this? Thank you.

Screen Shot 2020-03-12 at 10 15 21 AM

Screen Shot 2020-03-12 at 10 11 15 AM

lresende:
Fixes #359

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/parser.py
+++ b/elyra/pipeline/parser.py
@@ -25,10 +25,13 @@ class PipelineParser(LoggingConfigurable):
 
     @staticmethod
     def parse(pipeline_definition) -> Pipeline:
-
-        # The pipeline blueprint enables defining multiple pipelines
-        # in one pipeline json file. For now we only support processing
-        # one (primary) pipeline.
+        """
+        The pipeline blueprint enables defining multiple pipelines
+        in one pipeline json file. When SuperNodes are enabled, each
+        super node references another flow/pipeline in the pipeline
+        blueprint. For now, we are not supporting SuperNodes and  we
+        are only support processing one (primary) pipeline.
+        """
 
         # Check for required values.  We require a primary_pipeline, a set of pipelines, and
         # nodes within the primary pipeline (checked below).
@@ -43,6 +46,7 @@ class PipelineParser(LoggingConfigurable):
             pipeline_id = p['id']
             if pipeline_id == primary_pipeline_id:
                 pipeline = p
+                break
 
         if not pipeline:
             raise SyntaxError("Primary pipeline '{}' not found.".format(primary_pipeline_id))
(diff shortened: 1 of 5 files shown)


lresende:
@kevin-bates I completely agree with you that this is not the final solution, and thus the comments on the PR description. The issue today is that when a pipeline contains a super_node it fails badly with an attribute not found an error. The current change is mostly a workaround while we wait for a release from canvas to enable us to disable super_nodes on the UI side (see https://github.com/elyra-ai/canvas/issues/22). As for front-end vs backend, backend handles "all cases" from "possible different clients". 

As for "a super node can (and probably will) contain a notebook operation", the structure of a super node is mostly a special node that points to a different blueprint (other than the primary one), and thus ignoring the super_node should work around this problem without side effects (at least that I can think of).

kevin-bates:
OK. The super-node explanation helps - thanks.

Can a blueprint's output (from a super-node) be used as inputs to a regular node?

What should the server do when we do support disabling super-nodes on the front-end? Would that implementation be useful for this case instead?

lresende:
Had a quick chat with @kevin-bates and we agreed to actually throw an exception if a pipeline that contains a super node is submitted. Will push updates shortly.

lresende:
Here is what the error message looks like when a multipipeline is submitted:
image

Error messages are:
- Multipipeline
```
        if len(pipeline_definition['pipelines']) > 0:
            raise SyntaxError("Multiple pipeline blueprints not supported (e.g. SuperNode).")
```
- SuperNode
```
            # super nodes are not supported
            if node['type'] == "super_node":
                raise SyntaxError('Super Node feature not supported')
```

elyra/pipeline/parser.py
```diff
@@ -25,24 +25,30 @@ class PipelineParser(LoggingConfigurable):
...
+        """
 
         # Check for required values.  We require a primary_pipeline, a set of pipelines, and
         # nodes within the primary pipeline (checked below).
         if 'primary_pipeline' not in pipeline_definition:
             raise SyntaxError("Required field: 'primary_pipeline' not found.")
         if 'pipelines' not in pipeline_definition:
             raise SyntaxError("Required field: 'pipelines' not found.")
+        if len(pipeline_definition['pipelines']) > 0:
+            raise SyntaxError("Multiple pipeline blueprints not supported (e.g. SuperNode).")
```
kevin-bates: We're using `SuperNode` and `Super Node` (line 67 below) so we should be consistent.  I would recommend using just `supernode` (or `Supernode` to start a sentence) since that's what the descriptive texts do in the [schema definition](https://github.com/elyra-ai/pipeline-schemas/blob/master/common-pipeline/pipeline-flow/pipeline-flow-v1-schema.json#L151-L164).

elyra/pipeline/parser.py
```diff
@@ -25,24 +25,30 @@ class PipelineParser(LoggingConfigurable):
...
+        are only support processing one (primary) pipeline.
+        """
 
         # Check for required values.  We require a primary_pipeline, a set of pipelines, and
         # nodes within the primary pipeline (checked below).
         if 'primary_pipeline' not in pipeline_definition:
             raise SyntaxError("Required field: 'primary_pipeline' not found.")
         if 'pipelines' not in pipeline_definition:
             raise SyntaxError("Required field: 'pipelines' not found.")
+        if len(pipeline_definition['pipelines']) > 0:
```
kevin-bates, line 42: Should this be `> 1`?  Perhaps this is why the tests are failing?

Is it possible to add a `test_no_supernodes()` test?  It would be something that is still useful when supernodes are supported, but disabled.

PR #389, opened 2020-03-31
Declare mixins first, move mixin to shared file

kevin-bates:
Mixins should be listed in class definition lists prior to the superclass.  Since other components may also need mixins, they have been moved to a new file.

The wrong mixin order was realized while reviewing #241.  That's where I noticed `AppUtilMixin` was getting duplicated as well, thus its move into `mixins.py`.

 
           in the file; or

           it.

### DIFF
--- a/elyra/metadata/mixins.py
+++ b/elyra/metadata/mixins.py
@@ -1,0 +1,29 @@
+#
+# Copyright 2018-2020 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


elyra/metadata/mixins.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+class AppUtilMixin(object):
```
lresende, line 18: Does it make sense to have more explicit Mixin names, like one being `LoggingMixin` and `ValidationMixin`?

kevin-bates: Generally speaking, probably so, but these two methods are specific to applications (use `print` and `exit`) and any class using the `ValidationMixin` would need to also derive from the `LoggingMixin` since the former calls the latter.  Just seems kinda messy and I'm not sure it serves a useful purpose.

I think there needs to be a place for shared utility methods.  Because these needed to behave as if part of the hosting class, they were added to a mixin and could not be written as standalone methods - unless you pass in the hosting object instance, etc.

However, it does seem that the method names, per python "convention", should not be prefixed with an underscore, and that seems like a change that should be made.

PR #392, opened 2020-04-02
Fix python editor interupted run bug

ajbozarth:
Current there is a multi-faceted bug in the python runner:
- The `Waiting for kernel to start...` prompt doesn't display when the
output tab is already open
- Closing the output tab mid-run pasues the run rather than stop it,
meaning when the user next clicks run it continues the previous run
instead of starting a new one
- The `Waiting for kernel to start...` prompt in displayed and not
cleared in this "continued" state
- clicking stop while the tab is closed "fixes" the above issues.

The solution is:
- Always stop the run before clearing the output (ie closing tab or
clicking run again mid-run)
- Display the The `Waiting for kernel to start...` prompt every time
the `displayOutputArea` is called, not just on initalization

 
           in the file; or

           it.

### DIFF
--- a/packages/python-runner/src/widget.tsx
+++ b/packages/python-runner/src/widget.tsx
@@ -179,6 +179,7 @@ export class PythonFileEditor extends DocumentWidget<
    * Function: Clears existing output area.
    */
   private resetOutputArea = (): void => {
+    this.runner.shutDownKernel();
     // TODO: hide this.layout(), or set its height to 0
     this.dockPanel.hide();
     this.outputAreaWidget.model.clear();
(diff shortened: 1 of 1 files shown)


ajbozarth:
I found this bug while testing my LabIcon update for the quick scroll

ajbozarth:
Based on a slack convo with @kevin-bates this has brought to light another edge case with the runner:
When `shutDownKernel` is called while `await this.startKernel(kernelSettings)` is still waiting in `runPython` behavior gets weird.
Specifically if a run is interrupted (either by closing the tab or clicking run again) while the kernel is still starting the shutdown doesn't do anything, this mean that once the kernel starts it will still run.
- If the output area doesn't exist (ie when you close the tab) the user sees no errors and the logs have a warning about the output stream not going anywhere.
- If the output area does exist though (ie clicking run during kernel start or clicking run again right after closing the tab) it will still output the "canceled" run's output, splicing it with the run that was mean to replace it.

marthacryan:
Played around with it and all of those behaviors seem to be fixed. LGTM

Issue #271, opened 2020-02-18
Create a service class for runtimes connections


ajbozarth:
Create a service class like `CodeSnippetManager` added in #241 for handling calls to `api/metadata/runtimes`. The new class would be located in the `application` package for use by the pipeline editor and the button.

PR #396, opened 2020-04-02
Add error dialog when no runtimes found

### LINKED ISSUE #393: Display error dialog when no runtimes configured
For both pipeline and notebook submission if there are no configured runtimes they will display an empty select field. We should instead display an error dialog to the user in the case where the runtimes array in a response is empty.

ajbozarth:
Currently when no runtimes are configured the submission forms
display and empty dropdown, which would eventually cause a backend
error upon submission.

This adds a check and error dialog in that case.

This will be further improved upon with the implementation of #271

Fixes #393

 
           in the file; or

           it.

### DIFF
--- a/packages/application/src/submission.tsx
+++ b/packages/application/src/submission.tsx
@@ -68,7 +68,7 @@ export class SubmissionHandler {
 
   static noMetadataError(metadataName: string): Promise<Dialog.IResult<any>> {
     return showDialog({
-      title: 'Error retrieving metadata',
+      title: 'Could not submit',
       body: <p>No {metadataName} metadata has been configured.</p>,
       buttons: [Dialog.okButton()]
     });
--- a/packages/application/src/submission.tsx
+++ b/packages/application/src/submission.tsx
@@ -66,6 +66,14 @@ export class SubmissionHandler {
     });
   }
 
+  static noMetadataError(metadataName: string): Promise<Dialog.IResult<any>> {
+    return showDialog({
+      title: 'Error retrieving ' + metadataName,
+      body: <p>No {metadataName} metadata has been configured</p>,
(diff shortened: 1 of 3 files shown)


kevin-bates:
I think the "customization" in the title makes the body feel a little redundant.  Perhaps a title of 'Error retrieving metadata' would help eliminate that feeling?

karlaspuldaro:
Agree with @kevin-bates with just a title here
Maybe 'Error retrieving runtimes metadata' to be more clear?

kevin-bates:
Sorry, but my point was to leave the customization out of the title since it becomes redundant with the body.   Perhaps we can drop the title altogether?

ajbozarth:
updated, new screenshot:
Screen Shot 2020-04-03 at 5 23 42 PM

kevin-bates:
@ajbozarth - thank you for updating the title.

I don't know if others agree, but should this particular dialog even have a title?  After I mentioned perhaps dropping the title altogether, the more I think that's actually the right thing to do...

1. Because we control the body of the message completely, there's really no "preamble" necessary.  Whereas the 'Error submitting pipeline' error dialog is presenting a relatively random error thrown from the server - so a title to introduce that error message is useful.
2. This is more of a message than an error.
3. The body implies the title but doesn't really present it as an error - there's just no metadata of this particular type and, really, no attempt has been made to even create that metadata.  For example, had runtimes metadata existed, but all of them had schema issues, the fetch would get an empty list back and no message would be presented.  Same goes if all runtimes had been removed (after their initial existence).

I know this is minor, but while we're talking about this, we should do what we feel is right.

ajbozarth:
updated the title text as discussed in scrum and I rebased on master and force pushed to fix the merge conflicts

PR #397, opened 2020-04-02
Add lock to kernel startup

### LINKED ISSUE #394: Clicking run twice splices output of two runs
This bug was found in #392 including the related comment below:

Based on a slack convo with @kevin-bates this has brought to light another edge case with the runner:
When shutDownKernel is called while await this.startKernel(kernelSettings) is still waiting in runPython behavior gets weird.
Specifically if a run is interrupted (either by closing the tab or clicking run again) while the kernel is still starting the shutdown doesn't do anything, this mean that once the kernel starts it will still run.

If the output area doesn't exist (ie when you close the tab) the user sees no errors and the logs have a warning about the output stream not going anywhere.
If the output area does exist though (ie clicking run during kernel start or clicking run again right after closing the tab) it will still output the "canceled" run's output, splicing it with the run that was mean to replace it.

### LINKED ISSUE #398: Python editor should indicate that run has been stopped after stop selected
After starting a run in the Python editor there is no indication that the run has been stopped after selecting the stop button.   Only if recurring output is happening can the user determine that the kernel has been stopped (as the output stops as well).

Removing the `*` in the bracketed prompt would help.  

If the run button was disabled during the entire run (see #397), then its enablement would also be an indicator that the run has stopped.

ajbozarth:
Currently if you try to start a new run while the kernel is starting
it wint stop the previous run and the two runs output will be spliced

This adds a "lock" around starting the kernel, the lock prevents
users from starting a new run by both disabling both the run button
and the `runPython()` function if statement.

Fixes #394
Fixes #398

 
           in the file; or

           it.

### DIFF
--- a/packages/python-runner/src/PythonRunner.ts
+++ b/packages/python-runner/src/PythonRunner.ts
@@ -18,22 +18,34 @@ import { Kernel } from '@jupyterlab/services';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import { Dialog, showDialog } from '@jupyterlab/apputils';
 
+const RUN_BUTTON_CLASS = 'elyra-PythonEditor-Run';
+
 /**
  * Class: An enhanced Python Script Editor that enables developing and running the script
  */
 export class PythonRunner {
+  id: string;
   kernel: Kernel.IKernel;
   model: CodeEditor.IModel;
   kernelSettings: Kernel.IOptions;
+  startingKernel: boolean;
 
   /**
    * Construct a new runner.
    */
-  constructor(model: CodeEditor.IModel) {
+  constructor(model: CodeEditor.IModel, id: string) {
+    this.id = id;
     this.kernel = null;
     this.model = model;
+    this.startingKernel = false;
   }
 
+  startKernelLock(lock: boolean): void {
+    this.startingKernel = lock;
+    (document.querySelector(
+      '#' + this.id + ' .' + RUN_BUTTON_CLASS
+    ) as HTMLInputElement).disabled = lock;
+  }
   /**
    * Function: Starts a python kernel and executes code from file editor.
    */
@@ -41,12 +53,14 @@ export class PythonRunner {
     kernelSettings: Kernel.IOptions,
     handleKernelMsg: Function
   ): Promise<any> => {
-    if (!this.kernel) {
+    if (!this.kernel && !this.startingKernel) {
       const model = this.model;
       const code: string = model.value.text;
 
       try {
+        this.startKernelLock(true);
         this.kernel = await this.startKernel(kernelSettings);
+        this.startKernelLock(false);
       } catch (e) {
         return showDialog({
           title: 'Error',
--- a/packages/python-runner/src/PythonRunner.ts
+++ b/packages/python-runner/src/PythonRunner.ts
@@ -18,34 +18,33 @@ import { Kernel } from '@jupyterlab/services';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import { Dialog, showDialog } from '@jupyterlab/apputils';
 
-const RUN_BUTTON_CLASS = 'elyra-PythonEditor-Run';
-
 /**
  * Class: An enhanced Python Script Editor that enables developing and running the script
  */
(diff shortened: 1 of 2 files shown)


packages/python-runner/src/PythonRunner.ts
```diff
@@ -18,35 +18,49 @@ import { Kernel } from '@jupyterlab/services';
...
   ): Promise<any> => {
-    if (!this.kernel) {
+    if (!this.kernel && !this.startingKernel) {
       const model = this.model;
       const code: string = model.value.text;
 
       try {
+        this.startKernelLock(true);
         this.kernel = await this.startKernel(kernelSettings);
+        this.startKernelLock(false);
```
kevin-bates, line 63: I think it might be better to keep the lock enabled for the duration of the run or until the 'stop' button is selected.  I don't really see any reason to enable the run button when the kernel is already running and its enablement just makes it easier to disrupt things.

Unrelated but thought I'd raise here...
I would recommend using unique text (however slight it is) in error messages so that the location of where the error occurred can be identified since what we're usually provided is a description or screenshot at best.  A couple lines below we have duplicate messages:  'Could not start kernel environment to execute script.'  I suggest changing one to differentiate it from the other.

ajbozarth: the reason the lock is only implemented here is because my previous PR added support for interrupting runs by starting a new run. The edge case this is addressing is just more common with that new feature.

kevin-bates:
I had this occur today when selecting the run button during a run.  This happened multiple times, but not always.  I had two runners going and it seemed to only happen on the second ('b') of the two. Will try with a single python runner.
image

Here's the code in each runner - with the character changed to 'a' in the other to differentiate:
```python
import time

i = 0
while i < 60:
    print('b: ' + str(i))
    time.sleep(1)
    i = i+1
print('b: ' + str(i))
```

ajbozarth:
Based on discussion at today's scrum I will be doing the following on this PR:
- Lock on the whole run, this mean a user can only run again after clicking stop or letting it finish
- Move handling of stopKernel introduced in #392 to the close tab callback only
- if appropriate, address #398 while updating how stops are handled.
- if appropriate address the separate error text issue raise in @kevin-bates comment

packages/python-runner/src/PythonRunner.ts
```diff
@@ -18,35 +18,49 @@ import { Kernel } from '@jupyterlab/services';
...
    */
-  constructor(model: CodeEditor.IModel) {
+  constructor(model: CodeEditor.IModel, id: string) {
+    this.id = id;
     this.kernel = null;
     this.model = model;
+    this.startingKernel = false;
   }
 
+  startKernelLock(lock: boolean): void {
```
karlaspuldaro, line 43: I don't think DOM manipulation on an element defined in the widget should be done in PythonRunner, since it's supposed to only deal with the backend. I was thinking some sort of callback like we did with `handleKernelMsg()` that notifies the UI, so that button lock is done from the widget.

ajbozarth: That's fair, I moved it in here because I needed to access it from multiple functions, but I realized I would hand it to the constructor as a callback function instead of passing in the id and it would work the same.

ajbozarth:
I've pushed an update that addresses the following:
- Lock on the whole run
- Move function that handles locking to widget and passes the function to the PythonRunner
- Created a helper function for handling error dialogs that makes sure to unlock on error
- Added a call to stopKernel for error case where the kernel exists but is in an unusable state
- Make a stopRun handler that removes the `[*]` on stop (conditional on the output being visible to prevent undefined errors in the log)
- Added a lock flag in addition to disabling the button since a disabled button does not prevent programmatic calls to handlers, meaning without the flag and a check for it in runPython the button would still work even though it's "disabled"

Note there is one odd edge case remaining with this fix, but it has a simple and obvious workaround and thus isn't worth addressing. If a user closes the output tab while the kernel is starting it will sometime not stop the kernel and the run will continue "in the background" leaving the run button disabled until it finishes. The obvious workaround for this is to just click the stop button, which then stops the "background" run and reenables the run button.

Issue #437, opened 2020-04-24
Issue running with Pandas image "kernel_manager_class’"


yil532:
While running my pipeline under pandas image. It says KeyError: ‘kernel_manager_class’. Could someone please help me?
Here is the error message
```
`Parsing Arguments.....
Get file Part 1 - Data Cleaning-404fc2a7-8fde-4e45-9652-55bcd230916b.tar.gz from bucket yil532
TAR Archive pulled from Object Storage.
Unpacking........
Unpacking Complete.
Executing notebook through Papermill: Part 1 - Data Cleaning.ipynb ==> Part 1 - Data Cleaning_output.ipynb

Executing:   0%|          | 0/47 [00:00<?, ?cell/s]
Executing:   0%|          | 0/47 [00:00<?, ?cell/s]
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'kernel_manager_class'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "bootstrapper.py", line 145, in <module>
    kernel_name="python3"
  File "/usr/local/lib/python3.7/site-packages/papermill/execute.py", line 106, in execute_notebook
    **engine_kwargs
  File "/usr/local/lib/python3.7/site-packages/papermill/engines.py", line 49, in execute_notebook_with_engine
    return self.get_engine(engine_name).execute_notebook(nb, kernel_name, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/papermill/engines.py", line 343, in execute_notebook
    cls.execute_managed_notebook(nb_man, kernel_name, log_output=log_output, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/papermill/engines.py", line 402, in execute_managed_notebook
    return PapermillNotebookClient(nb_man, **final_kwargs).execute()
  File "/usr/local/lib/python3.7/site-packages/papermill/clientwrap.py", line 36, in execute
    with self.setup_kernel(**kwargs):
  File "/usr/local/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.7/site-packages/nbclient/client.py", line 401, in setup_kernel
    self.start_kernel_manager()
  File "/usr/local/lib/python3.7/site-packages/nbclient/client.py", line 328, in start_kernel_manager
[...]

PR #458, opened 2020-04-29
Update docker images used in pipeline editor

akchinSTC:
This PR updates both 1.X Tensorflow images as well as the Pandas image to resolve various issues. See commit msgs
 
           in the file; or

           it.


etc/docker/tensorflow/1.15.3-gpu-py3/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM tensorflow/tensorflow:1.15.2-gpu-py3
+
+RUN apt install -y curl
```
kevin-bates, line 3: We should probably add a newline here otherwise diff tools have a hard time with things.

etc/docker/tensorflow/1.15.3-py3/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM tensorflow/tensorflow:1.15.2-py3
+
+RUN apt install -y curl
```
kevin-bates, line 3: Same - newline needed.

packages/application/src/services.ts
```diff
@@ -17,13 +17,13 @@
...
-  'tensorflow/tensorflow:1.15.2-py3': 'Tensorflow 1.15.2',
-  'tensorflow/tensorflow:1.15.2-gpu-py3': 'Tensorflow 1.15.2 w/ GPU',
+  'elyra/tensorflow:1.15.2-py3': 'Tensorflow 1.15.2',
+  'elyra/tensorflow:1.15.2-gpu-py3': 'Tensorflow 1.15.2 w/ GPU',
   'tensorflow/tensorflow:2.0.0-py3': 'Tensorflow 2.0',
   'tensorflow/tensorflow:2.0.0-gpu-py3': 'Tensorflow 2.0 w/ GPU',
   'pytorch/pytorch:1.2-cuda10.0-cudnn7-runtime': 'Pytorch 1.2 w/ CUDA-runtime',
   'pytorch/pytorch:1.2-cuda10.0-cudnn7-devel': 'Pytorch 1.2 w/ CUDA-devel',
-  'amancevice/pandas:0.25.0-jupyter': 'Pandas'
+  'amancevice/pandas:1.0.3-jupyter': 'Pandas'
```
kevin-bates, line 26: Not for this PR, but I would recommend we not use the '-jupyter' version of these images.

akchinSTC: +1 to moving to the non jupyter image

kevin-bates: The more I think about this the more correct it seems - especially since we're overwriting nearly all the jupyter packages anyway and this would mean less chance of additional conflicts.

@lresende - what do you think?

lresende: +0, I didn't want to get on the business of maintaining these community images

kevin-bates: Agreed, but isn't this just a stop-gap to resolve #437?  We should also use non-jupyter-related images as much as possible per my previous comment.

Also, isn't this entire PR a stop-gap to address two current issues with "community images" (lack of `curl` and wrong versions)?  Both of which have issues open to address in kfp-notebook, but I thought there was a higher level of urgency here - thus the stop-gap measures.

lresende: you are correct, we need to address the user issue with a higher level of urgency.

Issue #426, opened 2020-04-11
Dialogs should have the option to be expanded


lresende:
As an example, error dialogs might have stack trace information, which contents might span to multiple lines and not very very easily seen on the small fixed size of the current dialogs. It would be good to enable users to enlarge it (e.g. like the editable version of this GitHub textbox)

![image](https://user-images.githubusercontent.com/382917/79030275-0e401380-7b4d-11ea-9ec5-e7582cef025a.png)

lresende:
@vabarbosa this is very similar to the #431 where you enable the text field to be resizable... might need a little more css to enable the dialog to be a little bigger or expandable.

vabarbosa:
so after investigating we have a few ways to approach the issue:

1. wrap all Elyra dialog messages in a `<textarea>` before calling JupterLab's `showDialog()`

    ```javascript
    import React from 'react';
    import { showDialog, Dialog } from '@jupyterlab/apputils';

    const dialogMessage = 'Lorem ipsum dolor sit amet consectetur';

    showDialog({
      title: 'Some dialog title',
      body: <textarea defaultValue={dialogMessage}></textarea>,
      buttons: [Dialog.okButton()]
    });
    ```

2. create and use Elyra's own `showDialog()` which extends JupyterLab's `showDialog()`

    for example in a `elyrautils.tsx` file in `packages/application`:

    ```javascript
    import { showDialog as showJpDialog, Dialog } from '@jupyterlab/apputils';
    import React from 'react';

    export const showDialog = function<T>(
      options: Partial<Dialog.IOptions<T>> = {}
    ): Promise<Dialog.IResult<T>> {
      if (typeof options.body === 'string') {
        options.body = <textarea defaultValue={options.body}></textarea>;
      }
      return showJpDialog(options);
    };
    ```

    then all Elyra dialog messages would call Elyra's `showDialog()` instead of JupyterLab's `showDialog()`:

    ```javascript
    import React from 'react';
    import { showDialog } from '@elyra/application';

    const dialogMessage = 'Lorem ipsum dolor sit amet consectetur';

    showDialog({
      title: 'Some dialog title',
      body: dialogMessage,
      buttons: [Dialog.okButton()]
    });
    ```

3. Update JupyterLab's `showDialog()` to wrap dialog body in a `<textarea>` and not changes would e needed in Elyra

    in `jupyterlab/packages/apputils/src/dialog.ts`:

    ```javascript
    export function showDialog<T>(
[...]

PR #460, opened 2020-04-29
Update error dialogs to support resizing

vabarbosa:
This PR addresses #426. It makes so when the error dialog is resized the details in the dialog resize accordingly. The PR also does some minor CSS clean updates to the error dialog details.

Support for resizing dialogs was addressed in [Jupyter Lab PR #8301](https://github.com/jupyterlab/jupyterlab/pull/8301) ([issue #8300](https://github.com/jupyterlab/jupyterlab/issues/8300))

Before (as dialog is resized, details area stayed fixed dimensions):
image

After (as dialog is resized, details area grows/shrinks accordingly):
image

 
           in the file; or

           it.


vabarbosa:
@ajbozarth when the dialog is resized explicit width and height values get set by the browser on the `.jp-Dialog-content` DIV.

as a result, when you expand the call stack it will be contained within the `.jp-Dialog-content` DIV dimensions. however, if you expand the call stack before resizing dialog there is no width or height set for the `.jp-Dialog-content` DIV so it will automatically grow to fit the size of the expanded call stack.

we could go ahead and set an explicit/default width for the `.jp-Dialog-content` DIV so if the user expands the call stack w/o resizing the dialog it displays at a reasonable dimension. the one catch appears to be with Safari which seems to not allow the user to resize smaller than this set width (they can resize larger). there may not be much benefit in allowing users to resize the dialog to an arbitrarily small size, so this may be an acceptable behavior.

alternatively, we can go for a JS+CSS solution (above is CSS only) which should give more flexibility in getting the exact desired behavior.

in any case, i'll continue to investigate and we can discuss further

ajbozarth:
Given the "expand content" is handled by an `onClick` set in js I think a JS+CSS solution may be the best here. When the onClick runs we can have it check for set dimensions and if they don't already exist set them to some predefined size. And if that give safari issues we could have the close onClick remove them or set them to a smaller "minimum"

vabarbosa:
@ajbozarth changes have submitted based on your feedback

PR #475, opened 2020-05-01
Implement Code Snippet Insert

### LINKED ISSUE #278: Enable including code-snippets into Python Scripts or Notebook cells

ajbozarth:
Implements the insert feature for code snippets
 
Fixes #278 

           in the file; or

           it.


ajbozarth:
Opening a PR with my initial work for @karlaspuldaro to pick up and finish

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -16,11 +16,16 @@
 
 import '../style/index.css';
 
+import { PythonFileEditor } from '@elyra/python-runner-extension/lib/PythonFileEditor';
```
ajbozarth, line 19: @karlaspuldaro I just noticed this imported wrong, make sure to fix this in your next commit (prob will need to add it as a dependency in package.json too

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -42,7 +48,7 @@ const COPY_ICON_CLASS = 'elyra-copy-icon';
 const INSERT_ICON_CLASS = 'elyra-add-icon';
 
 /**
- * CodeSnippetTable props.
+ * CodeSnippetDisplay props.
  */
 interface ICodeSnippetProps {
```
ajbozarth, line 53: We prob want to rename this `I CodeSnippetDisplayProps` to match the class it's props for

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -72,31 +74,66 @@ class CodeSnippetDisplay extends React.Component<ICodeSnippetDisplayProps> {
...
       const fileEditor = (documentWidget.content as FileEditor).editor;
 
-      // Wrap in code block if inserting snippet into a markdown file
       if (PathExt.extname(widget.context.path) === '.md') {
+        // Wrap snippet into a code block when inserting it into a markdown file
         fileEditor.replaceSelection(
           '```' + snippet.language + '\n' + snippetStr + '\n```'
         );
+
+        // TODO: Change how python file editor is checked below to be more generic
```
ajbozarth, line 83: should probably move this into the if statement (indentation just looks off having it here)

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -72,31 +74,66 @@ class CodeSnippetDisplay extends React.Component<ICodeSnippetDisplayProps> {
...
+        notebookWidget.context.sessionContext.kernelPreference.language;
+      this.verifyLanguage(
+        snippet.language,
+        kernelLanguage,
+        notebookCellEditor,
+        snippetStr
+      );
     } else {
-      this.showErrorDialog('Code snippet insert failed: Unsupported widget');
+      this.showErrDialog('Code snippet insert failed: Unsupported widget');
```
ajbozarth, line 102: why the rename?

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -72,31 +74,66 @@ class CodeSnippetDisplay extends React.Component<ICodeSnippetDisplayProps> {
...
     }
   }
 
-  private showErrorDialog = (
-    errorMsg: string
+  private verifyLanguage = async (
```
ajbozarth, line 106: We should make this name more descriptive. On first read through I was confused where the code was actually inserted until I read this function. Maybe `verifiedInsert` or ` verifyAndInsert` or `verifyLanguageAndInsert`

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -72,31 +74,66 @@ class CodeSnippetDisplay extends React.Component<ICodeSnippetDisplayProps> {
...
+    snippetLanguage: string,
+    editorLanguage: string,
+    editor: CodeEditor.IEditor,
+    code: string
+  ): Promise<void> => {
+    if (snippetLanguage !== editorLanguage) {
+      const result = await this.showWarnDialog(editorLanguage);
+      result.button.accept && editor.replaceSelection(code);
```
ajbozarth, line 114: Though this shorthand is valid, I think we should use a full if statement for better readability by those that don't know js shorthands

packages/code-snippet/src/CodeSnippetWidget.tsx
karlaspuldaro: I made a separate show dialog function for warning since it behaves a little different

ajbozarth: I meant why `Error` > `Err`?

karlaspuldaro: Oh I probably kept it short by accident when having 2 separate functions again (at one point they've merged to a single function but didn't work :p )

vabarbosa:
nice work!

one thing i noticed is when using my non-Python kernel is i get the message with the _language_ missing:

```
The current code snippet is incompatible with . Continue?
```

image

not sure if this is something we want to look into or if it is beyond the scope of this issue/PR.

fyi, this is an iJavaScript kernel i have installed

kevin-bates:
Seems like a fall-back "language" place-holder is desired.  
Q: If the destination language cannot be determined (I'm not sure, but language might be optional on kernelspecs), do we go ahead with the insert, or do we only insert on definitive language matches?

Also, rather than say "The current code snippet is incompatible with..." would it be possible to re-affirm the selection with something like, "Code snippet 'is_even' is incompatible with ..."?   The current wording can leave the user wondering if they selected the wrong snippet or the software didn't respect their selection.  (For paranoid folks like myself, _we_ always like to see affirmation. :smile:)

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -37,20 +50,104 @@ const COPY_ICON_CLASS = 'elyra-copy-icon';
...
+    const snippetStr: string = snippet.code.join('\n');
+
+    if (
+      widget instanceof DocumentWidget &&
+      (widget as DocumentWidget).content instanceof FileEditor
+    ) {
+      const documentWidget = widget as DocumentWidget;
+      const fileEditor = (documentWidget.content as FileEditor).editor;
+
+      if (PathExt.extname(widget.context.path) === '.md') {
```
vabarbosa, line 78: while [JuyterLab current only recognizes the `.md` extension](https://github.com/jupyterlab/jupyterlab/issues/8387) for markdown. there are [other extensions commonly used for markdown](https://superuser.com/a/285878) should we perhaps check for some of those extensions as well? not suggestion we try to check for all of them but perhaps at a minimum the same set of [extensions that GitHub recognizes](https://github.com/github/markup/blob/b865add2e053f8cea3d7f4d9dcba001bdfd78994/lib/github/markups.rb#L1). just a thought

vabarbosa:
> Q: If the destination language cannot be determined (I'm not sure, but language might be optional on kernelspecs), do we go ahead with the insert, or do we only insert on definitive language matches?
> 

the behavior i am seeing is it does not do an insert if the language does not match or cannot be determined.

vabarbosa:
question: should the insert append a new line when inserting the code? reason i ask is when inserting code snippets back-to-back the first of the second snippet is merged with the last line of the first snippet. not sure how often someone would do that :-/ or perhaps we leave it up to the person creating the snippet to add the appropriate new line if desired:

image

ajbozarth:
> question: should the insert append a new line when inserting the code? reason i ask is when inserting code snippets back-to-back the first of the second snippet is merged with the last line of the first snippet. not sure how often someone would do that :-/ or perhaps we leave it up to the person creating the snippet to add the appropriate new line if desired:

I would leave including newlines in the snippet to the user, there's definitely cases for not having it.

ajbozarth:
As for the missing language value. I would say that's a bug. It would seem that `notebookWidget.context.sessionContext.kernelPreference.language` can potentially be an empty string, we should find out why and if we may be looking in the wrong place for this value. Or if empty string is a valid case for this we should handle it with a default value like @kevin-bates suggested.

Also I agree replacing `current code snippet` with the name of the snippet in the error might be more readable. Or we could update the message to be something like this instead:
`You are inserting a <code snippet language> snippet into a <kernel language> editor. Continue?` (maybe use `field` instead of `editor`?)

ajbozarth:
I just checked and @kevin-bates is correct that `notebookWidget.context.sessionContext.kernelPreference.language` is an optional field. So we should check if it exists and handle the case for no kernel language set. I can see paths to handle it:
1. Update the error message with a "default" value like `unknown language` when `language = ''` or
2. treat it like we do file editors and just allow any insertion of any code into notebooks with kernels without a set language without a language check.

I'm personally not sure which is better, @lresende opinion?

kevin-bates:
> Or we could update the message to be something like this instead:
You are inserting a <code snippet language> snippet into a <kernel language> editor. Continue? (maybe use field instead of editor?)

We should remember that grammatical issues will arise of the substitution item has a vowel-sounding beginning where 'a' should be replaced with 'an'.  As such, I think just single-quoting the code snippet name and stating its incompatible with '<language>'.

Then, in cases where the language cannot be determined, always allow insert (which is consistent with file editor behaviors) (option 2).  That way, you will only display the error dialog on definitive language mismatches.

vabarbosa:
more clarification on the language warning.
below is the behavior i see with my Python and JavaScript kernels:

1. create a new notebook and try to insert code
   ```
   The current code snippet is incompatible with . Continue?
   ```

1. create a new notebook, hit Save, then try to insert code
   ```
   The current code snippet is incompatible with . Continue?
   ```

1. create a new notebook, exit notebook, open notebook, then try to insert code
   ```
   The current code snippet is incompatible with . Continue?
   ```

1. create a new notebook, hit Save, exit notebook, open notebook, then try to insert code: **Success** 👍  code snippet is inserted

this is happening with both my JavaScript kernel and default Python kernel. in both cases, i confirmed `language` is in the kernelspec

karlaspuldaro:
Nice catch @vabarbosa , I have just tested inserting a snippet into a notebook just created before saving it and I get that same behavior.
Looks like before I save a new python notebook, `notebookWidget.sessionContext.kernelPreference` has`language: "" `and  `name: "python3"`. For some reason `language: "python"` only gets updated when a saved notebook is reopened.

Maybe we could use `name` field for our checks and 
1. be more specific when creating a code snippet metadata, or
2. allow kernel language verification to check for possible language matches, that would have to be done with all supported notebook languages though. 

And good point @kevin-bates on the message suggestion and default to allow code insertion when we can't determine a language match, considering we choose to use `name` field as proposed above.

kevin-bates:
The `name` field - assuming its the same as the `name` field in the kernelspec - represents the name of the kernelspec and probably can't be used as the language.  For example, the name of the kernelspec with display_name = "Spark - Python (YARN Client Mode)" is "spark_python_yarn_client" and won't always infer the underlying language.

The kernel is always launched based on information from its corresponding kernelspec (where `name`) is most likely an index into the set of kernelspecs.  Somewhere in the client, the kernelspec lurks.  

It sounds like we should always be pulling the "destination language" from the corresponding kernelspec, but if not there, try the notebook, and if not there fail the insert.  Its probably reasonable to switch those first two, although, theoretically speaking, a user could decide to change languages within a given notebook - which would imply using the language from the kernelspec first.

vabarbosa:
> 
> Maybe we could use `name` field for our checks and
> 

i am not sure using `name` would work (in my default Python environment `name` is set to `base`).

i vote for just inserting the code without checking for a match (or displaying a warning) if the language is unavailable

karlaspuldaro:
Just updated the warning message and markdown file extension case.

Just to clarify, I will also look into getting the kernel spec directly, since notebook kernelPreference by itself is not reliable. 
Then if language is still unavailable, I also think we should allow inserting the snippet regardless, with no warning, rather than failing.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -37,20 +50,111 @@ const COPY_ICON_CLASS = 'elyra-copy-icon';
...
+    }
+  }
+
+  // Handle language compatibility between code snippet and editor
+  private verifyLanguageAndInsert = async (
+    snippetLanguage: string,
+    snippetName: string,
+    editorLanguage: string,
+    editor: CodeEditor.IEditor,
+    code: string
```
ajbozarth, line 119: I think it might make more sense to pass an `ICodeSnippet` param instead of passing 3 values from it

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -37,20 +55,153 @@ const COPY_ICON_CLASS = 'elyra-copy-icon';
...
+      'editorLanguage: ' +
+        editorLanguage +
+        ', snippet.languange: ' +
+        snippet.language
+    );
+    const snippetStr: string = snippet.code.join('\n');
+    if (editorLanguage === '') {
+      // Allow code snippet insertion when editorLanguage is unavailable
+      editor.replaceSelection(snippetStr);
+    } else if (snippet.language !== editorLanguage) {
```
ajbozarth, line 164: rather than having an else if here we can combine the first two checks
```suggestion
    if (editorLanguage && snippet.language !== editorLanguage) {
```

ajbozarth:
@karlaspuldaro I pushed my updates addressing both my comment and fixing the issues with `instanceof PythonFileEditor`. I also did a merge with master since my fix was dependent on my recently merged update to like packages instead of install them. I had no luck trying to test the kernel issues since I kept running into separate issues just getting my env to let me test it. So I decided to push what I have so as to not block you in the morning before I start my day and get back to my testing.

kevin-bates:
@karlaspuldaro asked me to take this PR for a spin against a Spark/YARN-cluster kernel, so here are those results.

I still see this message on new notebooks...
image

For an existing notebook, I get this (looks like language isn't being compared case-insensitively)...
image

It looks like `+` inserts into whatever the current cell is.  Should we add a newline to the end of the snippet when inserting into a cell with other text?
image

When using the "copy", I think a little help text or something indicating that the code-snippet has been copied to the clipboard would be helpful.

Lastly, I thought we were going to change the message to indicate the selected code-snippet per https://github.com/elyra-ai/elyra/pull/475#issuecomment-625413989

karlaspuldaro:
Thanks @kevin-bates 
Will fix the case sensitivity in the language comparison.

[This comment ](https://github.com/elyra-ai/elyra/pull/475#issuecomment-625430768)addresses the new line suggestion, unless we have a strong opinion to include it.

I believe we've discussed about the copy confirmation msg not being necessary, but I agree it would be helpful. Maybe we can open a separate issue since it's not under the insert feature scope.

Just wondering if you pulled the latest changes before testing? You should be able to see the updated warning message as below:
image

ajbozarth:
Pushed a real fix for PythonFileEditor verification. It's a "dumb" fix, but after lots of research and playing with `instanceof` and `prototype` I realized that it's just a known broken feature for our exact use case. So after stumbling upon a simple hard-coded string compare solution I decided to use it.

kevin-bates:
I took the latest commits and reran against EG (not that that matters).  The python comparison is fixed - thank you.  I also try inserting a python code-snippet into an R kernel code cell and received the appropriate dialog using the '+', but when I used the copy/paste approach, it just pasted into the cell.  I can't recall if that's how things were supposed to work in this case. 

I then created a markdown cell in my R-based notebook.  When I went to insert (using '+') I also got the prompt about the wrong language.  I thought when inserting into markdown cells we'd just unconditionally insert with the 3-back-tics and the language of the code snippet.  (Perhaps that hasn't been implemented yet?)  

Overall, this is looking really cool and will be a great feature - nice work so far.

ajbozarth:
In response to @kevin-bates note about different cell types inside a notebook I added a check for cell type and implemented it the same as in file editor. As for my testing, it seems to work ok except for in one case, but it behaves the same as trying to run a cell in that case. (My Scala kernel is stuck in "initializing" and lab hangs with no error message if I insert into or run a cell)

@karlaspuldaro I left your current kernel related code all in the CodeCell case. I assume you plan to remove all the extra commented out code before we un-wip this.

Also @kevin-bates that is how the copy-paste behavior is expected to work, it's just a copy-paste of text. Though I do agree a nice follow up would be to add some sort of transient tooltip on copy (like the copy button when signing into ibmcloud cli using sso)

ajbozarth:
This LGTM (can't "approve" since I created the PR)

Issue #177, opened 2019-10-26
Investigate full project build time


lresende:

akchinSTC:
These times will vary depending on a users internet connection:
My current setup consists of 500/100 down/up connection.
This is using just `make install`
Time to full build with full download of node dependencies is ~3m 41s.
Subsequent incremental builds without download of node dependencies is ~2m 17s.

I think the build times have gotten down to a reasonable length.

PR #484, opened 2020-05-04
Remove folders then files during make clean

vabarbosa:
Speed up `make clean` by deleting specific folders before searching for individual files to delete

 (in support of #177)

This change updates the `make clean` script to deletes the folders first then the files that way searching for the files it takes less time as it will need to search through less folders.

The biggest improvement is seen when the `node_modules` is deleted first as it can contain many files and nested folders.

 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,15 +30,15 @@ help:
 
 clean: ## Make a clean source tree
 	rm -rf build *.egg-info yarn-error.log
-	rm -rf $$(find . -name dist)
+	rm -rf $$(find . -name node_modules -type d -maxdepth 3)
+	rm -rf $$(find . -name dist -type d)
+	rm -rf $$(find . -name lib -type d)
+	rm -rf $$(find . -name __pycache__ -type d)
 	rm -rf $$(find . -name *.tgz)
-	rm -rf $$(find . -name lib)
-	rm -rf $$(find . -name node_modules)
 	rm -rf $$(find . -name tsconfig.tsbuildinfo)
 	rm -rf $$(find . -name *.lock)
 	rm -rf $$(find . -name package-lock.json)
 	rm -rf $$(find . -name .pytest_cache)
-	rm -rf $$(find . -name __pycache__)
 
 # Prepares Elyra for build/packaging/installation
 yarn-install:
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,10 @@ help:
 
 clean: ## Make a clean source tree
 	rm -rf build *.egg-info yarn-error.log
-	rm -rf $$(find . -name node_modules -type d -maxdepth 3)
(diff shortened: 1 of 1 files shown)


vabarbosa:
in my environment doing multiple comparative runs, this made `make clean` about 3-5 seconds faster, which by itself may not be much. however, there are couple more areas within the the Makefile which we may be able to provide similar scrutiny and get similar minor improvements. individually it may not be much but collectively could possibly produce a noticeable difference.

Makefile
```diff
@@ -30,15 +30,15 @@ help:
 
 clean: ## Make a clean source tree
 	rm -rf build *.egg-info yarn-error.log
-	rm -rf $$(find . -name dist)
+	rm -rf $$(find . -name node_modules -type d -maxdepth 3)
+	rm -rf $$(find . -name dist -type d)
+	rm -rf $$(find . -name lib -type d)
```
kevin-bates, line 35: For the `node_modules`, `dist` and `lib` searches, I'd feel a little more comfortable if we could replace `'.'` with the set of directories in which we expect the named directories to reside. 

Hmm, should this be more like this:
```
        rm -rf node_modules lib dist
	rm -rf $$(find packages -name node_modules -type d -maxdepth 3)
	rm -rf $$(find packages -name dist -type d)
	rm -rf $$(find packages -name lib -type d)
        rm -rf $$(find . -name __pycache__ -type d)
```
otherwise its attempting to delete a bunch of folders that are already deleted.

I'm not concerned about `__pycache__` directories in the source tree (`elyra` or `docs`), but don't want to preclude us adding a directory like `lib` someday.

lresende: What's your concern here? These directories  (`node_modules`, `dist` and `lib`) areis generated by the build and in a `clean` target, it should be deleted from any Elyra subdirectory, no? Could you explain what do you mean when you say: `otherwise its attempting to delete a bunch of folders that are already deleted.`?

vabarbosa: good point @kevin-bates, i agree and have made the changes accordingly

kevin-bates: @lresende - a couple things.  
1. the removal starts at the root of the project - so includes docs and elyra and I don't want to build something under one of these directories - only to have it get deleted because I run `make clean`.  This is more future-thinking; there are no cases of that now, but as soon as I write 1000 lines of code under elyra/metadata/lib, then run `make clean`, I'll be a little upset. :smile:
2. The other is just efficiency, if our plan is to cut down on time, why attempt to remove dozens of directories that no longer exist.  The find is going to locate the top folder, say `node_modules`, then find, all others in node_modules, but the `rm` is going to work off everything the `find` found - which now includes all node_module directories under the top-level node_module directory. 

Thanks for making these changes Va - approving now...

Issue #501, opened 2020-05-05
Cleanup build of jupyterlab extensions


lresende:
We have made a few workaround (and extra steps) on Elyra build to workaround some pre-existent build issues on Jupyterlab 1.x that were preventing updates of linked packages with same version, but are now resolved on JupyterLab 2.x:

Some of the changes (as I recall) are related to:
* Yarn cache clean
* Uninstall / Install of Elyra packages
* Multiple JupyterLab clean/builds

A validation test after any proposed changes would be to
* Start from a clean environment (clean-jupyterlab.sh)
* Build Elyra
* Make a change on Application package
* Build Elyra and make sure Application package has the new contents

lresende:
Related issue #375, that could be handled together.

ajbozarth:
I opened #502 to handle #375 and update our packages to link instead of install. This address some of the issues raised here. That PR does not handle any removal of extraneous cleaning that was added due to the use of install instead of link.

ajbozarth:
Note: I assigned myself since I'm partially working on this but others can feel free to address other parts

vabarbosa:
@ajbozarth i could assist here. any specific part i should start with?

ajbozarth:
I would look at if we can remove any steps, such as "cleaning" steps outside the clean target in Makefile

vabarbosa:
waiting for PRs #502 and #505 (which currently have changes to `Makefile`) to get merged in then i can look more in depth at this. but from a quick read though, a couple possible considerations:

- remove `rm -f yarn.lock package-lock.json` in the **yarn-install** as it's already captured in **clean**
- move `yarn cache clean` to **clean** instead of  **yarn-install** if goal is to keep all the clean scripts in same target
- replace `yarn cache clean` with `rm -rf $(yarn cache dir)` as native `rm` should be faster than `rimraf` (that yarn clean is running)
- **lint-ui** (which runs `yarn run prettier` & `yarn run eslint`) is called after the extensions have been packaged. should it not be before packaging up extensions? fyi, this is addressed in #502 (where packaging extensions is moved to a separate **npm-dists** that should only be called when publishing a release)


(related to #177)

lresende:
After some investigation, here are some more details that I discovered:

Steps done:
0-create a new local branch and change the branch version to a non-existing release
1-build elyra
2-clean-jupyterlab.sh
[...]

PR #502, opened 2020-05-06
Add watch support to makefile

### LINKED ISSUE #375: Look into updating build to support watch
Currently we need to do a full make clean install to get a consistent experience during development. We want to look into what it would take to update our makefile to instead support `jupyter lab --watch` which would allow us to edit our front end packages and see the changes without a full rebuild.

ajbozarth:
Fixes #375
Partially addresses #501

Includes:
- Updates makefile to link packages instead of creating dists and
installing them (allows for use of `watch`)
- Add make  watch command that watches for changes and rebuilds
packages. To be used alongside `jupyter lab --watch`

Known issues:
- `jupyter lab --watch` cant pick up changes in dependencies within
a monorepo, meaning it will not pick up changes in `application`
when called within another extension. A simple workaround is to
stop and start lab, changes in application do get picked up on start

 
           in the file; or

           it.


ajbozarth:
@akchinSTC and @lresende I am unsure how removing the creation of the dists will affect other parts of the code. I was unable to figure out if those dists were leveraged anywhere other than the `install` step of the Makefile. If they are needed elsewhere I can add a new make target that builds them outside the make install flow.

ajbozarth:
With this PR you can utilize watch with the following steps:
1. Initial build with `make clean install`
2. Watch for changes with `make watch`
3. Run Lab with `jupyter lab --watch`

The caveats:
- Steps 2 and 3 above tail and need their own terminal tab/window (make sure the conda env is the same)
- Changes to utils in application that are called within other packages will not be picked up by `jupyter lab --watch` and require stopping and starting to pick up.
- When running `jupyter lab --watch` the start up of lab takes about 2-3x longer.

ajbozarth:
After some clarification from @akchinSTC offline I've added back in the functionality to package the extensions into dists since they are included when creating elyra releases. The added make target assumes `make install` has already been run.

On a separate note, I updated the `.PHONY` targets in this PR. new targets were added to Makefile but were missing from `.PHONY` also since it was getting long I split it across two lines.

ajbozarth:
Per discussion in scrum I will be adding documentation for this. @lresende where should that documentation go? In the README, in the new docs we just added, or both?

lresende:
The documentation should be added to where we describe the contributor workflow, currently, it's available in [README.md](https://github.com/elyra-ai/elyra#development-workflow) and on the [development workflow](https://github.com/elyra-ai/elyra/blob/master/docs/source/developer_guide/development-workflow.md)

kevin-bates:
Could we please have a single location for dev workflow and link README to the docs?  

Then promote the community meeting section header up a level and add a sentence indicating/inviting folks to attend if they're interested in contributing?

Let's also remove the single ticks around `Elyra` in the opening dev workflow paragraph since this is more about the project and not an object within the project.

If anyone feels this needs to be its own separate pull request, then I'm happy to submit a PR later today.

ajbozarth:
@kevin-bates if you want to open another pr for that then I will only update the docs in this pr

kevin-bates:
Thanks Alex - I'll a PR with the changes I proposed above.

ajbozarth:
@lresende should we address the handling of dists in #503 rather than handling it here? ie once I push the docs update this would be ready to merge?

ajbozarth:
I've pushed docs for review

marthacryan:
Just played around with this locally and it worked for me!

karlaspuldaro:
Worked for me too! This is great, gonna help a lot our dev process :)

docs/source/developer_guide/development-workflow.md
```diff
@@ -42,3 +42,25 @@ You can check that the JupyterLab extension was successful installed with:
...
+
+Elyra supports incremental development using `--watch`. This allows you to make code changes to
+front-end packages and see them without running `make install` again.
+
+After installation run the following to watch for code changes and rebuild automatically:
+```bash
+make watch
+```
+
+Then in a separate terminal using the same environment, start JupyterLab in watch mode:
```
vabarbosa, line 56: is it worth clarifying what you mean by _same environment_? i.e., same Conda or Python virtual environment

ajbozarth: I was trying to figure out how to word that without assuming they were using conda. would saying `python environment` make sense? Personally I never had a concept of separate environments in my dev until I started working on lab and installed conda.

vabarbosa: yes, i think `Python environment` makes sense

ajbozarth: updated

lresende:
I would prefer if we do `make install` to produce the necessary npm packages and the full wheel that is ready to install.  I am open to having something like `make dev` to do a simple build`. Otherwise, this looks good.

ajbozarth:
Addressed @lresende comments based on discussion during scrum. We will remove creation of dists in another pr when addressing build speed and/or release automation.

Issue #494, opened 2020-05-05
Merge Notebook Scheduler and Pipeline editor extensions


lresende:
We are looking into expanding the number of buttons available on Elyra to enable submitting KFP Argo YAML and KFP script DSL and we will probably have other additions that are in the intersection of these two extensions. 

Merging them would:
- Avoid the split of pipeline code into the application package
- Make it easier to share code
- Simplify/Speedup build/packaging

Some of the drawbacks would be:
- ?

ajbozarth:
I am interesting in working on this.

I have a few opinions on this:

- The merge makes sense if you start thinking of the "the button" (notebook-scheduler) as just a pipeline submission button to add to notebooks, like how the git button is included with the git extension.
- This would allow the movement of all pipeline specific utils out of application.
- Would we want to introduce subdirectories in `src` separating the two functionalities and utils?
- As for drawbacks, there is the issue of back compatibility, this update might warrant a 1.0 release due to the change.

lresende:
> * The merge makes sense if you start thinking of the "the button" (notebook-scheduler) as just a pipeline submission button to add to notebooks, like how the git button is included with the git extension.
> 

Exactly, with the additions of the other buttons, this is more and more looking as just drivers to submit  one or more`items` to the pipeline.

> * This would allow the movement of all pipeline specific utils out of the application.
> 

Definetly a plus.


> * Would we want to introduce subdirectories in `src` separating the two functionalities and utils?

Not necessarily, most likely we need to introduce one extra widget (submitSingleItem) type of thing, that would then be shared by all `single item`submissions?

> 
> * As for drawbacks, there is the issue of back compatibility, this update might warrant a 1.0 release due to the change.

What is your concern here? The only thing I could think of is the pre-existence of the notebook-scheduler extension, but we could document how to remove or enhance our install code to remove that if there?

ajbozarth:
Moving this into the current milestone since I will be starting work on it tomorrow.

Based on a conversation with @lresende we will be doing this in multiple steps, the first of which will be implemented prior to continuing my work on #232 / #271 / #400 / #403

Step 1: Move the files in `notebook-scheduler` into `pipeline-editor` as is and include functionality found in the index.ts of `notebook-scheduler` in the index.ts of `pipeline-editor`

Step 1.5: Handle utility refactor to address #232 / #271 / #400 / #403

[...]

PR #505, opened 2020-05-06
Merge notebook-scheduler into pipeline-editor

ajbozarth:
Phase 1 of merging the notebook-scheduler and pipeline-editor
extensions. This first step moves all functionality of
notebook-scheduler into pipeline-editor "as is"

Includes:
- moves non-index files into pipeline-editor from notebook-scheduler
- copies code in index.ts and index.css between packages
- updates build files with missing dependacies and imports
- updates docs to remove notebook-scheduler
- deletes notebook-scheduler from packages

This PR will be followed up with Phase 1.5 and 2 which will address,
respectively, updating and moving utils and then refactoring
pipeline-editor to unify the code copied in in this PR and further
modularize it.

Partially addresses #494 

 
           in the file; or

           it.


lresende:
While my understanding is that phases 1.5 and 2 are coming, I believe these are going to address mostly refactoring on the `common service` and applying promises. Then I would like to make this at least following the same pattern already applied to pipeline existing contents.

Some suggestions to be evaluated:
- [ ] Split multiple widgets into separate files
- [ ] Rename `SubmitNotebook.ts` to `PipelineNotebookSubmissionDialog.tsx`
- [ ] Evaluate if it makes sense to create a `ts` file with the logic to get dialog configuration, generate and submit the sigle node pipeline.

ajbozarth:
Note on this change: If you don't run `make clean` prior to checking out this PR it will leave an empty `notebook-scheduler` dir that will break the build. The solution is to either run clean before checking it out or to manually delete the dir

ajbozarth:
Updated with latest master.

PR #509, opened 2020-05-08
Enable schema retrieval via REST API

### LINKED ISSUE #482: Add ability to retrieve schemas for a given namespace via REST API
In order to expose the ability to create metadata instances via the REST API, users will want the corresponding schemas so they can drive instance creation via REST.  This will require improvements to the existing [`SchemaManager`](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/metadata.py#L353) class.

The endpoints will be similar to fetching metadata instances:

Instance retrieval example:
```
http://localhost:8888/api/metadata/runtimes  # get all instances for namespace runtimes
http://localhost:8888/api/metadata/runtimes/my_kfp_instance  # get instance my_kfp_instance in namespace runtimes
```

Schema retrieval example:
```
http://localhost:8888/api/metadata/schemas/runtimes  # get all schemas for namespace runtimes
http://localhost:8888/api/metadata/schemas/runtimes/kfp  # get schema kfp in namespace runtimes
```

Only `GET` requests will be supported against `/api/metadata/schemas`.

kevin-bates:
Changes necessary to enable the retrieval of schemas via the REST API (GET) for a given namespace.  If no schema name is provided, all schemas for the namespace will be returned.

Usage: 
To fetch all schemas for namespace 'runtimes':
```
GET base-url/api/schema/runtimes
```
To fetch the 'kfp' schema for namespace 'runtimes':
```
GET base-url/api/schema/runtimes/kfp
```

Also removed the api endpoint that returns swagger doc in JSON.  Only YAML is supported at this time.

Fixes #482

 
           in the file; or

           it.


vabarbosa:
just poking around and trying to get myself familiar

- should the `'500'` response code also get added to the various paths defined in the `elyra.yaml`? it looks like they could possibly raise a 500 HTTPError
- it may be useful to specify the response type (via the `content` keyword) when defining the responses in the `elyra.yaml`

kevin-bates:
Thanks for looking at this Va.  I thought the response codes were what was _expected_ and 500 is never expected.   However, a quick lookup yields this page: https://swagger.io/docs/specification/describing-responses/ and it looks like both of your points are valid.  

I'll make these changes - thank you for bringing this up.

vabarbosa:
regarding my second comment, i believe for swagger 2.0 the keyword is `produces` (instead of `content`): https://swagger.io/docs/specification/2-0/describing-responses/

kevin-bates:
Hmm - after adding content to the responses, I get the following errors...
```
Structural error at paths./api/metadata/{namespace}/{resource}.get.responses.200
should NOT have additional properties
additionalProperty: content
```
Turns out that because this is swagger 2.0, things are slightly different and this is the document for describing 2.0 responses:  https://swagger.io/docs/specification/2-0/describing-responses/

In this spec, I do not see the 5XX codes listed.  I also found this discussion: https://stackoverflow.com/questions/54238746/should-internal-error-500-be-part-of-swagger-api-documentation

Have you documented 500 codes in the past?  I'm leaning toward defining 500 since, as you point out, is more definitive.

vabarbosa:
> regarding the second point in my comment, i believe for swagger 2.0 the keyword is `produces` (instead of `content`): https://swagger.io/docs/specification/2-0/describing-responses/

sorry i missed it initially, i see there is a `produces` already defined in the yaml file. (was comparing with wrong swagger version spec)

in regards to documenting 500 codes, i haven't had to document them but i've come across many of API docs which have included them. as someone consuming an API i would want to know all possible responses

kevin-bates:
Thank you for the responses Va.  I have made the following changes in the last commit...

1. Updated the swagger document to use OpenAPI 3.1 (the latest)
2. Added 5XX error code indicators
3. Changed the schema endpoints from /api/metadata_schema to /api/schema
4. Added a namespace endpoint /api/namespace that returns the current namespaces - since both the metadata and schema endpoints require a namespace value.

elyra/__init__.py
```diff
@@ -35,14 +36,16 @@ def load_jupyter_server_extension(nb_server_app):
...
          YamlSpecHandler),
         (url_path_join(web_app.settings['base_url'], r'/api/scheduler'), SchedulerHandler),
         (url_path_join(web_app.settings['base_url'], r'/api/metadata/%s' % (namespace_regex)), MetadataHandler),
         (url_path_join(web_app.settings['base_url'], r'/api/metadata/%s/%s' % (namespace_regex, resource_regex)),
-         MetadataNamespaceHandler),
+         MetadataResourceHandler),
+        (url_path_join(web_app.settings['base_url'], r'/api/schema/%s' % (namespace_regex)), SchemaHandler),
+        (url_path_join(web_app.settings['base_url'], r'/api/schema/%s/%s' % (namespace_regex, resource_regex)),
+         SchemaResourceHandler),
+        (url_path_join(web_app.settings['base_url'], r'/api/namespace'), NamespaceHandler),
```
lresende, line 48: should these new apis be relative to metadata?  like /api/metadata/schema?

kevin-bates: They are currently, but that's not necessarily the case down the road.  However, the primary reason for the change was in talking with @vabarbosa we both felt the previous endpoint `/api/metadata_schema` was too long and using something like `/api/metadata/schema/{namespace}` would be ambiguous with the current `/api/metadata/{namespace}/{resource}` endpoint, resulting in an attempt to return the metadata instance named {resource} of namespace 'schema' instead of the schemas in namespace {resource}. I had considered using that endpoint approach initially. :smile:

elyra/api/elyra.yaml
```diff
@@ -1,152 +1,232 @@
...
-                type: array
-                description: The name of the namespace.
-                items:
-                  $ref: '#/definitions/MetadataResource'
-        '404':
+        200:
+          description: Returns the list of current namespaces
+          content:
+            application/json:
+              schema:
```
vabarbosa, line 37: i believe this path (`/api/namespace`) is missing `items`.
similar to how you have it defined for `/api/schema/{namespace}/{resource}` or `/api/metadata/{namespace}`. it would be something like:
```
items:
  $ref: '#/components/schemas/NamespaceResource'
```
where `NamespaceResource` would define the schema of what is returned by the endpoint.

elyra/api/elyra.yaml
```diff
@@ -1,107 +1,232 @@
...
+                type: object
+                properties:
+                  namespaces:
+                    type: array
+                    description: The name of the namespace.
+        5XX:
+          description: Unexpected error.
```
vabarbosa, line 44: i believe this path (`/api/namespace`) is missing `items`.
similar to how you have it defined for `/api/schema/{namespace}/{resource}` or `/api/metadata/{namespace}`. it would be something like:
```
items:
  $ref: '#/components/schemas/NamespaceResource'
```
where `NamespaceResource` would define the schema of what is returned by the endpoint.

kevin-bates: This is really weird. I just happened to be thinking about how we should create a namespace object (i.e., promote it from just the name that it is) so I had to respond.  Since it's just a name, I didn't implement `/api/namespace/{resource}` because there's simply nothing to return at this time.

I suspect we'll want a Namespace object.  They could already use display names and a list-valued schema property.  We'll also need a defaulting property here as well (once we talk about that) - and I'm sure we'll want other items upstream.

If you're okay, how about we leave this API as-is, pending further discussion, then add the `/api/namepace/{resource}` endpoint at that time?  The other option would be to go ahead and define the endpoint, and implement it to return an empty dict for the object (e.g., `{<namespace_name>: {}}`)

vabarbosa: sorry if i wasn't clear. i didn't mean to suggest to add/implement `/api/namespace/{resource}`. rather, what i was trying to say is that the `/api/namespace` returns some object. therefore, it should define the schema for that returned object (in `items`).

if you view the `elyra.yaml` with the Swaggwe UI you will notice the difference:

image



and clicking on the [INVALID button [-]](https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fraw.githubusercontent.com%2Felyra-ai%2Felyra%2F0e300afe808def25d9a9cd33e059bfd86e1b7662%2Felyra%2Fapi%2Felyra.yaml) at the bottom returns:

```
"attribute paths.'/api/namespace'(get).responses.200.content.schema.items is missing"
```

let me know if that makes sense

kevin-bates: I see - thank you for the clarification.  I hadn't used the swagger app to check things out - but that's a great idea - and I'll make the fix.  It sounds like a test to better-enforce the return value is required as well.  Thank you!

elyra/__init__.py
lresende: Can't the metadata handler check if `path` is schema and redirect to `MetadataSchema handler`?

kevin-bates:
@vabarbosa - The namespace response format was corrected such that swagger no longer complains.  I also changed the 5XX references to 500 since we do explicitly return 500 error codes in some cases and I felt that anything other than the 500 we raise should be treated as an anomaly.

One of the changes previously made to change the constant used for testing metadata was incomplete.  This, coupled with the fact that we should really be masking the metadata-tests namespace from runtime uses, led me down the road of making the complete set of changes now since this PR formalizes the use of SchemaManager - which is responsible for whether the testing namespace is returned.  If we want to have that namespace returned, the server should be started with the environment variable `METADTATA_TESTING` set to `1`.

Since the test schema associated with the metadata-tests namespace was previously named 'test', I also updated that to 'metadata-test' to be more explicit.  This also follows the pattern of single-schema namespaces using the singular form of the namespace as the schema name (e.g., code-snippets -> code-snippet, runtime-images -> runtime-image).

I apologize for the last-minute changes relative to the above, but it was the right thing to do relative to this PR.

elyra/__init__.py
kevin-bates: I suppose it could, although I suspect just "redirecting" to another handler is much more involved than that since the entire request context also needs to be conveyed. However, I don't think that's correct to do anyway since, although completely hypothetical, it prevents the ability to ever have a namespace of `schema` - per my previous comment.

Issue #528, opened 2020-05-14
Show pipeline node properties when hover over it


lresende:
When you have a large pipeline with multiple properties set on each node, it becomes hard to troubleshoot or even just see some details of multiple nodes. It would be good if we could display those properties when hovering over the pipeline node.

PR #537, opened 2020-05-15
Update pipeline node tooltip info

vabarbosa:
This change updates the pipeline node's tooltip to show the node's properties.

Fixes: #528 

Before:
image

After:
image

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -70,7 +70,7 @@ const NODE_TOOLTIP_CLASS = 'elyra-PipelineNodeTooltip';
 
 const TIP_TYPE_NODE = 'tipTypeNode';
 
-function NodeProperties(properties: any): React.ReactElement {
+const NodeProperties = (properties: any): React.ReactElement => {
   return (
     <dl className={NODE_TOOLTIP_CLASS}>
       {Object.keys(properties).map((key, idx) => {
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -66,6 +66,28 @@ import * as properties from './properties.json';
 import { PipelineSubmissionHandler } from './submission';
 
 const PIPELINE_CLASS = 'elyra-PipelineEditor';
+const NODE_TOOLTIP_CLASS = 'elyra-PipelineNodeTooltip';
+
+const TIP_TYPE_NODE = 'tipTypeNode';
+
+function NodeProperties(properties: any) {
+  return (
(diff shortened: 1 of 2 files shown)


vabarbosa:
A couple additional comments:

1- The TIP_TYPE_NODE constant is hardcoded (`const TIP_TYPE_NODE = 'tipTypeNode';`) because I did not see a way to import the `TIP_TYPE_NODE` variable from `@elrya/canvas`. I am  not sure if I am missing something but I would expect the constant to be accessible outside the component.

2- Properties are shown by displaying their key/value pairs. I did not try to map/replace the key with more friendly names (e.g., `Environment Variable` vs `vars`) because I wanted to keep it generic/reusable in case there are other types of nodes in the future which may have different properties. However, if people have strong feelings about this it, I can change it.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -66,6 +66,28 @@ import * as properties from './properties.json';
 import { PipelineSubmissionHandler } from './submission';
 
 const PIPELINE_CLASS = 'elyra-PipelineEditor';
+const NODE_TOOLTIP_CLASS = 'elyra-PipelineNodeTooltip';
+
+const TIP_TYPE_NODE = 'tipTypeNode';
+
+function NodeProperties(properties: any): React.ReactElement {
```
marthacryan, line 73:
```suggestion
const NodeProperties = (properties: any): React.ReactElement => {
```

marthacryan: This should fix the lint error.

vabarbosa: @marthacryan Thanks

lresende:
@vabarbosa this looks great. While I understand we want to make it generic, I was wondering what is influencing the order of how the properties are displayed?

lresende:
It also seems that for items like a checkbox (e.g. recursive dependency) it's not showing the actual property info (e.g. checked vc unchecked)

vabarbosa:
@lresende 

> @vabarbosa this looks great. While I understand we want to make it generic, I was wondering what is influencing the order of how the properties are displayed?

no sorting is being done, so the order is just how the data was received from the `@elyra/canvas` CanvasController:

https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L416-L417
https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L76-L77

> It also seems that for items like a checkbox (e.g. recursive dependency) it's not showing the actual property info (e.g. checked vc unchecked)

since this has been merged, will create a new issue and PR to resolve the checkbox issue

PR #539, opened 2020-05-15
Add support for writable metadata via REST and hierarchical location behavior

### LINKED ISSUE #483: Add ability to create metadata instances via REST API
We should expose the ability to create metadata instances via the REST API.  Whether all namespaces support this is something we should probably discuss.  However, for namespaces like `code-snippets`, it makes sense to allow users to create, replace, and delete(?) in such instances.

The endpoint to create an instance would be: `/api/metadata/<namespace>/<resource>` where the body consists of a single JSON entity that is the resource.  For example, the following could be the JSON-body used to create a code-snippet named `powers_of_2` in namespace `code-snippets`:

Request:
```
PUT HTTP://localhost:8888/api/metadata/code-snippets/powers_of_2
```
Request body:
```json
{
  "display_name": "Powers of 2",
  "metadata": {
    "language": "Python",
    "code": [
      "import time\n",
      "i = 2\n",
      "while i > 0:\n",
      "    i *= i\n",
      "    print(i)\n",
      "    time.sleep(0.5)"
    ]
  },
  "schema_name": "code-snippet"
}
```
We use `PUT`since it both creates and replaces resources and specifies that the resource name (e.g., `powers_of_2`) be part of the URI (which seems more natural to me). 

Should we also support `DELETE`?  If so, we should restrict the scope of where instances can be deleted.  Both `PUT` and `DELETE` should only write to local runtime directories.  

This topic probably warrants further discussion.

### LINKED ISSUE #538: Some invalid instances of valid schemas do not display in the output of `elyra-metadata list` CLI
When I run the `elyra-metadata list` command I get the following output:

```
elyra va$ elyra-metadata list code-snippets
[E 2020-05-15 14:57:33,386.386] Schema validation failed for metadata 'project-snippet' in namespace 'code-snippets' with error: 'display_name' is a required property.
Available metadata instances for code-snippets (includes invalid):
Schema         Instance        Resource                                                              
------         --------        --------                                                              
code-snippet   common-imports  /Users/va/Library/Jupyter/metadata/code-snippets/common-imports.json  
code-snippet   create temp     /Users/va/Library/Jupyter/metadata/code-snippets/create temp.json     
code-snippet   is_even         /Users/va/Library/Jupyter/metadata/code-snippets/is_even.json         
code-snippet   js-test         /Users/va/Library/Jupyter/metadata/code-snippets/js-test.json  
```

The `project-snippet` instance fails the validation check (with error: `'display_name' is a required property`) and the instance does not get listed in the table. I was expecting for it to be in the table but include the `**INVALID** (ValidationError)` disclaimer (as seen in #524).

### LINKED ISSUE #543: Fail to create specific metadata on a clean environment
On a clean system/environment install Elyra

Try to add a new metadata:
```
elyra-metadata install  runtime-images \
    --schema_name=runtime-image \
    --name=covid-anaconda \
    --display_name="COVID with Anaconda Python 3" \
    --image_name="lresende/anaconda3:latest"
```

Failure:
```
(/Users/some-user/pd/cn-demo/env) some-user@fuzzy:~/pd/cn-demo$ elyra-metadata install  runtime-images \
    --schema_name=runtime-image \
    --name=covid-anaconda \
    --display_name="COVID with Anaconda Python 3" \
    --image_name="lresende/anaconda3:latest"
The following exception occurred saving metadata instance 'covid-anaconda' for schema 'runtime-image': [Errno 2] No such file or directory: '/Users/some-user/Library/Jupyter/metadata/runtime-images/covid-anaconda.json'
Install a metadata instance into namespace 'runtime-images'.
Options
-------
--replace
	Replace existing instance
--schema_name=<string>
	The schema_name of the metadata instance to install
--name=<string>
	The name of the metadata instance to install
--display_name=<string>
	The display name of the Runtime Image; pattern: ^[a-zA-Z][a-zA-Z0-9-_. (){}]*[a-zA-Z0-9)}]$
--image_name=<string>
	Runtime Image description
(/Users/some-user/pd/cn-demo/env) some-user@fuzzy:~/pd/cn-demo$ 
```

Workaround before executing the command again:
```
mkdir -p /Users/some-user/Library/Jupyter/metadata/runtime-images/
```

kevin-bates:
These changes introduce support for creating, updating, and deleting metadata instances via the REST API.  It also builds on the changes first introduced in PR #462 by introducing the notion of a hierarchy of locations in which metadata instances are searched.

The REST endpoints are (note, the current body structures should be updated, not listing here):
`POST /api/metadata/{namespace}`
`PUT /api/metadata/{namespace}/{resource}`
`DELETE /api/metadata/{namespace}/{resource}`

Metadata instances can reside at three levels: Factory, Shared and User.

**Factory:** Metadata at the factory level will reside in the python environment's location and are created when installing elyra.  This location derives from the active Python's `sys.prefix` and will be found in `sys.prefix/share/jupyter/metadata`. These instances should be treated as read-only.  These metadata instances are returned relative to their respective namespace **only if same-named instances do not exist at any other level**.

**Shared:** Metadata at the shared level will typically reside in either of two locations (although their actual paths vary by platform type).  For Unix-based systems, this will typically be `/usr/local/share/jupyter/metadata` or `/usr/share/juptyer/metadata`, both are valid.  These instances are meant for sharing across users (e.g., code-snippets).  No tooling currently exists to create these instances, so administrators must copy instances to these locations.  These metadata instances are returned relative to their respective namespace **only if same-named instances do not exist at the user level**.

**User:** Metadata at the user level resides in a subdirectory relative to the user running the application.  These will reside in a subdirectory of the user's HOME which also varies by platform.  This is where tools write, update, and delete instances from.

Note: The hierarchy is hidden from the user, for the most part.  However, attempts to delete an instance that resides in either Shared or Factory levels will result in a `PermissionError` (403).  Likewise, attempts to create or update an instance where `replace=False` but instances do exist in Shared or Factory, will also fail with `FileExistsError` (409).

The REST API semantics for POST and PUT are essentially the same as create and update semantics.  That is, POST (create) requires that the _logical_ instance (at **any** level) not already exist and, similarly, PUT (update) requires that the _logical_ instance (at **any** level) **does** exist.  Under the covers this both call the `add()` method with `replace=False` and `replace=True`, respectively.

**TODO:**
- [x] Better define the body on POST and PUT
- [x] Add Swagger entries

Resolves #483 
Fixes #538 
Fixes #543

           in the file; or

           it.

### DIFF
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -235,6 +235,8 @@ paths:
             application/json:
               schema:
                 type: string
+        400:
+          description: The operation is not supported
         404:
           description: The resource to update was not found
         500:
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -138,6 +138,39 @@ paths:
           content: {}
         500:
           description: Unexpected error.
+    post:
+      tags:
+      - metadata
+      summary: Create a metadata instance in a given namespace
+      parameters:
+      - name: namespace
+        in: path
+        description: The name of the namespace
+        required: true
+        schema:
+          type: string
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetadataResourceBodyPost'
+
+      responses:
+        201:
+          description: Returns the resource (path) to the created instance
+          content:
+            application/json:
+              schema:
+                type: string
+        400:
+          description: An error (validation, syntax) occurred relative to the instance data
+        404:
+          description: Namespace not found
+        409:
+          description: Resource already exists
+        500:
+          description: Unexpected error.
 
   /api/metadata/{namespace}/{resource}:
     get:
@@ -171,9 +204,121 @@ paths:
         500:
           description: Unexpected error.
 
+    put:
+      tags:
+        - metadata
+      summary: Update a given metadata resource within a given namespace
+      parameters:
+        - name: namespace
+          in: path
+          description: The name of the namespace
+          required: true
+          schema:
+            type: string
+        - name: resource
+          in: path
+          description: The name of the resource in a given namespace
+          required: true
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetadataResourceBodyPut'
+
+      responses:
+        200:
+          description: Returns the resource (path) of the updated instance
+          content:
+            application/json:
+              schema:
+                type: string
+        404:
+          description: The resource to update was not found
+        500:
+          description: Unexpected error.
+
+    delete:
+      tags:
+      - metadata
+      summary: Delete a given metadata resource from a given namespace
+      parameters:
+      - name: namespace
+        in: path
+        description: The name of the namespace
+        required: true
+        schema:
+          type: string
+      - name: resource
+        in: path
+        description: The name of the resource in a given namespace
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: Returns the resource (path) of the deleted instance
+          content:
+            application/json:
+              schema:
+                type: string
+        403:
+          description: Deletion of the resource is not permitted
+        404:
+          description: The resource was not found
+        500:
+          description: Unexpected error.
+
+
 components:
   schemas:
+    MetadataResourceBodyPost:
+      description: The set of properties comprising the request body for POST
+      required:
+        - display_name
+        - name
+        - schema_name
+        - metadata
+      type: object
+      properties:
+        name:
+          type: string
+          description: The canonical name of the resource
+        display_name:
+          type: string
+          description: The display name of the resource
+        schema_name:
+          type: string
+          description: The schema name used to validate the resource
+        metadata:
+          type: object
+          properties: {}
+          description: A free-form dictionary consisting of additional information
+            about the resource.
+
+    MetadataResourceBodyPut:
+      description: The set of properties comprising the request body for PUT; no field is required
+      type: object
+      properties:
+        name:
+          type: string
+          description: The canonical name of the resource
+        display_name:
+          type: string
+          description: The display name of the resource
+        schema_name:
+          type: string
+          description: The schema name used to validate the resource
+        metadata:
+          type: object
+          properties: {}
+          description: A free-form dictionary consisting of additional information
+            about the resource.
+
     MetadataResource:
+      description: The set of properties comprising a metadata resource entity
       required:
       - display_name
       - name
@@ -198,8 +343,9 @@ components:
           properties: {}
           description: A free-form dictionary consisting of additional information
             about the resource.
-      description: The set of properties comprising a namespace resource entity
+
     SchemaResource:
+      description: The set of properties comprising a schema resource entity
       required:
       - name
       - namespace
@@ -216,7 +362,7 @@ components:
           properties: {}
           description: A free-form dictionary consisting of additional information
             about the resource.
-      description: The set of properties comprising a namespace resource entity
+
   parameters:
     namespace:
       name: namespace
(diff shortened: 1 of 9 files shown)


kevin-bates:
Thank you Va.  
> there are a few more references to mock_runtime_dir in test_utils.py. only calling it out because i noticed you renamed all references in test_metadata_app.py

I will look into updating this. There are some tests that still require NotebookTestBase and I'm not certain if this can be unilaterally replaced.  But, you're right, I probably just didn't update these.

> calling PUT /metadata/{namespace}/{resource} i am able to update any of the fields except name when updating i get a 200 OK response but the name is actually never changed. not if expected behavior is for name to actually get update or to return a failure when trying to change the name

You're right, there's no way to rename an instance. This gets tricky when we have multiple hierarchies involved because you can never delete an instance that's at a "protected" level.  I'm not sure what the best approach is other than perhaps returning a 403 stating that rename ops are not supported.  What do you think?

> calling DELETE /metadata/{namespace}/{resource} i get a 200 OK response when trying to delete a non-existent resource. i was expecting a 404 Not Found error

I'll fix this.

> i receive 404 Not Found (Schema 'random-schema-name' in namespace 'code-snippets' was not found!) and the instance does also gets created. since a failure was returned from the request i was expecting the instance not to be created.

You're correct.  No instance should be created - good catch.

vabarbosa:
@kevin-bates updates look good! 

there is just a minor inconsistency with 

`POST /api/metadata/{namespace}`
`PUT /api/metadata/{namespace}/{resource}`
`DELETE /api/metadata/{namespace}/{resource}` 

all three when successful return a single string (i.e., resource path):

`/Users/va/Library/Jupyter/metadata/code-snippets/print-hello.json`

however, their definitions suggest the returned object is of type `application/json`

https://github.com/kevin-bates/elyra/blob/metadata-writes/elyra/api/elyra.yaml#L163
https://github.com/kevin-bates/elyra/blob/metadata-writes/elyra/api/elyra.yaml#L235
https://github.com/kevin-bates/elyra/blob/metadata-writes/elyra/api/elyra.yaml#L266

the swagger doc could be updated to `plain/text` instead but i would vote for returning a JSON object. preferably the complete of JSON for the resource that was added, updated, or removed. however, if that is too much i would settle for the original resource path in a JSON object:
```
{
   "resource": "/Users/va/Library/Jupyter/metadata/code-snippets/print-hello.json" 
}
```
---

since you are already in this code there are a couple additional inconsistencies i noticed. if you prefer i can push these off to a separate issue:

- `GET /api/schema/{namespace}` is defined in [elyra.yaml](https://github.com/kevin-bates/elyra/blob/metadata-writes/elyra/api/elyra.yaml#L349-L366) as returning an array for `namespace`:
```json
{
  "namespace": [
    {
      "namespace": "string",
      "name": "string",
      "properties": {}
    }
  ]
}
```
however, when i call `api/schema/code-snippets` the response body is does not return the array
```json
{
    "code-snippets": {
        "code-snippet": {
            "name": "code-snippet",
            "namespace": "code-snippets",
            "properties": {}
        }
    }
}
```

- similarly `GET /api/metadata/{namespace}` is defined as returning an array of `namespace`
```json
{
  "namespace": [
    {
      "name": "string",
      "display_name": "string",
      "schema_name": "string",
      "resource": "string",
      "metadata": {}
    }
  ]
}
```
however, calling `api/metadata/code-snippets` the response body is does not return the array
```json
{
    "code-snippets": {
        "print-hello": {
            "display_name": "print hello",
            "metadata": {},
            "schema_name": "code-snippet",
            "name": "print-hello",
            "resource": "/Users/va/Library/Jupyter/metadata/code-snippets/print-hello.json"
        },
        "is_even": { ... }
}
```

i would say the swagger doc is the more accurate definition. however, i suspect it would be easier to to update the swagger doc to match implementation than vice versa.

kevin-bates:
> the swagger doc could be updated to plain/text instead but i would vote for returning a JSON object. preferably the complete of JSON for the resource that was added, updated, or removed.

This same thought had occurred to me as well while implementing these.  I think this would be a good change to make.

I also agree that we should adhere to swagger on the list vs. dict results and would like to get that change included as well.

Thank you for catching these!  They're the kinds of details that end up being expensive to address down the road.

PR #545, opened 2020-05-19
[WIP] Addressed build issues

ajbozarth:

           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@
 # limitations under the License.
 #
 
-.PHONY: help clean yarn-install test-dependencies lint-server lint-ui lint lerna-build npm-packages bdist install
-.PHONY: watch test-server test-ui test-ui-debug test docs-dependencies docs install-backend docker-image
+.PHONY: help clean test-dependencies lint-server lint-ui lint build npm-packages bdist install install-backend
+.PHONY: install-external-extensions dev watch test-server test-ui test-ui-debug test docs-dependencies docs docker-image
 
 SHELL:=/bin/bash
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,12 +41,15 @@ clean: ## Make a clean source tree
...
 	rm -rf $$(find . -name .pytest_cache)
-
-# Prepares Elyra for build/packaging/installation
-yarn-install:
-	rm -f yarn.lock package-lock.json
 	yarn cache clean
-	yarn
+	pip uninstall -y jupyterlab-git
+	- jupyter labextension uninstall --no-build @jupyterlab/toc
+	$(call UNLINK_LAB_EXTENSION,application)
```
lresende, line 47: Could we have an `unlink` target and make `clean` depend on it. Do we need to have `install` also depend on it in case of the user only issuing a `make install` ?

Makefile
```diff
@@ -78,17 +81,25 @@ bdist: npm-packages
...
 	$(call LINK_LAB_EXTENSION,application)
-	$(call LINK_LAB_EXTENSION,theme)
-	$(call LINK_LAB_EXTENSION,code-snippet)
-	$(call LINK_LAB_EXTENSION,pipeline-editor)
-	$(call LINK_LAB_EXTENSION,python-runner)
+	jupyter lab build
+	jupyter serverextension list
+	jupyter labextension list
+
+install-backend: ## Build and install backend
```
lresende, line 89: can we call this `install-server` to be consistent with `test-server`? The other way around is also ok.

Makefile
```diff
@@ -78,17 +81,25 @@ bdist: npm-packages
 
 install: bdist lint ## Build distribution and install
 	pip install --upgrade dist/elyra-*-py3-none-any.whl
```
lresende, line 83: The original idea here would be to enable handing a wheel to someone and he would be able to install/update elyra on his system, but after our conversation, it does not seem that this is possible unless the npm packages are published. We should have a conversation and work out what should we do and maybe get rid of these.

Makefile
ajbozarth: I did have a separate target but I didn't like the order it was executing so I switched back to all of it in one target. I have no string opinions on it though and can separate them again.

As for make install, I can test some more edge cases but I don't believe make install needs to depend on clean anymore, I believe that was a build issue in 1.x that was fixed, but again I'd need to do some more edge case testing to know for sure. @vabarbosa if you're willing to help test the edge cases since you've also been working on some of this.

Makefile
ajbozarth: Sure, I like that better as well. (This was an already existing target that I just reused so I didn't want to edit it)

Makefile
ajbozarth: Based on what we discuss in scrum, I have a couple ideas that would remove quite a bit from the make file and streamline this pr

Makefile
vabarbosa: count me in

ajbozarth:
Closing in deference to #548

PR #548, opened 2020-05-19
Clean up build

### LINKED ISSUE #501: Cleanup build of jupyterlab extensions

ajbozarth:
Alternate solution to #545 that removes the idea of including dists for frontend components in the wheel entirely.

This also addresses @vabarbosa comments in #501

Fixes #501 
 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-.PHONY: help clean-elyra uninstall clean test-dependencies lint-server lint-ui lint build install install-backend
+.PHONY: help purge uninstall clean test-dependencies lint-server lint-ui lint yarn-install build-ui build-server install-server
 .PHONY: install-external-extensions install watch test-server test-ui test-ui-debug test docs-dependencies docs docker-image
 
 SHELL:=/bin/bash
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@
 # limitations under the License.
 #
 
-.PHONY: help clean yarn-install test-dependencies lint-server lint-ui lint lerna-build npm-packages bdist install
-.PHONY: watch test-server test-ui test-ui-debug test docs-dependencies docs install-backend docker-image
+.PHONY: help clean test-dependencies lint-server lint-ui lint build npm-packages bdist install install-backend
+.PHONY: install-external-extensions dev watch test-server test-ui test-ui-debug test docs-dependencies docs docker-image
 
 SHELL:=/bin/bash
 
@@ -41,12 +41,15 @@ clean: ## Make a clean source tree
 	rm -rf $$(find . -name *.lock)
 	rm -rf $$(find . -name package-lock.json)
 	rm -rf $$(find . -name .pytest_cache)
-
-# Prepares Elyra for build/packaging/installation
-yarn-install:
-	rm -f yarn.lock package-lock.json
 	yarn cache clean
-	yarn
+	pip uninstall -y jupyterlab-git
+	- jupyter labextension uninstall --no-build @jupyterlab/toc
+	$(call UNLINK_LAB_EXTENSION,application)
+	$(call UNINSTALL_LAB_EXTENSION,theme-extension)
+	$(call UNINSTALL_LAB_EXTENSION,code-snippet-extension-experimental)
+	$(call UNINSTALL_LAB_EXTENSION,pipeline-editor-extension)
+	$(call UNINSTALL_LAB_EXTENSION,python-runner-extension)
+	jupyter lab clean
 
 test-dependencies:
 	@pip install -q -r test_requirements.txt
@@ -60,12 +63,12 @@ lint-ui:
 
 lint: lint-ui lint-server ## Run linters
 
-lerna-build: yarn-install
+build:
+	yarn
 	export PATH=$$(pwd)/node_modules/.bin:$$PATH && lerna run build
 
-npm-packages: lerna-build
+npm-packages: build
 	mkdir -p dist
-	$(call PACKAGE_LAB_EXTENSION,application)
 	$(call PACKAGE_LAB_EXTENSION,theme)
 	$(call PACKAGE_LAB_EXTENSION,code-snippet)
 	$(call PACKAGE_LAB_EXTENSION,pipeline-editor)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ help:
 # http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-clean: ## Make a clean source tree
+clean-elyra:
```
kevin-bates, line 32: I don't see why `client` was split into `clean-elyra` and `uninstall` when neither of these is used elsewhere.

I also don't like the `-elyra` in the name and, if these do need to be split, would prefer to keep them as separate targets and create a combined target like `purge` or something as the helper.

Makefile
```diff
@@ -60,35 +68,24 @@ lint-ui:
...
-	$(call UNLINK_LAB_EXTENSION,code-snippet-extension-experimental)
-	$(call UNLINK_LAB_EXTENSION,pipeline-editor-extension)
-	$(call UNLINK_LAB_EXTENSION,python-runner-extension)
-	jupyter lab clean
+
+install-external-extensions:
+	pip install --upgrade jupyterlab-git==0.20.0
+	jupyter labextension install --no-build @jupyterlab/toc@3.0.0
+
+install: lint build install-server install-external-extensions ## Build and install
```
kevin-bates, line 83: I'm wondering if we shouldn't split `install-server` into `build-server` and `install-server` and rename `build` to `build-ui`.  (It doesn't look like there can be separate _build_ and _install_ steps for the UI components, but if there were, then add `install-ui`.)

I also think the `lint` should be removed here and let `build` (or whatever it becomes) depend on `lint-ui` and let `build-server` depend on `lint-server`.  That way, devs working on "either side" will always perform the appropriate linting steps (assuming they build using the makefile).

Makefile
ajbozarth: this was a request of @lresende in #545 I'm ok putting them back together

ajbozarth: I can use `purge` instead if @lresende still wants them separate

Makefile
ajbozarth: I actually like this. I'll modularize this based on your ideas and push a new commit

ajbozarth:
I pushed an update based on @kevin-bates review further modularizing the make. This change separate the backend and front end builds and installs and runs their respective linters as part of their builds.

Makefile
lresende: To me clean will remove/delete stuff from the build env or modified by the build. The name, as I understand, has a de-facto standard as `clean`. Now, in the original case, we were using `clean` to also do unlink which before was being done from `install` as in JupyterLab 1.x same version extensions/packages were not being installed unless they were unlinked/uninstalled first, thus it was being forced during install. In case we still need to unlink before it can be updated, we need to make sure we perform that when `install` is called without `clean`.

Makefile
lresende: I just want us to be consistent like we had `install-backend` and `test-server`.

kevin-bates:
I only ever run `make lint-server` so I'd be okay with its removal since each of the separate targets are included in the more common "dev" targets.

ajbozarth:
I sometimes will run `make lint` when I make minor changes and don't want to build to lint. I think of it the same way as `make test` it's meant to be called directly.

kevin-bates:
Sounds good, it's a keeper then. 🎣

vabarbosa:
works for me 👍

ajbozarth:
Fixed the merge conflict via the UI, I need to check it out and double check it still builds as expected

ajbozarth:
tested locally, this is good to go @lresende

lresende:
I think there is still something weird going on with linked packages, I have my test branch where I have bumped the package version to `v0.12.0` and I was using that for testing. Note that this is just a local branch and nothing has been published.

I have switched to this PR branch now, and did a `make clean` and `make install` separately, and when the build finished it shows:

```
jupyter serverextension list
config dir: /Users/lresende/opt/anaconda/envs/dev/etc/jupyter
    elyra  enabled 
    - Validating...
      elyra 0.11.0.dev0 OK
    jupyterlab  enabled 
    - Validating...
      jupyterlab 2.1.2 OK
    jupyterlab_git  enabled 
    - Validating...
      jupyterlab_git 0.20.0 OK
    nbdime  enabled 
    - Validating...
      nbdime 2.0.0 OK
jupyter labextension list
JupyterLab v2.1.2
Known labextensions:
   app dir: /Users/lresende/opt/anaconda/envs/dev/share/jupyter/lab
        @elyra/application v0.12.0  enabled  OK
        @elyra/code-snippet-extension-experimental v0.11.0-dev  enabled  OK*
        @elyra/pipeline-editor-extension v0.11.0-dev  enabled  OK*
        @elyra/python-runner-extension v0.11.0-dev  enabled  OK*
        @elyra/theme-extension v0.11.0-dev  enabled  OK*
        @jupyterlab/git v0.20.0  enabled  OK
        @jupyterlab/toc v4.0.0  enabled  OK
        nbdime-jupyterlab v2.0.0  enabled  OK

   local extensions:
        @elyra/code-snippet-extension-experimental: /Users/lresende/opensource/elyra/elyra/packages/code-snippet
        @elyra/pipeline-editor-extension: /Users/lresende/opensource/elyra/elyra/packages/pipeline-editor
        @elyra/python-runner-extension: /Users/lresende/opensource/elyra/elyra/packages/python-runner
        @elyra/theme-extension: /Users/lresende/opensource/elyra/elyra/packages/theme

   linked packages:
        @elyra/application: /Users/lresende/opensource/elyra/elyra/packages/application

``` 

Note the **@elyra/application v0.12.0  enabled  OK** even though when I look at the source it is `  "name": "@elyra/application", "version": "0.11.0-dev",`

This is more like a comment for now while I have the result log, but I am still going to investigate it further.

lresende:
I saw the following when the extension was being built:
```
jupyter labextension unlink --no-build @elyra/application
An error occured.
ValueError: No linked package for @elyra/application
See the log file for details:  /var/folders/br/fh9knw197l3899rkb_lrybxm0000gn/T/jupyterlab-debug-bqdtw5tp.log
make: [uninstall] Error 1 (ignored)
```

This might have been a side effect of my tests that were trying to install and not link application. Will continue updating here

lresende:
After running `clean-jupyterlab.sh` I can back and forth between the two branches and all seems working. LGTM.

PR #551, opened 2020-05-20
Fix style of extensions on left and right pane

### LINKED ISSUE #550: Code Snippet right panel style need fix
When switching code snippet panel to open on the right side, this is how it looks:
image

Compared to the default left panel:
image

karlaspuldaro:
Fixes #550 
Screenshot with changes:
image

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/style/index.css
+++ b/packages/pipeline-editor/style/index.css
@@ -80,15 +80,6 @@
   max-height: 100vh;
 }
 
-.lm-Widget #jp-left-stack {
-  padding-left: unset;
-}
-
-.lm-Widget #jp-right-stack {
-  margin-top: 0px;
-  margin-right: 0px;
-}
-
 .lm-Widget a {
   color: #1970b7;
 }
--- a/packages/pipeline-editor/style/index.css
+++ b/packages/pipeline-editor/style/index.css
@@ -81,12 +81,12 @@
 }
 
 .lm-Widget #jp-left-stack {
-  padding-left: 8px;
+  padding-left: unset;
 }
 
 .lm-Widget #jp-right-stack {
-  margin-top: 25px;
-  margin-right: 5px;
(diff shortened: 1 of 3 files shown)


packages/pipeline-editor/style/index.css
```diff
@@ -81,12 +81,12 @@
...
 .lm-Widget #jp-left-stack {
-  padding-left: 8px;
+  padding-left: unset;
 }
 
 .lm-Widget #jp-right-stack {
-  margin-top: 25px;
-  margin-right: 5px;
+  margin-top: 0px;
+  margin-right: 0px;
```
ajbozarth, line 89: if this is unnecessary we should just remove it instead

karlaspuldaro: I agree. Still looking at how the deletion can be picked up, as I tried a clean build and clearing browser cache as suggested but it didn't help

ajbozarth: I traced back to the original commit https://github.com/elyra-ai/elyra/commit/b842f49ae095df4ef98b11f88d07e1a52356b259 and I'm not sure these ever were needed

ajbozarth: I just realized after noting that the pr number for that commit didn't add up that the commit is from when the button was still its own repo. Back then it had it's own ui that opened in the right sidebar. We must have not removed the css when we removed that part of the ui

ajbozarth: I tested with these deleted and it looks great.

karlaspuldaro:
Thx @ajbozarth also for helping in getting my clean script to properly work, that was the issue. 
Only when overriding css selections worked for me, because when deleting those lines, my browser would pick up an old css file from notebook-scheduler-extension, which was still installed in my env

Issue #403, opened 2020-04-03
Reuse ExpandableComponent in SubmissionHandler


karlaspuldaro:
`ExpandableComponent` is a reusable react component created for code snippets display, that takes a display name and a react element with extra info that is displayed upon expand button click event. 
It is currently located in `packages/code-snippets/`
A similar usage would be in SubmissionHandler `showDialog()`
We should move ExpandableComponent to packages/application or to some util folder, and also integrate it to SubmissionHandler.

vabarbosa:
related work: https://github.com/elyra-ai/elyra/pull/546

ajbozarth:
Taking my name off this so others can pick it up, I don't know when I'd get to it.

The goal for this issue is to try and merge `ExpandableComponent` and `ExpandableErrorDialog` or at least have one leverage the other in some way. The two classes current duplicate quite a bit of code. Both are located within `ui-componets` so any refactor would have minimal impact to the UI/UX

vabarbosa:
i can look into this

ajbozarth:
thanks @vabarbosa feel free to ping me with any questions. My expectation was to make `ExpandableErrorDialog` just extend `ExpandableComponent` in some way

PR #546, opened 2020-05-19
Add separate package for modular UI components

marthacryan:
In anticipation of creating a similar extension (see #186) to code snippets, 
this moves code that's already very modular into a separate package
called ui-components. To start, only ExpandableComponents.tsx is
moved, but moving forward, more will be added to the ui-components
package.
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

Issue #513, opened 2020-05-08
Update the `/api` endpoints to `/elyra`


vabarbosa:
`/api` is used as the starting path for Elyra endpoints. e.g., `/api/metadata/code-snippets`.

while `/api` is convenient and also easy to remember, after some discussion and playing around with the swagger docs i have a couple comments:

1. `/api` is very common. there is the possibility that some extension (or Jupyter functionality) comes along which also uses the `/api` path. this clash could cause issues.

1. `/api` is not clearly identifiable as an Elyra specific endpoint. someone could easily assume that `/api/metadata/runtimes` is a Jupyter endpoint and have no way of immediately realizing it is strictly for Elyra.

with that being said, i would think Elyra would want its own/unique path to clearly distinguish its endpoints. e.g.,  `/elyra`.

the possibility of some other 3rd party component or a Jupyter update being introduced that used that same `/elyra` path is much more unlikely. at the same time, the Elyra API endpoints would become clearly recognizable:

- `/elyra/metadata/code-snippets`
- `/elyra/metadata/runtimes`
- `/elyra/schema`
- etc.

it doesn't necessarily have to be `/elyra` (although i think that works well) but something more distinguishable and less common than `/api` should be considered.

i would expect changing/updating the endpoint path is a manageable task but then again i don't know how set in stone the Elyra APIs are to properly speak on how much work would actually be involved. so i leave this here for discussion and thoughts from others?

kevin-bates:
I agree.  In fact, all of the notebook endpoints begin with /api.  I think this change makes sense.

kjdoyle:
I agree with this as well. We always try to differentiate the different APIs with a different prefix.

lresende:
I am ok with either `/api/elyra` or just `/elyra`.

kevin-bates:
I asked about this in today's jupyter_server meeting and the convention is for the server extension to define its own "root".  They referenced both Lab (`/lab`) and Voila (`/viola`) as examples.  As a result, we should go with `/elyra`.

vabarbosa:
nice. thanks for checking.

kevin-bates:
@ajbozarth is currently refactoring front-end methods that, upon completion, will result in fewer areas that reference `/api/xxx` today.  As a result, we should wait until the following issues are closed: #271, #400

kevin-bates:
[...]

PR #554, opened 2020-05-21
Refactor of the submission utility

### LINKED ISSUE #232: Generalize status code handling
Per [this comment](https://github.com/elyra-ai/elyra/pull/231#pullrequestreview-339297287) we should probably relax some of the status code interpretation in order to better facilitate server-side changes or even other servers in general without having to update front-end code to handle the addition of success (and error) codes.

In particular, we should expand the set of success status codes from 200 (which is currently all that the elyra server returns) to also include 201 (when resources are created via POST), 204 (when resources are deleted), etc.

### LINKED ISSUE #271: Create a service class for runtimes connections

### LINKED ISSUE #400: Update submission util to be more generic
As discussed in #241 the new code-snippet package has brought to light issues with the submission utils in application.

copied from my comment in #241:
> The current util is too specific to runtimes and pipeline submissions and the errors read wrong for code snippets, relatedly I believe the issues causing us to use a signal is a self-created problem from using the existing submission util that assumes it's only called by a user taking an action, not as part of startup as code snippet uses it.

From that I believe we should make our error handling more generic, allowing more control over how errors are handled, like letting errors surface in ways other than a blocking dialog and making sure error text matches the error case.

ajbozarth:
The submission utility in application has become a mess since
it's creation due a limited view of it's usage at conception
and extending it's functionality over time.

This overhaul of the utility accomplishes the following:
- Moves from passing call back functions to using async utility functions
- Moves pipeline related utility functions into pipeline-editor (Part of #494)
- Update all calls to metadata end points to use the metadata service. Fixes #271
- Move ErrorDialogContent into ui-components. First step towards #403
- Update function, class, and file names to match actual functionality
- Modularized HTTP_STATUS success code verification. Fixes #232

Fixes #400

 
           in the file; or

           it.


ajbozarth:
Sorry the diff is poor. Due to all the renamed files/extensions it is hard to tell what's changed in those files. This pr will also need to make some updates once #546 is merged since I plan to move `ErrorDialogContent` since this PR should be followed up with a refactor merging `ExpandableComponent` and `ErrorDialogContent` to finish #403

packages/application/src/services.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return showDialog({
+      title: 'Error retrieving metadata',
+      body: <p>No {metadataName} metadata has been configured.</p>,
+      buttons: [Dialog.okButton()]
+    });
+  }
+
+  static async getMetadata(name: string): Promise<any> {
+    const metadataResponse: any = await RequestHandler.makeGetRequest(
+      'api/metadata/' + name,
```
kevin-bates, line 36: `name` is a _namespace_ - we should name it as such.

packages/code-snippet/src/CodeSnippetManager.ts
```diff
@@ -25,37 +25,25 @@ export interface ICodeSnippet {
...
+    const allCodeSnippets: ICodeSnippet[] = [];
+
+    for (const codeSnippetKey in codeSnippetsResponse) {
+      const jsonCodeSnippet = codeSnippetsResponse[codeSnippetKey];
+      const codeSnippet: ICodeSnippet = {
+        name: jsonCodeSnippet.name,
+        displayName: jsonCodeSnippet.display_name,
+        description: jsonCodeSnippet.metadata.description,
+        language: jsonCodeSnippet.metadata.language,
+        code: jsonCodeSnippet.metadata.code
```
kevin-bates, line 41: Hmm.  Should we have the server _model_ promote the various metadata-level properties to the same as `name` and `description`?  If I'm reading these correctly, you could then have `jsonCodeSnippet` implement `ICodeSnippet`.

packages/pipeline-editor/src/SubmissionHandler.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { showDialog, Dialog } from '@jupyterlab/apputils';
+import * as React from 'react';
+
+export class SubmissionHandler {
+  static async getRuntimeImages(): Promise<IDictionary<string>> {
+    const runtimeImages = await FrontendServices.getMetadata('runtime-images');
+    const images: { [key: string]: string } = {};
+    for (const image in runtimeImages) {
+      const imageName: string = runtimeImages[image]['metadata']['image_name'];
+      images[imageName] = runtimeImages[image]['display_name'];
```
kevin-bates, line 31: I think using the `name` field would probably make for a better index than `image_name`.  `name` is derived from the [instance's filename](https://github.com/elyra-ai/elyra/tree/master/etc/config/metadata/runtime-images), whereas `image_name` may contain ["unfriendly" characters](https://github.com/elyra-ai/elyra/blob/master/etc/config/metadata/runtime-images/tensorflow_2x_gpu_py3.json#L4).

packages/pipeline-editor/src/SubmitNotebook.ts
```diff
@@ -73,52 +73,33 @@ export class SubmitNotebookButtonExtension
...
-
-              SubmissionHandler.submitPipeline(
-                pipeline,
-                result.value.runtime_config,
-                'notebook'
-              );
-            });
-          }
-        );
+    const runtimes = await FrontendServices.getMetadata('runtimes');
```
kevin-bates, line 81: `getMetadata('runtimes')` is called from `handleExport()` and `handleRun()`, as well as here.  I can't tell if `FrontendServices.getMetadata()` unconditionally issues a request to the server or not, but its good that we have a central place in case we want to cache metadata in general.

packages/application/src/services.tsx
ajbozarth: @lresende wrote this in a previous pr but I agree with you so I'll update it

packages/code-snippet/src/CodeSnippetManager.ts
ajbozarth: This was written by @karlaspuldaro originally with @lresende help. Iiuc I believe `ICodeSnippet` was written to capture what the UI currently needs, but the back end schema (what the api returns) was written with the expectation of adding more metadata in the future and includes fields we may not yet use in the UI.

packages/pipeline-editor/src/SubmissionHandler.tsx
ajbozarth: Sadly we didn't get to pick the index. The json returned here is actually copied as is into a canvas schema. So it has to match the expected format :(

ajbozarth: addendum to my previous comment, we do have a function that inserts the json into that schema file so we could always make this cleaner and then "revert" is to the necessary format in that function. If you really feel the service function should return a better formatted dict.

packages/pipeline-editor/src/SubmitNotebook.ts
ajbozarth: The original intention of #271 was to create a service class for runtimes, so I initially started writing a function like `getRuntimeImages` but I stop when I realized that it would just be:
```
static async getRuntimes(): Promise<any> {
    const runtimes = await FrontendServices.getMetadata('runtimes');
    return runtimes;
}
```
and I felt that was silly. Unlike our other metadata runtimes returns already in the format we need to use it.

Since we currently don't cache metadata I didn't think it was worth it, but if we want to enable an easy way to do that in the future I can create the `getRuntimes()` function in SubmissionHandler to be used in those the there uses.

packages/application/src/services.tsx
ajbozarth: pushed an update for this

ajbozarth:
@kevin-bates thanks for the quick review. As for the api endpoint changes, that should be easy to do in a follow up pr since with this pr we only write out each of the three endpoints once on the front end.

packages/code-snippet/src/CodeSnippetManager.ts
kevin-bates: Ok - if there's no advantage to removing the `metadata` level, I'm good.  This is the kind of thing we need to do prior to 1.0.  I suppose there's some advantage to keeping the metadata-specific properties contained.

packages/pipeline-editor/src/SubmissionHandler.tsx
kevin-bates: Could you please provide a reference to the schema file?  I'm sure it's from https://github.com/elyra-ai/pipeline-schemas but never know which is what in there.

Changing the index sounds like more effort than its worth since we don't really control things here.  I figured `images[]` was our thing.

packages/pipeline-editor/src/SubmitNotebook.ts
kevin-bates: Why wouldn't we just add the caching into `getMetadata` and allow all metadata namespaces to use the same caching framework - a cache per namespace - when that time comes?

kevin-bates:
Correct - this is exactly why #513 is [pending the merge of this PR](https://github.com/elyra-ai/elyra/issues/513#issuecomment-628763087).

packages/application/src/services.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  static async getMetadata(namespace: string): Promise<any> {
+    const metadataResponse: any = await RequestHandler.makeGetRequest(
+      'api/metadata/' + namespace,
+      false
+    );
+
+    if (Object.keys(metadataResponse[namespace]).length === 0) {
+      return this.noMetadataError(namespace);
```
vabarbosa, line 41: i don't think an empty list necessarily means an error retrieving metadata. for example, i could have a `code-snippets` namespace but just not have any code snippets at the moment. in this case an empty list is returned which i think should be valid and not considered an error. i think the error dialog should be presented only if the request returns an actual error.

in addition the use of a dialog here can become tedious. if i tend to leave the code-snippet panel open by default, each time i refresh or relaunch lab i am prompted with the dialog. instead of opening an error dialog return the empty list and leave it up to the caller to decide how to handle the empty list. for example, code snippets could just display some placeholder text (i.e., 'No code snippets have been configured') in the panel when there are no snippets.

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,137 @@
...
+    requestOptions: any,
+    displayWaitDialog: boolean
+  ): Promise<any> {
+    // use ServerConnection utility to make calls to Jupyter Based services
+    // which in this case are the in the extension installed by this package
+    const settings = ServerConnection.makeSettings();
+    const requestUrl = URLExt.join(settings.baseUrl, requestExt);
+
+    console.log(
+      'Sending a ' + requestOptions.method + ' request to ' + requestUrl
```
vabarbosa, line 97: more of a nit (or personal preference on my part). i see this (string concatenation) throughout. any particular reason for not using template literals instead? i tend to find them more readable. for example

```javascript
console.log(`Sending a ${requestOptions.method} request to ${requestUrl}`)
```

packages/pipeline-editor/src/SubmissionHandler.tsx
ajbozarth: This is the template we insert it into https://github.com/elyra-ai/elyra/blob/1947597c3b3e63e7377ac9bffae6f1c083f2fe9d/packages/pipeline-editor/src/pipeline-template.json

packages/application/src/services.tsx
ajbozarth: This is a good argument for having a separate `getRuntimes` function like @kevin-bates and I discussed in the other thread above. Then we could take this check out of the getMetadata call and move it into the specific namespace wrappers that require it (runtimes/runtimeImages). I'll make the changes

packages/application/src/requests.tsx
ajbozarth: This was old code, but I agree, I'll update it

packages/pipeline-editor/src/SubmissionHandler.tsx
kevin-bates: So the _keys_ of `images` wind up filling in the [`{{image}}`](https://github.com/elyra-ai/elyra/blob/1947597c3b3e63e7377ac9bffae6f1c083f2fe9d/packages/pipeline-editor/src/pipeline-template.json#L16) in the template?

And does `display_name` get used to map to the image in a dropdown somewhere?

ajbozarth: whoops that was the wrong file https://github.com/elyra-ai/elyra/blob/87f864b6951c1fc5281ccfd9166e7f5945098825/packages/pipeline-editor/src/properties.json

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,137 @@
...
+              resolve(result);
+            },
+            (reason: any) => {
+              // handle 404 if elyra server extension is not found
+              return this.server404();
+            }
+          );
+        }
+      );
+    });
```
karlaspuldaro, line 132: I find the flow here confusing, and there might be exceptions not being caught, then serverResponse could never be resolved. I believe at the end of every possible end point should be either a resolve or reject, instead of returns.
- ServerConnection.makeRequest doesn't have a catch for promise rejection
- if response.json fails, does it fall into this.server404? 
- if response.json gets a result with a server error status, shouldn't line 120 resolve the promise instead of using return?

packages/application/src/ErrorDialogContent.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const width = this.dialogNode.clientWidth;
+    const height = this.dialogNode.clientHeight;
+
+    if (
+      expanded &&
+      (width < ERROR_DIALOG_WIDTH || height < ERROR_DIALOG_HEIGHT)
+    ) {
+      this.dialogNode.style.width = ERROR_DIALOG_WIDTH + 'px';
+      this.dialogNode.style.height = ERROR_DIALOG_HEIGHT + 'px';
```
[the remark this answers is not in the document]
ajbozarth: @vabarbosa could explain better, but I believe it's a bit of a hack and it wouldn't work using plain css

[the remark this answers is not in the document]
vabarbosa: good question. it a result of the way [CSS resize](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) works.

when an HTML element is resizeable browsers will explicitly set its the width and height attributes inline as it's being resized. therefore using CSS classes for width and height would not work as the inline style would supersede it, so the way to reset/change the element's size is by updating the inline style.

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+import { showDialog, Dialog } from '@jupyterlab/apputils';
+import { URLExt } from '@jupyterlab/coreutils';
+import { ServerConnection } from '@jupyterlab/services';
+
+import * as React from 'react';
+
+import { ErrorDialogContent } from './ErrorDialogContent';
+
+const HTML_STATUS_SUCCESS = [200, 201];
```
lresende, line 25: HTTP response status codes indicate whether a specific HTTP request has been successfully completed. Responses are grouped in five classes:

- Informational responses (100–199),
- Successful responses (200–299),
- Redirects (300–399),
- Client errors (400–499),
- Server errors (500–599).


Within successful responses these are defined on the RFC 

- 200 ok
- 201 created
- 202 accepted
- 203 Non-Authoritative information
- 204 No Content
- 205 Reset Content
- 206 Partial Content
and a few other WebDAV related

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export class RequestHandler {
```
lresende, line 27: Can we make this visible only to application package, and expose this via Application.invokeService or something like that ?

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  static server404(): Promise<Dialog.IResult<any>> {
+    return showDialog({
+      title: 'Error contacting server',
+      body: <p>Elyra service endpoint not found.</p>,
+      buttons: [Dialog.okButton()]
+    });
+  }
+
+  static async makeGetRequest(
```
lresende, line 63: I believe that, with the latest @kevin-bates updates, we need to perform all the CRUD operations. Is the expectation that other type of requests fallback to makeServerRequest?

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,135 @@
...
+    displayWaitDialog: boolean
+  ): Promise<any> {
+    return this.makeServerRequest(
+      requestExt,
+      { method: 'POST', body: requestBody },
+      displayWaitDialog
+    );
+  }
+
+  static async makeServerRequest(
```
lresende, line 86: we can use some documentation here on the parameters

packages/application/src/requests.tsx
```diff
@@ -0,0 +1,135 @@
...
+    requestExt: string,
+    requestOptions: any,
+    displayWaitDialog: boolean
```
lresende, line 89: Could we have a separate request, like `makeServerLongRequest` that would set this as true. We could then implement things like timeout, retries, etc?

packages/application/src/services.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { Dialog, showDialog } from '@jupyterlab/apputils';
+import * as React from 'react';
+
+import { RequestHandler } from './requests';
+
+/**
+ * A utilities class for various elyra services.
+ */
+export class FrontendServices {
```
lresende, line 25: Jupyter already has `JupyterFrontEnd`  class which I really think that it wasn't a good name choice, and with this new one, with a similar name, could confuse users? What about just calling this like `Application` and then we would have `Application.getMetadata()`?

packages/application/src/services.tsx
```diff
@@ -0,0 +1,42 @@
...
+    });
+  }
+
+  static async getMetadata(namespace: string): Promise<any> {
+    const metadataResponse: any = await RequestHandler.makeGetRequest(
+      'api/metadata/' + namespace,
+      false
+    );
+
+    return metadataResponse[namespace];
```
lresende, line 40: Is the dict format returned here the best option to return, or should we use a list?
Understanding the usage pattern by the UI components would help on this decision (e.g if all consumers loop trough and populate UI components, list would probably make more sense... if there are scenarios where lookups are used, then dict might actually be the best option. Another approach is to enable something like getMetadata().asDict and getMetadata.asList()

packages/pipeline-editor/src/SubmissionHandler.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FrontendServices, RequestHandler } from '@elyra/application';
+import { showDialog, Dialog } from '@jupyterlab/apputils';
+import * as React from 'react';
+
+export class SubmissionHandler {
```
lresende, line 21: Again, this looks good, but it would be good to actually have a facade service here as well, with something like PipelineService.submitPipeline()which will somewhat follow the pattern we are trying to establish throughout the project.

packages/application/src/ErrorDialogContent.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { caretDownIcon, caretRightIcon } from '@jupyterlab/ui-components';
+
+import * as React from 'react';
+
+const MESSAGE_DISPLAY = 'elyra-pipelineSubmission-messageDisplay';
+const ERROR_DISPLAY_BUTTON = 'elyra-pipelineSubmission-errDisplayButton';
+const ERROR_DETAILS = 'elyra-pipelineSubmission-errDetails';
+const ERROR_DETAILS_VISIBLE = 'elyra-pipelineSubmission-error-visible';
+const ERROR_DETAILS_HIDDEN = 'elyra-pipelineSubmission-error-hidden';
```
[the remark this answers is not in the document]
ajbozarth: nice catch, that actually just poor naming since the function that utilizes it used to be called `pipelineSubmission`. I'll update the css names to be generic

packages/application/src/requests.tsx
ajbozarth: I figured we could add more valid cases here as we have them. Currently these are the only two valid success statuses that the elyra api returns. I can make this check for any value in the 200–299 range instead, but I though it'd be better to be specific because if we get a "success" code that shouldn't be possible we should treat it as an error.

packages/application/src/requests.tsx
ajbozarth: But this is meant to be a utility class, though I would agree that making all but `makePostRequest` and `makeGetRequest` private would be prudent. I'm unsure how we would make these two functions available outside of application otherwise.

packages/application/src/requests.tsx
ajbozarth: will do, after talking to @karlaspuldaro I plan to add more documentation to this file in general

packages/application/src/requests.tsx
ajbozarth: I'm not against this but how exactly would `makeServerLongRequest` be called. currently this function is only called by `makePostRequest` and `makeGetRequest` so would we need to add a `long` param to them?

packages/application/src/services.tsx
ajbozarth: This currently just returns the metadata "as is" and leaves the formatting to the service functions that call it. Each service function handles error cases and formatting to match how that metadata namespace is used by the frontend.

packages/application/src/services.tsx
ajbozarth: I also am not a fan our this name, but I don't think Application is any better. I's be more in favor of just naming it `MetadataService` since that's what it is.

packages/pipeline-editor/src/SubmissionHandler.tsx
ajbozarth: I'll update this to be `PipelineService` I almost named it that during development anyway.

packages/application/src/requests.tsx
ajbozarth: that was my original intention yes

packages/application/src/requests.tsx
ajbozarth: thinking about this more I'm unsure how creating a wrapper function around this setting this to true would enable things like timeout, retries, etc. `makeServerRequest` handles any "errors" itself. If we wanted it to be more extendable we would have to let `makeServerRequest` return errors and add error handling to every service that calls it, most of which would be duplicate code that currently resides in `makeServerRequest`

packages/application/src/requests.tsx
ajbozarth: @karlaspuldaro and I discussed this in detail over Webex to get both our head around how it work and I will be adding in-line documentation to explain it, no code changes are required.

ajbozarth: added comments

packages/application/src/requests.tsx
ajbozarth: added doc comments

packages/pipeline-editor/src/SubmissionHandler.tsx
ajbozarth: renamed

ajbozarth:
@lresende I've pushed some updates address some of your comments as well updating to latest master and moving ErrorDialogContent.tsx to ui-components.

packages/application/src/requests.tsx
lresende: @vabarbosa Please definetlly educate the `backend engineers`

packages/application/src/requests.tsx
lresende: I would prefer that the accepted range be (200-299) but would be ok with (200-2010) for now. @kevin-bates thoughts here? what are the success status code range you have been using?

packages/application/src/services.tsx
lresende: Yes, the "as is" is a dict. And then all the consumers have to convert that to list, at least on couple places I saw.

packages/application/src/requests.tsx
```diff
@@ -82,7 +109,18 @@ export class RequestHandler {
       displayWaitDialog
     );
   }
-
+  /**
+   * Make an request to the jupyterlab server.
+   *
+   * @param requestExt - The url for the request.
+   *
```
lresende, line 115: the `Ext` here really made me confused, could we make a better name here? maybe requestURL, endpoint, servicePath...

packages/application/src/requests.tsx
```diff
@@ -82,7 +109,18 @@ export class RequestHandler {
...
+   * @param requestOptions - The initialization options for the request.
+   *
```
lresende, line 117: What are these options? Valid values?

packages/application/src/requests.tsx
```diff
@@ -82,7 +109,18 @@ export class RequestHandler {
...
+   * @param displayWaitDialog - Whether or not to display a dialog warning
+   * warning the use the request may take time.
+   *
```
lresende, line 120: If we don't make a wrapper, could we make this a more generic name that flags the call as expensive, which will take a while to be submitted, and would for now require the dialog to show up, but could enable other things to be handled in this special case?

packages/application/src/services.tsx
```diff
@@ -20,9 +20,9 @@ import * as React from 'react';
 import { RequestHandler } from './requests';
 
 /**
- * A utilities class for various elyra services.
+ * A utility class for handling elyra metadata calls.
  */
-export class FrontendServices {
+export class MetadataService {
```
lresende, line 25: This should be a facade for multiple common things available from the Application package and usable for multiple Elyra packages, so metadata service is not a good name. I rather stay with FrontendServices if we are not going to use Application

packages/application/src/requests.tsx
kevin-bates: We're only using 200 and 201.  204 (No Content) could be used on delete if we were to choose not to return the deleted resource, but since we return content we use 200 on deletes.

packages/application/src/services.tsx
kevin-bates: If this call is just calling `GET /api/metadata/{namespace}` we return a JSON object of the form: [`{"runtimes": [ array of metadata instances ]}`](https://github.com/elyra-ai/elyra/blob/master/elyra/api/elyra.yaml#L126-L134), so it's both (a dict and a list).

packages/application/src/requests.tsx
karlaspuldaro: I still think this is an important piece since it's being called by any get and post requests, and as agreed in our discussion, it at least requires more testing for edge cases to make sure we don't have problems maintaining this in the future. 

I tried running  the prototype below in node to see what happens when something goes wrong somewhere inside, await doesn't seem to catch all exceptions if they're not properly handled:
```
const asyncFunc = async () => {
	// First promise
	const p1 = new Promise((res, rej) => {
		
		// Second promise
		new Promise((res2,  rej2) => { 
		setTimeout(
			() => {
				// Let's say there is some sort of error in this inner promise
				console.log("rejecting second promise");
				rej2("exception: rej2");
			}
			, 1000)}
		)
		.then(
			result => {
				if(!2===2){
					return console.log("some server error");
					// why not resolve
				}
				resolve(result);
			},
			error =>  {
				// Exception catch block
				console.log("404");
				// do we want to resolve or reject here?
			}
		);

	}); // no exception catch block for p1
	

	const result = await p1;
	console.log("p1 done! p1 res = "  + result + "\nReturning...");
	return result;
}


asyncFunc();
console.log("waiting for promise...");
```
And my output was:

```
waiting for promise...
rejecting second promise
404
```
In this case, makeServerRequest never gets resolved and its return is never called, even if it called 404 callback and the error bubbled up.
Even if we only treat the specific cases we want our flow to fall into and not expose anything unnecessary to the client, maybe we should make sure every possible promise path is being handled.
I'm not totally sure what are the consequences of hanging promises, or even if our flow is somehow controlled and changes are not required, test scenarios can answer this question.

ajbozarth: ok so your argument is that we should "handle" all the possible error cases, even if we should reach them. (I would just raise an error of some sort in those cases, not a dialog)

packages/application/src/requests.tsx
ajbozarth: though I would normally want to only handle expected cases as success, since this is a util I will update this to handle all 200-299 as success to follow html standard

packages/application/src/requests.tsx
ajbozarth: this part of the url is traditionally refereed to as the `path` so how about I update it to `requestPath`?

packages/application/src/requests.tsx
ajbozarth: this is passed as is to `makeRequest` but that function calls it `init` and defines it as `The initialization options for the request` should we change our wording to match?

packages/application/src/requests.tsx
ajbozarth: hmmm, I guess we can. I'll change the name to `longRequest`, it's more generic but still gets its purpose across.

packages/application/src/services.tsx
ajbozarth: Ok, I'll change it back to `FrontendServices` then, though if you prefer `ApplicationServices` I believe I would be ok with that. I just want to the classname to be clear it's a "services" class

ajbozarth:
@lresende @kevin-bates @karlaspuldaro I believe I've addressed everyones's review comments and managed to "rebase" on master (the commit history is a mess, but that'll be fixed by a squash before merge anyway). 

I also should note that a few open PRs will step on the toes of this one and will require updates depending on what merges first: #544 #560 #561

ajbozarth:
Once #561 is merged I'll update this to address the changes in #544 #560 and #561

ajbozarth:
I just pushed an update that merges in the latest master and addresses some issues raised in the merge. This should be good to go now if everyone could take a look and review

packages/application/src/requests.tsx
lresende: As a consumer, how do I figure out what valid options that I should use here? Is this based on some other function parameter field requirement/definition? is there any documentation available to help consumers?

packages/application/src/requests.tsx
lresende: I think more documentation is good, but I am really asking for [typedoc](https://typedoc.org/guides/doccomments/) for the function objective and expected inputs/outputs

packages/ui-components/src/ErrorDialogContent.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface IErrorDialogProps {
+  reason: string;
+  message: string;
+  timestamp: string;
+  traceback: string;
+  default_msg: string;
+}
+
+export class ErrorDialogContent extends React.Component<
```
lresende, line 38: Is this more like an `ErrorDialogWidget` or `ErrorDialogComponent`

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -0,0 +1,127 @@
...
+      pipeline: pipeline,
+      export_format: pipeline_export_format,
+      export_path: pipeline_export_path,
+      overwrite: overwrite
+    };
+
+    await RequestHandler.makePostRequest(
+      'api/pipeline/export',
+      JSON.stringify(body),
+      true
```
lresende, line 118: this should be false, we don't have to create/upload any dependencies etc when exporting

packages/ui-components/src/ErrorDialogContent.tsx
ajbozarth: I just kept the name, but following our other dialog body naming conventions we could change this to just `ErrorDialog`, `ServerErrorDialog`, or ExpandableErrorDialog` depending on how we want to classify it. I think the last one would be the most verbose and clear

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth: I can make this `false` if that's what you want, I was just keeping current functionality, which is that exporting also displayed the dialog. As far as my experience, exporting still took enough time that it was worth having the dialog though.

lresende:
I think I am done with my review, and I would like a few things before approving:
* Name consistency: PipelineService.ts and CodeSnippetService.ts
* Updated documentation on services exposed by Application package

Other things could be handled in future prs.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -40,7 +39,7 @@ import { Widget } from '@lumino/widgets';
 
 import React from 'react';
 
-import { CodeSnippetManager, ICodeSnippet } from './CodeSnippet';
+import { CodeSnippetManager, ICodeSnippet } from './CodeSnippetManager';
```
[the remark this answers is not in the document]
ajbozarth: I already fixed this, and whether a service file is ts or tsx would only depend on whether it has to leverage react in it. Some do, some don't.

PR #560, opened 2020-05-25
Add output location to pipeline submitted dialog

lresende:
Enhance pipeline submitted dialog with more information about where the experiment outputs are located

image

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -22,7 +22,7 @@ import autopep8
 from datetime import datetime
 
 from elyra.metadata import MetadataManager
-from elyra.pipeline import PipelineProcessor
+from elyra.pipeline import PipelineProcessor, PipelineProcessorResponse
 from elyra.util.archive import create_temp_archive
 from elyra.util.cos import CosClient
 from kubernetes.client.models import V1EnvVar
@@ -44,6 +44,8 @@ class KfpPipelineProcessor(PipelineProcessor):
 
         runtime_configuration = self._get_runtime_configuration(pipeline.runtime_config)
         api_endpoint = runtime_configuration.metadata['api_endpoint']
+        cos_endpoint = runtime_configuration.metadata['cos_endpoint']
+        cos_bucket = runtime_configuration.metadata['cos_bucket']
 
         with tempfile.TemporaryDirectory() as temp_dir:
             pipeline_path = os.path.join(temp_dir, f'{pipeline_name}.tar.gz')
@@ -76,7 +78,12 @@ class KfpPipelineProcessor(PipelineProcessor):
                                       pipeline_id=kfp_pipeline.id)
 
             self.log.info("Starting Kubeflow Pipeline Run...")
-            return "{}/#/runs/details/{}".format(api_endpoint, run.id)
+
+            return PipelineProcessorResponse(
+                run_url="{}/#/runs/details/{}".format(api_endpoint, run.id),
+                object_storage_url="{}".format(cos_endpoint),
+                object_storage_path="/{}/{}".format(cos_bucket, pipeline_name),
+            )
 
         return None
 
--- a/elyra/scheduler/handler.py
+++ b/elyra/scheduler/handler.py
@@ -37,17 +37,10 @@ class SchedulerHandler(HttpErrorMixin, APIHandler):
 
         pipeline = PipelineParser.parse(pipeline_definition)
 
-        run_url = PipelineProcessorManager.process(pipeline)
-        json_msg = json.dumps({"status": "ok",
-                               "message": "Pipeline successfully submitted",
-                               "url": run_url})
+        response = PipelineProcessorManager.process(pipeline)
+        json_msg = json.dumps(response.to_json())
 
         self.set_status(200)
-        self.write(json_msg)
+        self.set_header("Content-Type", 'application/json')
+        self.finish(json_msg)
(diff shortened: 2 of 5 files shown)


elyra/pipeline/processor_kfp.py
```diff
@@ -76,7 +78,12 @@ def process(self, pipeline):
                                       pipeline_id=kfp_pipeline.id)
 
             self.log.info("Starting Kubeflow Pipeline Run...")
-            return "{}/#/runs/details/{}".format(api_endpoint, run.id)
+
+            return PipelineProcessorResponse(
+                experiment_url="{}/#/runs/details/{}".format(api_endpoint, run.id),
```
kevin-bates, line 83: I'm not sure what _handle_ would be used to perform restarts or retry failed runs - guessing `pipeline_name` for example - but we should include that in the response since that's what would be used in a subsequent request (PUT /api/scheduler/{pipeline_name}) for retrying or continuing runs.

I'm not suggesting this must be displayed in the response dialog, but it needs to be available for future calls.

elyra/scheduler/handler.py
```diff
@@ -42,10 +42,13 @@ def post(self, *args, **kwargs):
...
         else:
-            run_url = PipelineProcessorManager.process(pipeline)
+            response = PipelineProcessorManager.process(pipeline)
             json_msg = json.dumps({"status": "ok",
                                    "message": "Pipeline successfully submitted",
-                                   "url": run_url})
+                                   "experiment-url": response.experiment_url,
+                                   "object-storage-url": response.object_storage_url,
+                                   "object-storage-path": response.object_storage_path
+                                   })
```
kevin-bates, line 51: I feel like this response is adding too much about _presentation_.  Why is `status` and `message` part of the response payload at all?  Status should derive from the status code and message should derive from the endpoint.

This should also be described in the [swagger yaml](https://github.com/elyra-ai/elyra/blob/master/elyra/api/elyra.yaml) once settled.

packages/application/src/submission.tsx
```diff
@@ -176,10 +177,25 @@ export class SubmissionHandler {
...
+              Check the experiment results and outputs at{' '}
+              <a
+                href={data['object-storage-url']}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                object storage.
+              </a>
+              <br />
+              Located in the {data['object-storage-path']} working directory
```
kevin-bates, line 198: This statement feels like part of the previous sentence since its not a complete sentence.  Seems like it should be...

_Check the experiment results and outputs at {object-storage-url} object storage and located in the {object-storage-path} working directory._

or 

_Check the experiment results and outputs at {object-storage-url} object storage.
These files are located in the {object-storage-path} working directory._

Or does the 'Located in the ...' statement refer to a _different_ set of files entirely?

elyra/pipeline/processor_kfp.py
lresende: As we don't have the `experiment list` available on Elyra today, it would be a little hard. KFP is also not a good citizen in this area (at least in my opinion). Having said that, once the user is on the KFP UI, he can manage all that.

packages/application/src/submission.tsx
lresende: The new dialog looks like this now:

image

packages/application/src/submission.tsx
```diff
@@ -167,6 +167,7 @@ export class SubmissionHandler {
       JSON.stringify(pipeline),
       submissionType,
       (data: any) => {
+        console.log(data);
```
ajbozarth, line 170: you forgot to remove your log line

packages/application/src/submission.tsx
```diff
@@ -176,10 +177,25 @@ export class SubmissionHandler {
           dialogTitle = 'Job submission to ' + runtime_config + ' succeeded';
           dialogBody = (
             <p>
-              Check the status of your run at{' '}
-              <a href={data.url} target="_blank" rel="noopener noreferrer">
-                Run Details
+              Check the status of your experiment at{' '}
```
ajbozarth, line 180: experiment feel like the wrong word here, why was `run` bad? Wouldn't submission be a better replacement since we use that term elsewhere in our messaging whereas we use experiment nowhere else.

packages/application/src/submission.tsx
```diff
@@ -176,10 +177,25 @@ export class SubmissionHandler {
...
+              <br />
+              The experiment results and outputs are in{' '}
+              <a
+                href={data['object-storage-url']}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                object storage
+              </a>{' '}
+              <br /> and located in the {data['object-storage-path']} working
```
ajbozarth, line 197: I feel like the English here is a bit off:
`The experiment results and outputs are in object storage and located in the /cos/dir working directory.`

I think grammatically this would be better (again note I'm not a fan of the work `experiment` in the dialog):
`The results and outputs are located in the /cos/dir directory in object storage.`

elyra/scheduler/handler.py
ajbozarth: In this case I agree with @kevin-bates , when the return json has expected return values only those values should be in the response.

packages/application/src/submission.tsx
lresende: image

lresende:
Latest dialog version

image

PR #561, opened 2020-05-26
Refactor Pipeline REST API endpoint

lresende:
Move pipeline submission endpoint from
api/scheduler to api/pipeline/schedule
 
           in the file; or

           it.


elyra/pipeline/handlers.py
```diff
@@ -51,8 +56,41 @@ def post(self, *args, **kwargs):
...
+        self.log.debug("Pipeline SchedulerHandler now executing post request")
+
+        pipeline_definition = self.get_json_body()
+
+        self.log.debug("JSON payload: %s", pipeline_definition)
+
+        pipeline = PipelineParser.parse(pipeline_definition)
+
+        if pipeline.export:
+            PipelineProcessorManager.export(pipeline)
```
kevin-bates, line 82: We should throw an exception here.  The client is calling the wrong endpoint.

The same goes for `PipelineExportHandler`.  It should confirm that `pipeline.export` is True.

elyra/pipeline/handlers.py
```diff
@@ -51,8 +56,41 @@ def post(self, *args, **kwargs):
...
+            json_msg = json.dumps({"status": "ok",
+                                   "message": "Pipeline successfully exported"})
         else:
-            return ','.join(pipeline_array)
+            response = PipelineProcessorManager.process(pipeline)
+            json_msg = json.dumps({"status": "ok",
+                                   "message": "Pipeline successfully submitted",
```
kevin-bates, line 88: Per this review comment https://github.com/elyra-ai/elyra/pull/560#discussion_r429986432 I don't think we should include status and dialog text in the response.  Let the UI deal with that (and it doesn't look like the UI even checks `status` or `message` from the response anyway).

elyra/pipeline/handlers.py
```diff
@@ -51,8 +56,41 @@ def post(self, *args, **kwargs):
...
+                                   "experiment-url": response.experiment_url,
+                                   "object-storage-url": response.object_storage_url,
+                                   "object-storage-path": response.object_storage_path
+                                   })
+
+        self.set_status(200)
```
kevin-bates, line 94: We should add a try/except block similar to the [`metadata handlers`](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/handlers.py#L57-L64) so checked exceptions don't all get mapped to `500` codes.

These will also be reflected in the swagger doc.

elyra/pipeline/handlers.py
```diff
@@ -51,8 +56,41 @@ def post(self, *args, **kwargs):
...
+            json_msg = json.dumps({"status": "ok",
+                                   "message": "Pipeline successfully exported"})
         else:
-            return ','.join(pipeline_array)
+            response = PipelineProcessorManager.process(pipeline)
+            json_msg = json.dumps({"status": "ok",
+                                   "message": "Pipeline successfully submitted",
+                                   "experiment-url": response.experiment_url,
+                                   "object-storage-url": response.object_storage_url,
+                                   "object-storage-path": response.object_storage_path
```
kevin-bates, line 91: I recommend adding a `to_json()` method on `PipelineProcessorResponse`, the change this to:

```python
json_msg = json.dumps(response.to_json())
```

elyra/pipeline/handlers.py
ajbozarth: echoing my response on another comment about this, iirc this is an API "good practice" and should be set whether or not our front end uses it.

elyra/pipeline/handlers.py
ajbozarth: +1

elyra/pipeline/handlers.py
lresende: that has been removed from here with `rename variables and clean up` pr

Issue #409, opened 2020-04-06
Add code mirror on code snippet display


karlaspuldaro:
Ideally, we want code snippets to display syntax highlighting as below:
![image](https://user-images.githubusercontent.com/25207344/78600940-e20d5580-7821-11ea-8f3d-be566eae0bcb.png)

vabarbosa:
Resolved by #589

PR #589, opened 2020-05-29
Create metadata editor

### LINKED ISSUE #556: Add ability to create code-snippets from the UI
Now that metadata instance can be managed via the REST API, we should minimally add the ability for users to create, update, and delete code-snippets from the web frontend.

marthacryan:
Opens a dialog that allows user to edit the name / description /
language / code of a code snippet. 

Still need to implement/fix:
- [x] deleting code snippets 
- [ ] Known bug that when editing code in the text area of the dialog, the dialog submits when the return key is pressed (might try to add code mirror and see if that helps)

Fixes #556 

Screenshots to come after implementing delete. 

 
           in the file; or

           it.


packages/ui-components/src/icons.tsx
```diff
@@ -16,6 +16,7 @@
 
 import { LabIcon } from '@jupyterlab/ui-components';
 
+import addSnippetSvg from '../style/icons/addSnippet.svg';
```
ajbozarth, line 19: since this is a pretty generic icon that could be used elsewhere I would suggest not using "snippet" in the name

ajbozarth: that should have applied to our pipeline icons as well

marthacryan: Do you think that [this line](https://github.com/elyra-ai/elyra/pull/589/files/ed11b87edce40a9277c9bc2d612a085c6192a56d#diff-075ae968e6c9905859aaa6eb9c77d6f4R34) makes it unique though? The variable name is just for retrieving it from the `style` directory.

ajbozarth: I meant all of it, the filename, vars names, style names. It's just an add icon, so it's name shouldn't include snippet. (Whereas the pipeline icons are a custom color for it's UI specifically)

marthacryan:
Screenshots of the current state of the UI:
image
Dialog when "New snippet" button is pressed:
image
Dialog when delete button is pressed:
image

Still need to fix the bug for pressing "enter" and also want to talk about the UI a bit more.

lresende:
Does this bring a new dialog for creating the code-snippet? How hard would it be to make the dialog to take control of the code-snippet panel and let the user create a new thing from there, this way he can easily navigate and copy+paste code from a notebook or python file?

vabarbosa:
@marthacryan nice progress on this so far. i would tend to agree with @lresende that have the add/edit form in the sidebar would be more helpful than in a new dialog.

i understand the PR is a WIP (and you are probably aware of these) but wanted to mention a couple things i noticed at first pass-through:

1. the Add code snippet icon does not give any visual indicator that it is actionable when you hover over it. maybe the cursor can be changed and/or highlight it's background
1. when successfully deleting a code snippet i had to refresh to see it was removed from sidebar
1. the multiple **+** icons can be confusing. i associate the  **+** with adding something which makes sense in the context of adding a new code snippet. however, it does not when you are inserting a code snippet into a notebook or file. since you are working in this area, i would vote you replace the insert icon with something more like the **Paste** or **Import** icons ([from Carbon Design System](https://www.carbondesignsystem.com/guidelines/icons/library)):
image
image

kevin-bates:
+1 to Luciano's idea, although I think we should still have a dialog option as well.

I guess I'd like a wider dialog.  Perhaps a layout that resembles something akin to this...
```
Name:   ----------------------------------------      Language:  ------------------------- V 
Description (optional):
------------------------------------------------------------------------------------------
Code: 
------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------
```
I know that's poorly conveyed, but you get the idea:
- Name and Language can share  a line,
- Language gets populated with a set of languages from existing kernelspecs (and code-snippets?),
- 'python' is marked as the default language, if it exists in kernelspecs or when there are zero languages in kernelspecs (rare).  If there are kernelspecs languages and none are python, I suppose we just choose the first in the list?
- The code portion is visibly wider and taller giving the impression that it can be pasted into.

Should we let users bring their own language?  Seems like we should. In that case, the language seems like it would need to be a combination edit/dropdown thing if that's possible.

marthacryan:
@lresende @kevin-bates I could definitely make it available in the code snippets panel, although I think we'd need to choose between opening the dialog and making it available in the code snippets panel, right? We could talk about this at the scrum if you think we should have a conversation about it.

marthacryan:
@vabarbosa Thoughts on this one for inserting code? 
image

vabarbosa:
> @vabarbosa Thoughts on this one for inserting code?
>

@marthacryan that one could also work

marthacryan:
Still need to clean up code and fix merge conflicts. Screenshots to come after cleanup.

marthacryan:
Here's what the most current UI looks like: 
For editing snippets:
image
For adding snippets:
image
Changes should automatically update in the left side pane when the "Save Snippet" button is pressed. 
I also added code mirror to the left side which fixes #409. 
Close-up of the snippet preview in the left side panel (might need to change color of the trash icon): 
image
On clicking the language dropdown:
image
If a language isn't present in the dropdown:
image
Reopens tabs on reloading the page (has a couple of issues with this: doesn't reopen unsaved code snippet editing tabs, also reverts the code snippet back to original fields even though the metadata has been updated)

marthacryan:
I want to make the form less hard-coded to the fields of code snippets, not sure if that needs to be in this PR?

kevin-bates:
Whoa - those screenshots look really cool!  

My only suggestions would be:
- Drop the large header in the code-snippet window.  I don't think they are necessary.  If anything, perhaps a reminder its a code-snippet like: "New Code-Snippet" and "Edit Code-Snippet".
- Make the 'Name' edit control about half its size so 'Description' moves over - or align them vertically.

Is that language dropdown list some newfangled, button-based dropdown?  I've never seen that format.

packages/ui-components/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,197 @@
...
+            this.metadata.language != ''
+              ? this.metadata.language
+              : '(No selection)'
+          }
+        />
+      </Select>
+    );
+    return (
+      <div className={ELYRA_METADATA_EDITOR_CLASS}>
+        <h1> Edit {this.metadata.name} Metadata </h1>
```
karlaspuldaro, line 164: Not sure about this header content, maybe 'New code snippet' if we are creating a new one.
The text size is a bit too large as well.

packages/code-snippet/src/CodeSnippetDisplay.tsx
```diff
@@ -0,0 +1,263 @@
...
+  };
+
+  // Display warning dialog when inserting a code snippet incompatible with editor's language
+  private showWarnDialog = async (
+    editorLanguage: string,
+    snippetName: string
+  ): Promise<Dialog.IResult<string>> => {
+    return showDialog({
+      title: 'Warning',
+      body:
```
vabarbosa, line 140: this may be easier to read using template literals, i.e.,
```
body: `Code snippet "${snippetName}" is incompatible with ${editorLanguage}. Continue?`
```

packages/code-snippet/src/CodeSnippetDisplay.tsx
```diff
@@ -0,0 +1,263 @@
...
+      ).then((response: any): void => {
+        this.props.updateSnippets();
+      });
+    });
+  };
+
+  // Render display of code snippet list
+  private renderCodeSnippet = (codeSnippet: ICodeSnippet): JSX.Element => {
+    const displayName =
+      '[' + codeSnippet.language + '] ' + codeSnippet.displayName;
```
vabarbosa, line 182: i would go with template literal here as well:
```javascript
const displayName = `[${codeSnippet.language}] ${codeSnippet.displayName}`;
```

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -214,37 +52,164 @@ export class CodeSnippetWidget extends ReactWidget {
...
+        snippetEditor.endpoint + '/' + newSnippet.name,
+        { method: 'PUT', body: newSnippetString },
+        false
+      ).then((response: any): void => {
+        this.updateSnippets();
+      });
+    }
+  }
+
+  async openCodeSnippetEditor(
```
vabarbosa, line 162: just something to think about...
rather than have callers pass in each property individually could this accept an object containing appropriate fields that you would deconstruct:
```javascript
  async openCodeSnippetEditor({name, description, language, code, newFile=true})
```
or if empty string is preferred default value instead of `undefined`:
```javascript
  async openCodeSnippetEditor({name='', description='', language='', code='', newFile=true})
```

that way each time `openCodeSnippetEditor` is called the caller does not have to worry about provided every single field individually and in the correct order.
for example, in line 79 you could just do something like:
```javascript
this.openCodeSnippetEditor(args)
```

and in line 126 you could do this:
```javascript
this.openCodeSnippetEditor({});
```

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -214,37 +52,164 @@ export class CodeSnippetWidget extends ReactWidget {
...
+      newFile,
+      this.saveSnippet,
+      this.editorFactory,
+      CODE_SNIPPET_ENDPOINT
+    );
+    metadataEditorWidget.id = METADATA_EDITOR_ID;
+    if (newFile) {
+      metadataEditorWidget.title.label = 'New Snippet';
+    } else {
+      metadataEditorWidget.title.label = '[' + language + '] ' + displayName;
```
vabarbosa, line 185: this is a good candidate for template literals and avoiding having to have all the `+`:

```javascript
metadataEditorWidget.title.label = `$[{language}] ${displayName}`;
```

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -214,37 +52,164 @@ export class CodeSnippetWidget extends ReactWidget {
...
+  }
+
+  addCodeSnippet(): void {
+    this.openCodeSnippetEditor('', '', '', '', true);
+  }
+
+  saveSnippet(snippetEditor: MetadataEditor): void {
+    const newSnippet = {
+      schema_name: 'code-snippet',
+      name: snippetEditor.metadata.name,
```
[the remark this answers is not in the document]
vabarbosa: also, the `name` field is more restrictive than `display_name`. for example, you cannot have a space in the `name` field but it is acceptable in `display_name`. may want to think about normalizing the the `name` field (i.e., replace spaces with underscoring or removing unacceptable characters). the user does not necessarily see the `name` but they do see the `display_name` so keeping the `display_name` as they entered but modifying the `name` should suffice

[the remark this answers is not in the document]
kevin-bates: I agree that this makes sense to check on the client.  FYI - The server will apply the schema file against the sent payload to validate all fields (including required) anyway. 

For now, it's probably fine to use "hard-coded" required values and forgo application of the schema.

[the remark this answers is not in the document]
vabarbosa: @kevin-bates the more i think about it,  am i correct in my interpretation of the `name` field as sort of a unique ID field for the metadata?

if so, given that assumption, would it be worth allowing submission to the server without a `name` field but include `display_name` (i.e., you need a `name` and/or a `display_name`)? in cases where only the `display_name` is provided the server would generate a `name` (e.g., normalizing the `display_name` or generating a unique id).  

and this would be only applicable for creating not editing. since you shouldn't be allowed to edit the ID/name but you still able to edit the `display_name`.

that way
- the client doesn't have to be cognizant of underlining ID/name mechanism
- the end user just sees/creates the `display_name` without having to think about or make sure they are abiding by the underlining ID/name requirements
- if  ID/name validation requirements ever changed it can all be handled in server rather than having to make updates in server and all clients calling the POST API.

your thoughts? maybe a discussion for a later time...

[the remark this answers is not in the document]
kevin-bates: Hi @vabarbosa - this all makes sense.  I agree that the name displayed here should only be the display name.

Yes, the `name` is actually the filename of the persisted metadata - so in that sense - it is unique.  Keep in mind though, that _this_ uniqueness spans the hierarchy and is essentially how we allow factory data to be "updated" (although that update equates to a file's creation in another directory).

Looking at how we [currently deal with `name`](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/metadata.py#L266-L272) we also support hyphens and already enforce lowercase, so preserving hyphens, replacing spaces with underscores and lowercasing characters will be compatible.  We'd then just need to figure out what to do with 8-bit+ characters (and I'm guessing we'd just strip them).  And of course, we'd need to tolerate it being `None`.

packages/application/src/services.tsx
```diff
@@ -39,4 +39,33 @@ export class FrontendServices {
 
     return metadataResponse[namespace];
   }
+
+  static async postMetadata(namespace: string, requestBody: any): Promise<any> {
+    const metadataResponse: any = await RequestHandler.makePostRequest(
+      'elyra/metadata/' + namespace,
```
vabarbosa, line 45: maybe make `elyra/metadata` a variable in the class and just reference the variable in each of the requests

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  onCloseRequest(msg: Message) {
+    this.dispose();
+    super.onCloseRequest(msg);
+  }
+
+  saveMetadata(): void {
+    if (this.newFile) {
+      this.metadata.name = this.metadata.displayName.split(' ').join('_');
```
vabarbosa, line 68: fyi, PR #629 is in review and once merged you shouldn't need this.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,248 @@
...
+    const newSnippetString = JSON.stringify(newSnippet);
+
+    if (this.newFile) {
+      FrontendServices.postMetadata(this.endpoint, newSnippetString).then(
+        (response: any): void => {
+          this.updateSignal();
+        }
+      );
+      this.newFile = false;
+      this.title.label = `[${this.metadata.language}] ${this.metadata.displayName}`;
```
vabarbosa, line 89: should these changes be made after the new code snippet has been successfully created? e.g., `newFile` should change to `false` only when the POST call returns a success

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,248 @@
...
+    this.editorFactory = editorFactory;
+    this.endpoint = endpoint;
+    this.newFile = newFile;
+    this.updateSignal = updateSignal;
+    this.handleNameChange = this.handleNameChange.bind(this);
+    this.handleDescriptionChange = this.handleDescriptionChange.bind(this);
+    this.handleLanguageChange = this.handleLanguageChange.bind(this);
+  }
+
+  onCloseRequest(msg: Message) {
```
karlaspuldaro, line 61:
```suggestion
  onCloseRequest(msg: Message): void {
```
to fix the check warnings

ajbozarth:
@marthacryan I updated the branch with latest master

kevin-bates:
Taking the latest branch, I see the following when going to create a code-snippet. The Code box appears to be two controls.  However, once resized, they become one - so its some sort of initial display bug:
image

kevin-bates:
We should confirm the code-snippet name in this dialog...
image

kevin-bates:
Hmm - deletion worked...
```
[D 07:42:52.660 LabApp] MetadataHandler: Deleting metadata instance 'power_of_y' in namespace 'code-snippets'...
[D 07:42:52.661 LabApp] Namespace 'code-snippets' is using metadata directory: /Users/kbates/Library/Jupyter/metadata/code-snippets from list: ['/Users/kbates/Library/Jupyter/metadata/code-snippets', '/usr/local/share/jupyter/metadata/code-snippets', '/usr/share/jupyter/metadata/code-snippets', '/opt/anaconda3/envs/elyra-dev/share/jupyter/metadata/code-snippets']
[I 07:42:52.661 LabApp] Removing metadata resource 'power_of_y' from namespace 'code-snippets'.
[D 07:42:52.662 LabApp] Loading metadata resource from: '/Users/kbates/Library/Jupyter/metadata/code-snippets/power_of_y.json'
[D 07:42:52.669 LabApp] 204 DELETE /elyra/metadata/code-snippets/power_of_y?1591972968877 (::1) 3788.95ms
```
But I got the dialog indicating that the Elyra endpoint was not found.  Not sure if it was relative to the DELETE request or another request, but the logs don't show any kind of error.

Upon return, the set of code-snippets was not refreshed, so the delete snippet remained until I refreshed the browser.

kevin-bates:
When sending metadata to the server, the fields MUST adhere to the schema for that metadata.  Issue #637 is occurring because the fields in the payload have their first characters upper-cased so `language` is sent via `Language` - which results in a schema violation because `language` is a required field and appears to be missing.

kevin-bates:
Regarding issues with POST, I've been running lab in a debugger with breakpoints in all the metadata handlers and metadata POST requests are NOT getting to the elyra server.  I have no idea where there are going.

Makefile
```diff
@@ -53,6 +53,7 @@ uninstall:
 	$(call UNLINK_LAB_EXTENSION,@elyra/ui-components)
 	$(call UNINSTALL_LAB_EXTENSION,@elyra/theme-extension)
 	$(call UNINSTALL_LAB_EXTENSION,@elyra/code-snippet-extension-experimental)
```
ajbozarth, line 55: @lresende should we remove experimental from code-snippet-extension in this PR?

packages/code-snippet/src/CodeSnippetService.ts
```diff
@@ -59,4 +63,26 @@ export class CodeSnippetService {
 
     return codeSnippetsByLanguage;
   }
+
+  deleteCodeSnippet(
+    codeSnippet: ICodeSnippet,
+    updateSnippets: () => void
+  ): void {
+    showDialog({
+      title: `Delete "${codeSnippet.displayName}" snippet?`,
```
kevin-bates, line 72: I think placing the snippet name on the end might read better since the display name could read like a sentence in some cases.  Not sure if we can remove the double-quotes (as suggested below) or not.  That's more aesthetics so we'll just need to feel that out.
```suggestion
      title: `Delete snippet: ${codeSnippet.displayName}?`,
```

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -32,36 +38,60 @@ import { DocumentWidget } from '@jupyterlab/docregistry';
...
 /**
  * The CSS class added to code snippet widget.
  */
 const CODE_SNIPPETS_CLASS = 'elyra-CodeSnippets';
 const CODE_SNIPPETS_HEADER_CLASS = 'elyra-codeSnippetsHeader';
+const CODE_SNIPPETS_HEADER_BUTTON_CLASS = 'elyra-codeSnippetHeader-button';
 const CODE_SNIPPET_ITEM = 'elyra-codeSnippet-item';
 
+export const defaultLanguageChoices = [
+  'Python',
```
kevin-bates, line 64: Should `Scala` be represented here?

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -176,10 +200,61 @@ class CodeSnippetDisplay extends React.Component<ICodeSnippetDisplayProps> {
...
+                label: 'Description',
+                value: codeSnippet.description,
+                type: 'TextInput',
+                schemaField: 'description'
+              },
+              {
+                label: 'Language',
+                value: {
+                  choice: codeSnippet.language,
+                  defaultChoices: defaultLanguageChoices
```
kevin-bates, line 230: At first, I was thinking we'd only use languages that are present in the currently configured kernels, but we should allow code snippets in any language since they won't necessarily be tied to a kernel.

Can a user bring their own language and, if so, should the `defaultLanguageChoices` be dynamic to include all languages from _existing_ code snippets that might not be in the hard-coded list?
Since we don't validate the language (against known languages), the server doesn't care what this value is so long as it meets the general criteria of the schema.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,253 @@
...
+    this.metadata = metadata;
+    this.editorFactory = editorFactory;
+    this.namespace = namespace;
+    this.newFile = newFile;
+    this.updateSignal = updateSignal;
+    this.handleTextInputChange = this.handleTextInputChange.bind(this);
+    this.handleDropdownChange = this.handleDropdownChange.bind(this);
+    this.tracker = tracker;
+    if (fileName) {
+      this.fileName = fileName;
```
kevin-bates, line 82: What is fileName?  Should this be called `name`?  Even though the server equates `name` to a file, we don't ever send the filename over the REST API, so I'm not sure what `fileName` is referring to.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,253 @@
...
+      });
+    } else {
+      this.dispose();
+      super.onCloseRequest(msg);
+    }
+  }
+
+  saveMetadata(): void {
+    const newMetadata: any = {
+      schema_name: 'code-snippet',
```
kevin-bates, line 111: This file sounds generic and I wonder if the code-snippet editor should subclass `MetadataEditor` and let its `saveMetadata()` method be the thing that sets the schema-name.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,253 @@
...
+      metadata: {}
+    };
+
+    for (const field of this.metadata) {
+      if (field.type == 'TextInput') {
+        newMetadata.metadata[field.schemaField] = field.value;
+      } else if (field.type == 'DropDown') {
+        newMetadata.metadata[field.schemaField] = field.value.choice;
+      } else if (field.type == 'Code') {
+        newMetadata.metadata[field.schemaField] = field.value.split('\n');
```
kevin-bates, line 123: Same comment here.  If this is a generic base class, let the specific subclass deal with class-specific behaviors like 'Code'.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,253 @@
...
+      } else if (field.type == 'DropDown') {
+        inputElements.push(
+          <DropDown
+            field={field}
+            handleDropdownChange={this.handleDropdownChange}
+          ></DropDown>
+        );
+      }
+    }
+    let headerText = `Edit "${this.getFormItem('Name').value}" Metadata`;
```
kevin-bates, line 224: Would it look better to drop `Metadata` from the header here?

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  onCloseRequest(msg: Message): void {
+    if (this.dirty) {
+      showDialog({
+        title: 'Close without saving?',
+        body: (
+          <p>
+            {' '}
+            {`Metadata "${this.fileName}" has unsaved changes, close without saving?`}{' '}
```
kevin-bates, line 93: Same here, I think we should drop `Metadata` here.

packages/code-snippet/src/CodeSnippetWidget.tsx
marthacryan: I'm changing it to check for other languages in existing snippets, but use python, R, java, scala, and markdown as the default language options.

packages/metadata-editor/src/MetadataEditor.tsx
marthacryan: Yeah, I'm just gonna remove this field, you're right.

packages/application/src/requests.tsx
```diff
@@ -197,7 +257,13 @@ export class RequestHandler {
...
             // handle 404 if the server is not found
             (reason: any) => {
-              return this.server404(requestPath);
+              if (response.status >= 200 && response.status < 300) {
+                resolve();
+              } else if (response.status == 404) {
+                return this.server404(requestPath);
+              } else {
+                return this.serverError(reason);
+              }
```
kevin-bates, line 266: I'm inclined to **only** handle success status `204` in this block since any other successes are bugs because it means they returned content.
```suggestion
            
             if (response.status == 404) {
                return this.server404(requestPath);
             } else if (response.status == 204) {
                resolve(); 
             } else {
                return this.serverError(reason);
             }
```

ajbozarth: is 204 the only success case with no content, or the only one we support currently? If it's the former then I agree with this, but if it's the later then I would argue that returning nothing and letting the services that call this function handle that "unexpected empty" error case is better.

kevin-bates: Yes - its the only success status _we_ return w/o content and I believe that's the case in general: https://restfulapi.net/http-status-codes/.   It looks like 304 also requires the response to be empty, but I think we could cross that bridge as necessary (I doubt we'll need to do ever deal with 304 status codes TBH).

Technically speaking, I think we should cap our "success range" to < 300 since 3xx codes are "redirection" and not "success" codes anyway.

ajbozarth: In that case I'm ok with this change. @marthacryan if you want to merge this when you get started in the morning.

marthacryan:
image
image
Two options for the "Insert Code" icons - any thoughts?

elyra/metadata/schemas/code-snippet.json
```diff
@@ -23,11 +23,18 @@
         },
         "language": {
           "description": "Code snippet implementation language",
-          "type": "string"
+          "type": "string",
+          "ui_hints": {
+            "field_type": "dropdown",
```
kevin-bates, line 28: I thought these were going to have a `label` field as well?

elyra/metadata/schemas/code-snippet.json
```diff
@@ -23,11 +23,18 @@
...
+            "defaultChoices": ["Python", "Java", "R", "Scala", "Markdown"]
```
kevin-bates, line 29: I'd prefer we use consistent formats.  Currently, all schema fields are lower-case with underscores.
```suggestion
            "default_choices": ["Python", "Java", "R", "Scala", "Markdown"]
```

elyra/metadata/schemas/code-snippet.json
ajbozarth: ah yes, I forgot that

elyra/metadata/schemas/code-snippet.json
ajbozarth: fixed

elyra/metadata/schemas/code-snippet.json
ajbozarth: fixed

marthacryan:
Here are screenshots in context and with the right fill color for those two options:
image
image

elyra/metadata/schemas/code-snippet.json
```diff
@@ -19,20 +19,25 @@
       "properties": {
         "description": {
           "description": "Code snippet description",
-          "type": "string"
+          "type": "string",
+          "ui_hints": {
```
kevin-bates, line 23: Sorry, but I just discovered the use of [`uihints` in properties.json](https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/properties.json#L31) and I think we should use the same format here as well (although I think `ui_hints` is better, possession is nine-tenths of the law :smile:).
```suggestion
          "uihints": {
```

ajbozarth: ill update it

elyra/metadata/schemas/metadata-test.json
```diff
@@ -62,14 +62,21 @@
         "enum_test": {
           "description": "Property used to test properties with enums",
           "type": "string",
-          "enum": ["elyra", "rocks"]
+          "enum": ["elyra", "rocks"],
+          "uihints": {
+            "field_type": "dropdown",
+            "defaultChoices": ["elyra"]
```
kevin-bates, line 68: let's apply the same convention here as well.
```suggestion
            "default_choices": ["elyra"]
```

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -214,37 +296,73 @@ export class CodeSnippetWidget extends ReactWidget {
...
+  openCodeSnippetEditor(args: any): void {
+    this.app.commands.execute(`${METADATA_EDITOR_ID}:open`, args);
+  }
+
   render(): React.ReactElement {
     return (
       <div className={CODE_SNIPPETS_CLASS}>
         <header className={CODE_SNIPPETS_HEADER_CLASS}>
-          {'</> Code Snippets'}
+          <p> {'</> Code Snippets'} </p>
```
karlaspuldaro, line 349: I know this has been like this for a while, maybe we should use code snippet icon instead of `</>`?

packages/metadata-editor/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@jupyterlab/fileeditor": "^2.0.2",
+    "@jupyterlab/launcher": "^2.0.2",
+    "@jupyterlab/logconsole": "^2.0.2",
+    "@jupyterlab/mainmenu": "^2.0.2",
+    "@jupyterlab/outputarea": "^2.0.2",
+    "@jupyterlab/ui-components": "^2.0.2",
+    "@jupyterlab/settingregistry": "^2.0.1",
+    "@jupyterlab/toc": "^4.0.0",
+    "@lumino/messaging": "^1.3.3",
+    "@lumino/widgets": "^1.11.1"
```
karlaspuldaro, line 49: do we need all these dependencies here only the ones imported in metadata-editor/src files?

packages/metadata-editor/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const openMetadataEditor = async (args: {
+      metadata: FormItem[];
+      newFile: boolean;
+      namespace: string;
+      updateSignal: any;
+      name: string;
+      editor: CodeEditor.IEditor;
+    }): Promise<void> => {
+      console.log('Elyra - metadata-editor extension is activated!');
```
karlaspuldaro, line 46: this is already displayed on activate (line 36)

packages/ui-components/src/DropDown.tsx
```diff
@@ -0,0 +1,103 @@
...
+      onClick={handleClick}
+      shouldDismissPopover={false}
+    />
+  );
+
+  filterDropdown: ItemPredicate<string> = (
+    query,
+    value,
+    _index,
+    exactMatch
```
karlaspuldaro, line 59: again, is there a way to use partial match instead?

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -16,51 +16,68 @@
...
  * The CSS class added to code snippet widget.
  */
 const CODE_SNIPPETS_CLASS = 'elyra-CodeSnippets';
 const CODE_SNIPPETS_HEADER_CLASS = 'elyra-codeSnippetsHeader';
+const CODE_SNIPPETS_HEADER_BUTTON_CLASS = 'elyra-codeSnippetHeader-button';
 const CODE_SNIPPET_ITEM = 'elyra-codeSnippet-item';
 
+const METADATA_EDITOR_ID = 'elyra-metadata-editor';
+
+const defaultLanguages = ['Python', 'Java', 'R', 'Scala', 'Markdown'];
```
kevin-bates, line 62: How does this differ from the [`default_choices`](https://github.com/elyra-ai/elyra/blob/377f42df88a8e52201fa014961ff0e82b0a079bd/elyra/metadata/schemas/code-snippet.json#L33) in the schema file?  Seems like the schema should be used or the hint removed from the schema.

packages/code-snippet/src/properties.json
```diff
@@ -0,0 +1,35 @@
...
+        "label": "Description",
+        "value": "",
+        "type": "TextInput",
+        "schemaField": "description"
+      },
+      {
+        "label": "Language",
+        "value": {
+          "choice": "",
+          "defaultChoices": ["Python", "Java", "R", "Scala", "Markdown"]
```
kevin-bates, line 20: Same comment as earlier.  There are three places where the set of default languages exist. We should have one.

I think to have a generic editor, the uihints must be baked into the schema file - as is the case in this PR as well.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,271 @@
...
+    this.handleDropdownChange = this.handleDropdownChange.bind(this);
+
+    this.initializeMetadata();
+  }
+
+  async initializeMetadata() {
+    const schemas = await FrontendServices.getSchema(this.namespace);
+    for (const schema of schemas) {
+      if (this.schemaName == schema.name) {
+        this.schema = schema;
```
kevin-bates, line 88: Could this use a `break;` statement?

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,271 @@
...
+      }
+    }
+
+    this.allMetadata = await FrontendServices.getMetadata(this.namespace);
+    if (this.name) {
+      for (const metadata of this.allMetadata) {
+        if (this.name == metadata.name) {
+          this.metadata = metadata;
+          this.title.label = this.metadata.display_name;
```
kevin-bates, line 97: Could this use a `break;` statement?

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,271 @@
...
+    this.update();
+  };
+
+  handleDirtyState(dirty: boolean): void {
+    this.dirty = dirty;
+    if (this.dirty && !this.title.className.includes(DIRTY_CLASS)) {
+      this.title.className += ` ${DIRTY_CLASS}`;
+    } else if (!this.dirty) {
+      this.title.className = this.title.className.replace(DIRTY_CLASS, '');
+    }
```
kevin-bates, line 190: I don't understand the utility for appending/masking `'jp-mod-dirty'` to/from the className of the title.  The className is never checked for this value that I can tell.

What do the back-ticks do on this statement?
```
this.title.className += ` ${DIRTY_CLASS}`;
```

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,271 @@
...
+          <DropDown
+            field={field}
+            handleDropdownChange={this.handleDropdownChange}
+          ></DropDown>
+        );
+      }
+    }
+    let headerText = `Edit "${this.getFormItem('Name').value}"`;
+    if (!this.name) {
+      headerText = 'Add new metadata';
```
kevin-bates, line 245: Do we have the schema corresponding to the metadata being edited at this point?  If so should we use the schema name instead of `metadata`?  E.g., `Add new code-snippet`

packages/metadata-editor/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    }): void => {
+      const metadataEditorWidget = new MetadataEditor({
+        ...args,
+        editorServices
+      });
+
+      if (args.name) {
+        metadataEditorWidget.title.label = args.name;
+      } else {
+        metadataEditorWidget.title.label = 'New Metadata';
```
kevin-bates, line 53: Similar to previous, can we use `schema` here?  `New ${schema}`.  (We'd probably want to upcase the first character in this case.)

packages/code-snippet/src/CodeSnippetWidget.tsx
ajbozarth: I haven't finished that yet, but yes that will be how this works

packages/metadata-editor/src/MetadataEditor.tsx
ajbozarth: good idea I'll add that

packages/metadata-editor/src/MetadataEditor.tsx
ajbozarth: I hadn't gotten to this code yet, but I can do this

packages/metadata-editor/src/index.ts
ajbozarth: Hmmm, that's a good idea, I would have to think a bit how best to "English-ize" the var name (it could also include dashes or underscores)

packages/metadata-editor/src/MetadataEditor.tsx
marthacryan: We're doing that because it adds an indicator over the "Close tab" button to show that there are unsaved changes to the metadata (this is implemented as part of the `DocumentWidget` code, but as long as it has that class name it can reuse that css for consistency). So before you make any changes to the metadata form, the tab would look like this:
image
Then after you make changes, it looks like this: 
image
And if you hover over that black circle it looks like this:
image
And if you press it, it opens a dialog to confirm you want to close without saving changes. 
You're right about the back ticks, I just removed them!

packages/ui-components/src/DropDown.tsx
marthacryan: I think right now the `exactMatch` variable isn't actually being used, I just needed to take that as an argument for this function. Partial matches work in the dropdown as far as I can tell:
image

packages/code-snippet/src/CodeSnippetWidget.tsx
marthacryan: Just added issue #683 to do this (but I don't want to hold up this PR by trying to figure that out)

packages/metadata-editor/src/MetadataEditor.tsx
kevin-bates: Thank you for the great explanation and I am ever more grateful that others deal with this level of detail than myself! :smile:

vabarbosa:
> image
> image
> Two options for the "Insert Code" icons - any thoughts?
   
   
first one makes me think of paragraph/text indent, but the second one can work.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -192,74 +183,124 @@ export class MetadataEditor extends ReactWidget {
...
+  renderField(fieldName: string) {
+    const uihints = this.schema[fieldName].uihints;
+    if (uihints == undefined) {
+      return;
+    } else if (
+      uihints.field_type == 'textinput' ||
+      uihints.field_type == undefined
+    ) {
+      return this.renderTextInput(
+        uihints.label,
```
ajbozarth, line 254: what if label is undefined? Since it's used as the key in `this.renderTextInput`

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -192,74 +183,124 @@ export class MetadataEditor extends ReactWidget {
...
       this.editor = this.editorServices.factoryService.newInlineEditor({
-        host: document.getElementById('Code:' + this.name),
+        host: document.getElementById('code:' + this.id),
         model: new CodeEditor.Model({
-          value: this.getFormItem('Code').value,
+          value: this.metadata['code'].join('\n'),
           mimeType: getMimeTypeByLanguage({
-            name: this.getFormItem('Language').value.choice,
-            codemirror_mode: this.getFormItem('Language').value.choice
+            name: this.metadata['language'],
```
ajbozarth, line 196: This makes the assumption that `language` and `code` exist for all code types, should there be a check for these values somewhere?

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -192,74 +183,124 @@ export class MetadataEditor extends ReactWidget {
     }
   }
 
-  onAfterShow(): void {
-    if (!this.editor) {
+  onUpdateRequest(msg: Message): void {
+    super.onUpdateRequest(msg);
+    if (!this.editor && this.metadata['code']) {
```
ajbozarth, line 188: to confirm, this implementation still only supports a single code field per metadata for now. I'm ok with that for an initial commit to get this in as long as we create a follow up to address it later

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -192,74 +183,124 @@ export class MetadataEditor extends ReactWidget {
...
+          schemaField={fieldName}
+          description={uihints.description}
+          choice={this.metadata[fieldName]}
+          defaultChoices={this.getDefaultChoices(fieldName)}
+          handleDropdownChange={this.handleDropdownChange}
+        ></DropDown>
+      );
+    } else if (uihints.field_type == 'code') {
+      return (
+        <>
```
ajbozarth, line 272: is this valid dom? shouldn't this need to be a div or span or something?

marthacryan: It doesn't seem to cause any issues, but I'll add `div` because it seems more reliable

packages/metadata-editor/src/MetadataEditor.tsx
marthacryan: Yes. This should definitely be addressed in a future PR but since this is high priority and we only have code snippets so far I think it can wait.

marthacryan:
## Current state of metadata editor UI
When the "Edit" button is pressed in an existing code snippet, this page appears in the main area:
image

When the '+' button is pressed (to the right of the 'Code Snippets' header below), this page opens up:
image
image

Before changes are made to the metadata, this is what the tab looks like:
image
After changes are made to the metadata, this is what the tab looks like:
image
And if that black circle is pressed, this dialog appears:
image


## Notes / Todo
* When the page is reloaded but there are unsaved changes, there is no warning, and the edits are lost. This can be a follow-up PR. 
* This currently assumes that you'll need at most one code editor. While it isn't immediately necessary to be able to have multiple editors in the same form, it should be implemented in a follow-up PR.

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -0,0 +1,325 @@
...
+    label: string,
+    description: string,
+    fieldName: string,
+    defaultValue: string
+  ) {
+    return (
+      <FormGroup
+        key={label}
+        label={label}
+        labelInfo="(required)"
```
ajbozarth, line 242: we should dynamically set this for every field based on the schema. Currently it's listed for all text inputs and drop down but not code. This is inaccurate since code is required for snippets but description (a text field) is not.

packages/metadata-editor/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+      if (openWidget) {
+        console.log(openWidget);
+        app.shell.activateById(metadataEditorWidget.id);
+      } else {
+        app.shell.add(metadataEditorWidget, 'main');
+      }
+    };
+
+    app.commands.addCommand(`${METADATA_EDITOR_ID}:open`, {
```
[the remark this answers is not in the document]
marthacryan: The id is used in other places in the file though, so I kept its definition and added this line below it, thoughts? 
```
const METADATA_EDITOR_ID = 'elyra-metadata-editor';
const commands = {
  OPEN_METADATA_EDITOR: `${METADATA_EDITOR_ID}:open`
}
```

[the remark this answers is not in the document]
ajbozarth: I don't know, is there any examples for this situation in lab code? otherwise I'd say leaving as it was is slightly better than this since it's just one command usage

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -46,21 +53,31 @@ import { CodeSnippetService, ICodeSnippet } from './CodeSnippetService';
...
+const CODE_SNIPPET_SCHEMA = 'code-snippet';
+
 /**
  * CodeSnippetDisplay props.
  */
 interface ICodeSnippetDisplayProps {
   codeSnippets: ICodeSnippet[];
   getCurrentWidget: () => Widget;
+  editorServices: IEditorServices;
+  openCodeSnippetEditor: any;
```
ajbozarth, line 72: set a specific type for this one too

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -87,6 +88,7 @@ export class MetadataEditor extends ReactWidget {
         if (this.name == metadata.name) {
           this.metadata = metadata['metadata'];
           this.displayName = metadata['display_name'];
+          this.requiredFields = metadata['required'];
```
kevin-bates: There are two levels of 'required' lists, one at the top level - as you have here, and another within the metadata stanza itself - which is probably more what your intention is.  The top level required fields will typically **always** be 'schema_name', 'display_name' and 'metadata' (i.e., system-level properties).  The required fields in the metadata stanza will be specific to what that metadata object is about.
```suggestion
          this.requiredFields = metadata['metadata']['required'];
```
Note, since a metadata object may not require any fields, you'll probably need to check if the `required` key exists in `metadata['metadata']` before attempting to access it.

marthacryan: Good catch!

packages/metadata-editor/src/index.ts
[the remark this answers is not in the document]
marthacryan: I can't find any examples of just one command being used

packages/metadata-editor/src/MetadataEditor.tsx
marthacryan: It's no longer being used as a key in `renderTextInput`, so I think it's ok if this is just blank if there's no label defined?

ajbozarth: it's still being set as the key for the FormGroup, we just are using it to access the value anymore. I'm worried that FormGroup may fail to create/render if label doesn't exist (different than empty string)

packages/code-snippet/src/CodeSnippetService.ts
```diff
@@ -47,7 +51,7 @@ export class CodeSnippetService {
   }
 
   // TODO: Test this function
-  async findByLanguage(language: string): Promise<ICodeSnippet[]> {
+  static async findByLanguage(language: string): Promise<ICodeSnippet[]> {
```
karlaspuldaro, line 54: this is not being used anywhere

docs/source/getting_started/overview.md
```diff
@@ -78,9 +78,9 @@ them against local or cloud-based resources seamlessly.
 
 ![Enhanced Python Support](../images/python-runner.png)
 
-#### Reusable Code Snippets (Experimental)
+#### Reusable Code Snippets
```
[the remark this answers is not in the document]
marthacryan: Yes - a documentation follow-up is on the to-do list above

packages/code-snippet/src/CodeSnippetService.ts
marthacryan: Hm yeah I think that was already not being used - do you remember why that was added in the first place? Maybe for future searching capabilities in the code snippet display?

karlaspuldaro: Yeah I think that was the intent, we can keep it there then :)

vabarbosa:
@marthacryan great work and thanks for addressing the previous comments! i did stumble across an interesting issue while trying out the latest version of the code snippets. after you delete a code snippet and create a new one (with the same name) the code block is not displayed in the sidebar (when expanding the newly created code snippet). you can still copy/paste, insert and it works as expected. the sidebar just does not show the code. refreshing the browser resolves the issue and the code is again displayed:

code-snippet-recreate

marthacryan:
@vabarbosa Good catch! Just pushed a fix.

lresende:
Some issues I found:
- [ ] Deleting a code-snippet that is open in the editor should close the editor (which is a normal behavior in other extensions), otherwise save is causing issues.
- [ ] There is no form validation before submit

I also reached out to @ajbozarth regarding why the editor is its own extension, and the idea is that this is a shared component among multiple extensions that uses metadata, and being its own extension makes it easier to be shared by other consumers.

lresende:
@vabarbosa could you take this for a spin regarding the "dark mode support" now that your changes are in master. Thanks.

marthacryan:
@lresende Just tried it out and it definitely would need some work:
image

marthacryan:
This is what the editor looks like if you click "Save & Close" with an empty form:
image
Currently, if you fix one of the fields and press the "Save & Close" button, the warning will go away for that field. 
image

marthacryan:
image
Just pushed a change that makes the page legible in dark mode, but it still needs some work on the color of the dropdown / submit button (couldn't find a simple way of activating the dark mode of blueprint.js through CSS - apparently if you add the class `bp3-dark` it'd automatically set up the right css for dark mode, but I couldn't figure out how to do that without a bunch of extra code.

ajbozarth:
@vabarbosa on @marthacryan comment above, did you learn how to make css conditional to the current theme in your pr? (Ie add/remove `bp3-dark` if the theme changes)

vabarbosa:
@lresende @marthacryan @ajbozarth will take a look at dark mode this morning

vabarbosa:
> @vabarbosa on @marthacryan comment above, did you learn how to make css conditional to the current theme in your pr? (Ie add/remove `bp3-dark` if the theme changes)

the best way to do that would be to use JupyterLab's ThemeManager and listen to changes in theme and then programmatically add/remove the `bp3-dark`. i have added the code for that (https://github.com/elyra-ai/elyra/pull/589/commits/55642a6ab435999612348a870795bf427e7575db).

marthacryan:
@vabarbosa Thank you, it looks great! Just added a couple little changes as well (colors of icons / buttons added in this PR for the code snippets were a little off) and here's what it looks like in light / dark mode:
image
image
@lresende I think the PR is ready whenever you want to give it a last pass

marthacryan:
After last commit, header of code snippets looks like this: 
image

ajbozarth:
The vertical spacing looks a bit off in those screenshots, perhaps it has to do with the <p> around the text "Code Snippet"

ajbozarth:
or the icon's css may just need some tweaking to match the changes you made to the header in this pr

@vabarbosa insight?

marthacryan:
Here's a side-by-side of this PR (top) vs the current master (bottom) for reference:
image

image

vabarbosa:
> or the icon's css may just need some tweaking to match the changes you made to the header in this pr
> 
> @vabarbosa insight?

this is how the icons currently line up:

image


image

Issue #277, opened 2020-02-19
Enable searching for code-snippets


lresende:
Use JupyterLab search as an example

lresende:
Fixed by #985

PR #591, opened 2020-05-29
Add a UI for accessing runtimes history

### LINKED ISSUE #584: Add link to KubeFlow pipeline console
As a user, I want to have access to previous Kubeflow pipeline runs from within Elyra. Today, a link to the Kubeflow Pipelines is displayed in a dialog window when a pipeline is submitted. However, after the dialog with the link is closed the link is no longer available anywhere in the Elyra UI, forcing the user to manually enter the URL. Since multiple runtimes can be defined in an Elyra installation, there might be multiple kfp consoles that the user can choose from.

### LINKED ISSUE #734: Code snippet editor: heading should display a user friendly text
Current: `Add new <schema-name-string>` 
User friendly: `Add new code snippet`

image

### LINKED ISSUE #790: Password should be masked in the Metadata Editor
Currently we the Metadata Editor show files such as `cos_password` in clear text. We should add support for masking passwords. There are some UX and security questions around how to go about this, issues already discussed will be included in a comment below.

ajbozarth:
Currently a user is only given a link to a runtime instance when they make a submission, this precludes easy access to old submissions.

This PR adds a toolbar button that opens a sidebar widget with a list of runtime configs and lets the user open a link to a runtime or cos instance.

- [x] Create new metadata-common library package
- [x] Create basic MetadataWidget for viewing metadata in sidebar
- [x] Create RuntimesWidget that extends MetadataWidget
- [x] Add pipeline toolbar button to open RuntimesWidget
- [x] Update MetadataEditor to handle schemas without uihints
- [x] Update CodeSnippetWidget to extend MetadataWidget
- [x] Added support for masking password fields in the Metadata Editor
- [x] Add/update tests

Fixes #584
Fixes #734 
Fixes #790 
 
           in the file; or

           it.


ajbozarth:
Screenshots:
Note the new toolbar button
Screen Shot 2020-05-29 at 4 41 27 PM
closeup of button when hovered over
Screen Shot 2020-05-29 at 4 41 45 PM
dialog that's opened when the button is clicked
Screen Shot 2020-05-29 at 4 46 08 PM

ajbozarth:
Since canvas doesn't support drops downs in the toolbar I opted to just have a new button open a dialog. For the dialog I generally repurposed the dialog for running a pipeline. I'm not 100% confident on the language around this new UI and am open to changes. The icon is a slightly customized carbon icon (changed the color to match canvas)

@ptitzler what do you think of this implementation, given this was your request.

kevin-bates:
@ajbozarth - thank you for putting this together so quickly.  Yeah, I'm not sure about the wording, since what this is really providing is a way to access the dashboard associated with a given runtime configuration, but we can talk about that later.

What I don't see is the exposure of the `api_endpoint` property of the selected runtime.  Somewhere, we need to provide that value as a selectable link with some kind of text like "Runtime Dashboard" or something to that effect.  Is that still in the works?

ajbozarth:
@kevin-bates we definitely need to work on wording then cause that's all there already. You select the runtime config you want from the drop down then click the Open button and it opens the `api_endpoint` for that config in a new tab/window.

kevin-bates:
Regarding terminology, all the other buttons are related to "pipeline" so perhaps coming at things from that angle: "Pipelines History"?

Or should we surface this feature as more of a "dashboard" thing (although I really like the counterclockwise icon for history)?

vabarbosa:
hi, i just had a couple questions, which perhaps you can clarify for me. and my expectations would differ depending on the answers.

not sure if makes sense but can a pipeline be submitted to multiple/different runtimes or would you need to create a separate/duplicate pipeline files to submit to different runtimes? i am assuming after you submit and the job finishes, you can resubmit to different pipeline and in these cases, does the pipeline know (or keep track of) all runtimes it has been submitted to in the past?

if the answer is yes to these questions then i think this implementation is OK (with the @kevin-bates' suggestions) and i would include limiting the options in the dropdown to only the runtimes the pipeline has been submitted to.

if the answer to these is no, then i  would suggest the button immediately launches the current runtime dashboard associated with pipeline and the hover would just display the actual URL/endpoint to the dashboard.

also does canvas support adding "comment boxes" to the flow which are basically text nodes that are not part of the flow but allow you add text info/documentation into the flow diagram?

ajbozarth:
updated wording
Screen Shot 2020-06-01 at 12 45 59 PM
Screen Shot 2020-06-01 at 12 46 18 PM

vabarbosa:
not trying to complicate things, but after giving this more thought.... 
in lieu of having a full history sidebar widget, i think having an "**Info**" button which opens a modal showing basic runtime info (i.e. config, link, etc) may be an approach to consider.

something more along these lines (of course wording and styling would need to cleaned up a bit and decision made on what to include):

image

but the above provides these additional points:

1. the user can see actually see the links (in case they are interested in just copying the URL) and also quickly know the destination of the runtimes all without having launch the dashboard
1. the user can launch multiple dashboards by clicking each individual link and not have to re-open the dialog, select the dropdown, and launch the dashboard if they wanted to go to multiple dashboards
1. i tend to associate buttons with triggering an action (i.e., Run pipeline, Copy snippet, etc) but for navigating/opening another pages i would expect a link so the **Open button** in the current dialog would be misleading (you're being redirected to a new page) as to where an link/anchor tag would be more accurate in the context.

with that i said, the call is already being made to retrieve the runtimes metadata so all the data to put in an **Info** dialog for the user should already be available. furthermore, if it is decided to add or surface more info about the runtime, the content of the dialog can more easily be expanded to accommodate/include those changes whereas the current implementation does as easily allow for that.

kevin-bates:
I like this idea @vabarbosa!  Having the link(s) "in hand" is really nice.  It becomes a "dashboard launch area".

ajbozarth:
Based on discussion over a Webex call, I will be reimplementing this as a sidebar widget rather than a dialog. The sidebar widget will display the runtime config info similar to @vabarbosa suggestion above and how code snippets work. When that new PR is ready I will close this one.

ajbozarth:
I've moved this to draft and pushed my initial working draft of the runtimes ui. My intention is to create a generic metadata ui that is meant to be "extended" with specific requirements. As such I've based my implementation on code snippets with the expectation that to would eventually leverage the common metadata ui code chunks.

Note that this is still very WIP as I still need to clean things up, separate out the generic bits from the runtime bits, as well as actually customize the runtimes url links to be clickable rather than just displaying the json as is.

ajbozarth:
I've finally finished my work on updating this PR. I'll update the top description to match the new design and will be attaching new screen shots. I've removed the dialog and replaced it with a sidebar widget.

This PR now contains a new abstract `MetadataWidget` that can be extended to display metadata in a sidebar. To start I created two classes that extend the new abstract class:

- `RuntimesWidget` is added to the sidebar at launch as suggested by @lresende and can also be opened (brought to focus) via a toolbar button in the pipeline editor. It displays links for the runtime and cos endpoints and has the ability to add edit and delete runtimes
- `DefaultMetadataWidget` is a transient sidebar widget that can be created via a command for any given metadata and will display the metadata field of each metadata as a json dump. It also includes add edit and delete options. Currently this command is not invoked anywhere but instead exists as both a starting point and example of extending the abstract class.

TODO:
- `CodeSnippetWidget` should be updated to extend `MetadataWidget` as well.

ajbozarth:
Screen Shot 2020-07-17 at 5 39 22 PM
Screen Shot 2020-07-17 at 5 39 56 PM
Screen Shot 2020-07-17 at 5 40 13 PM

packages/metadata-editor/src/MetadataEditor.tsx
```diff
@@ -304,17 +304,17 @@ export class MetadataEditor extends ReactWidget {
   }
 
   renderField(fieldName: string): React.ReactElement {
-    const uihints = this.schema[fieldName].uihints;
+    let uihints = this.schema[fieldName].uihints;
```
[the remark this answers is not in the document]
ajbozarth: My changes in this file do exactly that. Currently the MetadataEditor does not populate fields without a `uihints` This updates it so if `uihints` isn't in the schema for the specific fields then it creates a default label using the filed name. This change is not needed for Runtimes since I add `uihints` to the kfp schema in this PR, but it does fix the missing functionality we meant to include previously

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -742,6 +751,12 @@ export class PipelineEditor extends React.Component<
     });
   }
 
+  handleOpenRuntimes(): void {
+    this.app.shell.activateById(
+      `elyra-metadata:${RUNTIMES_NAMESPACE}:${KFP_SCHEMA}`
```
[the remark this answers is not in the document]
ajbozarth: It is overly specific, I used this naming for Runtimes so it matched the generic naming created when using `DefaultMetadataWidget`, meaning if the `elyra-metadata:open` command was called with the `runtimes` namespace and `kfp` schema it would open the Runtimes UI rather than creating a new generic one that shows the json dump rather than the links.

I'm not particularly attached to this idea though, its mostly there as a (cleaned up) artifact of my dev process rather than an intentional design.

ajbozarth:
@kevin-bates I responded to your inline comments with some explanation. As for your review comments:

> Regarding the proposed "widget hierarchy", I like the idea in spirit, but I guess I'm not following what functionality having such a hierarchy provides. At first thought, it seems like the default widget should be the base class (non-abstract) that provides the basic "JSON dump" functionality. Then, subclasses could be implemented to enhance the default implementation. With the abstract class approach (are some methods also abstract, such that subclasses are required to implement them?) it seems like you're requiring subclasses to be implemented. I think it would be important to have some fallback implementation in the base class so things still work and I suspect that's the idea behind the default widget class, although I'm not sure it's necessary to be a separate and sibling class to other (custom) subclasses as opposed to providing their base implementation (by default).

This was what our extended scrum discussion on Friday was about. This implementation was my original plan, but it would require the base widget to have access to the service utils found in `application`, but making `ui-components` depend on `application` would create a circular dependency since those same service utils depend on `ui-components` for their error handling (`ExpandableErrorDialog`). This means that if we wanted to stick with that implementation we would either need to move `MeatadataWidget` into `application` (a place it doesn't really make sense to belong) or create a new package to hold it (way overkill). The solution we decided in that discussion was to instead use a abstract class in `ui-components` that would be extended and not have a default fallback. Due to already having created a transient default version and tied it to a command during my interactive dev process I decided to just leave `DefaultMetadataWidget` available even though we don't use it directly in our code currently.

> For example, it seems like the COS password should have some knowledge that its a password field and not display its clear text (or provide the eyeball thing to allow users to toggle visibility) and even allow for confirmation with a sibling edit control. However, if that is the idea - that custom classes have such knowledge, then it seems to eliminate the need for having uihints in the schema at all. (That said, I believe the schema will need additional meta-properties that imply behavior for things like password fields though because the server needs them as well - and those should not be uihints.)

I'm a bit confused on how this connects to your previous paragraph, but taking this as a critique of the MetadataEditor and not this PR, I agree that we need to create a follow up issue that can specify if a schema field is a password (or protected in some way). It sounds like the backend would also want to make some changes to how such fields are handled, and the front end could be updated to obscure such fields in some way (know to do so would be straight forward since the front end has access to the schema and metadata both in their entirety.

> How does the code know which subclass to use, is there some form of registration? (Sorry for my lack of typescript familiarity.)

Exactly, an implementation of `MetadataWidget` needs to be created and registered in an extension. In our case `RuntimesWidget` is initialized in the `pipeline-editor` index.ts, just like CodeSnippetWidget is in `code-snippet`. `DefaultMetadataWidget` on the other hand is created by a command call. The command is created in the index.ts of `metadata-editor`.

Hopefully my answers provide some clarity, iiuc assuming my explains make sense and based on your comments the following items should be discussed or followed up on:

- Create a follow up issue that handles obscuring password type fields (both front and backend)
- Should the naming of the `RuntimesWidget` be simplified and separated from the naming of `DefaultMetadataWidget` instances?
- Should we include `DefaultMetadataWidget` at all given it's not used.

kevin-bates:
Alex - I apologize for not being attentive enough to realize that the packaging discussion in our extended scrum dictated the need for an abstract base class.  What I still don't follow, and this is probably the same discussion, is why there can't still be a subclass of the abstract base class that provides all the basic (generic) functionality from which actual metadata classes derive - like Runtime and CodeSnippets. Ah, this is the key...
>  In our case RuntimesWidget is initialized in the pipeline-editor index.ts, just like CodeSnippetWidget is in code-snippet. 

 Ok, that's unfortunate, but I think I get it now. 

>>For example, it seems like the COS password should have some knowledge...

> I'm a bit confused on how this connects to your previous paragraph, but taking this as a critique of the MetadataEditor and not this PR

I had edited my response yesterday several times and I knew the second paragraph was not "connecting" very well - sorry about that. However, what I didn't realize is that the MetadataEditor and this Widget stuff are two different things.  I now realize that the recently added screenshots are the _result_ of performing an action on the left-hand panel - which _is_ the widget, and the result of that action launches the corresponding editor, is that correct?  

If so, and because we now require _every_ `MetadataWidget` have its own implementation representing its metadata instance, then there should be a way for the widget to be used to "seed" the corresponding editor with necessary UI hints for that instance (among other things).  I.e., the widget is the "container" for UI items (like labels) that are not inherent in the metadata schema.

Another item to discuss:
- Can the widgets be used to house `uihints` for use by their corresponding editors?

ajbozarth:
> If so, and because we now require every MetadataWidget have its own implementation representing its metadata instance, then there should be a way for the widget to be used to "seed" the corresponding editor with necessary UI hints for that instance (among other things). I.e., the widget is the "container" for UI items (like labels) that are not inherent in the metadata schema.

This was actually how @marthacryan originally coded that PR until we revisited it and moved that info into the schema. I'm not sure if that discussion was done in meetings or on the PR though for referencing. 

> Can the widgets be used to house uihints for use by their corresponding editors?

Though we could do this I'm not sure it's better. Though we currently only open the MetadataEditor via a MetadataWidget, it actually is created via a command call and is not directly tied to the widget. I'm hesitant to tie them together intrinsically , but not 100% against it. Though I may reevaluate this once I've worked on moving CodeSnippetWidget to extend the abstract MetadataWidget (I'm not sure how that process will go).

kevin-bates:
> I'm not sure if that discussion was done in meetings or on the PR though for referencing.

It was in a meeting with the three of us, but I don't recall discussing what the code was doing, only more about how to go about improving labels, and that a _general_ approach was preferred so we figured the schemas were the best location.  However, since we're now essentially revisiting this topic and requiring specific implementations anyway, I think that discussion should be reopened because labels and such should really remain in the front-end (IMHO).

If we stick with a schema approach, then I'd propose we change `uihints.label` to `display_name` and let each property have an optional display_name meta-property. 

I think checking how CodeSnippetWidget evolves would be a good idea before making any decisions.

ajbozarth:
As discussed in scrum earlier this week I've pushed changes that switch from using an abstract class in `ui-components` to a base class (a combination of `MetadataWidget` and `DefaultMetadataWidget` in my previous implementation) in a new `metadata` package. As part of this move I also moved the MetadataEditor to this new package as well.

A point of potential discussion: I renamed `metadata-editor` package to `metadata-extension` to match the new `metadata` library package and it more general handling of both editing and viewing metadata. @lresende it may make sense to make the name change from `metadata-editor` to `metadata-extension` in another small and quick PR that will make it into 1.0 so we wont be changing the name between versions. If you agree I can open a quick PR with those changes.

I also still have the previous TODO of moving `CodeSnippetWidget` to extend `MetadataWidget` and addressing any issues found in that move.

elyra/metadata/schemas/kfp.json
```diff
@@ -21,33 +21,51 @@
       "properties": {
         "description": {
           "description": "The description of this KFP runtime instance",
-          "type": "string"
+          "type": "string",
+          "uihints": {
+            "label": "Description"
+          }
```
kevin-bates, line 27: Please note that I just created #781 to use a `title` field to convey the display_name/label for each property.  This would allow the removal of the `uihints` stanzas until there are actual needs for edit-controls or other UI-specific behaviors should it be merged.  As a result, we might want to look at that PR soon and remove these updates if merged.

elyra/metadata/schemas/kfp.json
```diff
@@ -21,33 +21,51 @@
...
         },
         "cos_endpoint": {
           "description": "The Cloud Object Storage endpoint",
           "type": "string",
-          "format": "uri"
+          "format": "uri",
+          "uihints": {
+            "label": "Cloud Object Storage endpoint"
+          }
```
kevin-bates, line 43: It appears as though the property name portion of the label is lowercased and, IMO, should be capitalized.  Note that should #781 be used to address the label, these are capitalized.

elyra/metadata/schemas/kfp.json
ajbozarth: I'll review that asap, I agree that it makes sense for it to go in first, plus this is has a final todo on it anyway

elyra/metadata/schemas/kfp.json
ajbozarth: Fair enough, iirc I just copied the deception and removed the "the" for each of these. I think I meant to fox them later and forgot. Either way this should wait on #781 anyway

ajbozarth:
I've pushed updates to rebase on master and address the packing naming as discussed in scrum. 

Remaining TODOs:

- Update to use new titles instead of labels
- Update CodeSnippetsWidget to extend MetadataWidget

Note that both the TODOs can be done as follow up PRs if this PR is determined to be ready to be merged before those updates are ready

ptitzler:
Usability: https://github.com/elyra-ai/elyra/pull/591#issuecomment-660398196 If possible the layout should clearly separate/group Kubeflow properties from COS properties. Not sure if the screen captures were just a WIP or a proposal ...

ajbozarth:
@ptitzler those screen caps are "final" but this those files are dynamically populated. This PR doesn't actually create that editor, only surface it. Though the editor could be improved overall, I'm not sure exactly what you want, a line break between API endpoint and COS endpoint?

lresende:
@ptitzler the current implementation aims to be a generic UI editor that is auto-generated based on metadata schema. We could collect suggestions to ehnance the generated ui and tackle that in a separate pr.

marthacryan:
@ptitzler We could add the ability to "group" different fields into the metadata editor - I think that'd be better as a separate issue / PR though, I just opened #789.

ajbozarth:
@vabarbosa odd, I thought I got the color right, I wonder if it got messed up during one of my rebases. I'll fix it

packages/metadata-common/src/MetadataWidget.tsx
```diff
@@ -0,0 +1,250 @@
...
+  render(): React.ReactElement {
+    return (
+      <div>
+        <header className={METADATA_HEADER_CLASS}>
+          <div style={{ display: 'flex' }}>
+            <p> {this.props.display_name} </p>
+          </div>
+          <button
+            className={METADATA_HEADER_BUTTON_CLASS}
+            onClick={this.addMetadata.bind(this)}
```
vabarbosa, line 239: button here should include a `title` attribute so a tooltip can appear on hover. e.g.,

image

vabarbosa:
password fields should probably be masked:

image

packages/metadata-common/src/MetadataWidget.tsx
ajbozarth: I noticed this while working on moving Code Snippet to extend Metadata. It's an "easy" add, but I'm unsure what to put in the text without adding a new part just for the text. I already have a param for the header text, but that is usually plural. I could alway just have it say "Create new" but I don't think that's very good. Any ideas?

ajbozarth:
@vabarbosa I did some looking into this

> the runtimes icon is a different than the color than the other icons in the pipeline editor toolbar:

and I can't seem to get the icon to match both the sidebar icon color and the pipeline toolbar color at the same time. If you have any ideas on how to get that to work I'd love the help. It may be just adding an extra bit of css to one or the other (I only tried editing the svg file to fix it)

packages/metadata-common/src/MetadataWidget.tsx
vabarbosa: what about the title used in the form? or is that not accessible in the sidebar component:

image

ajbozarth: we can use that, but that isn't a title, it's just the schema name (which in Code Snippets case would be `code-snippet`) #734 is currently open to update that to use the new `display_name` value on a schema, but this code does't have access to the schema content, just the name. I'm hesitant to add an api call to grab the schema just to get a display name, but I'm not 100% opposed to it.

vabarbosa:
> @vabarbosa I did some looking into this
> 
> > the runtimes icon is a different than the color than the other icons in the pipeline editor toolbar:
> 
> and I can't seem to get the icon to match both the sidebar icon color and the pipeline toolbar color at the same time. If you have any ideas on how to get that to work I'd love the help. It may be just adding an extra bit of css to one or the other (I only tried editing the svg file to fix it)

@ajbozarth should be all set now. i pushed the change (https://github.com/elyra-ai/elyra/pull/591/commits/4bb1539eb8160ca16f9be9ae08401d244d338fd3) into this PR

vabarbosa:
should an the icon be added to the Runtimes sidebar:

image

similar to the Code Snippet:

image

ajbozarth:
> should an the icon be added to the Runtimes sidebar title:

There were quite a few things added to code snippet while this PR was in development. I will be adding those in as I adjust Code Snippet to extend Metadata. So any feature of Code Snippet that is missing will be added.

packages/metadata-common/src/MetadataWidget.tsx
kevin-bates: Seems like we pound the server for all instances of metadata - not sure how often for schemas, but I would argue schemas should be fetched exactly once at startup and be available to all front-end components.  Since they're static and we don't have any widgets relative to user-defined schema, this should be fine.

kevin-bates:
> password fields should probably be masked:

I agree.  For now though, I think it would be better to convey this via the `uihints`, although we'll need a _meta-property_ to convey that for backend use at some point.  To that end, I would recommend a boolean-valued hint of something like `masked`, since we could very likely need to mask fields besides passwords (and I'm not convinced we'll have password-related fields at all down the road).

packages/metadata-common/src/MetadataWidget.tsx
ajbozarth: It sound like I should add another call to `getSchema` here to deal with this then, but with the expectation that we update that function to cache the response of previous calls. 

I would suggest we deal with that cacheing in the refactor purposed #785 since currently that function can "return" error dialogs, which would make caching responses a hassle. But that purposed refactor would remove that problem. If tabling that cache is ok with you @kevin-bates then I'll add it to my refactor proposal.

As for how often `getSchema` is currently called, it is called each time a MetadataEditor is opened currently. This would add one addition call when a MetadataWidget is created (so based on our current code, once on startup)

ajbozarth:
>> password fields should probably be masked:

> I agree. For now though, I think it would be better to convey this via the uihints, although we'll need a meta-property to convey that for backend use at some point. To that end, I would recommend a boolean-valued hint of something like masked, since we could very likely need to mask fields besides passwords (and I'm not convinced we'll have password-related fields at all down the road).

I think this should be handled in a separate issue/pr though. I thought we had an issue open on this due to all of our discussion around the topic but I can't seem to find it.

packages/metadata-common/src/MetadataWidget.tsx
kevin-bates: Deferring the refactor is fine - thanks!

ajbozarth:
I opened #790 to handle the password masking issue

ajbozarth:
I pushed an update that added icon in the header, add tooltip, and fixed a handful of other issues.

@kevin-bates while doing some other work I realized that I could cache schemas without waiting for that refactor, so I added that

packages/code-snippet/src/CodeSnippetService.ts
```diff
@@ -15,48 +15,29 @@
...
 
-    return allCodeSnippets;
+    return codeSnippetsResponse;
   }
 
   // TODO: Test this function
-  static async findByLanguage(language: string): Promise<ICodeSnippet[]> {
-    const allCodeSnippets: ICodeSnippet[] = await this.findAll();
-    const codeSnippetsByLanguage: ICodeSnippet[] = [];
+  static async findByLanguage(language: string): Promise<IMetadata[]> {
```
vabarbosa, line 35: this isn't being used/called from anywhere that i can find. should it still be kept around? i suppose it will eventually come in handy if a filtering functionality is added.

ajbozarth: We've left it in since it will eventually be leveraged when we start work on #277 I believe the Cal Poly interns are also leveraging it in their POC based on our code.

packages/code-snippet/style/index.css
```diff
@@ -13,81 +13,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
vabarbosa, line 15: this file is now empty. should it still be kept around?

ajbozarth: iiuc we should always have an index.css file. If we remove it we would also need to remove style from the npm build so the build doesn't fail, and if we do that, adding it back it is a hassle. So leaving an empty css that's available to add to seems to be the standard.

marthacryan:
(I think this was already been an issue?) When the metadata editor shows a validation error for an empty field that's required but doesn't have a "description" field defined, this is the error that shows:
image
You could fix this by changing line 283 of `MetadataEditor.tsx` to `let helperText = description == undefined ? '' : description;`.

packages/code-snippet/src/CodeSnippetService.ts
vabarbosa: 👍

vabarbosa:
looks good!
just need to address the two comments from @marthacryan 
- `Save & Close` button layout
- Missing/Invalid field error message

ajbozarth:
Pushed updates:

- Added style to force Metadata Editor Save button to it's own line
- made description for Metadata Editor fields "optional" (this was how `DropDown` already handled description)
- added and updated tests

ajbozarth:
found a bug while taking screenshots, I'm figuring it out now

ajbozarth:
pushed the bug fix. JS Objects (like dicts) are pass by reference, so by adding caching to `getSchema` I inadvertently caused all calls to it to share the same reference. Fixed it by returning a deep copy of the cached schema on request.

ajbozarth:
Screenshots of the final version:

Screen Shot 2020-07-29 at 3 50 54 PM
Screen Shot 2020-07-29 at 2 22 02 PM
Screen Shot 2020-07-29 at 2 22 39 PM
Screen Shot 2020-07-29 at 2 23 01 PM
Screen Shot 2020-07-29 at 2 23 06 PM
Screen Shot 2020-07-29 at 3 52 02 PM

ptitzler:
I'm a bit concerned about terminology and the extensive use of the KFP acronym, which I believe is not officially used (at least in the Kubeflow Pipelines documentation). 
Food for thought, to avoid potential ambiguity:
-  "Create new KFP" button hint -> "Add Kubeflow Pipelines runtime"
-  "Add new KFP" text -> "Add Kubeflow Pipelines runtime"

To be more future proof it'd be good if we could try to make the UI a bit more generic:
- "Add new runtime" button (not runtime type specific text)
- "runtime specific configuration wizard (for the time being the wizard defaults to Kubeflow Pipelines, later on the user has the option to choose any supported runtimes)

kevin-bates:
Thanks Patrick.  This display value is coming directly from the [schema's display_name](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/kfp.json#L5), so this is just a matter of updating that schema value, or, to be consistent with the other properties, we tweak the schema's existing `title` meta-property and use `title` instead.  I'd be in favor of including this level of change in this PR since its touching schemas anyway. (@ajbozarth - thoughts?)

However, because the runtimes namespace can have multiple schemas taking the generic or wizard approach seems like a good way to ask _which_ runtime metadata instance should be created.

ajbozarth:
Based on discussion in the dev meeting today I have added the following:
- Added support for masking password fields in metadata editor
- Fixed the schema title displayed in the Add New tooltip, and the tab name and header for a new Metadata Editor

ajbozarth:
screen shots for latest changes:
Screen Shot 2020-07-30 at 12 08 43 PM
Screen Shot 2020-07-30 at 12 08 52 PM
Screen Shot 2020-07-30 at 12 09 03 PM
Screen Shot 2020-07-30 at 12 09 09 PM
Screen Shot 2020-07-30 at 12 09 17 PM

ptitzler:
- "Kubeflow Pipeline" => "Kubeflow Pipelines" (we need to use the formal name)
- Would you mind changing the "API Endpoint" label to "Kubeflow Pipelines API Endpoint" so that there's no doubt which endpoint we are referring to?

lresende:
Based on the screenshots above, has anyone actually made things work with metadata pointing to `http://cloning1.fyre.ibm.com:31380/pipeline`? By work I mean, click on the link on the left and access the pipelines? To me, I had to add a `/` to the URL, which makes the link work and then breaks other places.... and I tried on three different browsers with similar behavior, let me try to workaround this.

lresende:
Also, wasn't the extension going to be called `metadata-editor`?

packages/pipeline-editor/src/RuntimesWidget.tsx
```diff
@@ -34,15 +34,15 @@ export const KFP_SCHEMA = 'kfp';
  */
 class RuntimesDisplay extends MetadataDisplay<IMetadataDisplayProps> {
   renderExpandableContent(metadata: IDictionary<any>): JSX.Element {
+    const apiEndpoint = metadata.metadata.api_endpoint.endsWith('/')
+      ? metadata.metadata.api_endpoint
+      : metadata.metadata.api_endpoint + '/';
```
ajbozarth, line 39: thanks for the fix, should we do the same with the cos_endpoint?

ajbozarth:
> Why are we using `metadata` instead of `metadata-editor`?

Because the extension handles more than the metadata-editor, it also handles the metadata widget (viewer). So we renamed it to be more general. This was the point of #784

packages/pipeline-editor/src/RuntimesWidget.tsx
lresende: cos endpoint works ok, this is probably a behavior from kfp ui

PR #595, opened 2020-06-01
[WIP] Add completer to python file editor

### LINKED ISSUE #586: Elyra python file editor has non-functioning autocomplete feature (works in JL text editor)
Please provide a fix for the 1.x version as well.

**Steps to reproduce:**

- Open jupyterlab, click "python file"
-     enter import sys
-     on next line enter sys. and press tab. You will just get a tab.
-     Right click to "Create a console for editor", option is disabled.

Workaround - don't use "python file", use "text file" editor instead. Once in "text file", switch syntax highlighting to python, and create the console. Autocompletion then functions normally.

image

image

karlaspuldaro:
Fixes #586 

This is a work in progress.
Without the completer setup, the behavior of the python editor was to add a tab whenever tab key was pressed, while the expected was to trigger code auto-complete.
While these changes bind the tab key with the completer feature (no longer adds a tab), the box with code hints is not yet being displayed, even after a kernel start up.

           in the file; or

           it.


karlaspuldaro:
After doing some investigation, I found that our python editor never sets its own [session](https://jupyterlab.github.io/jupyterlab/apputils/classes/sessioncontext.html#session), which represents a live connection to a session kernel, basically a manager that owns the kernel.

In our current implementation, we create a python kernel and manually runs/stops it, only reading the input once at runtime, and exposing the output when it's done, there is never a live session running (removing the shutdown command after a run is not enough) - hence the option to create a python editor console is disabled from our extension.
Code completer needs to talk to the kernel through its session, so we need to set it up and expose it from the python runner.

lresende:
Applying "Jupyterlab 1.x" as this needs backport to 1.x branch. I have started the backport but needs a little more work as TOC stopped working (we might need a smaller cleanup on 1.x than in 2.x).

PR #597, opened 2020-06-01
Update Elyra docker image

### LINKED ISSUE #381: Create and publish elyra docker images that works with JupyterHub
We should publish new elyra images when jhubs singleuser image comes out with 0.9.0 since we are experiencing build failures with the 0.8.2 tag.

lresende:
Due to some build and dependency issues around npm linked
packages and JupyterLab extensions, the current version of
the docker image do a git clone and perform an Elyra build.

Fixes #381 

Some other todos that might come on this pr
 - [x] - Start from jovyan work dir
 - [x] - Select user via variable, to support different user (see other images)
 - [x] - Integrate with Jupyter Enterprise Gateway
 - [x] - add some contents on the plain image
   - This can be done by git clone sample repos by the user

 
           in the file; or

           it.


lresende:
This is working locally, and I am doing some tests on the JupyterHub environment

image

kevin-bates:
This looks good so far.  I'm not even sure what you'd do to trim things.  

It's too bad we don't have a public "sandbox" area where we could point some baked-in runtime metadata and add some sample code snippets, etc.  (I suppose the code-snippets could be added via the docker build.)

We'll need some documentation as well.  Like how to bring your own filesystem so that any metadata that is created exists across the duration of the container's lifetime, etc.

etc/docker/elyra/start-elyra.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export NB_PORT=${NB_PORT:-8888}
+export GATEWAY_HOST=${GATEWAY_HOST:-localhost}
+export KG_URL=${KG_URL:-http://${GATEWAY_HOST}:${NB_PORT}}
+export KG_HTTP_USER=${KG_HTTP_USER:-jovyan}
```
kevin-bates, line 6: Should the default value here be `${NB_USER}`?  A `docker inspect` of the base image shows it will be in the environment:
```
           "Env": [
                "PATH=/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "DEBIAN_FRONTEND=noninteractive",
                "CONDA_DIR=/opt/conda",
                "SHELL=/bin/bash",
                "NB_USER=jovyan",
  ```

File: etc/docker/elyra/start-elyra.sh [hunk 2]
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export NB_PORT=${NB_PORT:-8888}
+export GATEWAY_HOST=${GATEWAY_HOST:-localhost}
+export KG_URL=${KG_URL:-http://${GATEWAY_HOST}:${NB_PORT}}
```
kevin-bates, line 5: Use of the `KG_` envs is obsolete - assuming you're only supporting the embedded gateway.  `KG_` should be replaced with `JUPYTER_GATEWAY_`.

etc/docker/elyra/start-elyra.sh
```diff
@@ -0,0 +1,17 @@
...
+export KG_HTTP_USER=${KG_HTTP_USER:-jovyan}
+export KG_REQUEST_TIMEOUT=${KG_REQUEST_TIMEOUT:-120}
+export KERNEL_USERNAME=${KG_HTTP_USER}
+
+echo "Gateway URL: " ${JUPYTER_GATEWAY_URL}
```
kevin-bates, line 10: You're never setting `JUPYTER_GATEWAY_URL`.

etc/docker/elyra/start-elyra.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export NB_PORT=${NB_PORT:-8888}
```
kevin-bates, line 3: NB_PORT is the notebook port for the local notebook, not the EG port.  If all defaults are taken, the gateway URL will be the same as the notebook URL.  Hmm, I wonder if that would just work?  I doubt it, but it would be like a loopback mode thing. :smile:

etc/docker/elyra/start-elyra.sh
```diff
@@ -0,0 +1,17 @@
...
+export KERNEL_USERNAME=${KG_HTTP_USER}
+
+echo "Gateway URL: " ${JUPYTER_GATEWAY_URL}
+echo "JupyterLab port: " ${NB_PORT}
+echo "Kernel user: " ${KERNEL_USERNAME}
+
+echo "${@: -1}"
+
+
+exec /usr/local/bin/start-singleuser.sh $*
```
kevin-bates, line 17: Based on this, there is no option to **not** use Enterprise Gateway.  Seems like we should allow for not using EG.

I would suggest using `JUPYTER_GATEWAY_URL` and make the user bring that as an env.

```
docker run ... -e JUPYTER_GATEWAY_URL=http://foo.bar:8888 ...
```

Things like request timeout (and user) can be setup unconditionally...
```
export JUPYTER_GATEWAY_REQUEST_TIMEOUT=${JUPYTER_GATEWAY_REQUEST_TIMEOUT:-120}
export JUPYTER_GATEWAY_HTTP_USER=${JUPYTER_GATEWAY_HTTP_USER:-${NB_USER}}
```

etc/docker/elyra/start-elyra.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
akchinSTC, line 1: Needs license header

etc/docker/elyra/start-elyra.sh
akchinSTC: +1 on the not requiring EG option

akchinSTC:
Should we come up with just a VERY simple hello world example notebook and pipeline for elyra? Could be useful for the binder as well.

etc/docker/elyra/start-elyra.sh
lresende: I will probably clean up this section, and remove some of these, but will have this in mind, which is mostly from JEG.

etc/docker/elyra/start-elyra.sh
lresende: Yes, `JUPYTER_GATEWAY...` is being set on the jupyter hub config.

etc/docker/elyra/start-elyra.sh
lresende: same comment as above, will cleanup this

etc/docker/elyra/start-elyra.sh
lresende: haven't we fixed this on lint @ajbozarth ?

lresende:
I was going to try to clone the python and the covid scenario here

etc/docker/elyra/start-elyra.sh
ajbozarth: the license header linting is only for the front end (.ts,.tsx,.js files), @kevin-bates wrote the back end lint and I'm unsure if he added that, it could also be because this script is not in the elyra dir and the backend lint only runs there (front end lint runs on the whole repo)

etc/docker/elyra/start-elyra.sh
kevin-bates: Sounds good.  Here's the [EG doc reference](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started.html#notebook-6-0-and-above).

etc/docker/elyra/start-elyra.sh
kevin-bates: Yes - currently flake8 runs on the `elyra` tree and only against `.py` files.

kevin-bates: Please add a header to https://github.com/elyra-ai/elyra/blob/master/etc/scripts/protect-metadata.sh while you're at it. :smile:

akchinSTC:
> I was going to try to clone the python and the covid scenario here

Would the covid scenario be a little too involved for someone just starting to use elyra? I was thinking of baby stepping it starting with like a 3 node pipeline

etc/docker/elyra/start-elyra.sh
lresende: I think the reference here is important, this is instantiating the notebook (JupyterLab/Elyra) and if gateway url is passed via env variable it will connect to it in a separate pod

etc/docker/elyra/start-elyra.sh
lresende: this should be provided otherwise we will use local kernels

etc/docker/elyra/start-elyra.sh
lresende: yes, and I was able to use that in hub both ways

lresende: i didn't play much with it as a standalone image tough.

etc/docker/elyra/start-elyra.sh
kevin-bates: Ok - thanks.  I guess that explains the comment below.  Why are we dealing with GATEWAY_HOST and KG_URL at all then if JUPYTER_GATEWAY_URL is already set?

etc/docker/elyra/start-elyra.sh
kevin-bates: ok - sounds good.  I think once the new variables are updated, you should be good.

lresende:
```
docker run -it -p 8888:8888 elyra/elyra:dev jupyter lab --debug
```

image

etc/docker/elyra/Dockerfile
```diff
@@ -16,24 +16,33 @@
...
+RUN chmod ugo+x /usr/local/bin/start-elyra.sh && \
+    apt-get update && apt-get install -y build-essential curl && \
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list && \
+    apt-get update && apt-get install -y yarn && \
+    rm -rf /var/lib/apt/lists/*
 
-RUN mkdir -p /home/jovyan/.local/share/jupyter/metadata/runtimes
+USER $NB_USER
+#USER jovyan
```
kevin-bates, line 33: Can we remove this commented out command?  If its purpose is to indicate that NB_USER will likely be jovyan, then we should add a more formal comment, but I don't think that is necessary.

lresende: sorry, i forgot to push the latest changes, including this one.

Issue #536, opened 2020-05-14
Add Python/R Editor front end tests


karlaspuldaro:
Initial test scenarios (ok to open separate issues/PRs for each  task or group below)

Create new python file
- [x] New python file available from main launcher/other ( #457 )
- [x] New python file available from file->new->python file ( #603 )

Icons
- [x] Editor icons visible from: launcher/other, file->new->python/r file, file browser when a `.py` or `.r` file is listed ( #2216 )

Open Python/R Editor
- [x] open blank Python/R Editor from launcher ( #457 )
- [x] open blank Python/R Editor from file->new->python file ( #603 )
- [x] open a `.py` or `.r` file should open Python/R Editor with expected content ( #2216 )
- [x] check content of menu when right click the editor tab (options should include `Python/R File`)( #2216 )

Python/R Editor toolbar
- [x] check for toolbar available ( #603 )
- [x] check for buttons and icons available: save, run, stop, select kernel drop down ( #603 )
- [x] verify all installed python (and only python) kernels listed on kernel select drop down ( #603) 

Output
- [x]  run code in the editor should open console output ( #2499 )
- [x]  console output scroll top/bottom buttons available ( #2499 )
- [x]  verify expected code output (#2507)
- [x]  run broken code should display error on console output (#2507)

ptitzler:
Certainly. Done!

Hema-Mathiyazhagan:
@karlaspuldaro I would like to work on few of this subtasks. could you please assign this one to me?

karlaspuldaro:
All tasks marked as done

PR #603, opened 2020-06-04
Python Editor tests

kjdoyle:
Basic Python Editor tests for #536 
 
           in the file; or

           it.

### DIFF
--- a/tests/integration/pythoneditor.ts
+++ b/tests/integration/pythoneditor.ts
@@ -57,6 +57,6 @@ describe('PythonEditor', () => {
   });
 
   it('check select kernel dropdown exists and has python3', () => {
-    cy.get('.elyra-PythonEditor .jp-Toolbar select > option[value=python3]');
+    cy.get('.elyra-PythonEditor .jp-Toolbar select > option[value*=python]');
   });
 });
--- a/tests/integration/pythoneditor.ts
+++ b/tests/integration/pythoneditor.ts
@@ -18,7 +18,45 @@ describe('PythonEditor', () => {
     cy.visit('?token=test&reset');
   });
 
-  it('opens blank python', () => {
+  it('opens blank python from launcher', () => {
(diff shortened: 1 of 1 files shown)


kjdoyle:
@marthacryan @karlaspuldaro if you could do a review on this WIP PR that would be great. Looking for feedback if this is the right starting approach for determining the elements to click.

tests/integration/pythoneditor.ts
```diff
@@ -18,7 +18,45 @@ describe('PythonEditor', () => {
...
+    cy.get('svg[data-icon="ui-components:run"]');
+  });
+
+  it('check stop button exists and icon', () => {
+    cy.get('button[title="Stop"]');
+    cy.get('svg[data-icon="ui-components:stop"]');
+  });
+
+  it('check select kernel dropdown exists and has python3', () => {
+    cy.get('.elyra-PythonEditor .jp-Toolbar select > option[value=python3]');
```
karlaspuldaro, line 60: Maybe the python kernel check could be more generic.
In cases of running elyra on a remote cluster, or if someday we use a cluster on a test environment , python kernels could look like [these](https://github.com/elyra-ai/elyra/blob/master/docs/source/images/python-runner.1.gif).
I would say checking if substrings have 'python' instead?

Also I think it's a good idea to have other kernel types installed to test if this dropdown only shows python ones

kjdoyle: Thanks for the tip. I will work to update it.

karlaspuldaro: Just a suggestion, once you're ready with this change, you can remove the WIP so we'd merge this PR with what you already have. The list of tests in the issue is a bit long, it can be delivered into smaller PRs, that way we could be having some of them earlier

kjdoyle: Thanks. That makes sense to break it up into smaller PRs. I have addressed this issue now.

PR #606, opened 2020-06-04
Update `/api` endpoints and swagger docs

### LINKED ISSUE #434: All endpoints should be documented in swagger docs
Just realizing that /api/scheduler is not documented in the swagger docs.  In addition, PR #433 may introduce /api/export.  Both of these need to be documented in the swagger docs.

### LINKED ISSUE #513: Update the `/api` endpoints to `/elyra`

vabarbosa:
Fixes #513 
by changing all endpoints from `/api` to `/elyra`

Fixes #434 
by adding `/elyra/pipeline/schedule` and `/elyra/pipeline/export` into the swagger documentation

 
           in the file; or

           it.

### DIFF
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -9,7 +9,7 @@ info:
 
 paths:
 
-  /api/elyra.yaml:
+  /elyra/elyra.yaml:
     get:
       tags:
       - api
@@ -24,7 +24,7 @@ paths:
         500:
           description: Unexpected error.
 
-  /api/namespace:
+  /elyra/namespace:
     get:
       tags:
       - namespace
@@ -45,7 +45,7 @@ paths:
         500:
           description: Unexpected error.
 
-  /api/schema/{namespace}:
+  /elyra/schema/{namespace}:
     get:
       tags:
       - schema
@@ -76,7 +76,7 @@ paths:
         500:
           description: Unexpected error.
 
-  /api/schema/{namespace}/{resource}:
+  /elyra/schema/{namespace}/{resource}:
     get:
       tags:
       - schema
@@ -107,7 +107,7 @@ paths:
         500:
           description: Unexpected error.
 
-  /api/metadata/{namespace}:
+  /elyra/metadata/{namespace}:
     get:
       tags:
       - metadata
@@ -177,7 +177,7 @@ paths:
         500:
           description: Unexpected error.
 
-  /api/metadata/{namespace}/{resource}:
+  /elyra/metadata/{namespace}/{resource}:
     get:
       tags:
       - metadata
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -449,19 +449,19 @@ components:
     PipelineScheduleResponse:
       description: The set of properties comprising a pipeline processor response entity
       required:
-      - run-url
-      - object-storage-url
-      - object-storage-path
+      - run_url
+      - object_storage_url
+      - object_storage_path
       type: object
       properties:
-        run-url:
+        run_url:
           type: string
           description: The runtime URL to access the pipeline experiment
-        object-storage-url:
+        object_storage_url:
           type: string
           description: The object storage URL to access the pipeline outputs
             and processed notebooks
-        object-storage-path:
+        object_storage_path:
           type: string
           description: The object storage working directory path where the pipeline outputs
             and processed notebooks are located
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -273,6 +273,51 @@ paths:
         500:
           description: Unexpected error.
 
+  /elyra/pipeline/schedule:
+    post:
+      tags:
+      - pipeline
+      summary: Execute pipelines as batch jobs
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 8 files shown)


vabarbosa:
@kevin-bates i have added  `/elyra/pipeline/schedule` and `/elyra/pipeline/export` into the swagger docs. please confirm i correctly interpreted the body and response for each request. thanks.

fyi, i also noticed the response for the `elyra/pipeline/schedule` return an object with key names that contain hyphen (i.e., `run-url`, `object-storage-url`, `object-storage-path`). however all other endpoints use underscore (e.g., `display_name`, `export_format`, etc.)

elyra/api/elyra.yaml
```diff
@@ -364,6 +409,84 @@ components:
...
+                type: object
+                properties: {}
+                description: Runtime configuration that should be used to submit the pipeline
+
+    PipelineScheduleResponse:
+      description: The set of properties comprising a pipeline processor response entity
+      required:
+      - run-url
+      - object-storage-url
+      - object-storage-path
```
kevin-bates, line 454: Per your previous question regarding the use of '-' in response keys, I think we need to be consistent and I view '_' more appropriate (and common) for "code".  My vote would be that these should be changed to underscores.

elyra/api/elyra.yaml
```diff
@@ -273,6 +273,51 @@ paths:
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    description: Pipeline export status
+                  message:
+                    type: string
+                    description: Pipeline export message
```
kevin-bates, line 319: Not a fan of redundant responses - especially since it's not providing any [contextual information](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/handlers.py#L52-L53) and, literally, not used by the frontend.

I'd rather see us return `204`, but that's probably a larger group decision.  If the response could contain some resource or **anything** related to this particular request, then `200` and that information would be fine.

@akchinSTC - what _is_ the result of an export?  How does the user access the exported pipeline?

Looking at the request body, it seems like `export_path` is probably what we want here.  If that doesn't include the filename on the request, the response should append the filename.

If that file _could be_ accessed via the content service, then the **best** response would be a URI relative to `/api/contents/`.  This would mean that `export_path` is a subdirectory under the notebook root_dir.

Are users given some help as to where `export_path` will be located - or do they have to type in the full thing?  If always relative to root_dir, then the contents URI is the answer (IMO).

akchinSTC: an export in the kfp processor returns the path to the file relative to the file browsers current directory. It gets this information from the pipeline editor widget.
https://github.com/elyra-ai/elyra/blob/6fd7a857ca6bcb2cc7010ba11bae244590885b8c/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L525-L532

kevin-bates: Excellent.  It looks like the export method already [returns the export_path](https://github.com/elyra-ai/elyra/blob/6fd7a857ca6bcb2cc7010ba11bae244590885b8c/elyra/pipeline/processor_kfp.py#L137), so we should just have the handler add that response value to `/api/contents/` as the response.

Here's an example using a pipeline file...
image

elyra/api/elyra.yaml
akchinSTC: +1

vabarbosa: thanks. i agree and have updated the keys to be more consistent (i.e., use `_`)

elyra/api/elyra.yaml
kevin-bates: I conflated things a bit in this discussion.  Its the _Location_ header that should represent the resource URI so I think we should set Location to the appropriate path relative to /api/contents... https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/handlers.py#L68

Then, just have the formal response be the `export_path`.

vabarbosa: thanks... yes i would agree that is better than status/message response... should i go ahead open a new issue/PR for this and work on it there?

PR #629, opened 2020-06-10
Update metadata API to require either name or display_name

### LINKED ISSUE #623: Metadata `name` should not be required if `display_name` is provided when creating a new metadata entry
When creating a new metadata entry, currently the API call requires a `name` while the `display_name` is optional. It may be worthwhile when creating a new metadata entry to allow submissions to the server without a `name` but include a `display_name` (i.e., you need a name and/or a display_name). In cases where only the `display_name` is provided the server would generate a `name` (e.g., normalizing the `display_name`).

This came up in the [discussion](https://github.com/elyra-ai/elyra/pull/589#discussion_r437650918) in review of #589 where creating a code snippet required the user to properly format the `name` (e.g., not using spaces or unacceptable characters) otherwise a server error was returned.

Rather than require the user to abide by the `name` restrictions and additionally enforce validation of the `name` in the client side, the server should accept a new entry if there is a `display_name` but no `name`. the server could then create a new `name` by properly formatting the `display_name`.

### LINKED ISSUE #632: Metadata display_name constraints should be relaxed
Currently, all metadata schemas restrict `display_name` fields to the regular expression: `^[a-zA-Z][a-zA-Z0-9-_. (){}]*[a-zA-Z0-9)}]$`.  This prevents display names with apostrophes, international characters, or starting with a digit - to name just a few.  (See [this comment](https://github.com/elyra-ai/elyra/pull/629/files#r438509890))

@vabarbosa will be updating the code-snippets schema in #629 but all of the other schemas should be similarly updated to allow for any characters.

vabarbosa:
Update the `/metadata/<namespace>` POST endpoint to allow missing `name` field if the `display_name` is provided

Fixes #623
Fixes #632 

~~FYI, the swagger docs continue to still indicate `name` is required. I did not update it because at the moment [swagger does not provide a mechanism for specifying any one of a set of parameters is required](https://github.com/OAI/OpenAPI-Specification/issues/256).~~ 
 
           in the file; or

           it.


elyra/metadata/handlers.py
```diff
@@ -74,11 +74,20 @@ def _validate_body(self):
         body = self.get_json_body()
 
         # Ensure name, schema_name and metadata fields exist.
```
kevin-bates, line 76: nit: we should probably remove 'name' from the comment.

EDIT: Actually, I suppose 'name' is still a required field - we just defer its "required-ness", so it's probably fine to leave that comment alone.

elyra/metadata/metadata.py
```diff
@@ -228,6 +228,20 @@ def __init__(self, namespace, **kwargs):
...
+        # remove all invalid characters
+        name = re.sub('[^a-z0-9-_]+', '', name)
+        # begin with alpha
+        if not name[0].isalpha():
+            name = 'a_' + name
+        # end with alpha numeric
+        if not name[-1].isalnum():
+            name = name + '_0'
+
+        return name
```
kevin-bates, line 243: Nice.

elyra/metadata/schemas/code-snippet.json
```diff
@@ -11,8 +11,7 @@
     },
     "display_name": {
       "description": "The display name of the Code Snippet",
-      "type": "string",
-      "pattern": "^[a-zA-Z][a-zA-Z0-9-_. (){}]*[a-zA-Z0-9)}]$"
+      "type": "string"
```
kevin-bates, line 14: Why is the pattern removed?  The pattern does require to begin with alpha and end with alphanumeric.

elyra/metadata/handlers.py
vabarbosa: good catch. i can remove it from comment since it is mentioned in the comment that immediately follows that section

elyra/metadata/schemas/code-snippet.json
vabarbosa: as it originally was, if i wanted to set the `diplay_name` for a code snippet as "**josé**" or "**test jupyter's kernel**" or "**5 sec delay**" these would all fail.

my understanding was the user should be able to enter any string for `display_name`. if it is necessary to enforce it starts with alpha and end with alphanumeric i can go along that, but do we want to also restrict the rest of the characters in between?

vabarbosa:
> Thanks Va - just a couple comments.
> 
> Regarding the swagger doc update, I guess I'm not following how removing 'name' as required from [`MetadataResourceBodyPost`](https://github.com/elyra-ai/elyra/blob/master/elyra/api/elyra.yaml#L326-L332) applies to the issue you list regarding query parameters. Once we support query parameters, I would expect them to apply to the various GET handlers anyway and I figured only the body for POST would require an update here. I'm sorry if I'm not understanding the issue.

@kevin-bates my apologies, the misunderstanding was on my part. not sure how or why i was confusing the query parameters with the request body. you can ignore that comment!

elyra/metadata/schemas/code-snippet.json
kevin-bates: Totally correct - that's what that field is for and the previous limitation was completely artificial.  All of the other schemas are this way and they'll need the same change - but the others can be handled separately.  Thanks for noticing that.

kevin-bates:
No worries at all Va!

kevin-bates:
Hi Va - your particular branch is causing build errors.  Not sure if its due to how you maintain your branches or not but I can't reproduce the build issue unless I pull and build your branch.  My system uses an older version of yarn than what GH Actions are using, but the results are the same...
```
yarn run eslint
yarn run v1.17.3
$ eslint . --fix --ignore-path .gitignore --ext .ts,.tsx,.js
✨  Done in 2.95s.
export PATH=$(pwd)/node_modules/.bin:$PATH && lerna run build
lerna notice cli v3.22.1
lerna info Executing command in 6 packages: "yarn run build"
lerna ERR! yarn run build exited 1 in '@elyra/ui-components'
lerna ERR! yarn run build stdout:
yarn run v1.17.3
$ tsc
../../node_modules/@blueprintjs/core/lib/esm/components/divider/divider.d.ts(258,28): error TS2707: Generic type 'UIEvent<T>' requires between 0 and 1 type arguments.
../../node_modules/@blueprintjs/core/lib/esm/components/divider/divider.d.ts(259,35): error TS2707: Generic type 'UIEvent<T>' requires between 0 and 1 type arguments.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build stderr:
error Command failed with exit code 1.

lerna ERR! yarn run build exited 1 in '@elyra/ui-components'
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
make: *** [build-ui] Error 1
```

vabarbosa:
thank you @kevin-bates, i did notice that and was looking into it. 

the failure is happening with `@blueprintjs` components which [`@jupyterlab/ui-components`](https://github.com/jupyterlab/jupyterlab/blob/master/packages/ui-components/package.json#L45-L46) depends on. and the `@elyra/ui-components` depend on `@jupyterlab/ui-components`. 

there was a new release of [`@blueprintjs`](https://www.npmjs.com/package/@blueprintjs/core) (right before i submitted the PR) which i believe is causing these errors. someone else has also [reported this error to them](https://github.com/palantir/blueprint/issues/4174). 

i will continue to look into this further to make sure i am not missing anything, but i suspect you should see the same error in any new clean build which ends up downloading the latest `@blueprintjs` components.

kevin-bates:
I see.  Thank you for the explanation.  I suppose this is just a way of life **_Dependency-land!_**  _Not quite the happiest place on earth._

vabarbosa:
i just pushed a change that should pin the `@blueprintjs` components version and resolve the build failures for now (until a fix/new version is available).

elyra/metadata/schemas/code-snippet.json
vabarbosa: change https://github.com/elyra-ai/elyra/pull/629/commits/c978ef93f38a3759cc88c10d0cc9b770b659716b has been included in this PR to relax the `display_name` restrictions for other schemas

vabarbosa:
> Thanks Va - this is a nice enhancement for display names.
> 
> Does the update of the build-break change now need to be backed out of this PR or will it just get subsumed by git's conflict resolution stuff?

it should be resolved by git but i will merge the update and confirm. if necessary will back out the changes

elyra/metadata/metadata.py
```diff
@@ -228,6 +228,20 @@ def __init__(self, namespace, **kwargs):
         self.log.debug("Namespace '{}' is using metadata directory: {} from list: {}".
                        format(self.namespace, self.preferred_metadata_dir, self.metadata_paths))
 
+    def _get_normalized_name(self, name):
+        # lowercase and replaces spaces with undersocre
```
karlaspuldaro, line 232: typo
```suggestion
        # lowercase and replaces spaces with underscore
```

vabarbosa: thx

PR #636, opened 2020-06-11
Validate that runtime images meet minimum criteria

### LINKED ISSUE #631: Add mechanism to validate runtime-images delivered in releases
We have a requirement that images used in elyra pipelines have both `curl` and `python3`.  As we look at the setup of runtime-images we provide and update those in various releases, it would be helpful to have a means of ensuring the updated images meet our requirements.

We should add a target to the Makefile that validates runtime images in this manner.  This target could be part of each build or part of the test framework.  I suspect it might take a little time to validate each, so I'm initially thinking this be tied to testing - so long as formal CI builds fail if an image doesn't meet our criteria.

kevin-bates:
Currently validates that each image contains `curl` and `python3`, but that list of commands can be extended in the Makefile (or via a variable).

Because validation can take 10-15 minutes when no images are present, I've created a separate GitHub job to run in parallel with the other platform-specific jobs.  The image-validation job does not use a matrix approach - so the build time should not be much longer than it is now.

By default, the images remain.  This way quicker dev/testing can occur.  The GitHub action, on the other hand, goes ahead and removes each image in case space considerations are an issue.  Developers can quickly remote these images by running validation just like GH Actions:
```
make REMOVE_RUNTIME_IMAGE=1 validate-runtime-images
```

Resolves #631

           in the file; or

           it.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -64,3 +64,12 @@ jobs:
     - name: Collect logs
       if: failure()
       run: cat /tmp/jupyterlab-debug-*.log
+  image_validation:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        clean: true
+    - name: Validate runtime images
+      run: make REMOVE_RUNTIME_IMAGE=1 validate-runtime-images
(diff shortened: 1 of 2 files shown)


lresende:
@kevin-bates you saved me the rest of my day, as I had just created a new branch to work on a similar test. Really appreciate you help here.

PR #642, opened 2020-06-12
Address metadata update issues

### LINKED ISSUE #637: Metadata updates could lead to data loss if update contains invalid data
When updating metadata, the old file is deleted prior to the new file's "creation".  However, if the new data violates the schema, the previous file is no longer present - resulting in data loss.

We'll need to revisit this approach try to preserve transactional integrity.

kevin-bates:
This pull request contains several changes, most of which were inter-related...
- Prevent data loss on updates
Data loss of failed update attempts is now prevented.  We now validate the metadata against its schema before persistence.  In addition, we rename the original file and if there are issues writing the updated file or schema validation fails after the write, then the renamed original is restored and an exception is raised reflecting the failure.

- Adjust exception to status code mappings
ValidationError and ValueError exceptions now return 400 status codes.  Previously, they were associated with 404.  Now only FileNotFoundErrors will result in 404 codes - this includes things like missing namespace or schema. 

- Require full body on PUT requests
Previously, PUT didn't require required properties since it would just overlay the updates on a newly fetched copy.  However, that approach does not lend itself to _removing_ fields from a schema instance - since there's no way to distinguish a field that wasn't sent from one being removed.  As a result, PUT behaves just like POST relative to the body requirement.  Neither requires a `name` field since POST will derive `name` from `display_name` and PUT will infer name from the `resource` in the endpoint.  This leads to a more simplified Swagger doc in that there's now a single metadata resource body description.

- Schema is required for metadata
In that same vein, we now require all metadata to be associated with a schema and, since schema_name is required in the schema file, metadata missing a schema designation will be considered invalid.

- Fix test side-effect
I found some issues with tests where one test was side-effecting another because the static test data was shallowed copied.  Once the deep copy was performed the tests side-effects stopped.

Fixes #637

           in the file; or

           it.

### DIFF
--- a/elyra/metadata/metadata.py
+++ b/elyra/metadata/metadata.py
@@ -13,12 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import copy
 import io
 import json
 import jupyter_core.paths
 import os
 import re
 import shutil
+import time
 import warnings
 
 from abc import ABC, abstractmethod
@@ -169,8 +171,8 @@ class MetadataStore(ABC):
     def __init__(self, namespace, **kwargs):
         self.schema_mgr = SchemaManager.instance()
         if not self.schema_mgr.is_valid_namespace(namespace):
-            raise ValueError("Namespace '{}' is not in the list of valid namespaces: {}".
-                             format(namespace, self.schema_mgr.get_namespaces()))
+            raise FileNotFoundError("Namespace '{}' is not in the list of valid namespaces: {}".
+                                    format(namespace, self.schema_mgr.get_namespaces()))
 
         self.namespace = namespace
         self.log = log.get_logger()
@@ -295,13 +297,18 @@ class FileMetadataStore(MetadataStore):
             raise ValueError("Name of metadata must be lowercase alphanumeric, beginning with alpha and can include "
                              "embedded hyphens ('-') and underscores ('_').")
 
+        # Validate the metadata prior to persistence.  We'll validate again after persistence.
+        self._validate_metadata(name, metadata.to_dict(trim=True))
+
         metadata_resource_name = '{}.json'.format(name)
         resource = os.path.join(self.preferred_metadata_dir, metadata_resource_name)
 
         # Handle replacement behavior for hierarchy.
+        renamed_resource = None
         if os.path.exists(resource):
-            if replace:
-                os.remove(resource)
+            if replace:  # If we're replacing (updating), we need to rename the current file to allow restore on errs
+                renamed_resource = resource + str(time.time())
+                os.rename(resource, renamed_resource)
             else:
                 msg = "Metadata resource '{}' already exists.".format(resource)
                 self.log.error(msg)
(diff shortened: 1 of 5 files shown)


elyra/metadata/metadata.py
```diff
@@ -169,8 +171,8 @@ class MetadataStore(ABC):
     def __init__(self, namespace, **kwargs):
         self.schema_mgr = SchemaManager.instance()
         if not self.schema_mgr.is_valid_namespace(namespace):
-            raise ValueError("Namespace '{}' is not in the list of valid namespaces: {}".
-                             format(namespace, self.schema_mgr.get_namespaces()))
+            raise FileNotFoundError("Namespace '{}' is not in the list of valid namespaces: {}".
```
lresende, line 174: I believe this should still be ValueError, as this is a generic store and not specific to FileMetadataStore.

lresende:
In general, I would not like to see FileMetadataStore specific exceptions (e.g. FileNotFound) being propagated as a service contract. What happens if you are using a MongoMetadataStore.

kevin-bates:
Understood. Please open an issue that indicates all the classes of exceptions we should return.  I suspect we'll probably group them similar to how they are mapped to status codes.  We'll want a "not found" category, a "permission" category, etc.

elyra/metadata/metadata.py
kevin-bates: ok

elyra/metadata/tests/test_metadata.py
```diff
@@ -403,6 +403,74 @@ def test_manager_hierarchy_update(tests_hierarchy_manager, factory_dir, shared_d
     assert byo_2.resource.startswith(str(shared_dir))
 
 
+def test_manager_update(tests_hierarchy_manager, metadata_tests_dir):
+
+    # Create some metadata, then attempt to update it with a known schema violation
```
vabarbosa, line 408: minor nit. this comment is not applicable to this method.

elyra/metadata/tests/test_handlers.py
```diff
@@ -273,12 +274,40 @@ def test_update_instance(self):
         instance = r.json()
         assert instance['metadata']['number_range_test'] == 7
 
+        # Now attempt the update again, but with bad metadata and ensure previous still exists
```
vabarbosa, line 277: not a big deal but i would advocate for breaking this up into multiple test methods each focusing one task (i.e., update non-existent, update valid instance, update bad metadata)

elyra/metadata/metadata.py
```diff
@@ -441,6 +457,20 @@ def _get_schema(self, schema_name):
 
         return schema_json
 
+    def _validate_metadata(self, name, metadata_json):
+        """Validate the metadata.  If no schema name is present, False is returned.
```
vabarbosa, line 461: minor nit. according to docstring, the method may return False, True, or ValidationError but it looks like it every only returns True or ValidationError

elyra/metadata/tests/test_handlers.py
kevin-bates: You're making @lresende proud! :smile: Sounds good.

elyra/metadata/metadata.py
kevin-bates: Good catch - that was a part of a later change.

kevin-bates:
No worries Va - that was a good catch.  So that I can make this worth your time, I noticed that the use of periods was not consistent - so I changed that as well. :smile:

PR #643, opened 2020-06-13
Added endpoint to 404 dialog text

### LINKED ISSUE #639: Add the endpoint to the 404 message
It would be helpful if this message could include the endpoint that is not found.
image

ajbozarth:
Added the api endpoint that returns the 404 to the 404 dialog text

Fixes #639

Screen Shot 2020-06-12 at 5 16 51 PM

 
           in the file; or

           it.

### DIFF
--- a/packages/application/src/requests.tsx
+++ b/packages/application/src/requests.tsx
@@ -64,10 +64,14 @@ export class RequestHandler {
    *
    * @returns A promise that resolves with whether the dialog was accepted.
    */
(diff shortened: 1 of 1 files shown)


lresende:
Would the full endpoint make more sense here other then just the service path?

ajbozarth:
@lresende I tried that and it looked pretty ugly, I'm not sure we need it since it would just be the same as the current url

PR #650, opened 2020-06-16
[WIP] Serve pipeline configuration from server

### LINKED ISSUE #454: Implement more dynamic pipeline editor configuration
Currently, the pipeline editor has static JSON configuration files. We should enable
more dynamic configuration files via REST APIs. 

Files today:
https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/palette.json
https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/properties.json

fresende:
Fixes #454 
 
           in the file; or

           it.


elyra/pipeline/handlers.py
```diff
@@ -97,3 +99,27 @@ def post(self, *args, **kwargs):
...
+
+class PipelineConfigHandler(HttpErrorMixin, APIHandler):
+    """Handler to expose method calls to retrieve pipelines editor configuration"""
+
+    valid_resources = ["palette", "properties"]
+
+    @web.authenticated
+    @gen.coroutine
+    def get(self, resource):
+        if(resource in self.valid_resources):
```
kevin-bates, line 112: I think it might be a good idea to ["unescape" any parameters](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/handlers.py#L101-L102) before their use.

elyra/pipeline/handlers.py
```diff
@@ -97,3 +99,27 @@ def post(self, *args, **kwargs):
...
+            msg_json = self._read_config(resource)
+        else:
+            # invalid resource, throw an error
+            raise ValueError("Invalid configuration name {}".format(resource))
+        self.write(msg_json)
+        self.flush()
+
+    def _read_config(self, config_name):
+        config_dir = os.path.join(os.path.dirname(__file__), 'resources')
+        config_file = os.path.join(config_dir, config_name + ".json")
```
kevin-bates, line 122: `os.path.join` can take a list of directories, so we should combine these statements...
```suggestion
        config_file = os.path.join(os.path.dirname(__file__), 'resources', config_name + ".json")
```

fresende:
I am currently having some issues after i tried to integrate with the front end, and i would appreciate some help:

image

any help would be highly appreciated

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -163,33 +165,60 @@ export class PipelineEditor extends React.Component<
...
+
+    Promise.resolve([
+      RequestHandler.makeGetRequest('elyra/pipeline/config/palette', false)
+    ])
+      .then(([palette]) => {
+        console.log(palette);
+      })
+      .catch((reason: Error) => {
+        console.error(reason.message);
+      });
```
ajbozarth, line 195:
```suggestion
    const palette = await RequestHandler.makeGetRequest('elyra/pipeline/config/palette', false);
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -163,33 +165,60 @@ export class PipelineEditor extends React.Component<
...
+    Promise.resolve([
+      RequestHandler.makeGetRequest('elyra/pipeline/config/properties', false)
+    ])
+      .then(([properties]) => {
+        console.log(properties);
+        this.properties = properties;
+      })
+      .catch((reason: Error) => {
+        console.error(reason.message);
+      });
```
ajbozarth, line 218:
```suggestion
    this.properties = await RequestHandler.makeGetRequest('elyra/pipeline/config/properties', false);
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: note that the key here is `await`, this tells the code to wait for the Promise to resolve then set `palette` to its result.

lresende:
> As for moving properties.json to an api call, I don't agree with this. properties.json is not a fully functional template or schema. It is just a partially filled out instance of a Common Properties arg. As such I don't believe it is something that should be hosted by the backend. I am open to arguments against my option here, I just don't see it as a valid case at this time.

For this, could we have the discussion on the issue #454 instead of hijacking the pr?

ajbozarth:
@lresende sure, I'll recommend there

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: I don't think you can do await in the constructor because it needs to be in an async method.
Maybe you can move the suggestions above to componentDidMount(), where
```
async componentDidMount(): Promise<void> {
  // request palette and properties config using await
}
```

ajbozarth: ah good point @karlaspuldaro , @lresende had to do that with an async call in #662

elyra/__init__.py
```diff
@@ -46,5 +46,7 @@ def load_jupyter_server_extension(nb_server_app):
         (url_path_join(web_app.settings['base_url'], r'/elyra/namespace'), NamespaceHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/schedule'), PipelineSchedulerHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/export'), PipelineExportHandler),
+        (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/config/%s' % (resource_regex)),
```
kevin-bates, line 49: I think I'd prefer an endpoint like `elyra/component` or `elyra/pipeline/component` rather than `config`.

I think we should return a dictionary of two items/keys:  'palette': palette.json, and 'properties': {dict of properties.json per node-type}.
```JSON
{
    "response": {
        "palette": {
            palette.json
        },
        "properties": {
            "notebookOp": {
                notebook-properties.json
            },
            "pythonScriptOp": {
                python-script-properties.json
            }
        }
    }
}
```

lresende:
Closing stale pull request. Please rebase and reopen if you are still planning to work on this issue.

PR #656, opened 2020-06-17
Multiple pipeline editor updates

lresende:
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #662, opened 2020-06-18
Add version capabilities to Pipeline

### LINKED ISSUE #647: Add support for pipeline version
Add Elyra version to pipeline file
Make Pipeline editor to support different versions of the pipeline with auto-upgrade

lresende:
Add a numeric version field to the Pipeline json file
and enable in-place migration when reading an old
pipeline version
    
Fixes #647

Note: to be merged after #656
 
           in the file; or

           it.


lresende:
I believe the one open question I have is regarding `in-place auto update` versus `ask for confirmation before update`.

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -124,4 +129,54 @@ export class PipelineService {
       buttons: [Dialog.okButton()]
     });
   }
+
+  static convertPipeline(pipelineDefinition: any): any {
+    const pipelineJSON = JSON.parse(JSON.stringify(pipelineDefinition));
+
+    if (!Utils.hasPipelineAppdataField(pipelineJSON.pipelines[0], 'version')) {
```
kevin-bates, line 136: I recommend this be method that gets the version from the pipeline.  In this case, that method will return an integer of 0.

I would then move the contents of this block to a method named something like `convertV0ToV1()`.  This way, when version 2 comes along, this code would resemble something like (using Python):

```python
version = getPipelineVersion(pipelineJSON.pipelines[0])
if version < 1:
    pipelineJSON = convertV0toV1(pipelineJSON)
if version < 2:
    pipelineJSON = convertV1toV2(pipelineJSON)   
if version < 3:
    pipelineJSON = convertV2toV3(pipelineJSON)   
```
and each of the "convert" methods only needs to address the specific changes for that particular transition.

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -124,4 +129,54 @@ export class PipelineService {
...
+      // original pipeline definition without a version
+      console.log(
+        'Migrating pipeline to version: ' + this.CURRENT_PIPELINE_VERSION
```
kevin-bates, line 139: If taking the previous recommendation, this message should indicate both the from and to versions.

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -124,4 +129,54 @@ export class PipelineService {
...
+        Utils.renamePipelineAppdataField(
+          node,
+          'recursive_dependencies',
+          'include_subdirectories'
+        );
+      }
+
+      pipelineJSON.pipelines[0]['app_data'][
+        'version'
+      ] = this.CURRENT_PIPELINE_VERSION;
```
kevin-bates, line 177: And this would simply be the 'to' version, not necessarily the final version.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -173,7 +173,10 @@ export class PipelineEditor extends React.Component<
     this.canvasController.setPipelineFlowPalette(palette);
     this.widgetContext = props.widgetContext;
     this.widgetContext.ready.then(() => {
-      this.canvasController.setPipelineFlow(this.widgetContext.model.toJSON());
+      const pipelineFlow = PipelineService.convertPipeline(
```
kevin-bates, line 176: Kinda thinking this should just be called `openPipeline()` and let this method determine if conversion should occur or not.  Otherwise, it seems strange that the very first thing done on a pipeline is to convert when 99% of the time, nothing will be converted.

packages/pipeline-editor/src/PipelineService.tsx
kevin-bates: Should the pipeline file get saved if conversions are performed?

kevin-bates: After hearing the discussion about updates and warning for updates I'm okay with whatever approach.  I guess I'd prefer to not get prompted, but since this is an automated update, it makes sense to prompt.  If we're going to prompt at all, I think it should be to confirm/approve the update vs a warning/info dialog that their pipeline was updated.

lresende: for the save, i would prefer the user to save it, which gives a way for a real person validation.

lresende:
With the recent updates, we now ask for user confirmation before migrating the pipeline

image

The only remaining issue is to close the "tab" if user cancel the migration.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -561,6 +561,63 @@ export class PipelineEditor extends React.Component<
...
+              </p>
+            ),
+            buttons: [Dialog.okButton()]
+          });
+          return;
+        } else {
+          // in this case, pipeline was last edited in a "old" version of Elyra and
+          // it needs to be updated/migrated.
+          showDialog({
+            title: 'Migrate pipeline ?',
```
kevin-bates, line 591: Space before question mark.
```suggestion
            title: 'Migrate pipeline?',
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -561,6 +561,63 @@ export class PipelineEditor extends React.Component<
...
+            body: (
+              <p>
+                The current pipeline was created/edited in an older version of
+                Elyra and needs to be migrated.
+                <br />
+                Note that the current file will only be overridden upon being
+                saved on the pipeline on the editor.
+                <br />
+                Proceed with migration?
```
kevin-bates, line 600: How about...

_This pipeline corresponds to an older version of Elyra and needs to be migrated.  
Although the pipeline can be further edited and/or submitted after its update, the migration will not be completed until the pipeline has been saved within the editor._

_Proceed with migration?_

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -561,6 +561,71 @@ export class PipelineEditor extends React.Component<
...
+                <br />
+                <br />
+                Proceed with migration?
+              </p>
+            ),
+            buttons: [Dialog.cancelButton(), Dialog.okButton()]
+          }).then(result => {
+            if (result.button.accept) {
+              // proceed with migration
+              pipelineJson = PipelineService.convertPipeline(pipelineJson);
```
kevin-bates, line 613: Should we pass the pipeline version to the convert method and avoid having to look it up again?
```suggestion
              pipelineJson = PipelineService.convertPipeline(pipelineJson, pipelineVersion);
```

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -124,4 +127,51 @@ export class PipelineService {
       buttons: [Dialog.okButton()]
     });
   }
+
+  static convertPipeline(pipelineDefinition: any): any {
```
kevin-bates, line 131: Per previous comment. 
```suggestion
  static convertPipeline(pipelineDefinition: any, currentVersion: number): any {
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -561,6 +561,71 @@ export class PipelineEditor extends React.Component<
...
+            body: (
+              <p>
+                This pipeline corresponds to a more recent version of Elyra{' '}
+                <br />
+                and cannot be used until Elyra has been upgraded.
+              </p>
+            ),
+            buttons: [Dialog.okButton()]
+          });
+          if (this.app.shell.currentWidget) {
```
karlaspuldaro, line 585: We could make this a pipeline command `closePipelineEditor` to make it consistent with the other ones, then just call `this.app.commands.execute(commandIDs.closePipelineEditor);`.
Same for line 616

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -124,4 +127,51 @@ export class PipelineService {
...
+    const pipelineJSON = JSON.parse(JSON.stringify(pipelineDefinition));
+
+    const currentVersion: number =
+      +Utils.hasPipelineAppdataField(pipelineJSON.pipelines[0], 'version') || 0;
+    if (currentVersion < 1) {
+      // original pipeline definition without a version
+      console.log('Migrating pipeline to version 1');
```
kevin-bates, line 138: We should move this outside the set of if statements to avoid multiple log statements and just provide the destination (or simply 'current version').
```suggestion
 console.log('Migrating pipeline to the current version.');
```
If this is more for debug and you _do_ want this logged on each step, then the log statement should be moved into the respective step's method.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
lresende: I would prefer not to do this now, as the consumer not necessarily needs to know how the fields are handled. Having said that, we probably need a "pipeline utility class" to hide a lot of these details from the consumers.

kevin-bates: I'm confused by what you mean by the "consumer" here.  This code is already extracting the version info from the pipeline to determine that conversion is necessary in the first place.

Are you suggesting that pipelines might be "migrated" from another location and call `convertPipeline()` unconditionally?

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth, line 131: why are these functions in the service file and not the util file? They have nothing to do with the elyra api

packages/pipeline-editor/src/utils.ts
```diff
@@ -26,6 +26,9 @@ export default class Utils {
     return uuid4();
   }
 
+  /*
+   *
+   */
```
ajbozarth, line 31: you added a comment block but no comment

packages/pipeline-editor/src/utils.ts
```diff
@@ -52,4 +55,56 @@ export default class Utils {
 
     return template;
   }
+
+  /*
```
ajbozarth, line 59: If this is meant to be a doc comment its missing the `**` here. If it's not meant to be a doc comment then you should remove this blank comment line.

packages/pipeline-editor/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../tsconfig-base",
   "compilerOptions": {
+    "module": "commonjs",
     "outDir": "lib",
     "rootDir": "src",
-    "module": "commonjs",
+    "suppressImplicitAnyIndexErrors": true,
```
ajbozarth, line 7: why are we suppressing the error checks?

lresende:
The most recent dialogs:

Migration to recent version is needed:

image

Newer version found, Elyra needs to be upgraded:

image

packages/pipeline-editor/src/PipelineService.tsx
lresende: updated to say `Migrating pipeline to the current version.`

packages/pipeline-editor/src/PipelineEditorWidget.tsx
lresende: I was on the fence over this, as this is really only handling the "read and display" of the pipeline file, which to me was load, while open was more the whole opening widget thing. But at the end I have renamed it.

packages/pipeline-editor/src/PipelineService.tsx
lresende: Services in this case are not tied to an API, it's more like a place where "operations" that work on a given "thing" (in this case pipeline) are centralized. Also, I don't like what I did with Utils either :) We need a "wrapper" to manipulate "pipelines" on the front end, but this will probably come later.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -575,6 +575,66 @@ export class PipelineEditor extends React.Component<
     });
   }
 
+  async handleOpenPipeline(): Promise<void> {
```
kevin-bates, line 578: I think there was some miscommunication here.  Originally, I was referring to the `convertPipeline()` method name because, at the time, there was no prompt to ask the user if they should convert, and since it was called unconditionally (with no indication a conversion might be performed), the name `convertPipeline()` seemed better suited to be called `openPipeline()` (or even `loadPipeline()`).  

Since then you added the prompt so that `convertPipeline()` is ONLY called in cases where conversion is appropriate and desired.  And, in the current case, the name of the calling method doesn't need to be changed - so if you prefer `handleLoadPipeline()`, that makes perfect sense.  At this point, it's more about whether you want to change it back or not.  I just wanted to share that the context of my original comment was invalidated once the prompting logic was added.

Sorry about that.  That's a case where waiting on reviewing is best.

lresende: No worries, as we also added `handleClosePipeline` i think the `handleOpenPipeline` is ok :) these are no public APIs, so we can always change our minds :)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -575,6 +575,66 @@ export class PipelineEditor extends React.Component<
...
+        // pipeline version and current version are divergent
+        if (pipelineVersion > PIPELINE_CURRENT_VERSION) {
+          // in this case, pipeline was last edited in a "more recent release" and
+          // the user should update his version of Elyra to consume the pipeline
+          showDialog({
+            title: 'Load pipeline failed!',
+            body: (
+              <p>
+                This pipeline corresponds to a more recent version of Elyra{' '}
+                <br />
```
vabarbosa, line 592: any reason for the <br /> mid sentence? depending on window/dialog size the formatting is off:

image

(same for the **Migrate pipeline?** dialog below (line 612))

packages/pipeline-editor/src/utils.ts
```diff
@@ -52,4 +55,69 @@ export default class Utils {
...
+  }
+
+  /**
+   * Check if an application specific field from the pipeline defintion exists
+   * (e.g. pipelines[0][app_data][fieldName])
+   */
+  static hasPipelineAppdataField(node: any, fieldName: string): boolean {
+    let isPresent = false;
+    if (node['app_data']) {
+      if (node['app_data'][fieldName]) {
```
vabarbosa, line 91: if the `fieldName` is present but has a value of `0`, `false`, or an empty string, this function may not behave as intended. although, the field is present this would return false.

depending on what you are trying to achieve a better approach is to check the just the existence of a field regardless of it's value would be:

```javascript
  return Object.prototype.hasOwnProperty.call(node, 'app_data') &&
    Object.prototype.hasOwnProperty.call(node['app_data'], fieldName);
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
lresende: I guess it helps with a better visualization with the regular window size, but not much if you try to resize or do anything with the dialog. I will remove them.

Issue #659, opened 2020-06-17
pipelines store notebook path information that causes issues


ptitzler:
To recreate:
- Assume JupyterLab is launched in `/mydir`. `mydir` therefore is the working directory, which is empty. 
- Create a subdirectory in `mydir` named `mydir2` (`/mydir/mydir2`)
- Create a notebook `mynotebook` in `/mydir/mydir2`
- Create a pipeline `mypipeline` in `/mydir/mydir2`
- Add notebook `mynotebook` to the pipeline
- Save the pipeline
- Inspect the saved pipeline file. 

You'll notice that the notebook's `filename` is stored as `mydir2/mynotebook.ipynb`

```
[
        {
          "id": "424905cf-aded-438b-a357-5b9572efc847",
          "type": "execution_node",
          "op": "execute-notebook-node",
          "app_data": {
            "filename": "mydir2/mynotebook.ipynb",
          ...
```

As a result the pipeline is not portable:
- Quit JupyterLab
- Copy the pipeline file and the notebook file (which reside in the same directory) to a different directory, e.g. `/mydir3`.
- Open Jupyterlab
- Open the pipeline. It is now unusable because it references a notebook in the `mydir2` directory that does not exist.

ptitzler:
I do believe this is potentially a  _major_ adoption issue. I just tried to use a pipeline that another user created and couldn't do it without completely rebuilding it.

ptitzler:
The root cause of this issue is that the notebook path needs to be relative to the location of the pipeline file, not relative to the location it was stored in when it was created.

As is, it's very easy for a user to break things by simply changing the working directory:
- launch JL in `/dir/dir2`
- create a notebook and a pipeline in ` /dir/dir2`
- quit JL
- cd to `/dir`
- Launch JL in `/dir`
- in the file browser navigate to the pipeline and open it
- the notebook references (and likely other references too) are no longer valid

kevin-bates:
Only if you stay within the same root (notebook) directory between runs does a _moved_ pipeline (within the root) still work.  This is because paths are relative to the root-dir (or notebook-dir).

[...]

PR #664, opened 2020-06-18
Show user friendly names in tooltip

### LINKED ISSUE #657: [Fit and finish] Notebook "tooltip": display user friendly property names
If a user hovers over a notebook node in the editor the node's properties are displayed as follows:

image

Even though a user should not have a problem interpreting the currently displayed property names (which look like variables) it would be nice if the labels could match the labels shown in the properties view to make it look less debug-gy.

image

vabarbosa:
Display friendly name for property fields shown in pipeline node hover tooltip

Fixes #657 

FYI, using the default label from the properties info displays the full label (same as it appears in the properties dialog):

image

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -303,7 +303,7 @@ export class PipelineEditor extends React.Component<
         'runtime_image.' + runtimeImage + '.label'
       ] = runtimeImages[runtimeImage];
     }
-    properties.parameters[0].enum = imageEnum;
+    properties.parameters[1].enum = imageEnum;
 
     this.propertiesInfo = {
       parameterDef: properties,
@@ -319,6 +319,7 @@ export class PipelineEditor extends React.Component<
     const node_props = this.propertiesInfo;
     node_props.appData.id = node_id;
 
+    node_props.parameterDef.current_parameters.filename = app_data.filename;
     node_props.parameterDef.current_parameters.runtime_image =
       app_data.runtime_image;
     node_props.parameterDef.current_parameters.outputs = app_data.outputs;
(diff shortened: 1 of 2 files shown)


vabarbosa:
@ptitzler you thinking of something like this for #657?

kevin-bates:
@vabarbosa - this looks great, although seeing the one `filename` field in lowercase makes me inclined to suggest we should add the notebook filename to the dialog as a read-only Filename property, just so this tip has a consistent look. 

I think the user seeing the name of the notebook file might actually be helpful.

vabarbosa:
> @vabarbosa - this looks great, although seeing the one `filename` field in lowercase makes me inclined to suggest we should add the notebook filename to the dialog as a read-only Filename property, just so this tip has a consistent look.
> 
> I think the user seeing the name of the notebook file might actually be helpful.

@kevin-bates yeah, that bothers me as well, but i wanted to be keep the code generic so if the `app_data` property is not in the properties info i leave i left as is (for now). thinking about it for cases like this, if we want the function to be generic then we have these options:

1. show the field as is
1. not show the field if it is not in the properties info
1. add the field to the properties info (with proper label)
1. apply some basic formatting (i.e., uppercase first letter, replace `_` with spaces) the field

not sure what people's preferences are

kevin-bates:
I totally agree, so I'd be okay with options 2 or 3, with a preference toward 3, since I think it adds value.  But yes, adhering to the principle that everything is driven from the properties info is first priority.

ajbozarth:
@vabarbosa @kevin-bates I think this runs into what we discussed about addressing #659 specifically how we want to make "filename" updatable. So my suggestion would be to add it to commprops. I also think we should only display values in commprops anyway. So I would say go with option 2 AND 3

ptitzler:
> @ptitzler you thinking of something like this for #657?
Yep 👍 , except 
- label capitalization should be consistent (e.g. `Include Subdirectories in Dependencies to make it consistent across all labels`)
- no period at the end of label 2 and 4 because these are not valid sentences

ajbozarth:
@ptitzler good catch, those grammar error should be fixed for commprops anyway

vabarbosa:
Filename has been added to the properties and displayed as read only.

image

image

vabarbosa:
> This looks great Va - thank you.
> 
> Per @ajbozarth's [previous comment](https://github.com/elyra-ai/elyra/pull/664#issuecomment-646337717), `filename` may need to be editable although I'm not certain that gets us out of the woods with #659.

agreed. i left it read-only for now. being able to edit filename i think would be better addressed in a separate PR.

PR #666, opened 2020-06-19
Fix python editor output tab to be static

### LINKED ISSUE #635: Python code editor - console output "tab" should be static
If one runs a Python script the output is displayed in a "tab" on the canvas. This console log tab appears to be not a tab in the traditional sense that a user is supposed to move around. Currently one can move the tab, and if one does, the content is lost.

karlaspuldaro:
Fixes #635
 
           in the file; or

           it.

### DIFF
--- a/packages/python-runner/src/PythonFileEditor.tsx
+++ b/packages/python-runner/src/PythonFileEditor.tsx
@@ -44,7 +44,7 @@ import {
   stopIcon,
   TabBarSvg
 } from '@jupyterlab/ui-components';
-import { BoxLayout, DockPanel, PanelLayout, Widget } from '@lumino/widgets';
+import { BoxLayout, PanelLayout, Widget } from '@lumino/widgets';
 
 import { KernelDropdown } from './KernelDropdown';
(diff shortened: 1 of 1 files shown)


packages/python-runner/src/PythonFileEditor.tsx
```diff
@@ -131,7 +131,8 @@ export class PythonFileEditor extends DocumentWidget<
    */
   private createOutputAreaWidget = (): void => {
     // Add dockpanel wrapper for output area
-    this.dockPanel = new DockPanelSvg();
+    const dockPanelOptions: DockPanel.IOptions = { tabsMovable: false };
+    this.dockPanel = new DockPanelSvg(dockPanelOptions);
```
ajbozarth, line 135: is there a reason we can't just do this "inline" like below and avoid an extra import?
```suggestion
    this.dockPanel = new DockPanelSvg({ tabsMovable: false });
```

karlaspuldaro: I guess typescript got me used to define a type for everything 😄

PR #684, opened 2020-06-24
Validate pipeline form dialogs

### LINKED ISSUE #658: Usability: pipeline export should not be attempted if required parameters are missing
A user should not be allowed to export a pipeline if a required export setting is not provided.
Steps:

- Create a pipeline
- Add a notebook
- Click "Export Pipeline"
- **Don't** select a runtime configuration

image

- Click "OK".  The export operation fails.

Expected behavior: the user cannot click "OK" (initiate the export).

karlaspuldaro:
Fixes #658 

Exporting or submitting a pipeline now requires the user to populate all required form fields.
Required fields in dialog body should contain attribute `data-form-required`.
While no selection is made or required inputs are empty, the OK button remains disabled.
image

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -28,7 +28,8 @@ import {
   dragDropIcon,
   exportPipelineIcon,
   pipelineIcon,
-  savePipelineIcon
+  savePipelineIcon,
+  showFormDialog
 } from '@elyra/ui-components';
 
 import { JupyterFrontEnd } from '@jupyterlab/application';
@@ -65,7 +66,6 @@ import Utils from './utils';
 
 const PIPELINE_CLASS = 'elyra-PipelineEditor';
 const NODE_TOOLTIP_CLASS = 'elyra-PipelineNodeTooltip';
-const OK_BUTTON_CLASS = 'elyra-PipelineOkButton';
 
 const TIP_TYPE_NODE = 'tipTypeNode';
 const PIPELINE_CURRENT_VERSION = 1;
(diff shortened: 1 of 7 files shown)


ajbozarth:
I like this as a POC, but if we're going to add this functionality I'd rather we "wrap" showDialog in our own "utility" function that can handle this validation more generally based on supplied params.

lresende:
Can we also make sure Submit dialog does not have the same problem.

vabarbosa:
> I like this as a POC, but if we're going to add this functionality I'd rather we "wrap" showDialog in our own "utility" function that can handle this validation more generally based on supplied params.

i may be misunderstanding what you are asking, but we can't _"wrap" `showDialog` in our own "utility" function that can handle this validation_ because `showDialog` returns a promise that get resolved only after the user has clicked on a dialog button. at which time it would be too late to prevent (or control) any action by the user.

what could perhaps be done is take the approach used here and create a more generic function that is reusable by some of the other dialogs prompts in Elyra (and would replace Jupyter's default `showDialog`).

would you know if this is a functionality [JupyterLab would welcome](https://github.com/elyra-ai/elyra/issues/658#issuecomment-649080213)? in which case it be added to [JupyterLab's Dialog](https://github.com/jupyterlab/jupyterlab/blob/b58c2c545586cd8ed33bb2e1b40e1823e61879cf/packages/apputils/src/dialog.ts) instead.

ajbozarth:
> what could perhaps be done is take the approach used here and create a more generic function that is reusable by some of the other dialogs prompts in Elyra (and would replace using JupyterLab's default showDialog).

Sorry for the poor clarity, this is exactly what I meant

> would you know if this is a functionality JupyterLab would welcome? in which case it be added to JupyterLab's Dialog instead.

This is a good question, we should ask, this would be cleaner if done in lab.

karlaspuldaro:
As per conversation with @vabarbosa we came up with a generic `showFormDialog` method that would live in `packages/ui-compoments`, that could be used by any elyra component. 
The idea it to provide:
1. A generic validation, where form fields marked as **_required_** would be considered valid when populated
2. A way to handle more specific validations would be passed as a function ,so that the caller could do other things like checking for a proper string format in an input field.
Below is a prototype of how the utility would look like (initially it would only have the validate-and-show-dialog function):
```
  /*
   * Validate required form dialog fields upon display
   *
   * options - The dialog setup options
   * formValidationFunc - Optional validation function passed by the caller to provide their own logic
   *
   */
  async showFormDialog(options, formValidationFunc) {
  const dialogBody = options.body;
  const dialog = new Dialog(options);
  if (formValidationFunc && typeof formValidationFunc === 'function') {

    // pass dialog widget to the caller for more specific logic and other dialog behaviors,
    // such as being able to disable other buttons that aren't the default one
    formValidationFunc(dialog);
  } else {

    // expect required fields to contain attribute: data-elyra-required
    const requiredFields = dialogBody.node.querySelectorAll('[data-elyra-required]');
    if (requiredFields.length > 0) {

      // disable defaultButton as default
      // add appropriate change event listener to each required field
      // that enables the defaultButton when field is populated
    }
  }
  return dialog.launch();
}
```
How would others feel about this approach?

ajbozarth:
@karlaspuldaro that approach looks great to me, looking forward to seeing the detailed implementation

lresende:
Maybe don't use 'elyra` on the `data-elyra-required`, but otherwise look good

karlaspuldaro:
NOTE: The Dialog widget will skip any validation upon pressing 'enter' key (needs to resolve in lab)

The generic validation implemented only checks if all required fields are populated. In the case of pipeline `name` input field, we check for a string not null, while a more detailed string validation is already handled by the backend.

packages/ui-components/src/FormDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+                ? fieldsValidated.add(element)
+                : fieldsValidated.delete(element);
+
+              // Only enable defaultButton if all required fields are validated
+              fieldsValidated.size === requiredFields.length
+                ? enableDialogButton(defaultButton)
+                : disableDialogButton(defaultButton);
+            });
+          } else if (fieldType === 'input' || fieldType === 'textarea') {
+            element.addEventListener('keyup', (event: any) => {
```
vabarbosa, line 84: jupyterlab listens to `keydown` events. it may be a good idea to do the same.

packages/ui-components/src/FormDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+          .querySelector('.jp-Dialog-footer')
+          .getElementsByTagName('button')[defaultButtonIndex];
+
+        disableDialogButton(defaultButton);
+
+        requiredFields.forEach((element: any) => {
+          // First deal with the case the field has already been pre-populated
+          element.value
+            ? fieldsValidated.add(element)
+            : fieldsValidated.delete(element);
```
vabarbosa, line 67: since this code block is being called a couple times, i think it was good the way you had it before as its own function. also maybe use `element.value.trim()` here as well.

packages/ui-components/src/FormDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+            element.addEventListener('change', (event: any) => {
+              // Update fieldsValidated set accordingly
+              event.target.value
+                ? fieldsValidated.add(element)
+                : fieldsValidated.delete(element);
+
+              // Only enable defaultButton if all required fields are validated
+              fieldsValidated.size === requiredFields.length
+                ? enableDialogButton(defaultButton)
+                : disableDialogButton(defaultButton);
```
vabarbosa, line 81: same here. since this code block is being called a couple times, i think it was good the way you had it before as its own function

packages/ui-components/src/FormDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import { Dialog } from '@jupyterlab/apputils';
+import { Widget } from '@lumino/widgets';
+
+const DEFAULT_BUTTON_CLASS = 'elyra-DialogDefaultButton';
+/*
+ * Validate required dialog fields upon display
+ * - Provides a generic validation by checking if required form fields are populated
+ * - Expect required fields in dialog body to contain attribute: data-form-required
+ * - NOTE: The Dialog widget will skip any validation upon pressing 'enter' key and resolve the dialog as is
```
vabarbosa, line 27: you may be able to capture the `keydown` event and prevent propagation/default behavior if **Enter** key is pressed and all required are not filled

vabarbosa:
> NOTE: The Dialog widget will skip any validation upon pressing 'enter' key (needs to resolve in lab)
> 

for this i think you can listen to the `keydown` event and prevent default behavior/propagation if **Enter** key is pressed and validation is not satisfied

packages/ui-components/src/FormDialog.tsx
karlaspuldaro: For some reason I can't seem to get consistent values using `keydown`. I can get a valid key code on the event, but the element value hasn't been updated yet.

packages/ui-components/src/FormDialog.tsx
karlaspuldaro: Thank you for the directions, overriding the dialog eventHandler worked for this case :)

ajbozarth:
I'll be taking a look at this later today

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -534,45 +535,46 @@ export class PipelineEditor extends React.Component<
 
   async handleExportPipeline(): Promise<void> {
     const runtimes = await PipelineService.getRuntimes();
-
-    showDialog({
+    const dialogOptions: Partial<Dialog.IOptions<any>> = {
       title: 'Export pipeline',
       body: new PipelineExportDialog({ runtimes }),
       buttons: [Dialog.cancelButton(), Dialog.okButton()],
+      defaultButton: 1,
```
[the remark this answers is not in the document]
karlaspuldaro: Yeah I also don't like that, but that's how [Dialog Widget](https://github.com/jupyterlab/jupyterlab/blob/f0153e0258b32674c9aec106383ddf7b618cebab/packages/apputils/src/dialog.ts#L790) sets its default button. The idea was to get it from the widget but defaultButton is private, so I just recreated what they do.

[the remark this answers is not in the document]
ajbozarth: ick, oh well, then this is fine :/

packages/ui-components/src/FormDialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+        const fieldsValidated = new Set();
+
+        // Override Dialog.handleEvent to prevent user from bypassing validation upon pressing the 'Enter' key
+        const dialogHandleEvent = dialog.handleEvent;
+        dialog.handleEvent = (event: Event): void => {
+          if (
+            event instanceof KeyboardEvent &&
+            event.type === 'keydown' &&
+            event.keyCode === 13 &&
+            fieldsValidated.size !== requiredFields.length
```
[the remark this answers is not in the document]
karlaspuldaro: Not always since this is a callback for anytime a keyboard button is pressed on the dialog, `fieldsValidated` could already have been populated at some point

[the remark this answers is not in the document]
ajbozarth: ah, whoops, this is in the event handler, my bad

PR #698, opened 2020-06-29
Properly set pipeline version on new pipelines

lresende:

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -55,6 +55,7 @@ import * as React from 'react';
 
 import { IntlProvider } from 'react-intl';
 
+import { PIPELINE_CURRENT_VERSION } from './constants';
 import * as i18nData from './en.json';
 import * as palette from './palette.json';
 import { PipelineExportDialog } from './PipelineExportDialog';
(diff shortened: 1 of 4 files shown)


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -578,58 +578,68 @@ export class PipelineEditor extends React.Component<
...
-              this.canvasController.setPipelineFlow(pipelineJson);
-            } else {
-              this.handleClosePipeline();
-            }
-          });
+      if (pipelineJson == null) {
+        // creating new pipeline
+        pipelineJson = this.canvasController.getPipelineFlow();
+        if (Utils.isNewPipeline(pipelineJson)) {
+          pipelineJson.pipelines[0]['app_data']['version'] = 1;
```
kevin-bates, line 585: I expected this to just be baked into the template but I see that's just for creating single-node pipelines for batch notebook processing.  Seems like we should be consistent and add a `'version': 1` field to the template as well.

kevin-bates, line 585: We should use the constant here...
```suggestion
          pipelineJson.pipelines[0]['app_data']['version'] = PIPELINE_CURRENT_VERSION;
```

lresende: What do you mean by template?

kevin-bates: https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/pipeline-template.json

PR #699, opened 2020-06-29
Add http:// to user_guide/runtime-conf.md

ca-scribner:
Current example does not work.  Minio endpoint requires http://, otherwise you get an invalid endpoint error.

Sorry didn't have a chance to check development outlines, etc.  Hopefully with such a minor change nobody minds (or they implement it however they'd prefer) :)
 
           in the file; or

           it.


docs/source/user_guide/runtime-conf.md
```diff
@@ -35,7 +35,7 @@ elyra-metadata install runtimes --schema_name=kfp \
        --name=my_kfp \
        --display_name="My Kubeflow Pipeline Runtime" \
        --api_endpoint=https://kubernetes-service.ibm.com/pipeline \
-       --cos_endpoint=minio-service.kubeflow:9000 \
+       --cos_endpoint=http://minio-service.kubeflow:9000 \
```
lresende, line 38: could you please also update the "result json" right below.

docs/source/user_guide/runtime-conf.md
```diff
@@ -35,7 +35,7 @@ elyra-metadata install runtimes --schema_name=kfp \
        --name=my_kfp \
        --display_name="My Kubeflow Pipeline Runtime" \
        --api_endpoint=https://kubernetes-service.ibm.com/pipeline \
```
lresende, line 37: Looking into [recipes/deploying-kubeflow-locally-for-dev.md](https://elyra.readthedocs.io/en/latest/recipes/deploying-kubeflow-locally-for-dev.html), I was wondering if we should use the same pattern here as if you were deploying it locally and use `--api_endpoint=http://localhost:31380/pipeline`

ca-scribner:
oh yeah sorry didn't even read the json output in the next paragraph.  whoops.

docs/source/user_guide/runtime-conf.md
ca-scribner: That makes sense to me, but I can't say if localhost:31380 actually works (I have not run through the other examples - just using an existing kfp).  It does sound a lot more likely to work in an example than `kubernetes-service.ibm.com` though

PR #702, opened 2020-06-30
Improve dependency management and archive generation

### LINKED ISSUE #204: Pipeline submission should fail and inform user if file dependency is not found
- When a user adds a file dependency A in common properties for a component, the pipeline editor should include this file in the tar archive. 
- Currently when a user adds a non-existent file, the archive creation succeeds. This should not be the case.
- We should bubble up an error message when a file dependency is not found prior to pipeline submission

### LINKED ISSUE #572: Fix subdirectories checkbox when used in conjunction with wildcards
When does include subdirectories apply? Disable sub dir checkbox in certain instances?

kevin-bates:
These changes improve how dependencies are handled, generally eliminating the need to include subdirectories when adding dependencies from subdirectories.

- Allow for `'*'`, `'?'`, and `'[a-z]'` / `'[!a-z]'` (sequence) wildcards.
- Allow for embedded wildcards, including directory names (e.g., `inputs/ACS*` - to include all directories beginning with ACS in subdirectory `inputs`)
- Enforce that all dependencies must be matched (i.e., present in resulting archive)
- Updated tests to include additional wildcard testing
- Added timing metrics around pipeline compilation, archive creation, and upload - logged as DEBUG statements
- Minor refactoring to help with readability
- Once nodes are updated to not include subdirectories but only include what they need, we see ~70% improvement in the submission time of the COVID pipeline - 80 secs to 25 secs.  With no changes to the pipeline, the submission goes from 80 secs to 50 secs.

## Behavioral Changes

1. With these changes, it is no longer the case that **selecting 'include subdirectories' with NO files listed as dependencies results in all files being included!** Instead, no files will be included in that case.  One must add a dependency of '*' to have all files included.  If include-subdirectories is also selected, then all files in the folder hierarchy will be added.  This treats empty dependency lists the same (and expected) way.
1. As noted above, unmatched dependencies will now fail a pipeline's submission for run and export.

Fixes #204 
Fixes #572 
 
           in the file; or

           it.

### DIFF
--- a/elyra/util/archive.py
+++ b/elyra/util/archive.py
@@ -104,7 +104,7 @@ def create_temp_archive(archive_name, source_dir, filenames=None, recursive=Fals
             return None
 
     # If there's a '*' - less things to check.
-    include_all =  filenames and len(set([WILDCARDS[0]]) & set(filenames)) > 0
+    include_all = filenames and len(set([WILDCARDS[0]]) & set(filenames)) > 0
 
     processed_filenames = []
     temp_dir = create_project_temp_dir()
--- a/elyra/util/archive.py
+++ b/elyra/util/archive.py
@@ -73,7 +73,7 @@ def create_temp_archive(archive_name, source_dir, filenames=None, recursive=Fals
             # allow if found - except if a single '*' is listed (i.e., include_all) in
             # which case we don't want to add this directory since recursive is False.
             # This occurs with filenames like `data/util.py` or `data/*.py`.
-            elif not include_all and directory_in_list(tarinfo.name, filenames):
+            elif not include_all and directory_in_list(tarinfo.name, filenames_set):
                 return tarinfo
             return None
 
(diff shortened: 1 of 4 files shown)


elyra/util/archive.py
```diff
@@ -49,37 +69,58 @@ def tar_filter(tarinfo):
...
     temp_dir = create_project_temp_dir()
     archive = os.path.join(temp_dir, archive_name)
 
     with tarfile.open(archive, "w:gz") as tar:
         tar.add(source_dir, arcname="", filter=tar_filter)
 
-    if not archive:
-        raise RuntimeError('Internal error creating archive: {}'.format(archive_name))
+    if require_complete and not include_all:
+        # convert filenames and matched_filenames to sets and ensure they're the same.
```
karlaspuldaro, line 120: why not creating sets for both from the start instead of initiating them as arrays?

kevin-bates: Good question @karlaspuldaro.
Sets are immutable, so because `matched_filenames` gets built up during the processing, it can't be a `set` until we're done. `filenames`, on the other hand, comes into the method as a `list`, but I don't see it getting altered.  If python `sets` behave like `lists` in terms of iterables, then, yeah, `filenames` should be converted to a `set` first thing.
The other thing converting a `list` to a `set` does is remove duplicates - which I can't think of any reason why we wouldn't also want that.  Good idea.  I'll look at converting `filenames` to a `set` right away.  Thank you.

kevin-bates: Awesome, I've pushed the update.  It wound up simplifying indentation as well.  Thank you for the suggestion!
I'm going to mark this conversation as resolved.

kevin-bates: Turns out I was mistaken - Python sets are mutable - only `frozenset` is immutable.  Thank you to @karlaspuldaro for digging deeper into things!  I've just pushed a change that eliminates the need for conversion - thanks again Karla!

Issue #683, opened 2020-06-24
Add code snippet icon instead of '</>' in the code snippet header


marthacryan:
The code snippet icon is used for the "tab" in the left side panel, but in the actual widget, the icon isn't used; it is hard-coded to use the string '</>'. (see below)
![image](https://user-images.githubusercontent.com/6673460/85624820-1aa61980-b630-11ea-86db-ccff98a48af8.png)
It'd be more consistent to add in the icon instead.

PR #706, opened 2020-07-01
Add support for JupyterLab Dark theme

### LINKED ISSUE #625: Update Elyra to support Dark Mode
Currently the only elyra extension that supports dark mode is python editor since it extends the functionality of file editor.

We should add dark mode support for:
- All icons (including the elyra logo)
- code snippet (and any future side bar widgets)
- pipeline editor

This issue was split off from #127 

If new developers are interested in tackling this before I get to it just post here, otherwise I'll post here once I actually start coding

vabarbosa:
Updated Elyra components to better support and align with JupyterLab Dark & Light themes

- [x] Code Snippet sidebar widget
- [x] Code Snippet icon (in sidebar)
- [x] Python Runner console output (scroll buttons)
- [x] Pipeline Editor icon (in launcher & filebrowser)
- [x] Elyra logo
- [x] Dialogs (ExpandableErrorDialog)
- [x] Pipeline Editor (i.e., @elyra/canvas).

Fixes #625 

image

image

image

image

image

 
           in the file; or

           it.


ajbozarth:
Took a quick look and this is looking great so far, I'll do a full review once I'm back on Monday.

lresende:
Does this also address #683

vabarbosa:
> Does this also address #683 ?

this updates how the code snippet icon is displayed (themed). it however, does not add the icon to the code snippet header (#683). i expected that to be its own PR but if you are OK with it i can include it in this PR as well. with this PR the icon properly displays in Light and Dark modes so once it does get included in the header it show properly show when switching themes.

lresende:
> > Does this also address #683 ?
> 
> this updates how the code snippet icon is displayed (themed). it however, does not add the icon to the code snippet header (#683). i expected that to be its own PR but if you are OK with it i can include it in this PR as well. with this PR the icon properly displays in Light and Dark modes so once it does get included in the header it show properly show when switching themes.

It's ok to be it's own pr, just that I saw some icon related changes and thought it was covering that.

marthacryan:
Just tried this locally and it worked great for me! Should we open an issue for pipeline editor / reach out to canvas about adding a dark mode setting?

ajbozarth:
I'll be taking a look at this later today

vabarbosa:
all components have been updated. see screenshots in the description. let me know if i missed any component or need to update anything further.

vabarbosa:
@ajbozarth all components have been updated. see screenshots in the description. the one remaining issue are some of the icons in the toolbar which are being displayed using `` instead of `<svg>` so colors cannot be updated via CSS

vabarbosa:
@marthacryan we may need to coordinate this with your metadata #589 PR to make sure the new pages/dialogs also support the Dark theme

vabarbosa:
> Just tried this locally and it worked great for me! Should we open an issue for pipeline editor / reach out to canvas about adding a dark mode setting?

@marthacryan thanks. i was able to work around the canvas support for dark mode and Pipeline Editor should now support switching between light and dark JupyterLab themes. however, the update could have been easier/quicker if canvas did have a support for switching (CSS) themes dynamically. it may be worth opening an issue with them for future consideration.

vabarbosa:
> Awesome! LGTM
> Just a comment, maybe "Run pipeline" icon should stay blue on dark mode to be consistent with the other buttons added on canvas toolbar from elyra?

@karlaspuldaro thanks. yeah, that is a discussion (https://github.com/elyra-ai/elyra/pull/706#issuecomment-654388918) i wanted to have. until we can update canvas (and use React components as icons) the simplest solution is to go with blue icons. the downside to that is the blue icon on black toolbar background may be difficult for some to see.

packages/theme/style/index.css
```diff
@@ -13,3 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+.lm-DockPanel {
+  background: var(--jp-layout-color0);
```
ajbozarth, line 18: what exactly is this css for? I'm unsure if we should be making any blanket changes to lumino css classes. Does the targeted class have an elyra specific class we could target instead?

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -46,10 +46,8 @@ import { toArray } from '@lumino/algorithm';
 import { IDragEvent } from '@lumino/dragdrop';
 
 import '@elyra/canvas/dist/common-canvas.min.css';
-
-import '@elyra/canvas/dist/common-canvas.min.css';
-
 import 'carbon-components/css/carbon-components.min.css';
+import '../style/canvas.css';
```
ajbozarth, line 50: I assume this change is because you're editing the file to support dark mode? Is there a way to override the css value in our own file rather than copying and editing a external library's one? If not could you annotate the diff in this PR detailing what you changed in the file?

vabarbosa:
> A couple comments, but this looks great over all.
> 
> As for the discussion around the canvas toolbar icon colors. I believe if we intend to update to latest canvas "soon" then leaving it blue for now (and opening an issue for the "know bug") is fine (at least its better than no dark mode). 

alright, will keep the toolbar icons as blue

> And iiuc canvas does not "actually" support dark mode, you just copied and overrode their css. If that is the case I believe you should open a PR on the canvas repo to include you dark mode support. Then, hopefully, we can get it into a release alongside the new jsx icon support that elyra can update to.

yes, canvas does not support a dark mode (or switching CSS themes dynamically). they do recommend importing the canvas `scss` files and overriding the variables to create your own theme: https://github.com/elyra-ai/canvas/wiki/5.0-Styling
however Elyra does not use Sass which means a Sass loader would need to be added to the build to process/compile the Sass to CSS and then import into JupyterLab. from [JupyterLab docs](https://jupyterlab.readthedocs.io/en/stable/developer/extension_dev.html?highlight=scss#themes):

**_Note that if you want to use SCSS, SASS, or LESS files, you must compile them to CSS and point JupyterLab to the CSS files._**

packages/theme/style/index.css
vabarbosa: i believe there may be a dark mode issue with JupyterLab (which i was going to open an issue in JupyterLab). since it it a simple enough, i included here in the meantime. this change removes the white background in the main dock panel when in dark mode:

before:

image

after:

image

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: this is not an external file from canvas. i created the new `canvas.css` file just to keep all the changes needed for canvas in a single document. i could add them into the existing `index.css` instead but i believe it is better left separate. and if canvas does change it default theme or adds support for dynamically switching themes, most of the work needed for Elyra would be contained in this file.

packages/theme/style/index.css
ajbozarth: ooooo, pretty, I like it. and you should definitely open a PR fixing this in Lab

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: Why are you removing `import '@elyra/canvas/dist/common-canvas.min.css';` above then. I had assumed that this replaced that since it was deleted. I agree though that keeping canvas.css separate from index.css is a good idea in this case.

vabarbosa:
> I assume this change is because you're editing the file to support dark mode? Is there a way to override the css value in our own file rather than copying and editing a external library's one? If not could you annotate the diff in this PR detailing what you changed in the file?

this is not an external file from canvas. i created the new canvas.css file just to keep all the changes needed for canvas in a single document. i could add them into the existing index.css instead but i believe it is better left separate. and if canvas does change it default theme or adds support for dynamically switching themes, most of the work needed for Elyra would be contained in this file.

the new `canvas.css` file added here does not change any of the values for the default canvas theme. rather it places all color values into CSS variables so they can easily be switched when changing themes. the file only contain color updates (i.e., `fill`, `stroke`, `color`, background`, `border`). the colors of the following components are updated in this file when switching themes. 

- comment box
- node
- toolbar
- properties modal
- tooltip
- context menu

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: the line:
```
import '@elyra/canvas/dist/common-canvas.min.css'
```
was actually included twice ([line 48 and 50](https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L48-L50)) so i removed the duplicate.

ajbozarth: oh >.< nice catch

vabarbosa:
> ooooo, pretty, I like it. and you should definitely open a PR fixing this in Lab

opened an issue in JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/8652

Issue #518, opened 2020-05-09
[Proposal] Meta-behaviors


kevin-bates:
**TL;DR:** This issue outlines a proposal for how we can apply system functionality across metadata instances by the use of system-defined meta-properties.  It does via a concerted example for how defaulting behaviors might be implemented.

NOTE: This proposal probably gets into implementation too much, but that is only to illustrate the kinds of considerations that must be taken into account.  It does not intend to impose a specific implementation.

# Introduction
On occasion, there is a need to be able to apply functionality or behavior across all instances (or their properties) of a given schema or namespace.  This kind of functionality is referred to as _Meta-behaviors_.  Some examples of meta-behaviors might be:

1. _Defaulting_, where a single instance among several is considered _the default_.  Only one instance, from the set of instances, can be considered _the default_.
2. _Data sensitivity_, where a given property is considered to hold sensitive data, that property's value should be encrypted or masked to prevent exposure of sensitive information.
3. _Uniqueness_, where a given property should be unique across the same property of other instances.  Today we "manually" impose uniqueness constraints on `display_name`, and let the filesystem impose this constraint on the `name` property since that property is used as the base name of the persisted instance.

This issue proposes how we might go about introducing meta-behaviors and uses _Defaulting_ as a concrete example of how this might be implemented, illustrating the items that must be considered along the way.

# Meta-behaviors

Meta-behaviors will be associated with _system-owned_ schema properties.  As a result, we will identify such properties by a common prefix.  For the sake of this proposal, we'll specify a prefix of `schema_` but could just as well be something like `elyra_`.  The reason `schema_` is being proposed is that meta-behaviors are essentially _schema wide_.  I.e., they span instances of a given schema (or namespace of for that matter).

Because these special schema properties are _system-owned_, we can associate functionality (i.e., _behaviors_) to each such property.  As noted above, defaulting behaviors would then be associated with a boolean-valued _instance_ property named `schema_default`, meaning that instances marked with `schema_default = True` are considered _the default_ instance for that schema.

Other schema properties indicating meta-behaviors won't necessarily be _instance_ properties, but rather properties of the _schema_.  For example, if we wanted to associate encryption with instance properties that contain sensitive data, we could indicate such behavior via a _list-valued schema property_ named `schema_encrypted` which would be an enumerated list of properties upon which encryption should be applied to their values.  For instance, the `kfp` schema might define `schema_encrypted` as:
```json
{
  "schema_encrypted": ["cos_password"]
}
```
[...]

PR #717, opened 2020-07-08
Use config to determine metadata storage class

### LINKED ISSUE #709: Metadata storage instance should be determined via configuration
Currently, the kind of storage used for the metadata service is determined when a `MetadataManager` instance is initialized:
https://github.com/elyra-ai/elyra/blob/fe48b72cfae8e65f252025d2d95fe39964a90611/elyra/metadata/metadata.py#L131-L138

However, the caller of each instance doesn't necessarily know how the system is configured (nor should they).  Instead, the `store` parameter should be removed and a `metadata_store_class` configurable trait should be added to `MetadataManager`.  As is the case today, the default value of `metadata_store_class` should be `FileMetadataStore`.

kevin-bates:
The class to use for the `MetadataStore` when creating a `MetadataManager` instance is now pulled from the `metadata_store_class` trait, which requires a class of `MetadataStore` and defaults to `FileMetadataStore`.

Differing storage classes can still be associated during `MetadataManager` creation by using parameter `metadata_store_class="class-name"` like so...
```python
    metadata_mgr = MetadataManager(namespace=METADATA_TEST_NAMESPACE, 
                                   metadata_store_class="elyra.metadata.tests.test_utils.MockMetadataStore")
```
This will be useful for building migration or distribution applications.  For example, one `MetadataManager` using `FileMetadataStore` and fetching instances, while another `MetadataManager` uses `MongoDBMetadataStore` that creates the fetched instances in the destination.  (Note that using `FileMetadataStore` on both ends won't work today (unless you were changing namespaces) because the filesystem locations are derived from within the process - not provided as parameters - but we could probably add that capability if necessary (although cp -r would work fairly easily in that case as well).)

Down the road, one could see a `MetadataStore` instance that corresponds to the specific metadata instance based on schema values.  This would allow for applications like Telemetry that consume JSON-validated records if, for example, the Metadata Store was implemented to "store" its instances wherever the telemetry "sink" is located.  We'd need to add concepts of write-only and perhaps read-only instances, but that would be just another [_meta-behavior_](https://github.com/elyra-ai/elyra/issues/518).

A `MockMetadataStore` has been created and the metadata manager and metadata store tests now run against both it and `FileMetadataStore` for testing.  This helps keep file-based items out of the metadata manager logic and can serve as an example of a non-file based store manager.

Resolves #709

 
           in the file; or

           it.

### DIFF
--- a/elyra/metadata/manager.py
+++ b/elyra/metadata/manager.py
@@ -43,11 +43,11 @@ class MetadataManager(LoggingConfigurable):
 
     def __init__(self, namespace: str, **kwargs):
         """
-        Generic object to read Notebook related metadata
-        :param namespace: the partition where it is stored, this might have
-        a unique meaning for each of the supported metadata storage
-        :param store: the metadata store to be used
-        :param kwargs: additional arguments to be used to instantiate a metadata store
+        Generic object to manage metadata instances.
+        :param namespace (str): the partition where metadata instances are stored
+        :param kwargs: additional arguments to be used to instantiate a metadata manager
+        Keyword Args:
+            metadata_store_class (str): the name of the MetadataStore subclass to use for storing managed instances
         """
         super(MetadataManager, self).__init__(**kwargs)
 
--- a/elyra/metadata/manager.py
+++ b/elyra/metadata/manager.py
@@ -39,7 +39,7 @@ class MetadataManager(LoggingConfigurable):
     metadata_store_class = Type(default_value=FileMetadataStore, config=True,
                                 klass=MetadataStore,
                                 help="""The metadata store class.  This is configurable to allow subclassing of
-                                the MetadataManager for customized behavior.""")
+                                the MetadataStore for customized behavior.""")
 
     def __init__(self, namespace: str, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


elyra/metadata/manager.py
```diff
@@ -21,25 +21,27 @@
...
+    metadata_store_class = Type(default_value=FileMetadataStore, config=True,
+                                klass=MetadataStore,
+                                help="""The metadata store class.  This is configurable to allow subclassing of
+                                the MetadataStore for customized behavior.""")
 
-    def __init__(self, namespace: str, store: Optional[MetadataStore] = None, **kwargs):
+    def __init__(self, namespace: str, **kwargs):
         """
         Generic object to read Notebook related metadata
         :param namespace: the partition where it is stored, this might have
```
vabarbosa, line 47: the `:param store:` (line 49) is no longer applicable/required

kevin-bates: Damn - great catch.  Thank you.

kevin-bates: I found a few things incorrect about that area and have submitted the change - thanks for your review.

PR #729, opened 2020-07-09
Allow code editor to be resized

### LINKED ISSUE #728: Code Editor components are not scrollable
In the Code Snippet sidebar and Metadata Editor form, the code editor component is not scrollable when filled with large amount of content. The scrollbar does appear however, it is still not scrollable. As a result, content (e.g., `Save` button) can get pushed off screen and become unreachable:

image

vabarbosa:
Updated the code editor component in the code snippet sidebar and the code editor component in the metadata editor form 

- [x] code editor component can be scrolled using the scrollbars
- [x] code editor component opens at an appropriate default size so other components do not immediately get pushed off screen

Fixes #728

**Before**:

image

**After**: 

image

 
           in the file; or

           it.

### DIFF
--- a/packages/code-snippet/style/index.css
+++ b/packages/code-snippet/style/index.css
@@ -55,6 +55,11 @@
   fill: var(--jp-ui-font-color1);
 }
 
+.elyra-codeSnippetsHeader + div:first-of-type {
+  overflow-y: auto;
+  height: calc(100vh - 95px);
+}
+
 .elyra-codeSnippet-item {
   border-bottom: var(--jp-border-width) solid var(--jp-border-color2);
   display: flex;
@@ -63,18 +68,18 @@
   padding: 0;
 }
 
-.elyra-codeSnippet-item .CodeMirror-wrap {
+.elyra-codeSnippet-item .elyra-expandableContainer-details-visible {
+  background-color: var(--jp-cell-editor-background);
+  resize: vertical;
+  height: 100px;
+}
+
+.elyra-codeSnippet-item .CodeMirror.cm-s-jupyter {
   background-color: inherit;
   border: none;
   font-family: var(--jp-code-font-family);
   font-size: var(--jp-code-font-size);
   line-height: var(--jp-code-line-height);
-  height: 100%;
-  width: 100%;
-  resize: vertical;
-  white-space: pre;
-  word-wrap: normal;
-  overflow-x: scroll;
 }
 
 .elyra-codeSnippet-saveButton {
--- a/packages/metadata-editor/src/MetadataEditor.tsx
+++ b/packages/metadata-editor/src/MetadataEditor.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { FormGroup, Intent } from '@blueprintjs/core';
+import { FormGroup, Intent, ResizeSensor } from '@blueprintjs/core';
 
 import { FrontendServices, IDictionary } from '@elyra/application';
 import { DropDown } from '@elyra/ui-components';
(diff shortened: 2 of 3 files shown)


ajbozarth:
Two points on this:

1. Does this address the issue from this comment on the original PR? https://github.com/elyra-ai/elyra/pull/589#pullrequestreview-433532524 @marthacryan and I could never recreate it, but checking the css the actual code editor is still only 1 line despite the Dom element having a min height of 100px. I would argue that the scrollability shouldn't require a min height.

2. This only kinda solve the issue as you raised it in #728 which is that the metadata editor widget doesn't scroll. If I shrink my window size vertically enough the save button will still go off screen. Though this fix is still important, we should make sure the widget itself can scroll as well since we don't control the quantity of metadata fields that may be displayed by other namespaces.

vabarbosa:
> Two points on this:
> 
>     1. Does this address the issue from this comment on the original PR? [#589 (review)](https://github.com/elyra-ai/elyra/pull/589#pullrequestreview-433532524) @marthacryan and I could never recreate it, 

i used to run into that issue a couple times early on but i have not seen it after the past couple commits to the metadata editor. i tried again now and not reproducing so while it looks like it is no longer happening i can't say for certain this update addressed it.

> but checking the css the actual code editor is still only 1 line despite the Dom element having a min height of 100px. I would argue that the scrollability shouldn't require a min height.
> 

the min-height is there just so when you open a new editor window you have a box that is clearly visible but if that is not desired it can be removed.


>     2. This only kinda solve the issue as you raised it in #728 which is that the metadata editor widget doesn't scroll. If I shrink my window size vertically enough the save button will still go off screen. Though this fix is still important, we should make sure the widget itself can scroll as well since we don't control the quantity of metadata fields that may be displayed by other namespaces.

just committed an updated which should also allow the widget itself doesn't scroll. you should now be able to scroll if your window is made small enough

ajbozarth:
Thanks for the quick updates, I just tested this locally and it works great for new snippet editor but it seems to have a weird bug when editing. The code box is not resizable when editing an existing snippet, but if you shrink the window size till the newly added widget scrolling appears the resize corn shows up (but is of course not resizable since the window is too small) 
Screen Shot 2020-07-09 at 3 07 53 PM
Screen Shot 2020-07-09 at 3 08 01 PM

vabarbosa:
> Thanks for the quick updates, I just tested this locally and it works great for new snippet editor but it seems to have a weird bug when editing. The code box is not resizable when editing an existing snippet, but if you shrink the window size till the newly added widget scrolling appears the resize corn shows up (but is of course not resizable since the window is too small)
> 

hmmm, i am not seeing that. which browser are you using? out of curiosity, if you put focus out and then back into the code editor does it re-appear?

ajbozarth:
I just checked and it's a bug unique to Safari (it works fine in FF and Chrome)

vabarbosa:
> I just checked and it's a bug unique to Safari (it works fine in FF and Chrome)

@ajbozarth the Safari issue should be addressed now

vabarbosa:
>     1. In Safari when you mouse over the resize corner the mouse pointer is a text cursor (just like the rest of the editor), whereas in chrome and FF its the resize cursor.
> 

all set. Safari should no longer display the text cursor. however it default to the regular cursor while Firefox/Chrome display the resize cursor.


>     2. In FF and Chrome the resize no longer respects the minimum height whereas Safari does, I don't really have an opinion on which is better but they should be consistent with each other.
> 

when using CSS `resize` Safari treats initial element height as a minimum height (Chrome, Firefox do not). i initially had `min-height` configured and all three behaved similar but there were previous comments about preferring not setting a `min-height`. in any case, i have added the `min-height` back so there is consistent behavior across browsers.

ajbozarth:
@vabarbosa thanks for the fixes, but now I'm no longer seeing the original feature:

> code editor component opens at an appropriate default size so other components do not immediately get pushed off screen

On both Chrome and Safari it always starts at the min height now, no matter the content

PR #741, opened 2020-07-09
Disable toolbar buttons on empty pipeline editor

### LINKED ISSUE #691: [Usability] A user should not be able to run, export or otherwise process an empty pipeline
1.0.0 beta 1

The UI should prevent the user from attempting a task that is guaranteed to fail.

Steps:
- create an empty pipeline
- for example, run the pipeline or export it -> fail

karlaspuldaro:
Fixes #691 
 
           in the file; or

           it.


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -222,31 +237,35 @@ export class PipelineEditor extends React.Component<
...
         iconEnabled: IconUtil.encode(clearPipelineIcon),
         iconDisabled: IconUtil.encode(clearPipelineIcon)
       },
       { divider: true },
-      { action: 'undo', label: 'Undo', enable: true },
-      { action: 'redo', label: 'Redo', enable: true },
-      { action: 'cut', label: 'Cut', enable: false },
-      { action: 'copy', label: 'Copy', enable: false },
+      { action: 'undo', label: 'Undo', enable: !this.state.emptyPipeline },
+      { action: 'redo', label: 'Redo', enable: !this.state.emptyPipeline },
```
ajbozarth, line 253: These probably should not be disabled on empty pipeline since the user might want to redo/undo adding or deleting the only node in the canvas.

When testing locally I noticed that they don't actually get disabled/enabled as I would have expected based on this line. They behave as I would expect them to (can still undo redo based on the "clipboard" not pipeline state) but when undoing or redoing the first/last node in the canvas the undo/redo doesn't trigger an update to whether the other buttons should be enabled/disabled. ie they stay in whichever state they were in prior to the undo/redo

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -287,9 +306,11 @@ export class PipelineEditor extends React.Component<
...
 
   updateModel(): void {
-    this.widgetContext.model.fromString(
-      JSON.stringify(this.canvasController.getPipelineFlow(), null, 2)
-    );
+    const pipelineFlow = this.canvasController.getPipelineFlow();
+
+    this.widgetContext.model.fromString(JSON.stringify(pipelineFlow, null, 2));
+
+    this.setState({ emptyPipeline: Utils.isNewPipeline(pipelineFlow) });
```
ajbozarth, line 313: does setState only update the given values or the whole state (ie will this remove `showPropertiesDialog` and `propertiesInfo` from the state)?

When testing it locally it doesn't seem to cause issues, but I'm still curious to the answer

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: Screen Shot 2020-07-09 at 1 44 04 PM

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: It updates the part of the state we want to update, so it will merge the object passed to setState with the current state, we don't lose the previous values

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: Good catch, gonna take a look

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -222,31 +239,35 @@ export class PipelineEditor extends React.Component<
...
-      { action: 'copy', label: 'Copy', enable: false },
-      { action: 'paste', label: 'Paste', enable: false },
+      { action: 'undo', label: 'Undo' },
+      { action: 'redo', label: 'Redo' },
+      { action: 'cut', label: 'Cut' },
+      { action: 'copy', label: 'Copy' },
+      { action: 'paste', label: 'Paste' },
       { action: 'addComment', label: 'Add Comment', enable: true },
-      { action: 'delete', label: 'Delete', enable: true },
+      { action: 'delete', label: 'Delete' },
```
karlaspuldaro, line 260: We don't need to set the `enable` property on these commands, they're handled by context on canvas already

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -714,6 +738,13 @@ export class PipelineEditor extends React.Component<
       this.handleSavePipeline();
     } else if (action == 'clear') {
       this.handleClearPipeline();
+    } else if (
+      action == 'undo' ||
+      action == 'redo' ||
+      action == 'delete' ||
+      action == 'addComment'
```
karlaspuldaro, line 745: Added listeners for canvas commands that need to update pipeline-editor buttons. 
`Cut` and `paste` already triggers model update.

packages/pipeline-editor/src/utils.ts
```diff
@@ -59,16 +59,24 @@ export default class Utils {
...
-    }
+    return Object.keys(pipelineDefinition.pipelines[0].nodes).length === 0;
+  }
+
+  /**
+   * Check if the provided pipeline is clear of nodes and comments
+   *
+   * @param pipelineDefinition
+   */
+  static isEmptyCanvas(pipelineDefinition: any): boolean {
```
karlaspuldaro, line 75: There might be cases we want to check if canvas is completely clean of nodes and comments in order to enable/disable action buttons (eg. we only disable clean pipeline button if canvas is empty and there is nothing else left to clean)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: `undo/redo` is fixed, also similar cases noticed with `delete` and `addComment`. Note we want `clear pipeline` enabled if there are comment boxes but no nodes.
`cut/paste` already trigger model update, didn't need to add listeners for those (in case we cut the only node left, or paste a node on an empty pipeline)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: not an issue but if you wanted you could simplify this to one line
```javascript
} else if (['undo', 'redo', 'delete', 'addComment'].includes(action)) {
```
or if you prefer
```javascript
} else if (['undo', 'redo', 'delete', 'addComment'].indexOf(action) > -1) {
```

karlaspuldaro:
@vabarbosa the undo/redo issue when clearing a pipeline is existing, looks like we never noticed it, or never properly handled it. I opened #756 where we can investigate as a separate PR, as discussed today.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -246,31 +263,35 @@ export class PipelineEditor extends React.Component<
...
         action: 'arrangeHorizontally',
         label: 'Arrange Horizontally',
-        enable: true
+        enable: !this.state.emptyPipeline || !emptyCanvas
       },
-      { action: 'arrangeVertically', label: 'Arrange Vertically', enable: true }
+      {
+        action: 'arrangeVertically',
+        label: 'Arrange Vertically',
+        enable: !this.state.emptyPipeline || !emptyCanvas
```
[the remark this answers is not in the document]
ajbozarth: upon further testing this is needed but should prob just be `!this.state.emptyPipeline` since comments are not included when the arrange buttons are used, so leaving it enabled when there's only comments does not make sense

ajbozarth:
I tested locally things seem to work great, I did open #757 with separate issues I found while testing this. I also ran the ui tests locally and they passed. I also check and this looks good in dark mode as well (except of course my previously mentioned issue).

So once my previous code comments are address this LGTM

lresende:
@karlaspuldaro I see approvals coming in for this pr. Are we ready to remove the `work in progress` tag and merge?

karlaspuldaro:
@lresende I wanted to add more tests for this but that can be done later, we already have basic tests up and a note for extra ones

Issue #733, opened 2020-07-09
Pipeline editor: dark theme uses low contrast color scheme


ptitzler:
Color scheme used by JupyterLab (relatively high contrast between font color and background):

![image](https://user-images.githubusercontent.com/13068832/87076651-e3f70400-c1d6-11ea-86cf-6e4901c92cf3.png)

Color scheme used by Elyra/pipeline editor (low contrast between font color and background):
![image](https://user-images.githubusercontent.com/13068832/87076494-b611bf80-c1d6-11ea-94be-090cfc6ed344.png)

ajbozarth:
This is blocked by our updating to a newer version of canvas. The version Elyra currently uses doesn't support svgs for the toolbar so we can't edit the color dynamically. This has been address in later versions of canvas and will be an easy fix once we update the canvas version (tracked in #645).

vabarbosa:
Icons passed to the current version of canvas being used displayed as `img` instead of `svg` and thus cannot be easily styled. An update has been added to the recent version of canvas which allows to passing of React components as icons which should allow for better and more easily styling the icons.

Per previous discussion (https://github.com/elyra-ai/elyra/pull/706#pullrequestreview-443437599), it was initially decided to leave the icons blue until the canvas version is upgraded. Upon further investigation, I was able to find workaround. Since image colors can't be changed via css I instead replace the images entirely when the theme is changed. This should suffice until canvas is upgraded.

See PR #743

PR #743, opened 2020-07-09
Replace Pipeline Editor toolbar icons when in dark mode

vabarbosa:
Icons passed to the current version of canvas being used displayed as `img` instead of `svg` and thus cannot be easily styled.

Per previous discussion (https://github.com/elyra-ai/elyra/pull/706#pullrequestreview-443437599), it was initially decided to leave the icons blue until the canvas version is upgraded which unblocks this issue. However after further investigation, I was able to find a workaround. Since color can't be changed in images with CSS, I instead just replace the images entirely with new ones when the theme is switched. This should suffice until canvas is upgraded. 

Fixes: #733 

**Before**:

image

**After**:

image

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -68,6 +68,10 @@ const NODE_TOOLTIP_CLASS = 'elyra-PipelineNodeTooltip';
 
 const TIP_TYPE_NODE = 'tipTypeNode';
 
+const SAVE_ICON_ID = 'toolbar-icon-save';
+const EXPORT_ICON_ID = 'toolbar-icon-export';
+const CLEAR_ICON_ID = 'toolbar-icon-clear';
+
 const NodeProperties = (properties: any): React.ReactElement => {
   return (
     <dl className={NODE_TOOLTIP_CLASS}>
@@ -109,6 +113,25 @@ export class PipelineEditorWidget extends ReactWidget {
     this.app = props.app;
     this.browserFactory = props.browserFactory;
     this.context = props.context;
+  }
+
+  themeChanged(theme: string, isLight: boolean): void {
+    [SAVE_ICON_ID, EXPORT_ICON_ID, CLEAR_ICON_ID].forEach((id: string) => {
+      const element = document.getElementById(id) as HTMLImageElement;
+      if (element) {
+        switch (element.id) {
+          case SAVE_ICON_ID:
+            element.src = IconUtil.encode(savePipelineIcon, !isLight);
+            break;
+          case EXPORT_ICON_ID:
+            element.src = IconUtil.encode(exportPipelineIcon, !isLight);
+            break;
+          case CLEAR_ICON_ID:
+            element.src = IconUtil.encode(clearPipelineIcon, !isLight);
+            break;
+        }
+      }
+    });
   }
 
   render(): React.ReactElement {
@@ -195,6 +218,7 @@ export class PipelineEditor extends React.Component<
 
   render(): any {
     const style = { height: '100%' };
+    const darkmode = document.querySelector("[data-jp-theme-light='false']");
     const emptyCanvasContent = (
       <div>
         <dragDropIcon.react tag="div" elementPosition="center" height="120px" />
@@ -216,22 +240,22 @@ export class PipelineEditor extends React.Component<
         action: 'save',
         label: 'Save Pipeline',
         enable: true,
-        iconEnabled: IconUtil.encode(savePipelineIcon),
-        iconDisabled: IconUtil.encode(savePipelineIcon)
+        iconEnabled: IconUtil.encode(savePipelineIcon, !!darkmode),
+        iconDisabled: IconUtil.encode(savePipelineIcon, !!darkmode)
       },
       {
         action: 'export',
         label: 'Export Pipeline',
         enable: true,
-        iconEnabled: IconUtil.encode(exportPipelineIcon),
-        iconDisabled: IconUtil.encode(exportPipelineIcon)
+        iconEnabled: IconUtil.encode(exportPipelineIcon, !!darkmode),
+        iconDisabled: IconUtil.encode(exportPipelineIcon, !!darkmode)
       },
       {
         action: 'clear',
         label: 'Clear Pipeline',
         enable: true,
-        iconEnabled: IconUtil.encode(clearPipelineIcon),
-        iconDisabled: IconUtil.encode(clearPipelineIcon)
+        iconEnabled: IconUtil.encode(clearPipelineIcon, !!darkmode),
+        iconDisabled: IconUtil.encode(clearPipelineIcon, !!darkmode)
       },
       { divider: true },
       { action: 'undo', label: 'Undo', enable: true },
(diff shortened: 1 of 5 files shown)


ajbozarth:
Even if this is temporary I'm not sure I want to hack this by card coding svg strings into css files. Based on your work in #589 (specifically 55642a6 ) could we instead dynamically replace the image instead. Then we could leverage the LabIcon object instead. We could grab the LabIcon instance for a icon, get/copy/update/set it's svg with the new color, then use our IconUtil to grab the new encoded string, which we could then set as the background for that css. More long winded but at least it'd be dynamic and not hard coded. This of course would rely on being able to update that css dynamically using the method you did previously. Feel free to slack me (or even Webex) if you want me to better explain my idea.

vabarbosa:
> Even if this is temporary I'm not sure I want to hack this by card coding svg strings into css files. Based on your work in #589 (specifically [55642a6](https://github.com/elyra-ai/elyra/commit/55642a6ab435999612348a870795bf427e7575db) ) could we instead dynamically replace the image instead. Then we could leverage the LabIcon object instead. We could grab the LabIcon instance for a icon, get/copy/update/set it's svg with the new color, then use our IconUtil to grab the new encoded string, which we could then set as the background for that css. More long winded but at least it'd be dynamic and not hard coded. This of course would rely on being able to update that css dynamically using the method you did previously. Feel free to slack me (or even Webex) if you want me to better explain my idea.

valid point, i can make that change. i was just trying to keep it simple and self contained within the CSS and not introduce additional JS since this should all have to be undone once canvas is upgraded

ajbozarth:
@vabarbosa you might be able to include most of the functionality I mentioned in one single util function in IconUtil in icons.tsx, similar to encode, just encodeDark. Then insert the string it returns on theme change

ptitzler:
Color change is good. How it's best implemented I have no opinion on. Do note that I didn't try out all Elyra features in dark mode. Therefore similar issues might be observed when a user completes other tasks.

vabarbosa:
> @vabarbosa you might be able to include most of the functionality I mentioned in one single util function in IconUtil in icons.tsx, similar to encode, just encodeDark. Then insert the string it returns on theme change

@ajbozarth PR has been updated to use ThemeManager and IconUtil to trigger and swap image instead of CSS only approach

packages/ui-components/src/icons.tsx
```diff
@@ -65,7 +65,35 @@ export const trashIcon = new LabIcon({
...
+    }
+    if (strokeColor) {
+      svgstr = svgstr.replace(
+        /stroke="[^(none)]+?"/gi,
+        `stroke="${strokeColor}"`
+      );
+    }
+
+    return new LabIcon({
+      name: `${icon.name}:color`,
```
ajbozarth, line 95: just in case this is eventually called with different colors we should make this name specific to the color params (ie: `${icon.name}:${fillColor}:${strokeColor}` or similar) Then we could also add a check to just return the icon if it already exists (I think it throws an error if you "recreate" a LabIcon)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -111,6 +115,25 @@ export class PipelineEditorWidget extends ReactWidget {
     this.context = props.context;
   }
 
+  themeChanged(theme: string, isLight: boolean): void {
```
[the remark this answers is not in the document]
vabarbosa: good catch, it was part of my initial approach. forgot to remove it. should be cleaned up now.

packages/ui-components/src/icons.tsx
```diff
@@ -65,7 +65,35 @@ export const trashIcon = new LabIcon({
  * A utilities class for handling LabIcons.
  */
 export class IconUtil {
-  static encode(icon: LabIcon): string {
-    return 'data:image/svg+xml;utf8,' + encodeURIComponent(icon.svgstr);
+  static encode(icon: LabIcon, darkTheme?: boolean): string {
+    return (
+      'data:image/svg+xml;utf8,' +
+      encodeURIComponent(
+        darkTheme ? this.colorize(icon, '#f4f4f4').svgstr : icon.svgstr
```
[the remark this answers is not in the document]
vabarbosa: should be all set now. the code has been refactored as suggested.

packages/ui-components/src/icons.tsx
vabarbosa: good point. should be set now

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -72,6 +72,11 @@ const SAVE_ICON_ID = 'toolbar-icon-save';
 const EXPORT_ICON_ID = 'toolbar-icon-export';
 const CLEAR_ICON_ID = 'toolbar-icon-clear';
 
+const getEncodedIcon = (icon: LabIcon, darkTheme: boolean): string => {
```
ajbozarth, line 75: this should be moved to `utils.ts`

packages/pipeline-editor/src/index.ts
```diff
@@ -149,6 +155,16 @@ const extension: JupyterFrontEndPlugin<void> = {
       command: commandIDs.submitNotebook,
       rank: -0.5
     });
+
+    const updateTheme = (widget: any): void => {
+      const pipelineEditor = widget.content;
+      pipelineEditor.themeChanged(themeManager.isLight(themeManager.theme));
```
ajbozarth, line 161: since this function has been simplified should this just be one line or do you think keeping it separate is cleaner?

packages/ui-components/src/icons.tsx
```diff
@@ -68,4 +68,29 @@ export class IconUtil {
...
   }
+
+  static colorize(
+    icon: LabIcon,
+    fillColor: string,
+    strokeColor?: string
+  ): LabIcon {
+    let svgstr = icon.svgstr;
+
+    if (fillColor) {
```
ajbozarth, line 79: `fillColor` is not listed as an optional param, so either this if is unnecessary or you should add the `?` to the param

packages/ui-components/src/icons.tsx
```diff
@@ -68,4 +68,29 @@ export class IconUtil {
...
+    if (strokeColor) {
+      svgstr = svgstr.replace(
+        /stroke="[^(none)]+?"/gi,
+        `stroke="${strokeColor}"`
+      );
+    }
+
+    return LabIcon.resolve({
+      icon: {
+        name: `${icon.name}:${fillColor || ''}:${strokeColor || ''}`,
```
ajbozarth, line 91: we should tweak this so the `:` only are added if `fillColor` or `strokeColor` exist. Otherwise we could get a name with extra colons

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: it has been moved

packages/ui-components/src/icons.tsx
vabarbosa: it has been made optional

packages/ui-components/src/icons.tsx
vabarbosa: done

packages/pipeline-editor/src/index.ts
vabarbosa: good catch. it is changed to single line

Issue #747, opened 2020-07-10
Add unit tests for Elyra pipeline processor 


akchinSTC:
Add tests for :
- [ ] KFP pipeline processor
- [ ] Pipeline handlers 
- [ ] Pipeline and operation constructors

kiersten-stokes:
@Michonster , here are a few pipeline `test_handlers.py` scenarios that we currently don't have coverage for that I think would be a good place to start:

1. Hitting the "export" API (`PipelineExportHandler`) for both a KFP pipeline and an Airflow pipeline (2 separate tests)
2. Hitting the "validate" API (`PipelineValidationHandler`) for both a KFP pipeline and an Airflow pipeline (2 separate tests)

We have some test resources already available in `elyra/tests/pipeline/resources` that can be used for these test scenarios too.

PR #751, opened 2020-07-13
Add additional tests for elyra pipeline module

akchinSTC:
Still a work in progress. Looking for input on better way of testing the cc pipeline function
- Moved testing resources for parsing to new dir
- Small fixes for pipeline class
- tests added for constructor and portion of the kfp processor

 
           in the file; or

           it.


elyra/pipeline/tests/test_kfp_processor.py
```diff
@@ -0,0 +1,165 @@
...
+                print(tarinfo.name)
+                tar_content.append(tarinfo.name)
+
+    assert sorted(correct_filelist) == sorted(tar_content)
+
+
+@pytest.mark.skip(reason='test should fail but does not, pipelines_test_file doesnt exist')
+def test_fail_generate_dependency_archive(processor):
+    pipelines_test_file = 'elyra/pipeline/tests/resources/archive/test.txt'
+    pipeline_dependencies = 'elyra/pipeline/tests/resources/archive/*'
```
kevin-bates, line 78: This won't throw any exception because 'include-all' is true - due to the wildcard.  I think we should probably test for the existence of filename prior to the creation of the archive (in the `_generate` method) since it will **always** be required.

akchinSTC: adding to existing open issue to address #747

lresende (approved):

PR #752, opened 2020-07-13
Add indicator for invalid node properties

### LINKED ISSUE #653: Usability: pipeline editor should indicate if there are problems in the flow configuration and where they are
Using the Beta release I've created a notebook flow and forgot to associate a runtime environment with each notebook. Running the flow the following user unfriendly stack trace is displayed in the error dialog:

```
  File "/Users/patti/anaconda3/envs/elyra_basic_37/lib/python3.7/site-packages/tornado/web.py", line 1703, in _execute
    result = await result
  File "/Users/patti/anaconda3/envs/elyra_basic_37/lib/python3.7/site-packages/tornado/gen.py", line 191, in wrapper
    result = func(*args, **kwargs)
  File "/Users/patti/anaconda3/envs/elyra_basic_37/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 91, in post
    pipeline = PipelineParser.parse(pipeline_definition)
  File "/Users/patti/anaconda3/envs/elyra_basic_37/lib/python3.7/site-packages/elyra/pipeline/parser.py", line 97, in parse
    parent_operations=links
  File "/Users/patti/anaconda3/envs/elyra_basic_37/lib/python3.7/site-packages/elyra/pipeline/pipeline.py", line 55, in __init__
    raise ValueError("Invalid pipeline operation: Missing field 'operation runtime image'.")
ValueError: Invalid pipeline operation: Missing field 'operation runtime image'.
```

The current behavior raises a couple of usability issues:
- The user needs to understand what the stack trace error is indicating in order to resolve the issue.
- Since it is known ahead of time that certain properties must be set (or the flow won't run as expected) the editor should flag problematic configuration issues, e.g. by visually marking a notebook node. 
- Imagine a flow that contains 20 notebooks. If the editor doesn't indicate which node (or nodes) are misconfigured, a user has to manually inspect every node in the flow to find out which ones are not correctly configured.

marthacryan:
Fixes #653. Adds a small red caret to nodes with invalid properties. Only checks
for having a runtime image chosen for now. Looks like this:
image
image

 
           in the file; or

           it.


lresende:
One thing here is the question of using `warning` vs something more like `error` as when the full validation is added this should/would disable pipeline submission if some of these are present. 

As for the icon, how about something more like a `red circle` or `red x` or `red *` on the top right corner aligned a little farther from the border? 

image

image

image

image


Note that the icon is more like a suggestion.

vabarbosa:
updated the icon used and it's position:

image

the icon can be replaced if desired by updating the `packages/ui-components/style/icons/error.svg`.
alternative icons can be found here: https://www.carbondesignsystem.com/guidelines/icons/library/
(just search for `error` or `alert`)

ajbozarth:
The update location of the icon is better, but looking at the carbon icons I think a "filled" icon would be slightly easier to see at that size, like "error filled" or "warning filled"

I'll do a more detailed review soon

lresende:
@ajbozarth Hovering on the node should show you the properties and what is wrong there.

ajbozarth:
@lresende that is correct, but if a user clicks on it before the tooltip appears it selects the node instead and the tooltip will not appear at that point until a user mouses off and on the node again

marthacryan:
Options for icon:
image
image
image
Thoughts? @vabarbosa @ajbozarth

ajbozarth:
Of those three options I would choose the first one (red dot). Is a "filled" version of the second one not an option though?

vabarbosa:
since the icon is small i would vote to keep it simple and go for the red dot.

vabarbosa:
for comparison purposes this is how it is done in canvas (Test harness)

- text color changed
- Error Filled icon near the node icon

image

marthacryan:
@vabarbosa @ajbozarth I've been playing around with the different options this morning - thoughts? 
image
image
image
image
image
image
image

vabarbosa:
> @vabarbosa @ajbozarth I've been playing around with the different options this morning - thoughts?
>

very nice! 

i like this with a red border:
image

or this as is:
image

marthacryan:
This PR now depends on #744 for the error alert at the top of the pipeline. 
Here's the updated behavior / look:
When you drag nodes onto the canvas, they look normal by default:
image

After pressing the run button, they look like this: 
image
(Wasn't sure how to phrase that error message at the top - it says "Invalid pipeline: some nodes have incomplete properties")

After fixing the properties for one of the nodes, it removes the error indicator for just that node:
image

After updating any of the properties of the nodes, it shows indicators for all nodes without the needed properties.

marthacryan:
Discussion: 
Should we be performing the same checks for exporting pipelines?

kevin-bates:
The screenshots and error message look great!

Yes, I believe this should apply to export as well. Anything that saves the user time is helpful.

vabarbosa:
looks great! 

a couple points. perhaps the message should read

```
Invalid pipeline: Some nodes have missing or invalid properties.
```

or something along those lines (other may have better suggestions for the phrasing).  

i would agree and say similar checks should be done on exporting. however (unless it will cause errors in the backend), i don't believe it should not prevent the user from exporting. just warn them so they are aware before they continue.

marthacryan:
@vabarbosa I just checked and if you go ahead with trying to export an invalid pipeline, it'll show an error dialog from the backend, so I think I'll just use the same behavior as the run button.

vabarbosa:
> @vabarbosa I just checked and if you go ahead with trying to export an invalid pipeline, it'll show an error dialog from the backend, so I think I'll just use the same behavior as the run button.

cool 👍 thanks

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -681,16 +834,98 @@ export class PipelineEditor extends React.Component<
...
+    }
+  }
+
+  /**
+   * Validates the properties of all nodes in the pipeline.
+   * Updates the decorations / style of all nodes.
+   *
+   * @returns true if the pipeline is invalid.
+   */
+  invalidPipeline(): boolean {
```
ajbozarth, line 902: the name of this function is not accurate to what it does. From the name and return value I would assume this returns if a pipeline is invalid, but the actual function does more than that, updating the pipeline base on the state's validity. This function should either be renamed to make its actual usage clear, or split into two separate functions, one for the update, and one for the boolean return.

marthacryan: Since I do the same thing in `updateDecorations` (i.e. both update the decorations and return the validity of the node) do you think maybe it should be called `updateAllDecorations`? And it just returns the validity of the pipeline?

marthacryan:
Last commit changes name of validation functions to `validateNode` and `validateAllNodes`. @lresende This should be good to merge as soon as @ajbozarth adds his approval and #744 is merged (already rebased off of those changes).

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -681,16 +814,103 @@ export class PipelineEditor extends React.Component<
...
+  /**
+   * Validates the properties of a given node.
+   *
+   * @param node: node to check properties for
+   *
+   * @returns a warning message to display in the tooltip
+   * if there are invalid properties. If there are none,
+   * returns null.
+   */
+  invalidProperties(node: any): string {
```
vabarbosa, line 870: along with the other function name changes this should be called be called `validateProperties` or `validateNodeProperties`

marthacryan: Good idea!

packages/pipeline-editor/style/index.css
```diff
@@ -20,6 +20,10 @@
   text-align: center;
 }
 
+.elyra-tooltipError {
+  color: var(--jp-error-color0);
```
vabarbosa, line 24:
```suggestion
  color: var(--jp-error-color1);
```

vabarbosa: `--jp-error-color1` is  a little bit more readable in dark mode

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -681,16 +814,103 @@ export class PipelineEditor extends React.Component<
...
+          x_pos: 20,
+          y_pos: 3
+        }
+      ]);
+      const pipelineId = this.canvasController.getPrimaryPipelineId();
+      const stylePipelineObj: any = {};
+      stylePipelineObj[pipelineId] = [node.id];
+      const styleSpec = {
+        body: { default: 'stroke: #d32f2f;' },
+        label: { default: 'fill: #d32f2f;' }
```
vabarbosa, line 851: would it be possible to set these to `var(--jp-error-color1)`?
also when the node is selected there is still the thinker blue stroke around it. would it be possible to set that to red as well?

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -879,22 +884,22 @@ export class PipelineEditor extends React.Component<
    *
    * @returns true if the pipeline is invalid.
    */
-  invalidPipeline(): boolean {
-    let invalidPipeline = false;
+  validateAllNodes(): boolean {
+    let invalidPipeline = true;
```
ajbozarth, line 888: given you inverted the meaning of this value the var name should probably be updated as well (same with the jsdoc above)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -885,16 +885,16 @@ export class PipelineEditor extends React.Component<
    * @returns true if the pipeline is invalid.
```
ajbozarth, line 885:
```suggestion
   * @returns true if the pipeline is valid.
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
marthacryan: Done

vabarbosa:
Looks good. thanks.

marthacryan:
New feature based on case brought up in #744 - checks for circular references on startup and adds an indicator for all links in a cycle, as well as an error message.

## Code changes
Adds a function `validateAllLinks` and a function `validatePipeline`. `validatePipeline` returns a tidy error message that collects the error messages from both `validateAllNodes` and `validateAllLinks`, and returns null if everything is valid. 

## Behavior changes
Here's what it looks like when there are both invalid links and invalid nodes and you try to run, export, or just open the pipeline:

image

After fixing the cycle, it automatically removes the style from other links:

image

vabarbosa:
@marthacryan this looks great! thank you for the quick turn around.

PR #759, opened 2020-07-15
Update notebook node to allow multiple input links

### LINKED ISSUE #750: Pipeline editor prevents multiple input links on first node added after open
For whatever reason, I can no longer associate multiple nodes as inputs to a single node like the following:
image
Instead, I see this...
NoDoubleLinks

I'm using master with a last commit of Friday evening: a52f2704bd8d7d95968834553cc9bbd8af8a84a6

Although this is also happening on @marthacryan's prevent-circular branch while looking at #744.

@lresende also confirmed.

vabarbosa:
Fixes #750

This addresses the issue of cardinality incorrectly being set to a `max: 0` and thus not allowing multiple input links on the notebook node.

The [second observation](https://github.com/elyra-ai/elyra/issues/750#issuecomment-657894923)(cardinality being deleted after first node) was addressed in canvas (https://github.com/elyra-ai/canvas/issues/162) and should be part of the next canvas release.
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/palette.json
+++ b/packages/pipeline-editor/src/palette.json
@@ -18,7 +18,7 @@
                 "ui_data": {
                   "cardinality": {
                     "min": 0,
-                    "max": 1
+                    "max": -1
                   },
                   "label": "Input Port"
                 }


lresende:
Do we know what happens if we don't specify cardinality info for the nodes? Are there an acceptable default?

ajbozarth:
As far as we've seen having no cardinality behaves the same as we would like it to

kevin-bates:
Although we _see_ a behavior of min: 0, max: -1, the [schema imposes an "at least 1" default](https://github.com/elyra-ai/pipeline-schemas/blob/master/common-pipeline/pipeline-flow/pipeline-flow-ui-v3-schema.json#L52-L65) while the _comment_ in the description indicates an "exactly 1" (min: 1, max: 1) behavior for inputs and an "at least 1" (min: 1, max: -1) behavior for outputs.  So they might have a bug when cardinality is not specified (IMO).

All the more reason to update the canvas for 1.0.

lresende:
The answer to my original question:
> The original intent for palettes was to load selected operators from an operator repository. The operator definition includes “cardinality” for the “port_definition”: https://github.com/elyra-ai/pipeline-schemas/blob/master/common-pipeline/operators/operator-v3-schema.json.
The description for “cardinality” is:
“Property to capture how many connections are allowed for this port, e.g., min: 1, max:1 implies you must supply 1 and only 1; min:0, max:1 implies it is optional and a max of one, min:0, max:-1 means it is optional and you can have any number of connections. The default value is 1:1 for inputs and 1:-1 for outputs.”

kevin-bates:
Sounds like what I said and referenced.    I should win a prize: :tada: :smile:

For ports with cardinality min > 0, when does that get enforced?  Is there some call to make to "validate a pipeline" prior to its submission?

PR #767, opened 2020-07-17
Update clean-jupyterlab to Install Debugger and Xeus python on dev env

lresende:

           in the file; or

           it.

### DIFF
--- a/etc/scripts/clean-jupyterlab.sh
+++ b/etc/scripts/clean-jupyterlab.sh
@@ -30,6 +30,7 @@ echo " "
 
 set -e
 echo "Uninstalling old packages"
+conda uninstall -y xeus-python -c conda-forge
 pip uninstall -y elyra || true;
 pip uninstall -y jupyterlab-git || true;
 pip uninstall -y jupyterlab-server || true;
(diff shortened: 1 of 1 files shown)


lresende:
Waiting for https://github.com/jupyter-xeus/xeus-python/issues/308

lresende:
@marthacryan @karlaspuldaro @ajbozarth 
Could a few of you give this a try and let me know if it works or if you see any issues? Seems ok for me now...

karlaspuldaro:
It works for me :)
Just curious why debugger is being installed in clean-jupyterlab and not in Makefile?

ajbozarth:
>Just curious why debugger is being installed in clean-jupyterlab and not in Makefile?

@karlaspuldaro I asked @lresende about that in a call a while back and the idea is this is kinda an early test. Debugger will be part of core in 3.0 so this would just be a stop gap until then for testing. Also even if it wasn't going into core, we didn't want it to be "part of" elyra like git and toc, but something we as a dev team make sure we support for customers that want to use it

ajbozarth:
@lresende out of curiosity are we planning on extending debugger support to our python editor? (eventually)

lresende:
I believe we tested and it was working via console integration, but the idea is to look into more deep integration.

PR #776, opened 2020-07-24
Add initial support for supernodes

kevin-bates:
Supernodes are now supported at a basic level.  The UI has been minimally modified to enable their creation.  However, we'll likely want to enhance the use of supernodes.  For example, we might want the various properties, envs, dependencies, to pertain to all nodes in the supernode, etc.  Right now, you're not able to edit the properties of a supernode - at least relative to elyra.

The pipeline submitted to the backend runtime consists of a flattened. pipeline where disjoint pipeline definitions can be executed in parallel.  The presence of a supernode essentially equates to the flattening of its corresponding pipeline reconnected to any external input/output nodes.

Although differences appear significant, much is due to refactoring to handle multiple pipeline definitions and recursion, as well as simplification of helper functions and the addition of type hints.  As a result, you might find it easier to review looking at the code directly rather than the diffs.

Here's an example of a supernode that was executed:
Screen Shot 2020-07-24 at 2 29 48 PM

And here's the corresponding "flattened" representation in kubeflow...
image

 
           in the file; or

           it.


kevin-bates:
Moving to draft to add some better tests to validate link navigation.

kevin-bates:
Updated pipeline_with_supernode.json to contain a valid pipeline with a supernode taking inputs and producing outputs.  Then updating the supernode test to confirm its structure.

Here's what it looks like:
Screen Shot 2020-07-25 at 7 36 12 AM

lresende:
@marthacryan Could you take a quick look at this regarding the validation of pipelines when supernodes are enabled.

kevin-bates:
@marthacryan - I'm not sure what all the front-end validation consists of, but I'd be happy to help you navigate the weeds of super and binding nodes. It took a fair amount of sitting down with the pipeline JSON to track things down.

marthacryan:
Ok I added some changes to address the issues with validation - someone should definitely play around with it though.

marthacryan:
Here are some pictures of the behavior:
- Collapsed supernode containing invalid nodes (including tooltip and highlighted edges for selected node):
image
- Expanded supernode containing invalid nodes (has the same tooltip message):
image
- Properties option is removed from the context menu of the supernode. Properties option for the supernode's children is still accessible from the context menu of the expanded node as well as in the complete view of the supernode:
image
And the same behavior of automatically re-validating after the properties are updated still works:
image

PR #777, opened 2020-07-24
Fix support for Object Storage hosted on IBM Cloud

### LINKED ISSUE #778: Error submitting pipeline when using IBM Cloud Object Storage: BucketAlreadyOwnedByYou execption
Configure you `Elyra Runtime metadata` to point to IBM COS
Create a pipeline, and submit it using the runtime pointing to IBM COS
This will fail with:
```
minio.error.BucketAlreadyOwnedByYou: BucketAlreadyOwnedByYou: message:
Your previous request to create the named bucket succeeded and you already own it.
```

This is happening because the the `bucket_exists` API from the minio client is always returning `False` even when the bucket already exists, and then when the backend code tries to create the bucket the `BucketAlreadyOwnedByYou` exception is raised

lresende:
Workaround for the fact that minio API to check if `bucket_exists` does not work on IBM COS

Fixes #778 
 
           in the file; or

           it.

### DIFF
--- a/elyra/util/cos.py
+++ b/elyra/util/cos.py
@@ -55,11 +55,12 @@ class CosClient(LoggingConfigurable):
         try:
             if not self.client.bucket_exists(self.bucket):
                 self.client.make_bucket(self.bucket)
-        except BucketAlreadyOwnedByYou as ex:
+        except BucketAlreadyOwnedByYou:
+            # Note: the check for `bucket_exists` does not work in the IBM Cos
+            # so we should not re-raise the exception
             self.log.warning("Object Storage bucket already owned by you", exc_info=True)
-            raise ex from ex
         except BucketAlreadyExists as ex:
-            self.log.warning("Object Storage bucket already exists", exc_info=True)
+            self.log.error("Object Storage bucket already exists", exc_info=True)
             raise ex from ex
         except ResponseError as ex:
             self.log.error("Object Storage error", exc_info=True)


elyra/util/cos.py
```diff
@@ -55,9 +55,10 @@ def __initialize_object_store(self):
         try:
             if not self.client.bucket_exists(self.bucket):
                 self.client.make_bucket(self.bucket)
-        except BucketAlreadyOwnedByYou as ex:
+        except BucketAlreadyOwnedByYou:
```
kevin-bates, line 58: This code doesn't make sense to me in the first place.  If we're checking for the bucket's existence (assuming the existence check works), why would we get EITHER of the exceptions `BucketAlreadyOwnedByYou` and `BucketAlreadyExists`??

In my opinion, and given the existence check can't be trusted in the first place, we should call `make_bucket` unconditionally and ignore the two Bucket exceptions.

elyra/util/cos.py
```diff
@@ -55,9 +55,10 @@ def __initialize_object_store(self):
...
+            # Note: the check for `bucket_exists` does not work in the IBM Cos
+            # so we should not re-raise the exception
             self.log.warning("Object Storage bucket already owned by you", exc_info=True)
```
kevin-bates, line 61: This code doesn't make sense to me in the first place.  If we're checking for the bucket's existence (assuming the existence check works), why would we get EITHER of the exceptions `BucketAlreadyOwnedByYou` and `BucketAlreadyExists`??

In my opinion, and given the existence check can't be trusted in the first place, we should call `make_bucket` unconditionally and ignore the two Bucket exceptions.

elyra/util/cos.py
```diff
@@ -55,9 +55,10 @@ def __initialize_object_store(self):
...
-            raise ex from ex
         except BucketAlreadyExists as ex:
             self.log.warning("Object Storage bucket already exists", exc_info=True)
             raise ex from ex
```
kevin-bates, line 64: If we're going to raise this exception then it should be logged as an Error, not a Warning.

elyra/util/cos.py
lresende: I believe creating a bucket is probably a much heavier operation than checking for its existence, and I don't want to penalize the instances where the check for the bucket exists works. Based on that, I would be more inclined to leave the code as-is.

elyra/util/cos.py
lresende: I see this as not an error, you call a function and the end result is to have a bucket created and owned by you. In this case, we are doing more work, as we are trying to create a bucket that already exists, but the end result is the same, which then we should not log as an error.

lresende:
> I left a couple comments.
> 
> Could you please add a description to the PR? That is the first place people look if they're trying to determine the history and list the reason why the change is being made would be helpful.

I have created issue #778 with the description and details for it

elyra/util/cos.py
kevin-bates: I agree. I figured we'd either leave as is or build a wrapper. I think we should still catch/ignore the exceptions either way.

elyra/util/cos.py
kevin-bates: I agree this is a warning situation, but why raise the error then?  This is going to raise an exception during the creation of CosClient instances.  Seems like treating it like AlreadyOwnedByYou is the way to go too.

kevin-bates:
Thanks for creating the issue Luciano.

elyra/util/cos.py
lresende: nvm, I was confusing with the exception we are actually changing....

kevin-bates: I see.   I suppose this case is only necessary for IBM COS, since we can't trust the existence check, and if they don't throw `BucketAlreadyOwnedByYou `, but DO throw `BucketAlreadyExists `, then that implies the bucket is NOT owned by you - is that the thought here?  

If so, I wonder if adding a clause like: `"Object Storage bucket already in use by another user"` would be more helpful - because if it already exists and is owned by me - then why fail the `CosClient` creation and prevent me from using my bucket?

Do we know if that's the case?  Seems like an odd way to let someone know the operation is not allowed.

lresende:
See details on #778

PR #782, opened 2020-07-27
Elyra launcher

lresende:
Use our own Launcher implementation to properly order Elyra category/section and display project logo

Todos:
- [x] Make this the default Launcher when Elyra is installed
- [x] Change the Elyra section logo to be project logo

image

 
           in the file; or

           it.

### DIFF
--- a/packages/theme/src/index.ts
+++ b/packages/theme/src/index.ts
@@ -16,21 +16,42 @@
 
 import { elyraIcon } from '@elyra/ui-components';
 import {
+  ILabShell,
   JupyterFrontEnd,
   JupyterFrontEndPlugin
 } from '@jupyterlab/application';
+import { ICommandPalette, MainAreaWidget } from '@jupyterlab/apputils';
 import { IMainMenu } from '@jupyterlab/mainmenu';
+import { launcherIcon } from '@jupyterlab/ui-components';
 
+import { toArray } from '@lumino/algorithm';
+import { JSONObject } from '@lumino/coreutils';
+import { Widget } from '@lumino/widgets';
+
+import { ILauncher, LauncherModel, Launcher } from './launcher';
 import '../style/index.css';
+
+/**
+ * The command IDs used by the launcher plugin.
+ */
+namespace CommandIDs {
+  export const create = 'launcher:create';
+}
 
 /**
  * Initialization data for the theme extension.
  */
-const extension: JupyterFrontEndPlugin<void> = {
+const extension: JupyterFrontEndPlugin<ILauncher> = {
   id: 'elyra-theme',
   autoStart: true,
-  requires: [IMainMenu],
-  activate: (app: JupyterFrontEnd) => {
+  requires: [IMainMenu, ILabShell],
+  optional: [ICommandPalette],
+  provides: ILauncher,
+  activate: (
+    app: JupyterFrontEnd,
+    labShell: ILabShell,
+    palette: ICommandPalette | null
+  ): ILauncher => {
     console.log('Elyra - theme extension is activated!');
 
     // Find the MainLogo widget in the shell and replace it with the Elyra Logo
@@ -52,7 +73,59 @@ const extension: JupyterFrontEndPlugin<void> = {
       widget = widgets.next();
       console.log(widget);
     }
+
+    //register custom command for using an Elyra custom launcher
+    const { commands } = app;
+    const model = new LauncherModel();
+
+    commands.addCommand(CommandIDs.create, {
+      label: 'New Launcher',
+      execute: (args: JSONObject) => {
+        const cwd = args['cwd'] ? String(args['cwd']) : '';
+        const id = `launcher-${Private.id++}`;
+        const callback = (item: Widget) => {
+          labShell.add(item, 'main', { ref: id });
+        };
+        const launcher = new Launcher({ model, cwd, callback, commands });
+
+        launcher.model = model;
+        launcher.title.icon = launcherIcon;
+        launcher.title.label = 'Launcher';
+
+        const main = new MainAreaWidget({ content: launcher });
+
+        // If there are any other widgets open, remove the launcher close icon.
+        main.title.closable = !!toArray(labShell.widgets('main')).length;
+        main.id = id;
+
+        labShell.add(main, 'main', { activate: args['activate'] as boolean });
+
+        labShell.layoutModified.connect(() => {
+          // If there is only a launcher open, remove the close icon.
+          main.title.closable = toArray(labShell.widgets('main')).length > 1;
+        }, main);
+
+        return main;
+      }
+    });
+
+    if (palette) {
+      palette.addItem({ command: CommandIDs.create, category: 'Launcher' });
+    }
+
+    return model;
   }
 };
+
+/**
+ * The namespace for module private data.
+ */
+namespace Private {
+  /**
+   * The incrementing id used for launcher widgets.
+   */
+  // eslint-disable-next-line
+  export let id = 0;
+}
 
 export default extension;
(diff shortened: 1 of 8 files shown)


packages/theme/src/index.ts
```diff
@@ -16,21 +16,43 @@
...
+import { Widget } from '@lumino/widgets';
+
+import { CustomLauncher } from './launcher';
 import '../style/index.css';
 
+/**
+ * The command IDs used by the launcher plugin.
+ */
+const CommandIDs = {
+  create: 'launcher:create-new'
```
marthacryan, line 38:
```suggestion
  create: 'launcher:create'
```

marthacryan: This needs to match the command that's normally used for the launcher (and the launcher extension has to be disabled) for this to work.

lresende:
This seems to be working now, it would be nice to have the icon on `Elyra section`  changed to the actual `Elyra icon`  instead of the `Pipeline editor`  one.

packages/theme/src/index.ts
```diff
@@ -50,9 +72,61 @@ const extension: JupyterFrontEndPlugin<void> = {
...
+    // Use custom Elyra launcher
+    const { commands } = app;
+    const model = new LauncherModel();
+
+    commands.addCommand(CommandIDs.create, {
+      label: 'New Elyra Launcher',
+      execute: (args: any) => {
+        const cwd = args['cwd'] ? String(args['cwd']) : '';
+        const id = `launcher-${Private.id++}`;
+        const callback = (item: Widget) => {
```
vabarbosa, line 86:
```suggestion
        const callback = (item: Widget): void => {
```

packages/theme/src/launcher.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    // Assemble the final ordered list of categories
+    // based on KNOWN_CATEGORIES.
+    each(KNOWN_CATEGORIES, (category, index) => {
+      React.Children.forEach(launcherCategories, cat => {
+        if (cat.key === category) {
+          categories.push(cat);
+        }
+        if (cat.key === 'Elyra') {
+          // change the icon here...
```
vabarbosa, line 61: the category icon is taken from the [first item in a category](https://github.com/jupyterlab/jupyterlab/blob/master/packages/launcher/src/index.tsx#L201,L210) hence the default to the pipeline editor icon. 

one possible option to change the icon is to use the `iconClass` (property of the command) and set the image accordingly via CSS rather than try to search & replace via the children props

vabarbosa:
would it be worth adding a `Code Snippet` card to the **Elyra** category, which would launch the _Add new code-snippet_ form?

packages/theme/src/launcher.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Launcher, ILauncher } from '@jupyterlab/launcher';
+
+import { each } from '@lumino/algorithm';
+
+import * as React from 'react';
+
+/**
+ * The known categories of launcher items and their default ordering.
+ */
+const KNOWN_CATEGORIES = ['Notebook', 'Console', 'Elyra', 'Other'];
```
vabarbosa, line 26: how about making `Elyra` the second category? any reason it shouldn't immediately follow the Notebook category?

lresende: I thought about even disabling the `Console`  category, but I am not sure how popular it is. The only reason they are together (Notebook and Console) is because they have a 1:1 relationship in terms of cards/tiles.

lresende:
> would it be worth adding a `Code Snippet` card to the **Elyra** category, which would launch the _Add new code-snippet_ form?

Not a bad idea, as long as the extension creates it, it will show up properly here on the launcher.

packages/theme/src/launcher.tsx
lresende: I am ok with that, are you volunteering to help on the css side?

vabarbosa: i can take a look

packages/theme/src/launcher.tsx
kevin-bates: I agree `Console` doesn't appear very useful to me.  I suppose it might provide a better "scratchpad" kind of experience.  I think moving `Elyra` up one slot would be good - although when there are lots of kernel types (which typically won't be the case) it could look a little odd since `Elyra` will break the 1:1 parallelism that folks are used to.  (But that shouldn't be a reason for making the Launcher more useful.)

packages/theme/src/launcher.tsx
vabarbosa: i have pushed the icon update (https://github.com/elyra-ai/elyra/pull/782/commits/35f26a2c379485b5a1b2220b876df5c25f63957a) into this PR:

---
image

---
image

lresende:
Below is how this looks like without console, which indeed makes it much more clean, but I don't feel confortable removing it unless the user can somehow re-enable it. 

image

vabarbosa:
> Below is how this looks like without console, which indeed makes it much more clean, but I don't feel confortable removing it unless the user can somehow re-enable it.
> 

i wouldn't recommend removing **Console** but rather just placing **Elyra** above it.

lresende:
After some discussion, looks like it's better just to move the `Elyra` section above `Others`  and leave everything else as is.

ajbozarth:
@vabarbosa I replaced your fix for the elyra icon with another solution I figured out. `iconClass` support was deprecated in Lab 2.0 and is set to be removed as soon as 3.0 so I would prefer we don't rely on it. As such I did some digging into react and figured out a way to replace the LabIcon itself.

lresende:
@vabarbosa are you ok with the suggested approach? From what @ajbozarth the CSS approach might be deprecated on Lab 3.0. We should also revisit the launcher interface on the lab side to make it more extensible I guess.

vabarbosa:
> @vabarbosa I replaced your fix for the elyra icon with another solution I figured out. `iconClass` support was deprecated in Lab 2.0 and is set to be removed as soon as 3.0 so I would prefer we don't rely on it. As such I did some digging into react and figured out a way to replace the LabIcon itself. 
> 

unless i am misunderstanding, `icon` is being deprecated in favor of `iconClass` or at least that's how i interpreted these comments:

https://github.com/jupyterlab/lumino/blob/master/packages/commands/src/index.ts#L194,L229

https://github.com/jupyterlab/jupyterlab/blob/master/packages/launcher/src/index.tsx#L199,L203

are those comments incorrect? is `iconClass` (not `icon`) what is actually being deprecated? if so, i guess i can go along with the approach.

ajbozarth:
@vabarbosa Ah I see where you're confused. 

`icon` has a messy history. A while back (before I started working with jupyter) `icon` was deprecated in favor of `iconClass` which behaved the same way, being a css class string, but was clearer in its actual meaning. 

But then when `LabIcon` was implemented for Lab 2.0 the deprecated `icon` was overloaded with the new functionality. So what's deprecated is the ability to set `icon` to a string, which would just be aliased to `iconClass`. The new functionality is that `icon` can also be a `LabIcon` (which extends `VirtualElement.IRenderer`). This way of handling icons is the new standard. You can learn more about LabIcon here: https://github.com/jupyterlab/jupyterlab/blob/5555c77a17d48b973c66cb32141844f2ce34e10f/packages/ui-components/CONTRIBUTING.md

ajbozarth:
Also if you want help with understanding how the code in your second link handle all three cases (really deprecated `icon` as a string, recently deprecated `iconClass`, and new `icon` as a `LabIcon`) then I can hope on a 5min call tomorrow to explain it (it took me some time to wrap my head around what those few line were actually accomplishing.

vabarbosa:
thanks for the clarification

lresende:
I guess we are in consensus here. Note that there is still a need to make this area a little more flexible on core and I will follow up with that later on.

PR #784, opened 2020-07-28
Rename metadata extension and other build fixes

ajbozarth:
I renamed `metadata-editor` to `metadata` in preparation of the
changes in #591 

I also included so build dependency fixes I found while documenting
the elyra build tree. 

           in the file; or

           it.


packages/python-editor/tsconfig.json
```diff
@@ -4,10 +4,5 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": ["src/*"],
-  "references": [
-    {
-      "path": "../application"
-    }
-  ]
```
lresende, line 12: is it ok to remove the reference to application here? I believe the `python-editor`  still requires application ?

ajbozarth: nope it doesn't, it's not in package.json at all, I'm not sure why this was here at all, might have been an overzealous copy during my --watch support work

lresende: good, I guess I got tricked by @jupyterlab/application

PR #792, opened 2020-07-29
Use get_event_loop instead of get_running_loop

### LINKED ISSUE #791: Error when submitting thru pipeline processor with python 3.6
Problem occurs when user python environment is 3.6. The asyncio.get_running_loop method is not avail until 3.7. 

      File "/opt/app-root/lib/python3.6/site-packages/tornado/web.py", line 1703, in _execute
        result = await result
      File "/opt/app-root/lib/python3.6/site-packages/elyra/pipeline/handlers.py", line 89, in post
      File "/opt/app-root/lib/python3.6/site-packages/elyra/pipeline/processor.py", line 70, in process
        res = await asyncio.get_running_loop().run_in_executor(None, processor.process, pipeline)
    AttributeError: module 'asyncio' has no attribute 'get_running_loop'

kevin-bates:
Although the docs recommend using `get_running_loop()`, it only exists in Python 3.7.  This changes that to use the `get_event_loop()` method which is essentially the same although the latter will create an event loop if one does not exist.  Since we're running in a tornado server, we're (virtually) guaranteed to have an event loop.

Fixes #791

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -67,7 +67,7 @@ class PipelineProcessorManager(SingletonConfigurable):
             raise RuntimeError('Could not find pipeline processor for [{}]'.format(pipeline.runtime))
 
         processor.root_dir = self.root_dir
-        res = await asyncio.get_running_loop().run_in_executor(None, processor.process, pipeline)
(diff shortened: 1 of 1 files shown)


lresende:
Should we enable tests for Py3.6 if they are not yet enabled?

Issue #704, opened 2020-07-01
Enhanced Troubleshooting in Jupyter: Detailed messages why kernel restarts happened and continous information on memory/cpu usage.


jschaeck:
When Jupyter is used in hosted cloud environments, in a lot of cases, CPU and Memory have certain limits. In these situation it essential that users understand certain characteristics when running their code.

One example is that right now, when the Jupyter Kernel gets an Out of Memory, the kernel is killed by Linux and Jupyter just shows a message that the kernel died with no indication of the root cause.

So better information would help to let the user understand what happened so that she can adjust her container settings. (For example increase memory)

On the other side, when the user uses to much memory, it would also be interesting to see how much percent of available memory is used. So when the user sees that only 20% from a 64GB environment is used, they can go down and save costs.

Perfect would be if after a notebook run, the peak usage could be measured.

Note: In most cases these system use container based approaches like Kubernetes. So it is essential that relative usage is computed with the container memory, not the node memory on which the container is running on.

lresende:
> Note: In most cases these system use container based approaches like Kubernetes. So it is essential that relative usage is computed with the container memory, not the node memory on which the container is running on.

We should also consider the case where the kernel is actually running in a remote/separate location, which then we should consider the remote system resources where the kernel is running, instead of the resources from where Jupyter is running.

PR #795, opened 2020-07-31
[WIP] Add System Monitor extension

ajbozarth:
Adds a system monitor to the right of the JupyterLab menu bar

This adds three external extensions to the Elyra build:

1. [nbresuse](https://github.com/yuvipanda/nbresuse) serverextension - api for accessing memory and cpu usage for the current jupyter lab instance
2. [jupyterlab-topbar-extension](https://github.com/jtpio/jupyterlab-system-monitor) labextension - enables adding widgets to the topbar, the area to the right of the mainmenu
3. [jupyterlab-system-monitor](https://github.com/jtpio/jupyterlab-topbar) labextension - adds a topbar widget that displays the memory and cpu usage returned by the nbresuse api

A user can set a memory_limit in their `jupyter_notebook_config.py` file which will add a percentage bar to the memory UI. An example is included in the default `jupyter_notebook_config.py` file added in this PR.

This is an initial step to address #704
 
           in the file; or

           it.


ajbozarth:
Some quick notes here:

Based on my initial research the extensions added in this PR address our immediate needs and, outside of some potential tinkering with css styling or surfacing of configs better, I'm not sure it's worth the dev hours to redo ourselves.

When we do have an api for accessing kernel metrics in the same way that nbresuse does for the jupyter instance we will want to start from scratch on a UI for that. That UI though may replace or work in parallel to this one, and any decision on that can happen then.

Points of discussion:
- Are we ok leveraging these extensions?
   - None of these extensions are open governance and are all owned by a specific person
   - We will have to write own own UI once the Kernel metric API is available
- Do we want to improve the UI these extensions currently give us?
   - We could submit fixes to the extension repo
   - We could add css fixes to our theme package

ajbozarth:
Screenshot:
Screen Shot 2020-07-31 at 4 33 19 PM

setup.py
```diff
@@ -44,6 +45,7 @@
     author="Elyra Maintainers",
     license="Apache License Version 2.0",
     data_files=[('etc/jupyter/jupyter_notebook_config.d', glob(auto_extension_path)),
+                ('etc/jupyter', glob(notebook_config_path)),
```
lresende, line 48: this will replace any configuration/customization that the user might have done on their notebook as the `jupyter_notebook_config.py` is widely used

Makefile
```diff
@@ -101,6 +104,9 @@ install-ui: build-ui
 install-external-extensions:
 	pip install --upgrade jupyterlab-git==$(GIT_VERSION)
 	jupyter labextension install --no-build @jupyterlab/toc@$(TOC_VERSION)
+	pip install --upgrade nbresuse==$(NBRESUSE_VERSION)
+	jupyter labextension install --no-build jupyterlab-topbar-extension@$(TOPBAR_VERSION)
+	jupyter labextension install --no-build jupyterlab-system-monitor@$(MONITOR_VERSION)
```
lresende, line 109: I would prefer not to depend on a user-specific extension, to avoid the issue that happened with `phosphor`.

Makefile
```diff
@@ -101,6 +104,9 @@ install-ui: build-ui
 install-external-extensions:
 	pip install --upgrade jupyterlab-git==$(GIT_VERSION)
 	jupyter labextension install --no-build @jupyterlab/toc@$(TOC_VERSION)
+	pip install --upgrade nbresuse==$(NBRESUSE_VERSION)
```
lresende, line 107: See https://github.com/yuvipanda/nbresuse/issues/62

setup.py
ajbozarth: actually I tested that, it will not override or replace any settings in ~/.jupyter/jupyter_notebook_config.py`

Makefile
ajbozarth: In that case we will need to write the API as well because `nbresuse` is also a user specific extension

setup.py
```diff
@@ -31,6 +31,7 @@
 
 npm_packages_path = "./dist/*.tgz"
 auto_extension_path = "./etc/config/jupyter_notebook_config.d/*.json"
+notebook_config_path = "./etc/config/*.py"
```
kevin-bates, line 34: Could use just use a .json file (I suspect it would be `nbresuse.json`) and include it in `auto_extension_path` along with the others?  This would be much preferred over this approach.  Isn't something like this needed to indicate the server extension is enabled anyway?

I would expect it to look something like:
```json
{
  "NotebookApp": {
    "nbserver_extensions": {
      "nbresuse": true
    },
    "ResourceUseDisplay": {
      "track_cpu_percent": true
    }
  }
}
```

test_requirements.txt
```diff
@@ -3,3 +3,4 @@ pytest-tornasync
 pytest-console-scripts
 flake8
 hacking
+requests-unixsocket
```
kevin-bates, line 6: This can be removed now that Notebook 6.1.1 is available and addresses this issue.

setup.py
ajbozarth: I tried using a json file like this, but I wasn't able to get it to work

ajbozarth:
Updated with my wip work prior to discovering that lab core already show the memory info in the bottom task bar by default when nbresuse is installed. This will be on hold for now as the short term need is addressed by just installing nbresuse

lresende:
Closing for now as nbresuse seems to be providing info on the lab status bar.

PR #805, opened 2020-08-03
fixed docker image

romeokienzler:

           in the file; or

           it.

### DIFF
--- a/etc/docker/elyra/Dockerfile
+++ b/etc/docker/elyra/Dockerfile
@@ -16,7 +16,7 @@
 #
 
 # Ubuntu 18.04 LTS - Bionic
-FROM jupyterhub/k8s-singleuser-sample:0.9.0
+FROM jupyterhub/k8s-singleuser-sample:0.9.1
 
 USER root
 
@@ -39,9 +39,11 @@ RUN python -m pip install --upgrade pip && \
     npm install -g yarn && \
     npm install -g npm && \
     cd /tmp && git clone https://github.com/elyra-ai/elyra.git && \
-    cd /tmp/elyra && git checkout tags/v1.0.0rc3 -b v1.0.0rc3 && make install
+    cd /tmp/elyra && git checkout tags/v1.0.0rc2 -b v1.0.0rc2 && make install
 
 WORKDIR /home/jovyan/work
 
-CMD ["/usr/local/bin/start-elyra.sh"]
+# set password to "password"
+ADD jupyter_notebook_config.json /home/jovyan/.jupyter/
+CMD ["jupyter-lab"]
 
--- a/etc/docker/elyra/Dockerfile
+++ b/etc/docker/elyra/Dockerfile
@@ -29,6 +29,9 @@ RUN chmod ugo+x /usr/local/bin/start-elyra.sh && \
     apt-get update && apt-get install -y yarn && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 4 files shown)


etc/docker/elyra/Dockerfile
```diff
@@ -39,9 +39,11 @@ RUN python -m pip install --upgrade pip && \
     npm install -g yarn && \
     npm install -g npm && \
     cd /tmp && git clone https://github.com/elyra-ai/elyra.git && \
-    cd /tmp/elyra && git checkout tags/v1.0.0rc3 -b v1.0.0rc3 && make install
+    cd /tmp/elyra && git checkout tags/v1.0.0rc2 -b v1.0.0rc2 && make install
```
lresende, line 42: this should be rc3, you might need to update your master

etc/docker/elyra/Dockerfile
```diff
@@ -39,9 +39,11 @@ RUN python -m pip install --upgrade pip && \
...
 
 WORKDIR /home/jovyan/work
 
-CMD ["/usr/local/bin/start-elyra.sh"]
```
lresende, line 46: this would break the integration with JupyterHub and JEG

etc/docker/elyra/Dockerfile
```diff
@@ -39,9 +39,11 @@ RUN python -m pip install --upgrade pip && \
...
+# set password to "password"
+ADD jupyter_notebook_config.json /home/jovyan/.jupyter/
```
lresende, line 47: how would this help? does this work on your system because you have this key stored/saved on your browser? What happens when someone tries this for the first time and that it does not see the URL with the key to use to access the notebook UI?

romeokienzler: he just uses "password" as password, without the need to copy-paste the token. also in "deamon" mode, getting the token is even more complicated. that way it can be stored in the browser.

etc/docker/elyra/Dockerfile
romeokienzler: done

etc/docker/elyra/Dockerfile
romeokienzler: let me fix this

etc/docker/elyra/Dockerfile
romeokienzler: please let me know if you disagree and we can go pack to the token based auth again

romeokienzler: now moved back to jupyterhub and set user to jovyan and password to jupyter which apears to be the default, is this ok?

etc/docker/elyra/Dockerfile
romeokienzler: What is JEG? I've changed from jupyter-lab to jupyterhub, but I can't run jupyter-lab from there, only jupyter notebooks...

romeokienzler:
@lresende let me rollback and start over @ptitzler gave me the correct comand to run the image, now it works

`docker run -it -p 8888:8888  elyra/elyra:1.0.0rc2 jupyter lab --debug`

Will add kfp config capabilities to the existing one and update the documentation

lresende:
Note that with rc3, we have added the UI to configure runtimes.

romeokienzler:
@lresende wow this is great! .. just fyi - tried it out and it didn't work - maybe I've done smth. wrong, but just in case filed an [issue](https://github.com/elyra-ai/elyra/issues/827)

PR #810, opened 2020-08-03
Metadata Editor and Widget Bug Fixes

### LINKED ISSUE #798: [Metadata editor regression?] Only first required property is checked by UI
1.0.0 rc3
I've noticed a UI behavior change that might be a regression:

- Create a new code snippet and enter only a name (there are more properties marked as required) OR
 create a new runtime and enter only a name (there are more properties marked as required)
- Previously the metadata editor would label each input field with "This field is required." (as shown for the name below)

image

- Click "save & close"

An error dialog is displayed:

```
Error making request
Validation failed for instance 'ddd' using the kfp schema with error: 'api_endpoint' is a required property.
Check the JupyterLab log for more details at 2020-08-03 09:11:19
```

For the sake of consistency the behavior should be the same for all input fields.

### LINKED ISSUE #802: [Fit and finish] Runtimes widget
- Display runtime type (e.g. Kubeflow Pipelines or Apache Airflow). While this might not seem that useful until another runtime is supported we should incorporate the information into the design.
 - Change "Runtime history" label to "Pipeline run history" since there isn't really a runtime history, only a pipeline run history.
image

### LINKED ISSUE #806: Runtimes widget disappears when no runtime is defined
1.0.0.rc3

- do not create a runtime

**Recreate 1:**

 - create a basic hello world notebook (any notebook will do)
 - create a new pipeline that includes the notebook
 - save the pipeline
 - click on the open runtimes button in the canvas
 - an error message is displayed ("no runtimes found")
  - the runtimes widget disappears

**Recreate 2:**
- Click on the runtimes icon on the left hand side

image

### LINKED ISSUE #814: [Fit and finish] Pipeline submission confirmation should display `display_name` instead of `name`
1.0.0. rc3

It appears that the runtime configuration name that the user enters in the UI (e.g. `def;ghi` or `abc def`) is used to generate a "safe" string that is used as an internal identifier , as seen in the output of the `list runtimes ...` command:

```
{
  "name": "defghi",
  "display_name": "def;ghi",
  "metadata": {
    "description": "hello",
    "api_endpoint": "http://...:31380/pipeline/",
    "cos_endpoint": "http://...:31671",
    "cos_username": "...",
    "cos_password": "...",
    "cos_bucket": "..."
  }
```

This identifier is displayed in the confirmation message instead of the display name.

image

ajbozarth:
Various Bug Fixes as a followup to #591

- Fixed required validation in MetadataEditor. Update to strict
equality check cause the `undefined` case to no longer be caught
by the `null` check
- Text updates to RuntimesWidget to clarify rutime type and api url
- Fixed support for RuntimeWidget when no runtimes are configured
- Focus on Name field when opening a Metadata Editor tab
- Update pipeline submission success dialog to use `display_name`

Fixes #798
Fixes #802
Fixes #806
Fixes #814 

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -123,7 +123,8 @@ export class MetadataEditor extends ReactWidget {
     for (const schemaField in this.schema) {
       if (
         this.requiredFields.includes(schemaField) &&
-        (this.metadata[schemaField] === null ||
+        (this.metadata[schemaField] === undefined ||
+          this.metadata[schemaField] === null ||
           this.metadata[schemaField] === '' ||
           this.metadata[schemaField] === [] ||
           this.metadata[schemaField] === '(No selection)')
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -110,13 +110,13 @@ export class MetadataEditor extends ReactWidget {
     this.update();
   }
 
-  private isFieldEmpty(schemaField: any): boolean {
+  private isValueEmpty(schemaValue: any): boolean {
     return (
-      schemaField === undefined ||
-      schemaField === null ||
-      schemaField === '' ||
-      schemaField === [] ||
-      schemaField === '(No selection)'
+      schemaValue === undefined ||
+      schemaValue === null ||
+      schemaValue === '' ||
+      schemaValue === [] ||
+      schemaValue === '(No selection)'
     );
   }
 
@@ -133,7 +133,7 @@ export class MetadataEditor extends ReactWidget {
     for (const schemaField in this.schema) {
       if (
         this.requiredFields.includes(schemaField) &&
-        this.isFieldEmpty(this.metadata[schemaField])
+        this.isValueEmpty(this.metadata[schemaField])
       ) {
         this.invalidForm = true;
         this.schema[schemaField].uihints.intent = Intent.DANGER;
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -55,6 +55,7 @@ export class MetadataEditor extends ReactWidget {
   requiredFields: string[];
   invalidForm: boolean;
   showSecure: IDictionary<boolean>;
+  widgetClass: string;
 
   schema: IDictionary<any> = {};
   allMetadata: IDictionary<any>[] = [];
@@ -67,6 +68,9 @@ export class MetadataEditor extends ReactWidget {
     this.schemaName = props.schema;
     this.onSave = props.onSave;
     this.name = props.name;
+
+    this.widgetClass = `elyra-metadataEditor-${this.name ? this.name : 'new'}`;
+    this.addClass(this.widgetClass);
 
     this.handleTextInputChange = this.handleTextInputChange.bind(this);
     this.handleDropdownChange = this.handleDropdownChange.bind(this);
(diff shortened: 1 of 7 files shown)


packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -123,7 +123,8 @@ export class MetadataEditor extends ReactWidget {
     for (const schemaField in this.schema) {
       if (
         this.requiredFields.includes(schemaField) &&
-        (this.metadata[schemaField] === null ||
+        (this.metadata[schemaField] === undefined ||
+          this.metadata[schemaField] === null ||
           this.metadata[schemaField] === '' ||
```
lresende, line 128: What if you do something like  `&& (this.metadata[schemaField] || this.metadata[schemaField] === '(No selection)')` where the first comparison should take care of null, undefined and empty string

kevin-bates: How about a local function `this.is_field_set(schemaField: str): boolean`?  Seems like there are way too many variations to deal with.

ajbozarth: @lresende we can't do that because 0 or false may be valid and are also considered falsy when evaluating `this.metadata[schemaField]`, also `[]` is considered truthy. I do like the idea of a separate function like @kevin-bates suggested, I'll do that.

ajbozarth: pushed an update

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -110,6 +110,16 @@ export class MetadataEditor extends ReactWidget {
...
 
+  private isFieldEmpty(schemaField: any): boolean {
+    return (
+      schemaField === undefined ||
+      schemaField === null ||
+      schemaField === '' ||
+      schemaField === [] ||
+      schemaField === '(No selection)'
+    );
+  }
```
kevin-bates, line 121: The way this is implemented, it's not necessary to be a class method but works.  I figured it would act on `this.metadata` and take the field name as a parameter.  

If we could at least change the parameter name to `schemaValue` then that might eliminate some potential confusion.  But then the question arises - should this be named `IsValueEmpty()`?  :smile:

ajbozarth: this would be a good function to put in a util, but I didn't think it was worth creating a util.ts just for this function (thus making it a private class method).

As for naming, you are correct, based on my current implementation it should be value not field. I'll push an update promptly

ptitzler:
@ajbozart just noticed a regression. The tool tip used to be correct.


image

ajbozarth:
Pushed an update that addressed the following issues @ptitzler raised:

- Focus on Name field when opening a Metadata Editor (including switching between tabs)
- Fixed race condition causing `Create new undefined` tooltip
- Fixed #814

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -110,6 +114,16 @@ export class MetadataEditor extends ReactWidget {
     this.update();
   }
 
+  private isValueEmpty(schemaValue: any): boolean {
+    return (
+      schemaValue === undefined ||
+      schemaValue === null ||
+      schemaValue === '' ||
```
karlaspuldaro, line 121: We can just use `!schemaValue` instead of these 3 value checks

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -110,6 +114,16 @@ export class MetadataEditor extends ReactWidget {
...
+      schemaValue === [] ||
```
karlaspuldaro, line 122: This is probably not gonna work since an array is an object, if we do [] === [] we get `false`
To be safe, we could do `Array.isArray(schemaValue) && schemaValue.length`

packages/metadata-common/src/MetadataEditor.tsx
ajbozarth: I mentioned this in a review comment from Luciano but we can't use that because it would evaluate 0 and false as falsy as well, but those might be valid content.

packages/metadata-common/src/MetadataEditor.tsx
ajbozarth: nice catch, I'll make that change

marthacryan:
Tried locally and it worked as expected for all of the issues mentioned. LGTM

PR #820, opened 2020-08-05
Add section for distributing pipeline best practice

### LINKED ISSUE #819: Add basic best practice documentation for distributing pipelines
As #608 points out, distributing/sharing pipelines can be problematic due to the current implementation.  We should produce minimal documentation that helps users navigate this scenario, but will address this more fully in a future release.

kevin-bates:
This adds a couple of paragraphs describing best practices for sharing/distributing pipelines after first describing the component relationships.

I also moved the developer-specific pipeline sections (file format, etc.) to a new Pipelines topic in the Developer Guide and updated its fields to correspond to version 1, as well as fixed a non-fatal build issue.

Resolves #819 
 
           in the file; or

           it.


docs/source/conf.py
```diff
@@ -88,7 +88,7 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    'description': "Elyra extends JupyterLab Notebooks with an AI centric approach",
+#    'description': "Elyra extends JupyterLab Notebooks with an AI centric approach",
```
lresende, line 91: Curious to what is the issue here?

docs/source/user_guide/pipelines.md
```diff
@@ -66,167 +66,32 @@ needed for the pipeline are present.
...
+
+#### Distributing Pipelines - Best Practice
+Prior to distributing your pipeline - which includes preserving the component relationships - it is best to commit these files (and directories) to a github repository.  An alternative approach would be to archive the files using `tar` or `zip`, but, again preserving the component relationships.
+
+When deploying a shared or distributed pipeline repository or archive, it is **very important** that the pipeline notebooks be extracted into the **same directory hierarchy relative to the hosting notebook server's working directory**.
+
+
+For example, if you archived your pipeline into a github repository named `my_analytics_pipeline` from the notebook server's working directory, then consumers of this archive would need to do the following:
+
+1. Clone your repository
```
lresende, line 83: I believe there are two options/best practices here: 
1 - The one you describe, which requires one single project per workspace
2 - publishing your project inside a folder, probably the same name as the repo, which should allow us to have multiple github projects in one workspace

Thoughts?  Should we also mention the number 2 ?

docs/source/conf.py
kevin-bates: 'description' isn't a valid html_theme option and the following is produced:
```
preparing documents... WARNING: unsupported theme option 'description' given
```

docs/source/user_guide/pipelines.md
kevin-bates: Yes - I agree.  It's all a matter of capturing the project relative to the notebook directory.  Thanks for the suggestion.

docs/source/conf.py
kevin-bates: @lresende - I left it commented out in case there's another place it should go, but would prefer it be removed.

lresende: We can remove it.

PR #835, opened 2020-08-07
Add command to palette to open Runtime Images UI

### LINKED ISSUE #829: Pipelines: As a user I want to define/use a custom runtime/Docker image to run notebooks in using the UI
A user can currently only select from a handful of predefined runtime images. Sometimes these images might not meet the users' requirements and therefore we should provide the ability to specify any image as a source.

Additional considerations:
 - Could images in private repositories be supported?

ajbozarth:
As part of the previous work on Metadata Widget we added a generic
command to open a UI for any metadata namespace/schema

This PR adds a command call to open the ui for runtime images to
the palette.

Fixes #829

Screen Shot 2020-08-06 at 4 46 26 PM
Screen Shot 2020-08-06 at 4 46 13 PM

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata/src/index.ts
+++ b/packages/metadata/src/index.ts
@@ -145,6 +145,7 @@ const extension: JupyterFrontEndPlugin<void> = {
 
     const openMetadataCommand: string = commandIDs.openMetadata;
     app.commands.addCommand(openMetadataCommand, {
+      label: (args: any) => args['label'],
       execute: (args: any) => {
         // Rank has been chosen somewhat arbitrarily to give priority
         // to the running sessions widget in the sidebar.
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -64,10 +64,13 @@ import { PIPELINE_CURRENT_VERSION } from './constants';
 import * as i18nData from './en.json';
 import * as palette from './palette.json';
 import { PipelineExportDialog } from './PipelineExportDialog';
-import { PipelineService } from './PipelineService';
+import {
+  PipelineService,
+  KFP_SCHEMA,
+  RUNTIMES_NAMESPACE
+} from './PipelineService';
 import { PipelineSubmissionDialog } from './PipelineSubmissionDialog';
 import * as properties from './properties.json';
-import { KFP_SCHEMA, RUNTIMES_NAMESPACE } from './RuntimesWidget';
 import Utils from './utils';
 
 const PIPELINE_CLASS = 'elyra-PipelineEditor';
@@ -107,6 +110,7 @@ interface IValidationError {
 
 export const commandIDs = {
   openPipelineEditor: 'pipeline-editor:open',
+  openMetadata: 'elyra-metadata:open',
   openDocManager: 'docmanager:open',
   newDocManager: 'docmanager:new-untitled',
   submitNotebook: 'notebook:submit'
--- a/packages/pipeline-editor/src/PipelineService.tsx
+++ b/packages/pipeline-editor/src/PipelineService.tsx
@@ -25,6 +25,11 @@ import * as React from 'react';
 
 import Utils from './utils';
 
+export const RUNTIMES_NAMESPACE = 'runtimes';
+export const KFP_SCHEMA = 'kfp';
+export const RUNTIME_IMAGES_NAMESPACE = 'runtime-images';
+export const RUNTIME_IMAGE_SCHEMA = 'runtime-image';
+
 export class PipelineService {
   /**
    * Returns a list of external runtime configurations available as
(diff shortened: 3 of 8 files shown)


ajbozarth:
@ptitzler does this handle what you wanted in #829 ?

ptitzler:
> @ptitzler does this handle what you wanted in #829 ?

Please see new comment in that issue about where/how it is exposed.

A few additional thoughts that are mostly specific to this editor instance (but some of the points raised are of generic nature and should be reviewed):

### Summary view
image

It's be great if the information could be displayed in a formatted way (which might include a link to the image in the public registry), not just as a JSON dump. This would be similar to how we currently display runtime configuration metadata.

### Add/edit view

image

- Not sure if the metadata editor already supports tool tips yet for each input property. If it doesn't I'd be happy to open a generic ER. A tool tip (or some other means to convey constraints) would enable the user to make a more informed decision what a valid value might be. 
- Label "Runtime image" => "Docker image name" (for clarity)
- See also https://github.com/elyra-ai/elyra/issues/838

### Input validation

Having tried adding a runtime image using the CLI I've noticed that no validation is performed and any bogus name is accepted. While there might be valid reasons why an image name cannot be resolved, there's also a chance that the user accidentally mistyped the name (or tag). As is, this problem would (I guess, need to test this to confirm) only surface when pipeline execution fails. Since that is pretty late in the workflow it'd be nice if the editor (or backend) could try to validate that the specified image can be accessed when a new runtime image is added or an image is edited. This way a user would know right away that there might be a problem and has the option to correct the issue or take the risk that things will fail.

### Dangling references

If I'm not mistaken this editor is the first of its kind that might cause dangling references. For example:
- create a custom runtime image
- reference this runtime image in a notebook pipeline
- delete the runtime image; the notebook pipeline is now invalid

We need to think about the implications:
- Should the editor display a warning that this runtime image _might_ be used somewhere and that something _might_ break as a result? (Elyra won't know where it is used.)
- Do the consumers of this piece of metadata already try to catch this error condition?

kevin-bates:
Regarding _dangling references_ and _changing image references_, these are somewhat conflicting solutions.  If you solve one, you have the other and vice versa.  I had noticed the same thing when I discovered that we only store the image name in the pipeline node, not the metadata instance reference.  When bringing this up, we decided to err on the side of reproducibility (and address potential dangling references) rather than resolve the _changing image references_.  I'm not sure which is better.  The other reason we have (currently) gone with the current approach is because that's what were doing already. :smile:

I do agree it would be nice to search through pipelines to see where a particular image or metadata instance might be referenced, however.

Regarding _input validation_ I also agree and is another side-effect of schema-driven approaches and _yet another example_ of needing [_meta-behaviors_](https://github.com/elyra-ai/elyra/issues/518) or [_custom metadata classes_](https://github.com/elyra-ai/elyra/pull/725) so that property-specific actions can be performed.

ajbozarth:
@ptitzler in response to both your comment here and on Issue https://github.com/elyra-ai/elyra/issues/829#issuecomment-670406564 

I don't currently have the bandwidth this sprint to address your concerns (I opened this PR as a "quick fix" while working on another issue). It does seem like most of your concerns on this though are of a "follow-up" level, that is that this PR does the job, but it should do it better. As such I will leave this as is for now, if other reviews are good we can merge this as part of 1.1 and address your concerns as a follow up PR in 1.2 next sprint when I have bandwidth again.

ptitzler:
> are of a "follow-up" level

Yes, that's fine. I just wanted to get the ball rolling. I will transfer the topics to separate issues where they can be discussed in more detail.

karlaspuldaro:
Works great. LGTM :)

vabarbosa:
how often do people actually look through or use the Commands palette? would it be worthwhile adding a corresponding tile in the Elyra section of the Launcher where it would more prominent?

ajbozarth:
> how often do people actually look through or use the Commands palette? would it be worthwhile adding a corresponding tile in the Elyra section of the Launcher where it would more noticeable?

AFAIK the Launcher is essentially a pretty UI for `File > New` and is designed for opening new doc files, not widgets in general

vabarbosa:
> > how often do people actually look through or use the Commands palette? would it be worthwhile adding a corresponding tile in the Elyra section of the Launcher where it would more noticeable?
> 
> AFAIK the Launcher is essentially a pretty UI for `File > New` and is designed for opening new doc files, not widgets in general

there is already  "Show Contextual Help" tile (in the Other section). i don't believe that is a `File > New` action.

ajbozarth:
@vabarbosa I didn't realize that.

I brought up the point in this mornings dev meeting and after a short discussion we came to this point: 
This functionality is too "low level" and minor to put on the very high visibility Launcher. Though adding things to the Launcher might make sense in the future, this is not the best case of one. There's also the fact that as of now there's no way to close a sidebar tab, meaning if a user opens the UI the only way to close it is to refresh the page.

The short of it all is that this is a very minor feature and wont be used much, the important part is the ability to use it at all. As such a command on the palette is the most straight forward location for it. This also allows for the addition of more such commands when future metadata instance are added that don't warrant a hight profile UI (like code snippets does)

PR #845, opened 2020-08-10
Add support to run pipelines in-place locally

### LINKED ISSUE #842: Enable running Notebook pipelines locally
We should investigate how we could enable running pipelines locally, assuming the environment works and the resources are there, to simplify some workflow that doesn't need the isolation and more powerful environments.

>  improved pipeline viewer (like Elyra, but for local execution highlighting which notebooks need to be re-run after a change has occurred; and allowing to re-run entire branch with a single click; this will be more of reproducible analysis flow tool than the pipeline as meant in data science).

lresende:
This provides support for running pipelines on your local environment without requiring availability/connectivity to a remote system. In this scenario, we will assume that the pipeline is composed of notebooks that already run locally and in-place just using the pipeline to identify the proper order of execution. Things like isolation, running on a different environment (e.g. docker image) and all these details are out of the scope of this PR.

TODO:
- [x] Refactor the UI to enable submitting for multiple runtimes
- [x] Run in-place

Co-authored-by: Kevin Bates <kevin.bates@us.ibm.com>
Fixes #842 
 

## Current UI:

Submission Dialog:
image

Success status:
image

Error executing pipeline:
image

Updated notebook after an error:
image

           in the file; or

           it.


packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -34,6 +34,11 @@ export class PipelineService {
   static async getRuntimes(showError = true): Promise<any> {
     const runtimes = await FrontendServices.getMetadata('runtimes');
 
+    console.log(runtimes);
+
+    const local_runtime: any = { name: 'local', display_name: 'local' };
+    runtimes.unshift(JSON.parse(JSON.stringify(local_runtime)));
+
```
ajbozarth, line 41: Putting this here for the quick POC makes sense, but a comment on a better plan for the final version:

The addition of local to the list should not happen in this service function, but in a helper function in PipelineEditor since SubmitNotebook also calls this function.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+    processor. If a notebook don't run from your regular `run all` it will fails on the
+    same way when processed by this processor. Also, any relationship or specific capabilities
+    associated with a particular runtime is not supported by local mode and any additional properties
+    other then the specific file to be executed is ignored by the local processor.
+
+    Note: Execution happens in-place and a lodger of runs are available at $TMPFILE/elyra/pipeline-name-<timestamp>
+    """
+    _type = 'local'
+
+    def __init__(self, root_dir=None):
```
kevin-bates, line 40: Should call superclass initializer...
```suggestion
    def __init__(self, root_dir=None):
        super(LocalPipelineProcessor, self).__init__(root_dir=root_dir)
```

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+        if root_dir:
+            self._root_dir = root_dir
```
kevin-bates, line 42: Since `self._root_dir` will be initialized to `None`, there's no need to check `root_dir`'s value first.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+        return self._type
+
+    def process(self, pipeline):
+        timestamp = datetime.now().strftime('%m%d%H%M%S')
+        pipeline_name = f'{pipeline.name}-{timestamp}'
+
+        pipeline_work_dir = self._get_pipeline_work_dir(pipeline_name)
+        self.log.info(f'Processing Pipeline : {pipeline.name} at {pipeline_work_dir}')
+
+        # TODO: consider ordering
```
kevin-bates, line 55: Ordering seems like an issue.  Does the canvas provide any kind of help in that respect?  If not, it seems like you'd need to locate all nodes that do not have inputs and execute those first.  After that, it gets complicated. :smile:

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+        for operation in pipeline.operations.values():
+            notebook = self.get_absolute_path(operation.filename)
+            assert os.path.isfile(notebook), "Invalid '{}'".format(notebook)
```
kevin-bates, line 58: Do we want an `AssertionError` here, or `FileNotFoundError`?  A message like "Invalid 'xyz.ipynb'" implies an invalid format of sorts, when the file is just not present.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+
+            self._execute_file(pipeline_work_dir, notebook)
+
+        object_storage_path = f'{pipeline_work_dir}'
```
kevin-bates, line 62: Why does `pipeline_work_dir` need to be formatted?  Why not just pass that variable as the 3rd argument to the `PipelineProcessorResponse()` constructor?

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+        return PipelineProcessorResponse('', '', object_storage_path)
```
kevin-bates, line 63: Now that the arguments no longer take default values, it seems like they should default to the empty string.  Using `None` is cleaner and the response's consumer will need to make a check either way.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def export(self, pipeline, pipeline_export_format, pipeline_export_path, overwrite):
+        raise NotImplementedError('Local pipelines does not support export functionality')
```
kevin-bates, line 66: Hmm - seems like a local export would be a fantastic way to facilitate capture of a "project" for sharing with others!

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _get_pipeline_work_dir(self, pipeline_name):
+        log_dir = os.path.join(tempfile.tempdir, 'elyra', pipeline_name)
+        if not os.path.exists(log_dir):
+            os.mkdir(log_dir)
```
kevin-bates, line 71: This can be simplified via [`os.makedirs`](https://docs.python.org/3/library/os.html#os.makedirs):
```suggestion
        os.makedirs(log_dir, exist_ok=True)
```
Since this could contain sensitive information, should the mode of the directory be `0700` like it is for metadata instances?

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+        return log_dir
+
+    def _execute_file(self, work_dir, filename):
```
kevin-bates, line 74: I find `filename` too generic and a bit misleading since this also contains a path portion.  Would `notebook_path` work for you - or something similar?

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+            try:
+                # Try executing with configured python kernel
+                papermill.execute_notebook(
+                    input_path=filename,
+                    output_path=filename,
+                    stdout_file=stdout_file,
+                    stderr_file=stderr_file,
+                    cwd=notebook_dir
+                )
+            except NoSuchKernel:
```
kevin-bates, line 92: Let's not introduce a dependency on `jupyter_client` just for this - which will break when kernel providers are used anyway.  Since `NoSuchKernel` is a subclass of `KeyError`, could we just catch `KeyError` and, if you really need to, check the class name of the error and compare that to the string 'NoSuchKernel'?

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+                # force default python kernel
+                papermill.execute_notebook(
+                    input_path=filename,
+                    output_path=filename,
+                    stdout_file=stdout_file,
+                    stderr_file=stderr_file,
+                    cwd=notebook_dir,
+                    kernel_name="python3"
```
kevin-bates, line 100: Actually, per the large comment for the class, this shouldn't happen.  In addition, it will fail for non-python kernels anyway.  As stated above: _If a notebook doesn't run using `run all` it will fail in the same way when processed by this processor._ so we shouldn't make any attempts to run the notebook on any error.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,107 @@
...
+                )
+            except Exception:
+                self.log(f'Internal error executing {filename}')
```
kevin-bates, line 103:
```suggestion
                self.log.error(f'Internal error executing {filename}')
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1049,6 +1052,9 @@ export class PipelineEditor extends React.Component<
     }
 
     const runtimes = await PipelineService.getRuntimes();
+    const local_runtime: any = { name: 'local', display_name: 'local' };
+    runtimes.unshift(JSON.parse(JSON.stringify(local_runtime)));
```
kevin-bates, line 1056: Aren't people going to ask - _How do I edit this 'local' runtime?_

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1066,13 +1072,14 @@ export class PipelineEditor extends React.Component<
     // prepare pipeline submission details
     const pipelineFlow = this.canvasController.getPipelineFlow();
 
+    const runtime_config = dialogResult.value.runtime_config;
+    const runtime =
+      PipelineService.getRuntimeName(runtime_config, runtimes) || 'local';
```
kevin-bates, line 1077: Does this mean that 'local' will be the default runtime so if no runtimes have been configured, they will ALWAYS be able to run 'local'?

Are there any places where the front-end looks up the runtime to get more information?

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -64,6 +64,15 @@ export class PipelineService {
...
   }
 
+  /**
+   * The runtime name is currently based on the schema name (one schema per runtime)
+   * @param name
+   * @param metadataArr
+   */
+  static getRuntimeName(name: string, metadataArr: IDictionary<any>[]): string {
+    return metadataArr.find(r => r['name'] === name)['schema_name'];
+  }
```
kevin-bates, line 74: What is the schema for the local runtime?

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -83,28 +92,49 @@ export class PipelineService {
...
+            rel="noopener noreferrer"
+          >
+            object storage
+          </a>
+          .
+        </p>
+      );
+    } else {
+      // pipeline executed in-place locally
+      dialogTitle = 'Pipeline execution in-place locally succeeded';
```
kevin-bates, line 127: The temporary directory will contain failure information as well.  How does a user determine what went wrong with local execution?

elyra/pipeline/processor_local.py
lresende: yes, I believe that's the only remaining thing, and I am working on it.

elyra/pipeline/processor_local.py
lresende: Where on the hierarchy root_dir is considered? is that some configurable/traits constructor magic together with kwargs?

packages/pipeline-editor/src/PipelineEditorWidget.tsx
lresende: This means that, if it can find a selection on the list of runtimes, uses `local` which is the only item that is inserted to the list on the UI. There is another call `PipelineService.getDisplayName`

packages/pipeline-editor/src/PipelineService.tsx
lresende: Papermill updates the notebook, with failure details. The location of the execution logs is also displayed to the user on the response dialog box in case they need to find more info or investigate any issues.

kevin-bates: Sorry, I missed the fact that `_execute_file()` eats all errors and thought an exception would bypass the creation of the `PipelineProcessorResponse()` instance - like it does for the other runtime.

elyra/pipeline/processor_local.py
kevin-bates: You probably don't need to pass root-dir to the superclass, but when I have this in my IDE it warns that the superclass `__init__()` is not being called.  

However, fun fact I've recently learned, you can pass traits into constructors and they will _automagically_ get "connected" to their corresponding trait on that class. :smile:

elyra/pipeline/sort.py
```diff
@@ -0,0 +1,52 @@
...
+    print('>>>')
+    for operation in operations_by_id.values():
+        if operation not in ordered_operations:
+            print(f'>>> processing operation [{operation.name}]')
+            # operation is a root node
+            if not operation.parent_operations:
+                print(f'   adding root operation [{operation.name}] to ordered list')
+                ordered_operations.append(operation)
+            else:
+                if operation not in ordered_operations:
```
kevin-bates, line 31: This is redundant.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1049,6 +1052,9 @@ export class PipelineEditor extends React.Component<
     }
 
     const runtimes = await PipelineService.getRuntimes();
+    const local_runtime: any = { name: 'local', display_name: 'local' };
```
ajbozarth, line 1055: I'm not sure `local` is the best value for `display_name`, I feel like we should figure out something better.

Based on text in the results dialog perhaps `Run in-place locally` would be a good value

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -83,28 +92,49 @@ export class PipelineService {
...
-          object storage
-        </a>
-        .
-      </p>
-    );
+    let dialogTitle;
+    let dialogBody;
+    if (response['run_url']) {
+      // pipeline executed remotely in a runtime of choice
+      dialogTitle = 'Pipeline submission to ' + runtimeName + ' succeeded';
```
ajbozarth, line 99: This text is also used for notebook submission, we should not refer to pipelines here

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1049,6 +1052,9 @@ export class PipelineEditor extends React.Component<
     }
 
     const runtimes = await PipelineService.getRuntimes();
```
ajbozarth, line 1054: Given we're now adding a local runtime to this list we should update this so it will still work if no runtimes are configured
```suggestion
    const runtimes = await PipelineService.getRuntimes(false);
```

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -83,28 +92,49 @@ export class PipelineService {
...
+      dialogBody = (
+        <p>
+          Check the status of your pipeline at{' '}
```
ajbozarth, line 102: This text is also used for notebook submission, we should not refer to pipelines here

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth: The text here feels off, might be an issue with word ordering.

elyra/pipeline/processor_local.py
lresende: I created an issue and will solve this problem more globally as it's the same current behavior as in the kfp runtime

packages/pipeline-editor/src/PipelineService.tsx
lresende: it will fallback to || local on the invocation of this method

lresende:
Current UI:
image


image

elyra/pipeline/processor.py
```diff
@@ -86,15 +86,7 @@ async def export(self, pipeline, pipeline_export_format, pipeline_export_path, o
...
-    def __init__(self, run_url, object_storage_url, object_storage_path):
-        # validate that the response has all required properties
-        if not run_url:
-            raise ValueError("Invalid Processor Response: Missing field 'run_url'.")
-        if not object_storage_url:
-            raise ValueError("Invalid Processor Response: Missing field 'object_storage_url'.")
-        if not object_storage_path:
-            raise ValueError("Invalid Processor Response: Missing field 'object_storage_path'.")
-
+    def __init__(self, run_url='', object_storage_url='', object_storage_path=''):
```
kevin-bates, line 89: I still don't really understand the need to change from `None` to the empty string. Is this only because the client can only handle strings and can't handle `None`?  Don't they still need to check either way (length 0 vs no value)?

Also, it seems the local runtimes always return what used to be reserved for successful submissions - even in error cases.  I think I'd rather see exceptions raised in error cases and a link in the exception message to the temporary "ledger directory".  This way, the UX is the same.

cc: @ajbozarth

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+    processor. If a notebook doesn't run using `run all` it will fail in the
+    same way when processed by this processor. Also, any relationship or specific capabilities
+    associated with a particular runtime is not supported by local mode and any additional properties
+    other then the specific file to be executed is ignored by the local processor.
+
+    Note: Execution happens in-place and a ledger of runs will be available at $TMPFILE/elyra/pipeline-name-<timestamp>
+    """
+    _type = 'local'
+
+    def __init__(self, root_dir=None):
```
kevin-bates, line 39: Is `root_dir` really optional?  If it's `None`, `get_absolute_path()` will fail.  So we should remove `=None`

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+        super().__init__()
+        self._root_dir = root_dir
+
+    @property
+    def type(self):
+        return self._type
+
+    def process(self, pipeline):
```
kevin-bates, line 47: Please add help string.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+            self.log.debug(f'Execution of {notebook_name} took {duration:.3f} secs.')
+
+    @staticmethod
+    def _get_pipeline_work_dir(pipeline_name: str) -> str:
+        log_dir = os.path.join(tempfile.tempdir, 'elyra', pipeline_name)
+        os.makedirs(log_dir, mode=0o700, exist_ok=True)
+        return log_dir
+
+    @staticmethod
+    def _get_operations_by_dependency(operations_by_id: dict) -> list:
```
kevin-bates, line 109: This method name is not intuitive.  Seems like `_sort_operations` or `_get_sorted_operations` is more straightforward (and intuitive).

Please add help string.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+        pipeline_work_dir = LocalPipelineProcessor._get_pipeline_work_dir(pipeline_name)
+        self.log.info(f'Processing Pipeline : {pipeline.name} at {pipeline_work_dir}')
+
+        operations = LocalPipelineProcessor._get_operations_by_dependency(pipeline.operations)
+        for operation in operations:
+            notebook = self.get_absolute_path(operation.filename)
+            if not os.path.isfile(notebook):
+                raise FileNotFoundError(f'Could not find {notebook}')
+
+            self._execute_file(pipeline_work_dir, notebook)
```
kevin-bates, line 60: This should probably be called `_execute_notebook` since you'll have a different method for python scripts.  If the plan is to use the same method for both, then identification (and existence check) of the file (notebook or python script) should occur within `_execute_file`.

elyra/pipeline/processor_local.py
kevin-bates: Then please remove this error handling.  It should not be done.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+                )
+            except Exception:
+                self.log.error(f'Internal error executing {filename}')
+
+            t1 = time.time()
+            duration = (t1 - t0)
+            self.log.debug(f'Execution of {notebook_name} took {duration:.3f} secs.')
+
+    @staticmethod
+    def _get_pipeline_work_dir(pipeline_name: str) -> str:
```
kevin-bates, line 103: help-string

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+        ordered_operations = []
+
+        for operation in operations_by_id.values():
+            if operation not in ordered_operations:
+                # operation is a root node
+                if not operation.parent_operations:
+                    ordered_operations.append(operation)
+                else:
+                    LocalPipelineProcessor.\
+                        _visit_operation(operations_by_id, ordered_operations, operation)
```
kevin-bates, line 119: I believe this can be simplified to the following - please check...
```suggestion
        for operation in operations_by_id.values():
            LocalPipelineProcessor._visit_operation(operations_by_id, ordered_operations, operation)
```

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,130 @@
...
+
+        return ordered_operations
+
+    @staticmethod
+    def _visit_operation(operations_by_id, ordered_operations, operation):
```
kevin-bates, line 124: This seems more than a 'visit'.  I would argue its `_sort_operation`.
Nonetheless, please add some comments and a help string.

Could you add type hints here?  You might also need to pull things like `Dict` and `List` from the `typing` package if you need to use `Optional[]`, etc.

elyra/pipeline/tests/pipeline_complex.json
```diff
@@ -0,0 +1,555 @@
+{
```
kevin-bates, line 1: I would mix the order of these up a bit.  They look like they're already relatively ordered.

Also, can we trim out the `ui_data` stanzas?  (This could be done separately since I believe there are others the same way)

elyra/pipeline/tests/test_pipeline_processor_local.py
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from elyra.pipeline import PipelineParser
+from elyra.pipeline.processor_local import LocalPipelineProcessor
+from .util import _read_pipeline_resource
+
+
+def test_pipeline_execution_order():
+    expected_operation_names = ['a', 'b', 'c', 'd', 'e', 'f', 'x', 'y', 'g', 'h']
```
kevin-bates, line 22: As noted on the test pipeline, this is essentially the same order the file is in already.  Please mixup the nodes in the file.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1048,7 +1051,13 @@ export class PipelineEditor extends React.Component<
       return;
     }
 
-    const runtimes = await PipelineService.getRuntimes();
+    const runtimes = await PipelineService.getRuntimes(false);
+    const local_runtime: any = {
+      name: 'local',
+      display_name: 'Run in-place locally'
```
kevin-bates, line 1057: 'in-place locally' is redundant and one of the two should be picked.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
kevin-bates: Right, but when there are no runtimes configured, 'local' will always be present - correct?

Does this work (to reduce hard-coding)?
```suggestion
      PipelineService.getRuntimeName(runtime_config, runtimes) || local_runtime.name;
```

packages/pipeline-editor/src/PipelineService.tsx
kevin-bates: But doesn't the front-end need to the schema for certain operations?  What is the schema for the local runtime?

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -83,28 +92,49 @@ export class PipelineService {
...
+            rel="noopener noreferrer"
+          >
+            object storage
+          </a>
+          .
+        </p>
+      );
+    } else {
+      // pipeline executed in-place locally
+      dialogTitle = 'Job execution in-place succeeded';
```
kevin-bates, line 127: Doesn't this also get produced when the local job execution fails since you're not raising exceptions on failures?  How is the UI supposed to determine failure in these cases?

elyra/pipeline/processor.py
lresende: I thought None -> '' was your suggestion... 
See the issue description for UX details.

elyra/pipeline/processor_local.py
lresende: I mostly needed this for testing purposes... but if it's not optional, how does it get injected on the kfp runtime?

lresende: Anyway, removed this

elyra/pipeline/processor_local.py
lresende: I would like to have consistent behavior, so will remove when addressing the other issue if that is ok.

elyra/pipeline/processor_local.py
lresende: with this change, it does not process the root nodes

kevin-bates:
I'm unable to run a pipeline that runs using 1.0 or master.  I get the following...
```
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/tornado/web.py", line 1703, in _execute
    result = await result
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 87, in post
    pipeline = PipelineParser().parse(pipeline_definition)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/elyra/pipeline/parser.py", line 61, in parse
    self._nodes_to_operations(pipeline_definitions, pipeline_object, primary_pipeline['nodes'])
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/elyra/pipeline/parser.py", line 95, in _nodes_to_operations
    operation = PipelineParser._create_pipeline_operation(node, super_node)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/elyra/pipeline/parser.py", line 142, in _create_pipeline_operation
    parent_operations=parent_operations)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.7/site-packages/elyra/pipeline/pipeline.py", line 53, in __init__
    raise ValueError("Invalid pipeline operation: Missing field 'operation filepath'.")
ValueError: Invalid pipeline operation: Missing field 'operation filepath'.
```

elyra/pipeline/processor_local.py
kevin-bates: Sure it does.  All nodes are visited by virtue of the for loop and root nodes will simply skip the `for parent_operation_id in operation.parent_operations` loop in "visit_operation" - immediately adding themselves to the `ordered_operations` list.

elyra/pipeline/processor_local.py
```diff
@@ -0,0 +1,149 @@
...
+                        input_path=filepath,
+                        output_path=filepath,
+                        stdout_file=stdout_file,
+                        stderr_file=stderr_file,
+                        cwd=notebook_dir,
+                        kernel_name="python3"
+                    )
+                except Exception as ex:
+                    self.log.error(f'Internal error executing {filepath}')
+                    raise RuntimeError(f'Internal error executing {filepath}.'
```
kevin-bates, line 101: Please distinguish this exception message from the outer one. 
```suggestion
                    self.log.error(f'Retry execution of {filepath} failed.')
                    raise RuntimeError(f'Retry execution of {filepath} failed.'
```

elyra/pipeline/processor_local.py
kevin-bates: I don't really understand that.  The "other issue" is in a completely different repository and this is an error case that no one should rely on.

lresende:
Sorry for the error, aggressive ide refactoring, which is now reverted.

packages/pipeline-editor/src/PipelineService.tsx
lresende: No, this is only for success. Otherwise, the internal call show the error results with a different common dialog that displays the exception 

image

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: though English-wise I agree, in this case they clarify different things. "in-place" tells the user that the notebooks are being run where they currently are, and "locally" tells them that it's not sending the job to a remote service. If you remove one it changes the meaning, just "locally" could imply that it's running locally but using a local instance of a service, and just "in-place" is (at least to me) not clear enough on it's own for a user to know where it will run.

packages/pipeline-editor/src/PipelineService.tsx
lresende: there is no schema, and this "local runtime" is only visible for the run dialog...

lresende:
Running the covid-eu pipelines locally:
image

Running the covid-us pipelines locally:
image

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ptitzler: I do believe  the current description is too ambiguous. "Local" could refer to the machine where the web browser is running or the machine where JupterLab is running.

ptitzler:
1. When is the "job execution .. succeeded" message displayed? (after the pipeline run was started or after the pipeline run finished?) For Kubeflow Pipelines the corresponding message is displayed _after_ the pipeline was started

  ```
  Job submission to KFP dev instance succeeded
  Check the status of your pipeline at Run Details.
  The results and outputs are in the /ptitzler/www-0814103015 working directory in object storage.
  ```

2. The "local" execution message does not provide any information where the results and outputs are located (unlike it is done for Kubeflow Pipelines where COS bucket location is displayed). Could we add this information for clarity?

3. Would it make sense to treat "local" execution like any other runtime configuration from a management point of view? This way 
- it would be listed in the same place (runtime widget) and certain configuration settings could be customized by users 
- it could be removed if an administrator of a hosted JL instance wants to prevent  "local" pipeline execution
- we could provide a link to the log files (however, I assume links would only be valid for as long as the temp directory doesn't change)

kevin-bates:
> When is the "job execution .. succeeded" message displayed? 

This is displayed after the _execution_ of the pipeline. Should errors be encountered along the way, the first error would produce an error message and execution is stopped.

> The "local" execution message does not provide any information where the results and outputs are located (unlike it is done for Kubeflow Pipelines where COS bucket location is displayed). Could we add this information for clarity?

Seems like a good idea, although the outputs are really a function of where the producing notebooks reside, so they'll be located throughout the working area.  But perhaps some kind of message to that effect should be displayed.

> Would it make sense to treat "local" execution like any other runtime configuration from a management point of view?

This has been a point of contention.  Some wanted to treat local runs as merely a _mode of operation_ and not introduce a runtime, others felt a _virtual_ runtime was appropriate.  The implementation has proceeded with the latter.

You raise excellent points regarding consistency between runtimes.  I would also like to see something more permanent than a temporary directory used since many *nix platforms delete temporary file locations on reboots, etc.  Seems like utilizing the Jupyter DataDir for storing this kind of information would be appropriate, then a stable directory could be referenced that houses _all_ local runs.

lresende:
> it could be removed if an administrator of a hosted JL instance wants to prevent "local" pipeline execution

To me, this seems like disabling the ability to execute a notebook on JupyterLab, which is all that this is doing.

lresende:
After some investigation, the logs created by papermill are always empty and updates/failures are actually on the user notebooks so I have removed the logs which simplify the user experience and solve a few questions raised above.

lresende:
> I find this wording very odd. Does something like "In-place job execution succeeded" sound any better?

How about:
title: `Job execution succeeded`
body: `Your job has been executed in-place in your local environment.`

image

ajbozarth:
I think `Job successfully executed` is better than `Job execution succeeded` with the change to the word structure after taking out "in-place"

lresende:
> I think `Job successfully executed` is better than `Job execution succeeded` with the change to the word structure after taking out "in-place"

In this case, I was just using the same wording we already use for the remote execution mode

elyra/pipeline/processor_local.py
kevin-bates: fyi, `test_pipeline_execution_order` passes after making this change.

ptitzler:
> I would also like to see something more permanent than a temporary directory [...]  then a stable directory could be referenced that houses all local runs.
👍 

> To me, this seems like disabling the ability to execute a notebook on JupyterLab, which is all that this is doing.
True. Maybe I jumped the gun here.

kevin-bates:
FWIW - I've pushed support for envs to my fork based off this PR: https://github.com/elyra-ai/elyra/commit/16cb8a17de54e8a144d41c8c3bf23d2a4d008e5d

I can now run my node_type framework locally:
image

lresende:
> FWIW - I've pushed support for envs to my fork based off this PR: [16cb8a1](https://github.com/elyra-ai/elyra/commit/16cb8a17de54e8a144d41c8c3bf23d2a4d008e5d)

Really great. Can we then all agree and set the bar that for now, the only properties that we are going to support are `env vars`?

kevin-bates:
What other properties are there?  Inputs, outputs and dependencies are already "supported".  Envs were the only items that could not be "brought to the party" via the pipeline.

lresende:
> What other properties are there? Inputs, outputs and dependencies are already "supported". Envs were the only items that could not be "brought to the party" via the pipeline.

I have heard about docker already, but really don't want to enable that via this mode.

kevin-bates:
> I have heard about docker already, but really don't want to enable that via this mode.

I guess I don't view 'docker' as a "property", but I totally agree.  That's what the pipeline frameworks are for.  I think access to files is where "simple" docker support breaks down. We can a lot using COS in container land - and that seems like a line in the sand thing.

lresende:
@kevin-bates I have cherry-picked your env support to this PR. I guess we are now ready for your approval and we can merge and let others play with it.

elyra/pipeline/processor_local.py
kevin-bates: This seems like a nice way to simplify things - do you disagree?

elyra/pipeline/tests/test_pipeline_processor_local.py
kevin-bates: I still think the order of the operations need to change within the pipeline file.  Right now, you could not perform any sorting and get this same answer.  You will likely need to change the order of the expected names if this is the way you're going to validate.

lresende:
A just added another test scenario for the dependency order
image

kevin-bates:
Update the pipeline file so that the 'g' stanza is first or second, then we should be good.  The way it is now, again, if you skipped sorting, it would likely be the expected result.

elyra/pipeline/processor_local.py
lresende: With your suggested changes (unless i missed something) ,some nodes are skipped/ignored, tests are failing like:
```
Expected :['a', 'b', 'c', 'd', 'e', 'f', 'x', 'y', 'g', 'h']
Actual   :['a', 'b', 'c', 'd', 'e', 'f', 'x', 'y', 'g', 'x', 'y', 'h']
```

lresende: Adding another check on the top of the loops seems to resolve this, something like:
```
    @staticmethod
    def _visit_operation(operations_by_id: dict, ordered_operations: list, operation: Operation) -> None:
        """
        Helper method to the main sort operation function
        """
        if operation not in ordered_operations:
            for parent_operation_id in operation.parent_operations:
                parent_operation = operations_by_id[parent_operation_id]
                if parent_operation not in ordered_operations:
                    LocalPipelineProcessor.\
                        _visit_operation(operations_by_id, ordered_operations, parent_operation)
            ordered_operations.append(operation)
```
Want to go for this?

lresende:
> Update the pipeline file so that the 'g' stanza is first or second, then we should be good. The way it is now, again, if you skipped sorting, it would likely be the expected result.

I still don't understand, are you saying that I just need to re-arrange, manually, the JSON file representing the pipeline?

lresende:
@kevin-bates Please check if the changes on the `pipeline_dependency_simple` look ok

kevin-bates:
The order in the pipeline file is the same as the expected order: `['f', 'a', 'c', 'g']`.  The test that tests that sorting worked would likely still succeed if the sorting method were to be skipped altogether.  As a result, the test doesn't seem valid.  If you move the stanza containing node 'g' anywhere else, then the order in the pipeline file will be different from the expected order and you will essentially confirm that the sorting method has worked.  I would move 'g' between 'f' and 'a'. 

> Can we then all agree and set the bar that for now, the only properties that we are going to support are env vars?
> I have heard about docker already, but really don't want to enable that via this mode.

It occurred to me that `image_name` is a property you were referring to.  Correct, support for image name would really not be a "local" operation so we should say that we don't support `images` with local runtimes.

However, that leads me to another question (which can be dealt with separately).  Currently, `image_name` is a required property on nodes and is validated as such.  Seems like that validation should almost be deferred until the runtime has been selected.  But we can talk about that later, I could go either way on that.

elyra/pipeline/processor_local.py
```diff
@@ -90,27 +90,27 @@ def _sort_operations(operations_by_id: dict) -> list:
...
         """
-        for parent_operation_id in operation.parent_operations:
-            parent_operation = operations_by_id[parent_operation_id]
-            if parent_operation not in ordered_operations:
-                LocalPipelineProcessor.\
-                    _visit_operation(operations_by_id, ordered_operations, parent_operation)
-        ordered_operations.append(operation)
+        # Optimization: check if already processed
+        if operation not in ordered_operations:
+            # process each of the dependencies that needs to be executed first
```
kevin-bates, line 106: Good catch on this - thanks.

kevin-bates:
This might be a good change.  The first assertion makes sure the current order is not the expected order.  Essentially validating the test's validity. :smile:  It would be good to have more than one node out of place, if possible.
```python

    pipeline = PipelineParser().parse(pipeline_definitions)

    ordered_operation_names = _get_operation_names(pipeline.operations.values())
    assert ordered_operation_names != expected_operation_names

    operations = LocalPipelineProcessor.\
        _sort_operations(operations_by_id=pipeline.operations)

    ordered_operation_names = _get_operation_names(operations)
    assert ordered_operation_names == expected_operation_names
```

lresende:
It turns out that the additional assertion is already valid without any changes to pipeline file.

kevin-bates:
Yeah, I noticed that when adding that as a check.  The original only had one node ('g') out of place: `['a', 'b', 'c', 'd', 'e', 'f', 'g', 'x', 'y', 'h']` vs. the expected: `['a', 'b', 'c', 'd', 'e', 'f', 'x', 'y', 'g', 'h']`.  Would be nice to have more, but that's fine.

Issue #755, opened 2020-07-14
Access Kubeflow Pipelines behind Dex Authentication


cliveseldon:
I have installed Kubeflow via the Dex KfDef but this means Elyra can't access pipelines as it needs Auth.
It would be great to be able access Auth restricted Kubeflow environments.

lresende:
As a reference, this kfp issue might be useful:
https://github.com/kubeflow/kfctl/issues/140

PR #847, opened 2020-08-10
Support to KFP via Dex Authentication

karlaspuldaro:
Solves #755 

Initial support to authentication restricted kubeflow environment (using istio dex as kfdef)

- [X] Update kfp dependency version to 1.0.0
- [X] Add api endpoint credentials to kfp schema
- [X] Authenticate as client to kfp server and store session cookie
- [x] Double check edge cases
- [X] Update docs

Flow:
1. User creates a new kfp metadata and add (optional) values for `api_endpoint_user` and `api_endpoint_password`, when using kfp authentication restricted environment
2. User submits a pipeline through Elyra and select the auth restricted kfp endpoint schema
3. Elyra authenticates connection with the kfp server and stores its session cookie (no programatically authentication via KFP SDK available yet)
  **Kubeflow issue**: - Making a request auth to kfp passing an invalid session cookie produces too generic error: `No host specified`
4. Pipeline is uploaded, Elyra displays access link
5. User logs in from the browser and accesses kfp dashboard (find the new pipeline on the dashboard list)

image

Currently we run manual tests uploading pipelines to a dex auth configured kubeflow env. We should think about integrating tests in the future.

           in the file; or

           it.


elyra/pipeline/processor_kfp.py
```diff
@@ -67,8 +97,22 @@ def process(self, pipeline):
             self.log.info("Kubeflow Pipeline successfully compiled.")
             self.log.debug("Kubeflow Pipeline was created in %s", pipeline_path)
 
-            # Upload the compiled pipeline and create an experiment and run
-            client = kfp.Client(host=api_endpoint)
+            # Upload the compiled pipeline, create an experiment and run
+
+            if (api_endpoint_username is not None) and (api_endpoint_password is not None):
```
lresende, line 102: you can do similar to JS/TS here if(api_endpoint_username) which will cover for None and empty strings

elyra/pipeline/processor_kfp.py
```diff
@@ -37,6 +38,30 @@ class KfpPipelineProcessor(PipelineProcessor):
     def type(self):
         return self._type
 
+    def get_auth_session_cookie(self, url, api_endpoint_username, api_endpoint_password):
+        get_response = requests.get(url)
```
lresende, line 42: do we need this? or can we actually just assume you need to auth and avoid one extra hop to go directly to line 46?

elyra/pipeline/processor_kfp.py
```diff
@@ -37,6 +38,30 @@ class KfpPipelineProcessor(PipelineProcessor):
     def type(self):
         return self._type
 
+    def get_auth_session_cookie(self, url, api_endpoint_username, api_endpoint_password):
```
lresende, line 41: could you make this a private function more towards the bottom `def _get_auth_session_cokkie`... also, should this be called `__get_user_auth_session_cokkie` or `_authorize_user`?

setup.py
```diff
@@ -49,7 +49,7 @@
     packages=find_packages(),
     install_requires=[
         "jupyter_core>=4.0,<5.0",
-        "kfp==0.5.1",
+        "kfp==1.0.0",
         "urllib3==1.24.2",
```
lresende, line 53: @akchinSTC FYI, what are the implications here? I thought we pinned to 0.5.1 for some reason that I don't remember...

elyra/pipeline/processor_kfp.py
```diff
@@ -67,8 +97,22 @@ def process(self, pipeline):
...
-            # Upload the compiled pipeline and create an experiment and run
-            client = kfp.Client(host=api_endpoint)
+            # Upload the compiled pipeline, create an experiment and run
+
+            if (api_endpoint_username is not None) and (api_endpoint_password is not None):
+                endpoint = api_endpoint.replace('/pipeline', '')
+                session_cookie = self.get_auth_session_cookie(endpoint, api_endpoint_username, api_endpoint_password)
+
+                # TODO: handle case no session cookie found --> authentication error?
+                # When session cookie is none, we get a making request error, reason: No host specified
```
lresende, line 107: what happens if you return an empty string instead of none?

elyra/metadata/schemas/kfp.json
```diff
@@ -33,6 +33,19 @@
           "type": "string",
           "format": "uri"
         },
+        "api_endpoint_username": {
+          "title": "Kubeflow Pipelines API Endpoint Username",
+          "description": "The Kubeflow Pipelines API Endpoint username",
```
kevin-bates, line 38: We should probably lowercase endpoint in the description (same goes for the password property)
```suggestion
          "description": "The Kubeflow Pipelines API endpoint username",
```

elyra/pipeline/processor_kfp.py
kevin-bates, line 41: and simplify the variable names - no need to know their context within the function.
```suggestion
    def get_auth_session_cookie(self, url, username, password):
```

elyra/pipeline/processor_kfp.py
kevin-bates: Won't this distinguish between the endpoints that are not secured, yet users have added a user/pwd to the config and those that are?  I think its probably a good idea to ensure 'auth' is present.

elyra/pipeline/processor_kfp.py
```diff
@@ -37,6 +38,30 @@ class KfpPipelineProcessor(PipelineProcessor):
...
+            cookie_auth_key = 'authservice_session'
+            credentials = {'login': api_endpoint_username, 'password': api_endpoint_password}
+
+            # Redirect to post authentication request
+            post_response = requests.post(get_response.url, data=credentials)
+            cookie_auth_value = post_response.cookies.get(cookie_auth_key)
+
+            if cookie_auth_value is not None:
+                return cookie_auth_key + '=' + cookie_auth_value
+            else:
```
kevin-bates, line 55: This can be removed.

elyra/pipeline/processor_kfp.py
kevin-bates, line 102: Is a password required for authentication?  Just wondering.

elyra/pipeline/processor_kfp.py
```diff
@@ -67,8 +97,22 @@ def process(self, pipeline):
...
+
+                # TODO: handle case no session cookie found --> authentication error?
+                # When session cookie is none, we get a making request error, reason: No host specified
+
+                client = kfp.Client(
+                    host=api_endpoint,
+                    cookies=session_cookie
+                )
+            else:
+                client = kfp.Client(host=api_endpoint)
```
kevin-bates, line 114: Please return `None` and collapse both of these Client statements to a single:
```python
client = kfp.Client(host=api_endpoint, cookies=session_cookie)
```
You'll need to initialize `session_cookie = None` prior to the if user/pwd block.

elyra/pipeline/processor_kfp.py
karlaspuldaro: @kevin-bates I'm also wondering if an empty string is a valid pwd. According to [dex kfdef yaml](https://raw.githubusercontent.com/kubeflow/manifests/v1.0-branch/kfdef/kfctl_istio_dex.v1.0.2.yaml), this is the section we're looking at:
```
- kustomizeConfig:
      overlays:
      - istio
      parameters:
      - name: namespace
        value: auth
      - name: issuer
        value: http://dex.auth.svc.cluster.local:5556/dex
      - name: client_id
        value: kubeflow-oidc-authservice
      - name: oidc_redirect_uris
        value: '["/login/oidc"]'
      - name: static_email
        value: admin@kubeflow.org
      - name: static_password_hash
        value: $2y$12$ruoM7FqXrpVgaol44eRZW.4HWS8SAvg6KYVVSCIwKQPBmTpCm.EeO
      repoRef:
        name: manifests
        path: dex-auth/dex-crds
    name: dex
```
email and password can be edited, our environment is using the default kubeflow deployment.
We can try out with an empty pwd (only not sure what kind of hash it's being used but I can look it up)

kevin-bates: I'm not sure what I'm looking at here.  It looks like data values so I don't know how you'd determine if a password requires a value.  
> but I can look it up

Is there an API doc for the `authservice_session` thing?

elyra/pipeline/processor_kfp.py
karlaspuldaro: I agree with Kevin, I made this as some kind of validation and not only rely on user config. If the first response stops at a redirected url where auth is required, we go ahead and authenticate

elyra/pipeline/processor_kfp.py
karlaspuldaro: In my use case the cookie is found in the last item of response history, for some reason it wasn't carried over. Maybe I should try using requests.Session instead?

elyra/pipeline/processor_kfp.py
karlaspuldaro: same behavior

elyra/pipeline/processor_kfp.py
kevin-bates: Sorry, I meant the `else` statement can be removed and its block unindented since the `if` can only return.

elyra/pipeline/processor_kfp.py
karlaspuldaro: I wasn't able to find any specific documentation around that. 
I can try to add another auth user to our kfp env, where the pwd field is a bcrypt hash of an empty string. If that works, then in our code we would only check for `api_endpoint_username as you suggested. Otherwise if we also assume `api_endpoint_password` where empty string is actually its valid value, we won't authenticate when we were actually supposed to

karlaspuldaro:
> Don't we still need this portion from the previous 'else' block (just not _in_ an `else` block)?

No, because now we're using requests.Session() object which allows cookies to persist across all requests, no need to check the history anymore

elyra/pipeline/processor_kfp.py
```diff
@@ -67,8 +72,22 @@ def process(self, pipeline):
...
+            session_cookie = None
+
+            # TODO: double check if empty string is a valid pwd
+            if api_endpoint_username and api_endpoint_password:
+                endpoint = api_endpoint.replace('/pipeline', '')
+                session_cookie = self._get_user_auth_session_cookie(endpoint,
+                                                                    api_endpoint_username,
+                                                                    api_endpoint_password)
+
+            client = kfp.Client(
```
kevin-bates, line 86: Would you mind joining these lines into a single line for the statement?  I think it reads easier and, although I like whitespace, this seems to introduce a lot of it.

elyra/pipeline/processor_kfp.py
```diff
@@ -287,3 +306,21 @@ def _get_runtime_configuration(self, name):
...
+            # Authenticate user
+            session = requests.Session()
+            session.post(get_response.url, data=credentials)
+            cookie_auth_key = 'authservice_session'
+            cookie_auth_value = session.cookies.get(cookie_auth_key)
+
+            if cookie_auth_value:
+                return cookie_auth_key + '=' + cookie_auth_value
+
+        return None
```
kevin-bates, line 326: This return (None) statement is not necessary since that's the default behavior in Python.

docs/source/developer_guide/metadata.md
```diff
@@ -133,6 +135,8 @@ if not runtime_configuration:
     raise RuntimeError("Runtime metadata not available.")
 
 api_endpoint = runtime_configuration.metadata['api_endpoint']
+api_endpoint_username = runtime_configuration.metadata['api_endpoint_username']
+api_endpoint_password = runtime_configuration.metadata['api_endpoint_password']
```
kevin-bates, line 139: I'm wondering if these should be named simply `api_username` and `api_password`.  This would also be consistent with the naming conventions used for the `cos_` fields.  Since we can't easily change `cos_username` to `cos_endpoint_username`, it would make sense (for consistency's sake) to change the new fields.

karlaspuldaro: I like that suggestion and to make them shorter :)

elyra/pipeline/processor_kfp.py
karlaspuldaro: Tested this today, and confirmed that password requires a value

elyra/pipeline/processor_kfp.py
```diff
@@ -67,8 +72,18 @@ def process(self, pipeline):
...
+            session_cookie = None
+
+            if api_username and api_password:
+                endpoint = api_endpoint.replace('/pipeline', '')
+                session_cookie = self._get_user_auth_session_cookie(endpoint,
+                                                                    api_username,
+                                                                    api_password)
+
+            client = kfp.Client(host=api_endpoint, cookies=session_cookie)
+
```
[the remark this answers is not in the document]
karlaspuldaro: Thank you for the suggestion, I like this approach. 
This is how the error dialog looks like now:
image

setup.py
lresende: @akchinSTC said it's ok

PR #848, opened 2020-08-11
refactor pipeline editor dialogs

bourdakos1:

           in the file; or

           it.


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -61,15 +61,21 @@ import * as React from 'react';
...
-import { PipelineExportDialog } from './PipelineExportDialog';
+import PipelineExportDialog from './PipelineExportDialog';
 import { PipelineService } from './PipelineService';
-import { PipelineSubmissionDialog } from './PipelineSubmissionDialog';
+import PipelineSubmissionDialog from './PipelineSubmissionDialog';
 import * as properties from './properties.json';
 import { KFP_SCHEMA, RUNTIMES_NAMESPACE } from './RuntimesWidget';
 import Utils from './utils';
 
+export interface IRuntime {
```
ajbozarth, line 74: this interface would be better kept in `PipelineService.tsx`

packages/pipeline-editor/src/NotebookSubmissionDialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { IDictionary } from '@elyra/application';
+import * as React from 'react';
+
+import { IRuntime } from './PipelineEditorWidget';
+
+type Props = {
```
ajbozarth, line 22: is there a reason you're using `type` for Props throughout this PR instead of `interface`? As far as jupyterlab (and elyra) code goes it seems to always use `interface` for props

packages/pipeline-editor/src/NotebookSubmissionDialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+  runtimes: IRuntime[];
+  images: IDictionary<string>;
+  env: string[];
+};
+
+const chunkArray = <T extends {}>(arr: T[], n: number): T[][] => {
```
ajbozarth, line 28: this seems like a pure util function, should we put it in `utils.ts` instead?

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
```diff
@@ -14,65 +14,44 @@
...
-}
+import * as React from 'react';
+
+import { IRuntime } from './PipelineEditorWidget';
+
+type Props = {
+  runtimes: IRuntime[];
+};
+
+const PipelineSubmissionDialog = ({ runtimes }: Props): JSX.Element => {
```
ajbozarth, line 25: Given you've made all the dialogs `const` of `JSX.Element` instead of classes should we keep all the dialogs in a `dialogs.tsx` file instead of separate files? Especially since `PipelineSubmissionDialog.tsx` and `PipelineExportDialog.tsx` were named after the class they contained.

packages/pipeline-editor/src/dialogWidget.ts
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ReactWidget, Dialog } from '@jupyterlab/apputils';
+import { MessageLoop } from '@lumino/messaging';
+import { Widget } from '@lumino/widgets';
+
+export const dialogWidget = (
```
ajbozarth, line 21: since this seems to be specifically for dialogs containing forms I think a better name would be `formDialogWidget`

packages/pipeline-editor/src/SubmitNotebook.ts
```diff
@@ -1,313 +0,0 @@
...
-      // +'<br/>'
-      // +'<input type="text" id="gpus" name="gpus" placeholder="0" value="0"/>'
-      // +'</td>'
-      //
-      // + td
-      // +'<label for="memory">Memory:</label>'
-      // +'<br/>'
-      // +'<input type="text" id="memory" name="memory" placeholder="1Gb" value="1Gb"/>'
-      // +'</td>'
-      // +'</tr>'
```
ajbozarth, line 221: @lresende this PR will remove any trace of these commented out sections. Are you ok with that?

packages/pipeline-editor/src/SubmitNotebook.ts
```diff
@@ -1,313 +0,0 @@
...
-    this._runtimes = runtimes;
-    this._runtimeImages = runtimeImages;
-
-    this.node.appendChild(this.renderHtml());
-    (this.node.getElementsByClassName(
-      'elyra-form-runtime-config'
-    )[0] as HTMLSelectElement).value = '';
-    (this.node.getElementsByClassName(
-      'elyra-form-framework'
-    )[0] as HTMLSelectElement).value = '';
```
ajbozarth, line 154: @lresende this PR removes the code that explicitly sets the drop downs default values. IIUC the drop downs will use the first value as default after this change (I will double check when I test this locally). IS this change in behavior ok? We've discussed it in the past but I can't remember the details

packages/pipeline-editor/src/SubmitNotebookButtonExtension.tsx
```diff
@@ -0,0 +1,117 @@
...
+ *  - Attach button to notebook toolbar and launch a dialog requesting
+ *  information about the remote location to where submit the notebook
+ *  for execution
+ */
+export class SubmitNotebookButtonExtension
+  implements DocumentRegistry.IWidgetExtension<NotebookPanel, INotebookModel> {
+  private panel: NotebookPanel;
+
+  constructor(app: JupyterFrontEnd) {
+    this.app = app;
```
ajbozarth, line 43: I don't think this is used any more (probably remnant code), so we can probably remove this

packages/pipeline-editor/src/NotebookSubmissionDialog.tsx
bourdakos1: I normally always use `interface` too, not sure what led me to use `type` here haha

packages/pipeline-editor/src/SubmitNotebookButtonExtension.tsx
bourdakos1: I removed this already in a follow-up PR I'm working on

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
bourdakos1: Do you prefer classes? I can switch to classes or put all the dialogs in one file or both or neither haha, up to you 😄

I normally use function components instead of class components and I usually keep one component per file

I also use normal functions instead of arrow functions, but eslint didn't like that:
```tsx
function PipelineSubmissionDialog(props) {
  ...
}
```

ajbozarth: That makes sense, I personally don't have a preference, but I believe we and JupyterLab usually use classes so I think we should be consistent (unless making this a class would bloat the code, in which case lets keep it as is)

lresende: I would not recommend stashing all dialogs on the same file, I believe what I have seen on other extensions is a "component" per file, so a dialog or widget per file.

packages/pipeline-editor/src/utils.ts
```diff
@@ -33,27 +32,34 @@ export default class Utils {
...
     template.primary_pipeline = generated_uuid;
     template.pipelines[0].id = generated_uuid;
 
     template.pipelines[0].nodes[0].id = generated_uuid;
     template.pipelines[0].nodes[0].app_data.filename = filename;
-    template.pipelines[0].nodes[0].app_data.runtime_image = options.framework;
-    template.pipelines[0].nodes[0].app_data.env_vars = options.env;
-    template.pipelines[0].nodes[0].app_data.dependencies = options.dependencies;
+    template.pipelines[0].nodes[0].app_data.runtime_image = framework;
+    template.pipelines[0].nodes[0].app_data.env_vars = envArray;
```
lresende, line 57: we recently had some updates on trying to use same/similar names for properties to make reading much simpler,  could we use something like `envVars` or `envVarsArray` here?

packages/pipeline-editor/src/utils.ts
```diff
@@ -33,27 +32,34 @@ export default class Utils {
...
+    template.pipelines[0].nodes[0].app_data.dependencies = dependencies;
 
     template.pipelines[0].app_data.name = artifactName;
     template.pipelines[0].app_data.runtime = 'kfp';
```
lresende, line 61: You might need to rebase here, I believe we removed the hard code in master.

ajbozarth: As a follow up to this, I believe we recently touched other code you've moved to different locations as well. I would double check changes to files this PR "deletes"

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -62,14 +62,17 @@ import { IntlProvider } from 'react-intl';
 
 import { PIPELINE_CURRENT_VERSION } from './constants';
 import * as i18nData from './en.json';
+import { formDialogWidget } from './formDialogWidget';
 import * as palette from './palette.json';
-import { PipelineExportDialog } from './PipelineExportDialog';
+
+import PipelineExportDialog from './PipelineExportDialog';
```
ajbozarth, line 68: imports should be wrapped in `{ }`, I'm not sure why this is passing lint, in fact this change chuck is unnecessary in general.

packages/pipeline-editor/src/PipelineExportDialog.tsx
```diff
@@ -14,78 +14,55 @@
...
-      '</select>' +
-      '<input type="checkbox" id="overwrite"/>' +
-      '<label for="overwrite">Replace if file already exists</label>' +
-      br;
+interface IProps {
+  runtimes: IRuntime[];
+}
 
-    htmlContent.innerHTML = content;
+const FILE_TYPES = ['yaml', 'py'];
```
ajbozarth, line 25: Just for consistency I think we usually put global `const`s like this above the `interface`s in files

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -25,6 +25,11 @@ import * as React from 'react';
 
 import Utils from './utils';
 
+export interface IRuntime {
```
ajbozarth, line 28: same note about the ordering of `const` and `interfaces` in files

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -62,14 +62,17 @@ import { IntlProvider } from 'react-intl';
...
 import {
   PipelineService,
   KFP_SCHEMA,
   RUNTIMES_NAMESPACE
 } from './PipelineService';
-import { PipelineSubmissionDialog } from './PipelineSubmissionDialog';
+import PipelineSubmissionDialog from './PipelineSubmissionDialog';
```
ajbozarth, line 74: same note as above

packages/pipeline-editor/src/SubmitNotebookButtonExtension.tsx
```diff
@@ -0,0 +1,117 @@
+/*
```
ajbozarth, line 1: Is there a reason git isn't recognizing this as a file name change from `SubmitNotebook.ts` -> `SubmitNotebookButtonExtension.tsx`? They seem to be the same file (with your updates/changes of course). It makes reading git history a bit difficult when files contents are copied to new file instead of renaming them and then editing.

packages/pipeline-editor/src/SubmitNotebookButtonExtension.tsx
ajbozarth: It's probably fine to remove it here as well, shouldn't break any diffs in your followup PR (I've done it before)

packages/pipeline-editor/src/index.ts
```diff
@@ -38,9 +38,10 @@ import {
   KFP_SCHEMA,
   RUNTIMES_NAMESPACE
 } from './PipelineService';
+
```
ajbozarth, line 41: not sure why you're adding a newline here

packages/pipeline-editor/src/NotebookSubmissionDialog.tsx
```diff
@@ -0,0 +1,138 @@
...
+              <td>
+                <br />
+                <input
+                  type="checkbox"
+                  id="dependency_include"
+                  name="dependency_include"
+                  size={20}
+                  defaultChecked
+                />
+                <label htmlFor="dependency_include">Include dependencies</label>
```
ajbozarth, line 115: missed this comment in my last review (even though I referenced it)

Previously this had a space before `Include dependencies`, this is the only visible change in this PR but I believe it is better this way without the space (makes it consistent with other checkboxes in forms)

bourdakos1: Yea, I removed the space to be consistent with the label in the other dialog. Should I leave it without the space?

packages/pipeline-editor/src/SubmitNotebookButtonExtension.tsx
bourdakos1: Maybe the diff was too big? I think most of the file changed

ajbozarth: if it wasn't intentional then that's probably the case, oh well

packages/pipeline-editor/src/NotebookSubmissionDialog.tsx
ajbozarth: yeah it's better this way

lresende:
@bourdakos1 could you please check my comment related to `template.pipelines[0].app_data.runtime = 'kfp';` still be hardcoded in this PR. Do we need a merge/rebase or is it a conflict that needs to be fixed?

bourdakos1:
@lresende it looks like that is hardcoded in master too. Should I be pulling from somewhere else?

bourdakos1:
https://github.com/elyra-ai/elyra/blob/b9fe9f46dee02a6a7785bc3a9f6a73d0416628f7/packages/pipeline-editor/src/utils.ts#L55

ajbozarth:
@lresende @bourdakos1 that particular line is still hard coded since it's used by the submit notebook button which doesn't have local support (doesn't need it). The other hard-coded lines we did remove are still removed in this PR as expected.

lresende:
Thanks for the clarification.

PR #861, opened 2020-08-14
Update pipeline editor to use relative paths

ajbozarth:
Currently nodes in a pipeline save thier paths relative to the
jupyter lab workspace. This update them to use paths relative to
the pipeline file instead. This enables moving a pipeline "project",
such as moving the directory a pipelines contents are in, checking
out a project from git, and starting lab from a different workspace.

A node's path is converted to be relative to the pipeline upon add
and that relative path is what's stored in the pipeline file. The
path is converted back to be relative to the workspace prior to
being submitted to the export or schedule api, or when a node is
"opened"

Note that this change is designed to work in conjunction with
validation of the node files and the ability to change node file
locations.

Fixes #823
Fixes #825 

Fixes #659 
Fixes #608 
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #867, opened 2020-08-17
Update node properties to allow changing file

### LINKED ISSUE #732: Pipeline editor: allow for changing of the notebook [name/location]
As a user I would like to change the notebook name or location of a node. Currently the node properties display the name/location but do not allow for editing.  Therefore if a user wants to change the name/location of a notebook the following steps need to be completed::

- create a new node for the "renamed" notebook
- select a runtime container
- copy and paste "file dependencies"
- copy and paste environment variables
- copy and paste "output files"
- remove old node
- link notebook nodes

With this proposed enhancement a user would only have to change the name/location of the notebook.

image

vabarbosa:
This PR adds support for changing the file associated with a node in the Pipeline Editor.

Fixes #732

To achieve this a new `BrowseFileDialog` was added into `@elyra/ui-components`. The `BrowseFileDialog` displays a file browser in a dialog window which resolves to the selected files when accepted. 

It is worth noting JupyterLab already contains a `FileDialog` in `@jupyterlab/filebrowser`. However, this does not allow flexibility and customization easily, hence the reason for creating the new `BrowseFileDialog`. Unlike the `FileDialog`, the `BrowseFileDialog` can be configured to allow for disabling/preventing multiselect and also allow for selecting directories (if needed). The `FileDialog` does includes the New Folder, Upload, and Refresh buttons (that appear in the default FileBrowser) but not options to opt out of displaying those buttons.

elyra-browse-file

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -466,9 +466,7 @@ export class PipelineEditor extends React.Component<
 
     if (app_data.filename !== propertySet.filename) {
       app_data.filename = propertySet.filename;
-      node.label = propertySet.filename
-        .replace(/^.*[\\/]/, '')
-        .replace(/\.[^/.]+$/, '');
+      node.label = Utils.getFilenameFromPath(propertySet.filename, true);
     }
 
     app_data.runtime_image = propertySet.runtime_image;
@@ -714,11 +712,7 @@ export class PipelineEditor extends React.Component<
             str => str + '='
           );
 
-          data.nodeTemplate.label = item.path.replace(/^.*[\\/]/, '');
-          data.nodeTemplate.label = data.nodeTemplate.label.replace(
-            /\.[^/.]+$/,
-            ''
-          );
+          data.nodeTemplate.label = Utils.getFilenameFromPath(item.path, true);
           data.nodeTemplate.image = IconUtil.encode(notebookIcon);
           data.nodeTemplate.app_data['filename'] = item.path;
           data.nodeTemplate.app_data[
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -466,7 +466,10 @@ export class PipelineEditor extends React.Component<
 
     if (app_data.filename !== propertySet.filename) {
       app_data.filename = propertySet.filename;
-      node.label = Utils.getFilenameFromPath(propertySet.filename, true);
+      node.label = path.basename(
+        propertySet.filename,
+        path.extname(propertySet.filename)
+      );
     }
 
     app_data.runtime_image = propertySet.runtime_image;
@@ -712,7 +715,10 @@ export class PipelineEditor extends React.Component<
             str => str + '='
           );
 
-          data.nodeTemplate.label = Utils.getFilenameFromPath(item.path, true);
+          data.nodeTemplate.label = path.basename(
+            item.path,
+            path.extname(item.path)
+          );
           data.nodeTemplate.image = IconUtil.encode(notebookIcon);
           data.nodeTemplate.app_data['filename'] = item.path;
           data.nodeTemplate.app_data[
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -30,6 +30,7 @@ import {
   pipelineIcon,
   savePipelineIcon,
   runtimesIcon,
+  showBrowseFileDialog,
   showFormDialog,
   errorIcon
 } from '@elyra/ui-components';
@@ -200,6 +201,7 @@ export class PipelineEditor extends React.Component<
   position = 10;
   node: React.RefObject<HTMLDivElement>;
   propertiesInfo: any;
+  propertiesController: any;
 
   constructor(props: any) {
     super(props);
@@ -233,6 +235,10 @@ export class PipelineEditor extends React.Component<
     this.applyPropertyChanges = this.applyPropertyChanges.bind(this);
     this.closePropertiesDialog = this.closePropertiesDialog.bind(this);
     this.openPropertiesDialog = this.openPropertiesDialog.bind(this);
+    this.propertiesActionHandler = this.propertiesActionHandler.bind(this);
+    this.propertiesControllerHandler = this.propertiesControllerHandler.bind(
+      this
+    );
 
     this.node = React.createRef();
     this.handleEvent = this.handleEvent.bind(this);
@@ -342,6 +348,8 @@ export class PipelineEditor extends React.Component<
     ];
 
     const propertiesCallbacks = {
+      actionHandler: this.propertiesActionHandler,
+      controllerHandler: this.propertiesControllerHandler,
       applyPropertyChanges: this.applyPropertyChanges,
       closePropertiesDialog: this.closePropertiesDialog
     };
@@ -456,6 +464,13 @@ export class PipelineEditor extends React.Component<
     }
     const app_data = node.app_data;
 
+    if (app_data.filename !== propertySet.filename) {
+      app_data.filename = propertySet.filename;
+      node.label = propertySet.filename
+        .replace(/^.*[\\/]/, '')
+        .replace(/\.[^/.]+$/, '');
+    }
+
     app_data.runtime_image = propertySet.runtime_image;
     app_data.outputs = propertySet.outputs;
     app_data.env_vars = propertySet.env_vars;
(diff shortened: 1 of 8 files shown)


kevin-bates:
@vabarbosa - this looks great!  Just curious if the same could be done for the File Dependencies section (where the start point is the location of the notebook file)?  If so, then that would eliminate folks typing in names/paths to existing files.  I'd be happy to open the issue if that's something we want to do.

vabarbosa:
> @vabarbosa - this looks great! Just curious if the same could be done for the File Dependencies section (where the start point is the location of the notebook file)? If so, then that would eliminate folks typing in names/paths to existing files. I'd be happy to open the issue if that's something we want to do.

@kevin-bates you read my mind, i was thinking the same thing as was going to bring it up for discussion during the scrum to see if it is something would like to have included

lresende:
This looks awesome!!!

marthacryan:
This is so cool!! Just one style thing - for some reason this is showing up a little misaligned for me (in chrome)
image
I didn't investigate a ton but it looks like it's being split in half between the filename and the button - any ideas? 
image

vabarbosa:
> This is so cool!! Just one style thing - for some reason this is showing up a little misaligned for me (in chrome)
> I didn't investigate a ton but it looks like it's being split in half between the filename and the button - any ideas?
> And in safari it's the same:

@marthacryan i've updated the button position so it is consistent across the browsers. the 50/50 split comes from the fact that i am using the `columnPanel` provided by canvas to display/group multiple items in the same row:

https://github.com/elyra-ai/canvas/wiki/3.3-Common-Properties-Controls#grouppanel-controls

marthacryan:
Just tried your new changes and style is working as expected! One (very) small thing - definitely doesn't need to be part of this PR, but it'd be great if you could select a file in that browser by double-clicking. Not sure how easy that'd be to add, just a note.

vabarbosa:
@marthacryan i used a different panel layout and worked around the 50/50 split. yes, i noticed the console errors when switching to chrome. looking into it now. and will look into enabling double-click to select also. thanks

vabarbosa:
> Sorry - just noticed that there's a console error that should probably be fixed before merging:
> 

@marthacryan i believe the error is coming from the carbon components modal (used by common canvas to display the properties dialog). more specifically, carbon components modal does not support stacked modals:

https://github.com/carbon-design-system/carbon/issues/5947
https://github.com/carbon-design-system/carbon/issues/582

in any case, i was able to figure out a workaround the issue and the error should no longer be thrown.

vabarbosa:
> Just tried your new changes and style is working as expected! One (very) small thing - definitely doesn't need to be part of this PR, but it'd be great if you could select a file in that browser by double-clicking. Not sure how easy that'd be to add, just a note.

@marthacryan, you should now be able to select a file by double-clicking

marthacryan:
@vabarbosa Thanks for addressing those! LGTM

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -456,6 +464,13 @@ export class PipelineEditor extends React.Component<
     }
     const app_data = node.app_data;
 
+    if (app_data.filename !== propertySet.filename) {
+      app_data.filename = propertySet.filename;
+      node.label = propertySet.filename
+        .replace(/^.*[\\/]/, '')
+        .replace(/\.[^/.]+$/, '');
```
kevin-bates, line 471: This is stripping off the directory name prefix and any suffix, leaving just the base filename.  Two questions...
1. Are there library methods that can do that is typescript (similar to Python `os.path`)?
2. If not, would you mind adding a simple comment to this effect so folks new to this can more easily maintain it?

I know folks will get this after a bit, but this is one of those places where a few words could save someone getting unnecessarily lost in the regex weeds.

ajbozarth: if this is just getting the filename then you should use `path.basename` we already use that on line 802 in this file (in the export function)

ajbozarth: also on a bit of looking we should probably make this change to use the `path` lib in a few other places in this file (it seems you took this regex from the add function).

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -456,6 +464,13 @@ export class PipelineEditor extends React.Component<
     }
     const app_data = node.app_data;
 
+    if (app_data.filename !== propertySet.filename) {
+      app_data.filename = propertySet.filename;
```
ajbozarth, line 468: As a note: this line will need to be updated to work with #861 is merged (in whichever PR is merged last)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: @kevin-bates there are libraries that do that but currently none are included as a dependency in the pipeline-editor package. since it is a short method similar code and it is being called in multiple places i have moved it into the `Utils` class (and updated and documented accordingly).

packages/pipeline-editor/src/utils.ts
```diff
@@ -146,4 +146,23 @@ export default class Utils {
...
   }
+
+  /**
+   * Return the last portion (filename) of a path. The file extension is
+   * stripped if `excludeExtension` is set to true.
+   *
+   * @param path
+   * @param excludeExtension
+   */
+  static getFilenameFromPath(path: string, excludeExtension?: boolean): string {
```
ajbozarth, line 157: There's no need to add this function. It already exists in the `path` library, in fact we do exactly this in the `handleExportPipeline` function. Look at the following on line 802:
```
    const pipeline_name = path.basename(
      pipeline_path,
      path.extname(pipeline_path)
    );
```
this accomplishes the same task and doesn't require adding a new util

vabarbosa: ah ok, will remove

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: i stand corrected. the library is already imported and being. will update accordingtl

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: Note this would specifically need to be updated to the following once #861 is merged or in #861 once this is merged.
```suggestion
      app_data.filename = PipelineService.getPipelineRelativeNodePath(
        this.widgetContext.path,
        propertySet.filename
      );
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: done https://github.com/elyra-ai/elyra/pull/867/commits/4dfe3be0a09a5964f7d5e9c6b55e888571b72314

packages/pipeline-editor/src/utils.ts
vabarbosa: done https://github.com/elyra-ai/elyra/pull/867/commits/4dfe3be0a09a5964f7d5e9c6b55e888571b72314

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -456,6 +464,11 @@ export class PipelineEditor extends React.Component<
...
```
ajbozarth, line 468: Note this would specifically need to be updated to the following once #861 is merged or in #861 once this is merged.
```suggestion
    const propertyFilename = PipelineService.getPipelineRelativeNodePath(
      this.widgetContext.path,
      propertySet.filename
    );
    if (app_data.filename !== propertyFilename) {
      app_data.filename = propertyFilename;
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -471,6 +484,28 @@ export class PipelineEditor extends React.Component<
...
+  propertiesControllerHandler(propertiesController: any): void {
+    this.propertiesController = propertiesController;
+  }
+
+  propertiesActionHandler(id: string, appData: any, data: any): void {
+    if (id === 'browse_file') {
+      const propertyId = { name: data.parameter_ref };
+      showBrowseFileDialog(this.browserFactory.defaultBrowser.model.manager, {
+        filter: (model: any): boolean => {
+          return model.type == 'notebook';
```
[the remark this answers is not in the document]
vabarbosa: yes, the filter is so you only show files that should be selectable. in this case, we only want to show `notebook` files (no point in showing all available files if user shouldn't be selecting them). and yeah, if the dialog is to be re-used in other context then you can configure what files do appear. for example, with File Dependencies (https://github.com/elyra-ai/elyra/issues/873) browser dialog would show more than just notebook files

PR #871, opened 2020-08-17
Add ability to group metadata

### LINKED ISSUE #789: Add ability to "group" fields in the metadata editor
In #591, it was noted that the metadata editor for runtimes would be better if it could group the cos fields separately from the kubeflow fields (see current display below). This could be added as an optional field in the ui hints and be picked up by the metadata editor. 
image

marthacryan:
Fixes #789. Using an optional "category" field in the "uihints" of a schema, adds a header when a new category encountered. 
image
 
First pass relies on order of schema properties to determine where to add the headers - planning to incorporate some sorting to make that less reliant on JSON ordering. 

           in the file; or

           it.


packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +102,32 @@ export class MetadataEditor extends ReactWidget {
         if (!this.name) {
           this.title.label = `New ${this.schemaDisplayName}`;
         }
+        this.sortedFields = [];
+        const categories: string[] = [];
+        // Find categories of all schema properties
+        for (const schemaProperty in this.schema) {
+          if (!this.schema[schemaProperty].uihints) {
+            this.schema[schemaProperty].uihints = {};
```
ajbozarth, line 110: I believe we do something similar to this in `renderField` might be worth updating that to remove any redundant code.

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +102,32 @@ export class MetadataEditor extends ReactWidget {
...
+          }
+          if (!this.schema[schemaProperty].uihints.category) {
+            this.schema[schemaProperty].uihints.category = null;
+          }
+          const category = this.schema[schemaProperty].uihints.category;
+          if (!categories.includes(category)) {
+            categories.push(category);
```
ajbozarth, line 117: ok so this took me some time to understand, but this add a `null` category doesn't it? That why it still grabs fields without a category. I'm not sure I like the idea of having a `null` in the array for this, would an empty string instead leave the same?

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +102,32 @@ export class MetadataEditor extends ReactWidget {
         if (!this.name) {
           this.title.label = `New ${this.schemaDisplayName}`;
         }
+        this.sortedFields = [];
+        const categories: string[] = [];
+        // Find categories of all schema properties
+        for (const schemaProperty in this.schema) {
```
vabarbosa, line 108: you should be able to simplify this `for` loop and remove the multiple `if` statements, like so:

```javascript
  const categories = new Set();

  for (const schemaProperty in this.schema) {
    categories.add(
        (this.schema[schemaProperty].uihints && this.schema[schemaProperty].uihints.category) || 
        null
    );
  }
```

and you can replace the `null` with whatever you want to represent no category (e.g., `''`, etc.)

vabarbosa: looking at this more closely. as a possible alternative, you could simplify further and actually replace the back-to-back `for` (line 108, line 120) with something like so:

```javascript
  const categories = { '_noCategory': [] }
  
  for (const schemaProperty in this.schema) {
    const cat = this.schema[schemaProperty].uihints && this.schema[schemaProperty].uihints.category;
    if (!cat) {
      categories['_noCategory'].push(schemaProperty)
    } else if (categories.hasOwnProperty(cat)) {
      categories[cat].push(schemaProperty)
    } else {
      categories[cat] = [schemaProperty]
    }
  }
```

this results in a `categories` that looks like so:
```json
{
  "_noCategory": [
    "description"
  ],
  "Kubeflow Pipelines": [
    "api_endpoint",
    "api_username",
    "api_password"
  ],
  "Cloud Object Storage": [
    "cos_endpoint",
    "cos_username",
    "cos_password",
    "cos_bucket"
  ]
}
```
which i believe gives you all the info to layout the fields in order by category...

marthacryan: That looks great! Thanks!

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +101,21 @@ export class MetadataEditor extends ReactWidget {
...
+        // Find categories of all schema properties
+        this.categories = { _noCategory: [] };
+        for (const schemaProperty in this.schema) {
+          const cat =
+            this.schema[schemaProperty].uihints &&
+            this.schema[schemaProperty].uihints.category;
+          if (!cat) {
+            this.categories['_noCategory'].push(schemaProperty);
+            // eslint-disable-next-line no-prototype-builtins
+          } else if (this.categories.hasOwnProperty(cat)) {
```
vabarbosa, line 113: an alternative to disabling the [eslint check is to write is as](https://eslint.org/docs/rules/no-prototype-builtins):

```javascript
} else if (Object.prototype.hasOwnProperty.call(this.categories, cat)) {
```

marthacryan: Thanks!

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +101,22 @@ export class MetadataEditor extends ReactWidget {
...
+        // Find categories of all schema properties
+        this.categories = { _noCategory: [] };
+        for (const schemaProperty in this.schema) {
+          const cat =
+            this.schema[schemaProperty].uihints &&
+            this.schema[schemaProperty].uihints.category;
+          if (!cat) {
+            this.categories['_noCategory'].push(schemaProperty);
+          } else if (
+            Object.prototype.hasOwnProperty.call(this.categories, cat)
```
ajbozarth, line 113: is there a reason we're using `Object.prototype` here and not just `if (this.categories[cat])` to check existence in the dict?

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +101,22 @@ export class MetadataEditor extends ReactWidget {
         if (!this.name) {
           this.title.label = `New ${this.schemaDisplayName}`;
         }
+        // Find categories of all schema properties
+        this.categories = { _noCategory: [] };
```
ajbozarth, line 105: Since this is now a dict containing all the fields this name isn't the most accurate anymore. Perhaps `schemaPropertyByCategory`, `sortedSchemaProperties`, or `categorizedSchemaProperties`. All are a bit wordy but better show that it is a dict of schema properties organized by category

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -100,6 +101,22 @@ export class MetadataEditor extends ReactWidget {
...
+        for (const schemaProperty in this.schema) {
+          const cat =
```
ajbozarth, line 107: similar to my comment above, we could rename this `category`

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -436,8 +453,18 @@ export class MetadataEditor extends ReactWidget {
...
   render(): React.ReactElement {
     const inputElements = [];
-    for (const schemaProperty in this.schema) {
-      inputElements.push(this.renderField(schemaProperty));
+    for (const category in this.categories) {
+      if (category !== '_noCategory') {
+        inputElements.push(
+          <h4 style={{ flexBasis: '100%', paddingBottom: '10px' }}>
+            {' '}
+            {category}{' '}
```
ajbozarth, line 461: Two questions: Are the spaces needed here? and if so why is this not all on one line?

PR #876, opened 2020-08-18
Build a source distribution of Elyra when creating a release

akchinSTC:
Conda-forge packages require a source distribution to be used instead
of a wheel. This will need to be published alongside the wheel in Pypi

We will need to retroactively(if possible) publish a source dist of elyra for 1.0.0 using the tagged release into pypi to get the conda forge package going, or wait until the next 1.x.x release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ build-ui: yarn-install lint-ui ## Build packages
 	export PATH=$$(pwd)/node_modules/.bin:$$PATH && lerna run build
 
 build-server: lint-server ## Build backend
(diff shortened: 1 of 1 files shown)


lresende:
@akchinSTC have you actually tried to "install from source package" in a clean environment? Does that work?

akchinSTC:
@lresende - it works. used a clean conda env with python 3.7. Uses still need to run `jupyter lab build` but thats the same install flow as with our wheel.

lresende:
Also published a 1.0.1 release with both wheel and source tar.

PR #877, opened 2020-08-19
Add JSON UI and optional sort for metadata widget

### LINKED ISSUE #851: Runtime image editor: present metadata in a user-friendly way
Follow up to https://github.com/elyra-ai/elyra/pull/835 (Add editor for runtime images)

The summary view currently displays a JSON dump of the metadata:
image

It'd be great if the information could be displayed in a formatted way (which might include a link to the image in the public registry), not just as a JSON dump. This would be similar to how we currently display runtime configuration metadata and link to the "target":

image

ajbozarth:
Fixes #851
 
Adds a default JSON viewer to the MetadataWidget and adds an optional sort which default to alphabetical by display name.

           in the file; or

           it.


ajbozarth:
Screen Shot 2020-08-18 at 5 03 25 PM
Screen Shot 2020-08-18 at 5 03 34 PM
Screen Shot 2020-08-18 at 5 04 00 PM
Screen Shot 2020-08-18 at 5 04 14 PM

ajbozarth:
I moved the new react component into the `ui-components` package and cleaned it up a bit. This should be ready for review now

ptitzler:
I guess I'm missing something. The tab doesn't show up after I ran `make clean install` on your branch. Is it supposed to be there?

ptitzler:
- Command/Ctrl Shift C
- `Elyra` > `Show Runtime Images`

So I don't have to look it up again (:

ptitzler:
Very convenient to have as an interim solution for generic metadata!

Three thoughts (I realize the last two might be not specific to this PR but since you are already touching the metadata widget ...):

- Is there a need to display the `metadata` node? If it doesn't serve an purpose and we would only ever display child properties of metadata (which seems to be the case for runtime images) there shouldn't be a need for it.
image

- Could the widget automatically sort the metadata entries by name?  As the list of entries grows it gets harder and harder to find the entry ...
image

- We distinguish between user-defined and system-defined metadata. In above screen capture only the last one is  user-defined metadata. As a user I can only _delete_ user-defined metadata. Trying to delete a system-defined metadata an error is raised. Could we hide the delete icon for system-defined entries that cannot be deleted?

ajbozarth:
> - Is there a need to display the `metadata` node? If it doesn't serve an purpose and we would only ever display child properties of metadata (which seems to be the case for runtime images) there shouldn't be a need for it.

This was a side effect of how the JSON Component was written. I can play with it some and see if I can show only the contents of a node without showing a parent. (should be possible)

> - Could the widget automatically sort the metadata entries by name? As the list of entries grows it gets harder and harder to find the entry ...

Actually having an optional `sort` that defaults to alpha when enable and can be overridden would be a good feature, I can tack it in here if it seems straight forward to add (otherwise I'll open another issue and pr to do it).

> We distinguish between user-defined and system-defined metadata. In above screen capture only the last one is user-defined metadata. As a user I can only delete user-defined metadata. Trying to delete a system-defined metadata an error is raised. Could we hide the delete icon for system-defined entries that cannot be deleted?

I actually discussed this some with @kevin-bates a while ago and currently the api doesn't distinguish whether metadata is user-defined or system-defined until you call the delete api. I believe a "fix" for this on either the front or back ends would be overly complicated and messy for little benefit. Especially because you can "edit" a default to override it, at which point you could delete that edited version.

kevin-bates:
> We distinguish between user-defined and system-defined metadata. In above screen capture only the last one is user-defined metadata. As a user I can only delete user-defined metadata. Trying to delete a system-defined metadata an error is raised. Could we hide the delete icon for system-defined entries that cannot be deleted?

Only the server distinguishes between user and system-define metadata instances.  We would need to add a meta-property to indicate this - probably something like a boolean-valued _`system_owned`_ would be sufficient.  

However, because we allow system-owned instances to be modified and essentially _become_ user-defined instances - which can be deleted - then things get very confusing.  For example, if we were to set the `system_owned` meta-property to `False` when a user "updates" a system-owned instance, then the UI would show a delete icon and allow its deletion.  But, upon its deletion, the actual system-owned instance - of the same name - reappears and, this time, without a delete icon.

As a result, I'd be inclined to a) add a system-owned meta-property where necessary (i.e., schemas that can _span_ the user/system domain), and b) disallow updates to system-owned instances.  This would mean that a user wishing to "update" a system-owned instance would be required to create a new instance with a different name containing the updated properties.

ptitzler:
> then things get very confusing. 

Yep, I noticed that:
 - edit a system-defined metadata entry (OK)
- delete the same entry (which is now a user-defined entry, OK)
- but an entry is still visible, which is  (again) a system-defined, which cannot be deleted (FAIL)

> This would mean that a user wishing to "update" a system-owned instance would be required to create a new instance with a different name containing the updated properties.

For the sake of simplicity that's the option  I would favor, even though a case could be made to not restrict editing/deletion of system-defined metadata at all. (Ignoring for now that a user can already do that by manually accessing the file system with the appropriate permissions.)

ajbozarth:
I pushed a couple commits addressing @ptitzler commit's. I updated the display to not show the "root" node and just the json contents. I also added an optional default alphabetical sorting based on `display_name` for `MetadataWidget`, which I disabled in `CodeSnippetWidget` but enabled for `RuntimesWidget`. Is alphabetical sorting something we would like for the Runtimes tab? Would we want to add it to Code Snippets?

New screenshot:
Screen Shot 2020-08-24 at 4 23 12 PM

ptitzler:
Thank you  Alex! 

I do believe all lists should be sorted by default using a criteria that enables the user to quickly find relevant information for the majority of use case scenarios. A system-sorted order (whatever that means specifically as to how the sort order is implemented explicitly)  shouldn't be used unless it provides a clear advantage. Let's assume (dunno if that is the case) the current system-sort order for the metadata widgets is "by change date [of the underlying file], descending". Does the user derive any benefits from having the list presented in that order? How would the criteria help the user find the appropriate entry faster compared to an alphabetically sorted list?

kevin-bates:
Regarding sorting, I agree that we should have a consistent message regarding sorting - with the default being that items are sorted by `display_name`.   Code snippets, however, should be _grouped by_ `language`, then sorted by `display_name`, with the groupings sorted by `language`.

I don't know what the implementation of the Code Snippets list widget (not sure that's the correct term) is, but we could probably introduce a meta-property to the schemas named something like `group_by`.  This would be an optional string-valued meta-property (i.e., resides outside the `metadata` stanza).  For Code Snippets, it would have a value of `language`.  Applications processing these instances would know to first group Code Snippets instances by `language`, _then sort_ them by `display_name` within a grouping.  Schemas that do not use `group_by`, always sort by `display_name`.

The same rules should apply to the CLI tooling as well.

ajbozarth:
Based on the conversation above I updated to also sort code snippets using the method Kevin described

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -233,14 +233,22 @@ class CodeSnippetDisplay extends MetadataDisplay<ICodeSnippetDisplayProps> {
     ];
   };
 
+  getDisplayName(metadata: IMetadata): string {
+    return `[${metadata.metadata.language}] ${metadata.display_name}`;
+  }
+
+  sortMetadata(): void {
+    this.props.metadata.sort((a, b) =>
+      this.getDisplayName(a).localeCompare(this.getDisplayName(b))
```
ptitzler, line 242: We should ignore case like it is done in the JupyterLab File Browser

image

ajbozarth: Is it not doing that, the lab sort code looks exactly the same as this.

ajbozarth: I just checked and it's already behaving that way
Screen Shot 2020-08-25 at 12 58 41 PM

ptitzler: Ah yes, my bad. I misread the MDN spec. It does appear to sort case insensitive by default.

packages/ui-components/src/JSONComponent.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface IProps {
+  json: any;
+}
+
+/**
+ * A React Component for displaying a json object
+ *
+ * A slimmed down copy of the `Component` class in @jupyterlab/json-extension
+ */
```
lresende, line 50: I have updated the license to reflect the info above.

PR #881, opened 2020-08-19
Enable file browser for dependencies in pipeline node properties

### LINKED ISSUE #873: Allow for updating File Dependencies using a file browser
Per the [comment](https://github.com/elyra-ai/elyra/pull/867#issuecomment-674924411) in #867, updating a Pipeline Editor node's **File Dependencies** via a file browser should be supported.

vabarbosa:
This PR add support for selecting file dependencies using using a `BrowseFileDialog` from the Properties dialog of the Pipeline Editor.

Fixes #873

- Both files and directories are selectable
- Multiple files and directories can be selected at a time
- File Dependecies textarea can still be manually updated/edited
- BrowseFileDialog opens at the notebook path and prevents navigating to parents/grandparents directories (subdirectories are navigable)
- Selected files and/or directories are appended (relative to the notebook's parent directory) to the File Dependecies textarea

elyra-dependencies-2

 
           in the file; or

           it.


ajbozarth:
I'm still reviewing and testing, but I found a couple UI issues I figured I'd share now

The folder icon is missing from the crumbs when in a subfolder, but is still visible when in "root"
Screen Shot 2020-08-19 at 12 08 09 PM

The `Add Dependancies...` button overlaps withe the text box
Screen Shot 2020-08-19 at 12 08 42 PM

both issues are visible in your gif above

vabarbosa:
> I'm still reviewing and testing, but I found a couple UI issues I figured I'd share now
> 
> The folder icon is missing from the crumbs when in a subfolder, but is still visible when in "root"
> 
> The `Add Dependancies...` button overlaps withe the text box
> 
> both issues are visible in your gif above

thanks, i have resolved the button alignment.

as far as the folder icon, that is the link back to JupyterLab's working directory. if a notebook is within a subdirectory then my understanding is the user should not be allowed to navigate to JupyterLab's working directory (hence the reason i hid it). in any case, i have updated PR and took a different approach. rather hide/remove directories not accessible by users, the directories are just disabled and user not allowed to click and traverse to it:

image


not sure which approach is preferred so feedback is welcomed.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -506,9 +506,12 @@ export class PipelineEditor extends React.Component<
   }
 
   propertiesActionHandler(id: string, appData: any, data: any): void {
+    const propertyId = { name: data.parameter_ref };
+    const filename = this.propertiesController.getPropertyValue('filename');
```
ajbozarth, line 510: Once #861 is merged this will need to be updated to
```suggestion
    const filename = PipelineService.getWorkspaceRelativeNodePath(
       this.widgetContext.path,
       this.propertiesController.getPropertyValue('filename')
     );
```

packages/ui-components/src/BrowseFileDialog.tsx
```diff
@@ -31,32 +31,80 @@ export interface IBrowseFileDialogOptions {
...
+        if ((crumb as HTMLSpanElement).title.indexOf(this.rootPath) === 0) {
+          crumb.className = crumb.className.replace(
+            'elyra-BreadCrumbs-',
+            'jp-BreadCrumbs-'
+          );
+        } else {
+          crumb.className = crumb.className.replace(
+            'jp-BreadCrumbs-',
+            'elyra-BreadCrumbs-'
+          );
```
ajbozarth, line 76: I'm not sure why we need this `else` block

packages/ui-components/src/BrowseFileDialog.tsx
```diff
@@ -31,32 +31,80 @@ export interface IBrowseFileDialogOptions {
...
+      const breadcrumbs = document.querySelectorAll(
+        '.elyra-browseFileDialog .jp-BreadCrumbs > span[title]'
+      );
+
+      breadcrumbs.forEach((crumb: Element): void => {
+        if ((crumb as HTMLSpanElement).title.indexOf(this.rootPath) === 0) {
+          crumb.className = crumb.className.replace(
+            'elyra-BreadCrumbs-',
+            'jp-BreadCrumbs-'
+          );
```
ajbozarth, line 71: is there not a better way of handling this than changing the class? Like adding a class and making the css target more specific? (editing html like this just feels hack-y to me, personal opinion)

packages/ui-components/src/BrowseFileDialog.tsx
vabarbosa: the way the `BreadCrumbs` works is as you navigate, the directory nodes are actually just getting the [`textContent`](https://github.com/jupyterlab/jupyterlab/blob/master/packages/filebrowser/src/crumbs.ts#L359) and [`title`](https://github.com/jupyterlab/jupyterlab/blob/master/packages/filebrowser/src/crumbs.ts#L361) updated. therefore, since the `className` is getting changed, need to make sure to undo the changes when appropriate. otherwise the `className` would be apply to incorrect node.

packages/ui-components/src/BrowseFileDialog.tsx
vabarbosa: `BreadCrumbs` processes click events based off of the `className` for the DOM node.

the approach taken here is to prevent the click event by changing the `className` (so it won't get processed). 

initially, i did try to override the `handleEvent` in `BreadCrumbs` as a better way to prevent handling of specific click events. unfortunately, `BreadCrumbs` is not really extensible and typescript prevents accessing the necessary private functions (e.g., `_evtClick`) and private variables (e.g., `_crumbs`) declared in `BreadCrumbs`.

another option is to re-implement the entire breadcrumbs widget and duplicate/copy over most of the existing `BreadCrumbs` code, but i suspect this is overkill (just to be able to disable couple of links)

vabarbosa: `BreadCrumbs` processes click events based off of the `className` for the DOM node. adding a class and making CSS target more specific won't prevent handling of the click event

the approach taken here is to prevent the click event by changing the `className` (so it won't get processed). 

in any case, i have updated the PR just added a new class instead (then use css to prevent the click event)

packages/pipeline-editor/src/PipelineEditorWidget.tsx
vabarbosa: done

PR #885, opened 2020-08-21
Fix missing field value in Metadata Editor

### LINKED ISSUE #883: Runtime configuration wizard: existing runtime configuration information is "lost"
Follow-up to https://github.com/elyra-ai/elyra/pull/847

I had created a runtime configuration with an earlier version of Elyra. Opening the configuration using the dev build that includes above PR the name and description in that configuration is not displayed:

image

ajbozarth:
The latest release of JupyterLab updated to the latest release
of blueprintjs core, which included an async bugfix for`InputGroup`
which broke how our async update interacts with fields. Thus
causing any already rendered `InputGroup` to not update it's value
on an async call to `update()` such as the one we do after calling
the metadata api.

I solved this by making the rendering of the display_name field
conditional on the existance of an inital value, just like we
already do with the other fields.

Fixes #883

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -216,9 +216,9 @@ export class MetadataEditor extends ReactWidget {
     this.handleDirtyState(true);
     // Special case because all metadata has a display name
     if (schemaField === 'display_name') {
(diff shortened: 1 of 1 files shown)


ajbozarth:
note: I also included a small update remove a deprecated property we were using and replace it with the property it was aliasing.

PR #887, opened 2020-08-21
[WIP] Fix migrate dialog bug

### LINKED ISSUE #863: The migration dialog appears unexpectedly and is vague
After shutting down Jupyter Lab with two pipeline tabs and a notebook tab open - with the notebook tab being active when shutdown occurs, I ran a version of elyra that included a pipeline version update and was presented with the following dialog upon restart.  

image (1)

There are two issues with this.
1. This dialog is being presented when the "active" tab is a notebook.
2. This dialog does not include the name of the pipeline being upgraded/migrated.  As a result, I can't tell which of my two pipelines it's going to change. 

We should only present the dialog when a pipeline is "in focus" and include the name of the pipeline in the message to make it clear.

karlaspuldaro:
Fixes #863 

Before these changes, the migrate pipeline dialog would pop up on a page reload or lab restart, even when the pipeline in old version tab wasn't on focus.
Now `PipelideEditorWidget` listens to `activeChanged` shell event and only handles open pipeline when the requested one actually becomes active (on reload or tab change).
The active pipeline title was also added to the dialog to give it more context.

Example below on a page reload to a python editor widget and `untitled.pipeline` version is outdated:
migrate-bug-fix

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -942,8 +942,10 @@ export class PipelineEditor extends React.Component<
             // in this case, pipeline was last edited in a "old" version of Elyra and
             // it needs to be updated/migrated.
             const filePath = this.widgetContext.path;
-            const extension = path.extname(filePath);
-            const pipelineName = path.basename(filePath, extension);
+            const pipelineName = path.basename(
+              filePath,
+              path.extname(filePath)
+            );
 
             const migrateDialogResult = await showDialog({
               title: "Migrate pipeline '" + pipelineName + "'?",
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -970,7 +970,7 @@ export class PipelineEditor extends React.Component<
               // proceed with migration
               pipelineJson = PipelineService.convertPipeline(
                 pipelineJson,
-                this.widgetContext.path
+                filePath
               );
               this.canvasController.setPipelineFlow(pipelineJson);
             } else {
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -941,9 +941,9 @@ export class PipelineEditor extends React.Component<
           } else {
             // in this case, pipeline was last edited in a "old" version of Elyra and
             // it needs to be updated/migrated.
-            const path = this.widgetContext.path;
+            const filename = this.widgetContext.path.split('/').pop();
             const migrateDialogResult = await showDialog({
-              title: 'Migrate pipeline ' + path + '?',
+              title: 'Migrate pipeline ' + filename + '?',
               body: (
                 <p>
                   This pipeline corresponds to an older version of Elyra and
@@ -961,17 +961,15 @@ export class PipelineEditor extends React.Component<
               ),
               buttons: [Dialog.cancelButton(), Dialog.okButton()]
             });
-            if (migrateDialogResult) {
-              if (migrateDialogResult.button.accept) {
-                // proceed with migration
-                pipelineJson = PipelineService.convertPipeline(
-                  pipelineJson,
-                  this.widgetContext.path
-                );
-                this.canvasController.setPipelineFlow(pipelineJson);
-              } else {
-                this.handleClosePipeline();
-              }
+            if (migrateDialogResult && migrateDialogResult.button.accept) {
+              // proceed with migration
+              pipelineJson = PipelineService.convertPipeline(
+                pipelineJson,
+                this.widgetContext.path
+              );
+              this.canvasController.setPipelineFlow(pipelineJson);
+            } else {
+              this.handleClosePipeline();
             }
           }
         }
@@ -1292,20 +1290,18 @@ export class PipelineEditor extends React.Component<
 
     const labShell = this.shell as LabShell;
 
-    const activeWidgetListener = (_: any, args: any) => {
-      const activeWidget = args.newValue;
+    this.initPropertiesInfo().finally(() => {
+      const activeWidgetListener = (_: any, args: any): void => {
+        const activeWidget = args.newValue;
 
-      // Only handle open pipeline if PipelineEditorWidget is active
-      if (activeWidget && activeWidget.id === this.widgetId) {
-        this.initPropertiesInfo().finally(() => {
+        // Only handle open pipeline if PipelineEditorWidget is active
+        if (activeWidget && activeWidget.id === this.widgetId) {
           this.handleOpenPipeline();
-        });
-
-        labShell.activeChanged.disconnect(activeWidgetListener);
-      }
-    };
-
-    labShell.activeChanged.connect(activeWidgetListener);
+          labShell.activeChanged.disconnect(activeWidgetListener);
+        }
+      };
+      labShell.activeChanged.connect(activeWidgetListener);
+    });
   }
 
   componentWillUnmount(): void {
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -35,7 +35,7 @@ import {
   errorIcon
 } from '@elyra/ui-components';
 
-import { JupyterFrontEnd } from '@jupyterlab/application';
+import { JupyterFrontEnd, LabShell } from '@jupyterlab/application';
 import { showDialog, Dialog, ReactWidget } from '@jupyterlab/apputils';
 import {
   DocumentRegistry,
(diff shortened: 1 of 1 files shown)


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1238,9 +1244,22 @@ export class PipelineEditor extends React.Component<
     node.addEventListener('lm-dragover', this.handleEvent);
     node.addEventListener('lm-drop', this.handleEvent);
 
-    this.initPropertiesInfo().finally(() => {
-      this.handleOpenPipeline();
-    });
+    const labShell = this.shell as LabShell;
```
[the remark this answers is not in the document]
karlaspuldaro: `PipelineEditorfactory` requires shell `as JupyterFrontEnd.IShell` type. We are casting it to `LabShell` to be able to access `activeChanged` function

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -891,8 +895,9 @@ export class PipelineEditor extends React.Component<
           } else {
             // in this case, pipeline was last edited in a "old" version of Elyra and
             // it needs to be updated/migrated.
-            showDialog({
-              title: 'Migrate pipeline?',
+            const filename = this.widgetContext.path.split('/').pop();
+            const migrateDialogResult = await showDialog({
+              title: 'Migrate pipeline ' + filename + '?',
```
kevin-bates, line 900: We refer to this as the pipeline, yet display the pipeline file.  I think using just the basename here (no parent directory or extension) would be best, along with quoting of the basename:

`Migrate pipeline 'untitled'?`

or reference it as the pipeline _file_.  Although taking this route should probably lead to an update of pipeline references to 'pipeline _file_' references below.  So I think the former might be best.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1238,9 +1244,22 @@ export class PipelineEditor extends React.Component<
...
-      this.handleOpenPipeline();
-    });
+    const labShell = this.shell as LabShell;
+
+    const activeWidgetListener = (_: any, args: any) => {
+      const activeWidget = args.newValue;
+
+      // Only handle open pipeline if PipelineEditorWidget is active
+      if (activeWidget && activeWidget.id === this.widgetId) {
+        this.initPropertiesInfo().finally(() => {
```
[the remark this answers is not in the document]
karlaspuldaro: Good idea, that would speed up a tiny delay when rendering the pipeline for the first time

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: I agree, this will look cleaner. We already have a basename but it includes the extension. Will push a commit with this suggestion soon

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -941,8 +941,12 @@ export class PipelineEditor extends React.Component<
           } else {
             // in this case, pipeline was last edited in a "old" version of Elyra and
             // it needs to be updated/migrated.
-            showDialog({
-              title: 'Migrate pipeline?',
+            const filePath = this.widgetContext.path;
+            const extension = path.extname(filePath);
+            const pipelineName = path.basename(filePath, extension);
```
ajbozarth, line 946: This can all be one line since you only use `pipelineName` below

karlaspuldaro: I thought it would be a long line and not as readable:
`const pipelineName = path.basename(this.widgetContext.path, path.extname(this.widgetContext.path));`
If that's still ok I can make the change

ajbozarth: the linter will line wrap it and it'll be readable, that's what every other call to `path.basename` in our code looks like.

ajbozarth: though you could also do
```suggestion
            const filePath = this.widgetContext.path;
            const pipelineName = path.basename(filePath, path.extname(filePath));
```
instead, which would also be good since you call `this.widgetContext.path` below and that could then use `filePath` as well

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -959,18 +965,17 @@ export class PipelineEditor extends React.Component<
...
-                this.canvasController.setPipelineFlow(pipelineJson);
-              } else {
-                this.handleClosePipeline();
-              }
             });
+            if (migrateDialogResult && migrateDialogResult.button.accept) {
+              // proceed with migration
+              pipelineJson = PipelineService.convertPipeline(
+                pipelineJson,
+                this.widgetContext.path
```
ajbozarth, line 973: you should just use `filePath` here

karlaspuldaro: Oops missed that one :p

ajbozarth:
I checked out the latest code and I'm seeing an odd behavior, not sure when it was introduced, but I no longer see the pop up when opening an out of date file until I change tabs away and back to it.

karlaspuldaro:
> I checked out the latest code and I'm seeing an odd behavior, not sure when it was introduced, but I no longer see the pop up when opening an out of date file until I change tabs away and back to it.

Thanks for catching it. I couldn't reproduce it while debugging step by step, but apparently moving `initPropertiesInfo()` back inside `activeWidgetListener()` seemed to solve this corner case. Let me know if it works for you.

lresende:
The latest code works for old pipelines (e.g. version 1), after migration, it does not open the pipelines anymore. I will leave this for 1.2.0.

ajbozarth:
Ok now I'm seeing a different odd behavior. If a good pipeline is opened but not "in focus", either on a page reload or if you open multiple files and switch tabs to it, it wont load the pipeline contents into the canvas until you click inside the tab somewhere. This seems related to the slight delay intros load when opening a pipeline (which flickers empty before loading in the canvas contents

karlaspuldaro:
@ajbozarth The is a slight delay when opening the pipeline for the first time since we're only loading its properties when it's in focus, although I couldn't recreate the behaviour you noted above. This is what I see on reload and switch tab to a good pipeline:
open-pipeline

marthacryan:
Not able to reproduce the behavior @ajbozarth is seeing

marthacryan:
Just tried in Safari and was able to see a delay on loading the pipeline, but it wasn't dependent on clicking in the pipeline.

karlaspuldaro:
Interesting. I think we should investigate the extra delay in safari then.

ajbozarth:
So I've done some further investigation while on a call with @karlaspuldaro and discovered the following:

Note my method of testing was to select multiple pipeline files (from various versions) and open them at once.

- In FF and Chrome when you open multiple files it actually opens each file in a tab prior to opening the next one, whereas in Safari it will open the first one and "open" the rest in the background. This is the source of the inconsistency I we were seeing. In FF and Chrome this means that it behaves the same as it currently does on master (multiple dialogs open, not corresponding to the open tab) but now with the new clearer text that tells the user the file being referenced, so I believe this is an acceptable fix. The issue with Safari stem from the next bug.

- When a tab is opened but not put in focus the canvas will appear empty until a user either clicks inside (which doesn't always work) or click outside then inside again (like switching tabs or clicking the file browser). In FF and Chrome this happen on a page reload, but in Safari it also happens when opening multiple files since they are opened in the background and never "focused", specifically it happens when one of the files you open prompts the dialog and you select cancel, which closes the tab, and the next tab doesn't have the dialog and isn't "in focus"

The second bullet is a bug introduced in this PR and the real cause of the weird Safari behavior, whereas the first bullet is just a UX grievance and is acceptable to leave in place.

karlaspuldaro:
I tested it out and couldn't recreate the last behaviours noted. The flickering between loading an empty to an existing pipeline happens when they are open for the first time, it is slightly slower in safari. The dialog is triggered when the correct pipeline is in focus. I also tested reloading a good pipeline in different browsers and the pipeline loads ok.
Opening multiple pipelines (one is old) then reloading the page in
Safari:
safari

Firefox:
firefox

Chrome:
chrome

karlaspuldaro:
So the bug I still see happens in firefox only:
opening multiple pipelines at once, the migrate dialog pops up when a different pipeline tab is in focus. Cancelling the dialog closes the tab on focus (which is not the old pipeline)
firefox-bug

lresende:
Is this a jupyterlab issue that we are working around in Elyra? I just got this which is actually opening a `select kernel` dialog out of context.

image

karlaspuldaro:
> Is this a jupyterlab issue that we are working around in Elyra? I just got this which is actually opening a `select kernel` dialog out of context.
This is new to me. Looks like it could be another symptom of pipeline widget not always being tracked as the active one.

lresende:
BTW, if this is ready for final review/merge, please remove the `work in progress` tag.

karlaspuldaro:
@lresende it is not ready yet, we need further investigation to fix the racing condition bug we've noted:

> opening multiple pipelines at once, the migrate dialog pops up when a different pipeline tab is in focus. Cancelling the dialog closes the tab on focus (which is not the old pipeline)

This only happens to me on FF but @marthacryan was able to reproduce it in Chrome as well

lresende:
Please reopen if we are still going to pursue anything here.

PR #895, opened 2020-08-26
Avoid validating pipelines that needs to be migrated

lresende:
This was also causing the red error message when the migration dialog appeared as pipeline version 1 had absolute paths.
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -921,7 +921,6 @@ export class PipelineEditor extends React.Component<
       } else {
         // opening an existing pipeline
         const pipelineVersion: number = +Utils.getPipelineVersion(pipelineJson);
-        this.setAndVerifyPipelineFlow(pipelineJson);
 
         if (pipelineVersion !== PIPELINE_CURRENT_VERSION) {
           // pipeline version and current version are divergent
@@ -974,6 +973,8 @@ export class PipelineEditor extends React.Component<
               }
             });
           }
+        } else {
+          await this.setAndVerifyPipelineFlow(pipelineJson);
         }
       }
     });


ajbozarth:
Rather than adding a version check inside our validation why don't we just not validate prior to conversion?

As far as I can tell this line is extraneous and the source of this problem, we could just remove it:
https://github.com/elyra-ai/elyra/blob/2bfcf9a903e56fc08435ccd8b87d80fe0c5228ef/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L914

kevin-bates:
This strikes me as a bit of a slippery slope and will likely lead to additional version checks scattered across the codebase when only the initial load of a pipeline should detect, and respond to, version levels and the rest of the code should be able to assume everything is at the expected version. 

If this is an _order of operations_ thing, then shouldn't version checks (with the possible trigger of migration) be performed _before any_ validation?  If the user then chooses **not** to migrate, shouldn't they see the validation errors since their pipeline is indeed invalid?  I don't think we should give down-level users a false impression that their pipelines are valid when they're not (relative to the current software version).

vabarbosa:
currently, if the user chooses not to migrate the pipeline is not opened/showed. so the user has no option to open/view the pipeline without migrating. if this is the preferred behavior then we should validate only if no migration is needed or after a successful migration. @ajbozarth  i believe this should involve just moving the `this.setAndVerifyPipelineFlow(pipelineJson);` out of [line 914](https://github.com/elyra-ai/elyra/blob/2bfcf9a903e56fc08435ccd8b87d80fe0c5228ef/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L914) and into an `else` clause: 

```javascript
if (pipelineVersion !== PIPELINE_CURRENT_VERSION) {
  ...
} else {
   this.setAndVerifyPipelineFlow(pipelineJson);
}
```

lresende:
I did as suggested by @vabarbosa but actually inside the `setAndVerifyPipelineFlow` to avoid duplication

ajbozarth:
> I did as suggested by @vabarbosa but actually inside the `setAndVerifyPipelineFlow` to avoid duplication

This is still adding a version check inside validation code, just farther upstream. @kevin-bates point still stands. I'm not sure what you mean by "avoid duplication", why would handling this where we already do a version check be duplicating code?

lresende:
Ok, this now only changes the `opening validation phase`.  And let all other validations as is.

kevin-bates:
Hmm.  It still seems like we should perform validation even after opting out of migration and the **first** check made during validation is that the version is a match.  In this case, unlike initial load, no migration wizard is presented, but the pipeline is flagged as invalid and the _rest_ of the validation is not performed.  This would then display the pipeline to the user, allowing them to open notebooks, adjust properties, etc., but they **must** first go through the migration wizard before attempting to run.  (Of course, I'm sure I'm over-simplifying things here, but I'm hoping that kind of functionality could be achieved.)

Skipping validation entirely on a canceled migration still seems to give a false impression.  Perhaps I'm missing something.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -968,12 +967,14 @@ export class PipelineEditor extends React.Component<
                   pipelineJson,
                   this.widgetContext.path
                 );
-                this.setAndVerifyPipelineFlow(pipelineJson);
+                await this.setAndVerifyPipelineFlow(pipelineJson);
```
ajbozarth, line 970: this line doesn't build. You can't put an await inside a non-async function (which this `then` function is not)

lresende: sorry, had the change locally to test but did after the commit...

lresende:
> Hmm. It still seems like we should perform validation even after opting out of migration and the **first** check made during validation is that the version is a match. In this case, unlike initial load, no migration wizard is presented, but the pipeline is flagged as invalid and the _rest_ of the validation is not performed. This would then display the pipeline to the user, allowing them to open notebooks, adjust properties, etc., but they **must** first go through the migration wizard before attempting to run. (Of course, I'm sure I'm over-simplifying things here, but I'm hoping that kind of functionality could be achieved.)
> 
> Skipping validation entirely on a canceled migration still seems to give a false impression. Perhaps I'm missing something.

When you cancel a migration, you cancel the opening of the pipeline, and except for the async aspects of the behavior, you shouldn't have a UI to see the errors.

kevin-bates:
Ok. I thought the pipeline editor would still be functional after a cancelled migration. Thanks.

PR #897, opened 2020-08-27
add runtime image documentation to user guide

### LINKED ISSUE #894: Document how runtime images can be managed
I don't think we have documented yet how users can manage runtime-images.

- Using the CLI: `elyra-metadata [list | install | remove] runtime images`
- Using the UI:  Open the command palette <ctrl/cmd> <SHIFT> <c>  and select  _Elyra_ > _Show Runtime Images_

The content should be similar to https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html for runtime configurations.

ptitzler:
This PR adds a new topic to the user guide, which outlines how to manage runtime image metadata using the UI and the CLI. Please note the intentional deviations from  https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html, which I used as the initial baseline. 
Opinionated reasoning for making those changes:
- The instructions are intended to be less abstract. Therefore I tried to avoid using the term metadata as much as possible and replaced "[runtime image] metadata" with "[runtime image] configuration". While it makes perfect sense from a developers point of view to name it "runtime images metadata" because it's a "subclass" of metadata (just like "runtimes metadata"  is a "subclass") , I believe a _user_ generally doesn't need to be aware of the underlying implementation details. 
- The instructions cover the UI first and the CLI second, since the former is more user-friendly. 
- The structure of the generated JSON is not documented.  If one wants to learn more about the structure, one can display an existing instance. The main reason for omission is (1) that the structure might change in the future (2) it reduces the amount of content and (3) knowing how the information is persisted provides little (or no) value to a regular user.
- The document does mention the concept of user-managed runtime image configurations but does not go into details how they deviate from system-managed runtime image configurations as this is a topic that applies to all "metadata subclasses". 

Feedback very welcome.

Closes #894 
 
           in the file; or

           it.


docs/source/user_guide/runtime-image-conf.md
```diff
@@ -0,0 +1,125 @@
...
+
+##### Description (description)
+Desscription for this runtime image definition.
+
+Example: `My custom runtime Docker image`
+
+##### Image Name (image_name)
+The name of an existing Docker image in the public Docker Hub container registry that meets the stated prerequisites. This property is required.
+
+Example: `jdoe/my-image`
```
lresende, line 125: I would recommend using the version tag to make things more reproducible, otherwise, you might start getting new "latest" versions when those become available.

ptitzler: ah yes, of course. excellent point! Will change the text and the screen capture.

ptitzler:
I've also documented the internal `name` identifier as a required property for the CLI. This surfaces a bit of a usability issue because of the convention I've used to denote the property's UI label and CLI name:

image

If people find this bothersome I could change the layout as follows:
- Only use a descriptive property name to the heading. Caveat is that this property name might not match the UI label in cases where the property is not user provided.
- Add a bulleted list with the property name, like so below the heading (example for the `name` property):
  ```
  ##### Internal configuration identifier
  A unique internal identifier for the runtime image configuration. The property is required when the command line interface is used manage a configuration. An identifier is automatically generated from the user-friendly name when a configuration is added using the UI.
  - UI label: N/A
  - CLI option: name
  ```

PR #900, opened 2020-08-27
Update README.md

ptitzler:
- Use `latest` Docker image tag instead of specific version (was `1.1.0`). This reduces the need to update certain portions of the readme whenever a new release is published.
- Added link to change log in Binder section
- Added disclaimer that start time for development image might be long (PRs are frequently merged into the master branch invalidating any cached image Binder might have stored)

           in the file; or

           it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,19 +56,22 @@ Click on a link below to try Elyra, on a sandbox environment, without having to 
 
 #### Using Docker
 
-You can also try Elyra via docker.
+You can also try Elyra by running one of the docker images from [Docker Hub](https://hub.docker.com/r/elyra/elyra/tags):
+- `elyra/elyra:latest` has the latest released version installed.
+- `elyra/elyra:x.y.z` has a specific version installed.
+- `elyra/elyra:dev` is automatically re-built each time a change is committed to the master branch.
 
-The command below starts a clean Elyra environment:
+The command below starts the most recent development build in a clean environment:
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,14 +61,14 @@ You can also try Elyra via docker.
 The command below starts a clean Elyra environment:
 
 ```
-docker run -it -p 8888:8888 elyra/elyra:1.1.0 jupyter lab --debug
+docker run -it -p 8888:8888 elyra/elyra:latest jupyter lab --debug
 ```
```
lresende, line 65: I believe master should be using the `dev` tag which is built automatically by docker hub on updates. Other docs such as the `JupyterHub guide`.

ptitzler: Oooooh, I think see. I was under the assumption we would promote the latest stable version Docker image, whereas you are saying it should be the latest development version image? If that's the case I can change the tag from `latest` to `dev` and add a blurb that hopefully helps the user make an informed decision which image would be the right one.

lresende: So, currently:
`latest` is updated during a release and points to the latest released version
`dev` is auto-built by docker hub and should be updated after commits (time-wise it depends on build queue size)

Also, note that during "releases" the readme does get updated to the release version, which then is available on the tag.

ptitzler: I added a list of options along with a link to Docker Hub to address your feedback

PR #906, opened 2020-09-01
Create a button in palette for all schema

marthacryan:
Automatically detects all schema and creates a button to open the widget for a given schema.
image

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata/src/index.ts
+++ b/packages/metadata/src/index.ts
@@ -182,7 +182,8 @@ const extension: JupyterFrontEndPlugin<void> = {
       }
     });
     app.contextMenu.addItem({
-      selector: '[data-id^="elyra-metadata:"]',
+      selector:
+        '[data-id^="elyra-metadata:"]:not([data-id$="code-snippet"]):not([data-id$="runtimes"])',
       command: closeTabCommand
     });
 
--- a/packages/metadata/src/index.ts
+++ b/packages/metadata/src/index.ts
@@ -159,17 +159,22 @@ const extension: JupyterFrontEndPlugin<void> = {
     });
 
     const schemas = await FrontendServices.getAllSchema();
-    console.log(schemas);
     for (const schema of schemas) {
       let icon = 'ui-components:text-editor';
-      if (schema.uihints && schema.uihints.icon) {
-        icon = schema.uihints.icon;
+      let title = schema.title;
+      if (schema.uihints) {
+        if (schema.uihints.icon) {
+          icon = schema.uihints.icon;
+        }
+        if (schema.uihints.title) {
+          title = schema.uihints.title;
+        }
       }
       palette.addItem({
         command: commandIDs.openMetadata,
         args: {
-          label: `Show ${schema.display_name}`,
-          display_name: schema.display_name,
+          label: `Manage ${title}`,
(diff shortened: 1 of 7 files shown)


kevin-bates:
Hmm.  The plurality is off.  Since namespace names are plural and schema names are singular, I wonder if the namespace name is what is wanted here anyway.  Since 2 of the 3 are single-schema namespaces, they just work.  The other would should KFP and AIrflow "runtimes" leaving the headings with...

Show Code Snippets
Show Runtime Images
Show Runtimes

Actually, the namespace names are lowercase and values like "code-snippets", "runtime-images" and "runtimes".  So this is another great example of promoting namespaces to full objects where they too could have display_name values - leading to what I showed above (with proper case and spacing).  Until then, I don't think this is going to fly, unfortunately.

marthacryan:
@kevin-bates What if we add 'Widget' after each of them? So "Show Code Snippet Widget" or "Show Runtime Image Widget"

kevin-bates:
To show a "widget" sounds odd to me.  

Also, I think we should forgo using `display_name` from the **schema** because its more of an application property _defined in the schema_.  Instead, we added `title` everywhere and that's what should be used here instead.  When dealing with metadata _instances_ you would reference `display_name`.

marthacryan:
@kevin-bates Yeah good point - maybe "browser"?

packages/metadata/src/index.ts
```diff
@@ -155,6 +157,26 @@ const extension: JupyterFrontEndPlugin<void> = {
...
+    for (const schema of schemas) {
+      let icon = 'ui-components:text-editor';
+      if (schema.uihints && schema.uihints.icon) {
+        icon = schema.uihints.icon;
+      }
+      palette.addItem({
+        command: commandIDs.openMetadata,
+        args: {
+          label: `Show ${schema.display_name}`,
+          display_name: schema.display_name,
```
kevin-bates, line 172: We should switch to using `title` from the schema file.
```suggestion
          label: `Show ${schema.title}`,
          display_name: schema.title,
```

marthacryan:
How's this look? I made it so that it uses the `title` by default but if there's a `uihints.title` it'll use that instead. 
image

karlaspuldaro (approved):
LGTM :)

marthacryan:
Ok while working with this I figured it might be good to also implement that "Close tab" ability for metadata viewers in the left side pane since they can really add up and now it's easy to reopen any of them. Just pushed those changes but if you don't think it's good with this PR I can undo that. Thoughts?

marthacryan:
Just added a small change so that it won't show the "Close tab" option for code snippets or runtimes (because they're supposed to be more permanent). It was causing a small bug of creating a generic metadata widget for code snippets (which doesn't include the coding languages in the display and doesn't include the insert / copy buttons).

marthacryan:
So here's what the context menu looks like for code snippets / runtimes:
image

And here's the context menu for any other metadata schema present:
image

karlaspuldaro (approved):
Nice adding a close tab option. Re-ran this locally and looks good.

lresende (approved):
LGTM - Tested locally and all looks good.

PR #918, opened 2020-09-09
Unpin urllib3 package to version 1.24.2 or greater

### LINKED ISSUE #914: urllib3 dependency conflicts with python 3.8 in conda-forge
I also posted this in https://github.com/conda-forge/elyra-feedstock/issues/4 but since there are no other issues there, figured this repo was better monitored. When installing elyra via conda, using the conda-forge channel, the dependency on urllib3=1.24.3 prevents the use of python 3.8 b/c there are no python 3.8-compatible versions of urllib3=1.24.3 on conda-forge (or conda's main channel). Wondering if it's possible to make that a ">=" dependency or if there's something specific about urllib3 version 1.24.3 that is needed for elyra

akchinSTC:
Unpin to satisfy package resolution when using python 3.8 with
conda-forge feedstock

Fixes #914

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup_args = dict(
         'requests>=2.9.1,<3.0',
         'rfc3986-validator>=0.1.1',
         'traitlets>=4.3.2',
-        'urllib3==1.24.2',
+        'urllib3>=1.24.2',
     ],
     include_package_data=True,
     classifiers=(


lresende (approved):

lresende:
Please update the recipe as well (we can just change the build number and it will update the recipe)

PR #932, opened 2020-09-16
fix broken doc link introduced by PR #929

ptitzler:
Fixes a problem which was introduced by #929. The referenced document yielded a 404 because the file extension was invalid. Verified that the new linke works as expected.


kevin-bates (approved):
LGTM

PR #967, opened 2020-10-09
Add exported file location to response dialog

### LINKED ISSUE #872: Successful pipeline exports should convey where the output can be found
Exporting a pipeline can produce a YAML or python file as output.  Upon its completion, which can take a while to produce, the successful export message makes no mention of the output file or its location.  If the user happens to initiate the export from a different location than the pipeline file, they will not see any new file produced in the left-hand pane.  Seems like, at a minimum, Elyra should let the user know where they can find the file produced from export.

image

It looks like the [`elyra/pipeline/export` request handler](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/handlers.py#L50-L68) is formulating both the export path in the response **and** setting the location header for the corresponding `api/contents` request to fetch the export file.

lresende:
Fixes #872
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineService.tsx
+++ b/packages/pipeline-editor/src/PipelineService.tsx
@@ -174,7 +174,7 @@ export class PipelineService {
       overwrite: overwrite
     };
 
-    await RequestHandler.makePostRequest(
+    const response = await RequestHandler.makePostRequest(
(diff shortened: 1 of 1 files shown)


lresende:
image

PR #980, opened 2020-10-19
Code Snippet should insert new cell after inserting a snippet

lresende:
When the user inserts a code-snippet into an "existing" cell, it most likely will need to work on something else on a new cell. The idea here is to give the user the same env as he started before adding the snippet (with a new cell to work on), similar to how Notebook does when you are on the last cell of the notebook where it creates a new one.  

 
           in the file; or

           it.


marthacryan:
Just tried this locally and it works for Python snippets, but markdown snippets are inserted into the current cell and then create a new blank cell. Any idea why that might be happening?

marthacryan:
Oh actually this appears to be happening for all languages of code snippets.

marthacryan (approved):
Tested locally and works as expected. LGTM

lresende:
Thanks @marthacryan, I have updated the description with a little more description of the motivations of this change.

PR #955, opened 2020-10-04
Improve subprocess error handling

kevin-bates:
This change improves the messages that users receive when encountering issues during node operation failures.  

For both notebook and python processes, the `capture_output` parameter has been enabled which results in the raised error object instance representing the error in its `stderr` property.  This property value is then decoded to unescape newlines - and stderr becomes part of the message that hereto now only appeared in the server log file - as discussed on #951.

For notebook operations, it also turns off the progress bar output of the `papermill` command that had been captured in the log file and was included in the `stderr` property following the previous change.  

Fixes #951 

Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #981, opened 2020-10-20
Remove use of capture_output - not available in py 3.6

kevin-bates:
This is essentially a rollback of #955.  When making that change I failed to realize that the subprocess `run` argument `capture_output` was only available in Python versions >= 3.7.  I happened to run on Python 3.6 and got an error.  This also implies we need to fix up our local processor tests since we _should_ be able to run something locally.
 
           in the file; or

           it.


lresende:
@kevin-bates I guess we have a few of these at `create-release.py` as well. Should we care about removing it from there?

kevin-bates:
I don't think that's necessary. We should probably try to _build_ releases using the latest supported Python since it might have better multi-release support built in. But it's probably good to _develop_ in the oldest. Just my opinion.

lresende (approved):

PR #1020, opened 2020-10-29
Fix jest tests

marthacryan:
Fixes the jest test failures by updating the testutils version. Changes to using port 8890 by default for tests to avoid conflicts when JupyterLab is running. 

Edit: no longer changes the port used by tests because of #1024 
 
       By making a contribution to this project, I certify that: 799189

           in the file; or

           it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "devDependencies": {
     "@4tw/cypress-drag-drop": "^1.3.1",
     "@cypress/webpack-preprocessor": "^4.1.5",
-    "@jupyterlab/testutils": "^2.2.4",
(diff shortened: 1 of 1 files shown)


lresende:
Could we update this on master (on top of #1024).

marthacryan:
@lresende Just tested locally on master and the newer version of testutils doesn't work with JupyterLab 2.2.9 due to server changes in 3.0, so I think we should keep this as a PR to the jupyterlab3 branch

marthacryan:
Also note: just removed the changes from the last two commits because they were duplicates of #1024 - this PR now only changes the version of the JupyterLab testutils.

lresende (approved):

PR #1043, opened 2020-11-06
Code Snippet drag and drop feature

### LINKED ISSUE #532: Code snippet drag and drop
The insert code snippet button provides the current integration with editors.
Having a drag and drop from the snippet display list into the editor (file editors and notebooks) would enhance the code snippets UX

karlaspuldaro:
Drag and drop code snippet from display list into notebooks and file editors

Fixes #532 

Adapted from https://github.com/jupytercalpoly/jupyterlab-code-snippets/
drag-and-drop

           in the file; or

           it.


marthacryan:
Just tried out the current progress and it looks like it's working great! Just a thought after testing - I think maybe it'd be nice to have the blue bar that appears between cells appear below the last cell as well (see below)
Screen Shot 2020-11-10 at 11 55 05 AM

karlaspuldaro:
@marthacryan I think that is a default notebook cell behaviour on a drag move event. When you select any cell and start shifting its position, the blue bar is only displayed if you intend to move it to the top or between cells. I will do some more investigation to check if that can be overridden on our side.

marthacryan:
Another thought - could markdown snippets automatically be added to notebooks as "markdown cells" rather than "code cells"?

marthacryan:
Also - (this can definitely be in another PR because I think it's not a blocker) maybe it'd be nice to be able to drop snippets into an existing cell if you're hovering over that cell - and it could drop the text wherever your cursor is. (And you could still add them as cells if you're hovering between cells)

karlaspuldaro:
Thanks @marthacryan for your feedback, just documenting what was discussed today based on your suggestions above,
1- Let's create an enhancement issue in lab regarding indicating the notebook blue horizontal bar when dragging a cell over the last position
2- For now we can still use the `insert` button to add a code snippet to an existing notebook cell. We can create a separate issue to further investigate and enhance the drag and drop option.

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -359,3 +539,57 @@ export class CodeSnippetWidget extends MetadataWidget {
...
+  /**
+   * Create a new markdown cell widget.
+   */
+  createMarkdownCell(options: CellModel.IOptions): MarkdownCellModel;
+}
+
+/**
+ * The default implementation of an `IModelFactory`.
+ */
+export class ModelFactory {
```
ajbozarth, line 561: These two classes look like they were taken from the console package in lab code, I think rather than copying them we should use/extend those classes instead

ajbozarth (changes requested):
Looking at the drag&drop handling code it feels like it's doing a lot more than it should need to (I might just not have enough experience with event handlers though), but since it works I'm ok with it as is. I included a couple comments below as well.

ajbozarth (approved):
I'm going to approve this for now and open a new issue for the drag drop selection bug in safari since it affects both this and pipeline as well and it's only an annoyance and only on safari

ajbozarth:
Opened issue: https://github.com/elyra-ai/elyra/issues/1074

lresende (approved):

PR #1048, opened 2020-11-09
Update runtime-images so lint side-effect doesn't occur

kevin-bates:
The addition of tags recently added to runtime-images is not in a format that `make lint-ui` wants so it automatically updates each json file - which ends up getting marked as changed files.  This update makes the files compatible with `make lint-ui`.
 
           in the file; or

           it.


lresende (approved):

ajbozarth:
For reference this was caused by merging a PR that force bypassed the automated push script. We may want update CLI to fail on the petty command instead of applying it like it does locally

PR #1062, opened 2020-11-11
Create a dropzone component for dragging and dropping jupyter files into canvas

bourdakos1:
Notes:
- adds `eslint-plugin-react-hooks` for checking react hooks
- turns off `react/prop-types` because we use TypeScript and it seems like we don't use PropTypes

           in the file; or

           it.


karlaspuldaro (approved):
LGTM

lresende:
Thanks @bourdakos1,  overall it looks good to me. One question I have is about the actual dropzone, is this something specific to pipelines? Would code-snippet, which is enabling some darg-drop support be able to reuse this? This could help us decide overall where the location of this component should reside.

bourdakos1:
@lresende I wouldn't say this is specific to pipelines. However, I'm guessing this won't be useful for code-snippets, because this dropzone component only handles the "drop". Where code-snippets handles "dragging from", and I assume the Notebook itself handles the actual "drop". 

If we have another widget where we are dragging and dropping something onto the widget itself, it would be useful there.

ajbozarth:
In general this look good to me, I don't fully understand the details in some of the new files, but I assume it's just leveraging react more directly. I do have a few meta comments though:

1.  Why is LuminoDropzone split into multiple files and kept in a directory, from previous experience in lab usually components like this would just be in a single file?
2. I think this components belongs in the ui-components package rather than pipeline
3. Why the name LuminoDropzone? 

I'll take some time to check this out and test locally later today as well

ajbozarth:
tested locally and it works great

bourdakos1:
@ajbozarth I’m just used to a react components being a folder. The index file is the react component, `useLuminoDropzone` is a react “hook” that could be used directly in a functional component. (Our pipeline class component can’t use the hook and has to use <LuminoDropzone>)

I can definitely move it to ui-components

I called it LuminoDropzone because the events it listens for are from “@lumino/dragdrop”

ajbozarth:
> I’m just used to a react components being a folder. The index file is the react component, useLuminoDropzone is a react “hook” that could be used directly in a functional component. (Our pipeline class component can’t use the hook and has to use )

That makes sense, I figured it was a react related reason. Prior to starting dev on lab I preferred creating subduers like this anyway.

> I can definitely move it to ui-components

Thanks, I'll "approve" once you've move it

> I called it LuminoDropzone because the events it listens for are from “@lumino/dragdrop”

gotcha

marthacryan (approved):
Tried locally and it worked as expected. Code LGTM. Thanks!

ajbozarth (approved):

packages/ui-components/src/LuminoDropzone/useLuminoDropzone.ts
```diff
@@ -0,0 +1,90 @@
...
+
+interface IReturn {
+  getRootProps: () => void;
+}
+
+interface IRootProps {
+  ref: React.MutableRefObject<ILuminoElement>;
+}
+
+export const useLuminoDropzone = ({ onDrop }: IProps): IReturn => {
```
lresende, line 33: maybe rename to `useDropzone`

lresende (changes requested):
Also, if using a directory, could we name it maybe `dragdrop` as in `Lumino` or `dropzone`?

packages/ui-components/src/Dropzone.tsx
```diff
@@ -0,0 +1,100 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { IDragEvent } from '@lumino/dragdrop';
+import React, { useCallback, useEffect, useRef } from 'react';
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/interface-name-prefix
```
ajbozarth, line 20: We should not be disabling lint unless absolutely necessary, based on our lint rules the following interface should be called `IHTMLElementEventMap`, is there a reason it can't be?

bourdakos1: @ajbozarth We are overriding the global `HTMLElementEventMap` not creating a new interface

bourdakos1: The `HTMLElementEventMap` type doesn't have `lm-dragenter`, `lm-drop`, etc.. so this just adds those to the type so we don't get typescript errors

ajbozarth: gotcha, ok the this still LGTM

lresende (approved):

PR #1075, opened 2020-11-16
Cleanup Pipeline Editor integration tests

lresende:

           in the file; or

           it.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Set up Node
       uses: actions/setup-node@v1
       with:
-         node-version: '12.13'
+         node-version: '12.18'
     - name: Display dependency info
       run: |
         python --version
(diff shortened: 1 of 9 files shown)


marthacryan:
**Note**: My last two commits add an extra line `getFileByName('helloworld.ipynb').click()` immediately before `getFileByName('helloworld.ipynb').rightclick()`, because for some reason, cypress' `rightclick` function does not select the element that it right clicks. The context menu appears, but the file browser element is not selected. Our code uses the selected elements in the file browser to determine what to add into the pipeline, so it breaks our code when the right click doesn't select the element in the browser. For now, adding the extra `click` call does the trick, but it might be good to consider whether we should change our code to make sure that we're using the element that was right clicked on somehow.

lresende:
@marthacryan, that seemed to have resolved that issue, thank you.

marthacryan (approved):
LGTM

akchinSTC (approved):
LGTMT

Issue #870, opened 2020-08-17
Node validation should be sensitive to runtime selection


kevin-bates:
Currently, the pipeline validation consists of ensuring no circular dependencies, required properties are set on each node and the target (notebook) file exists.  However, with respect to the required properties, properties such as `runtime_image` do not apply to "local" runtimes and their non-existence should not prevent users from proceeding.  As a result, we should perhaps defer the pipeline's validation until after the runtime has been selected.  If 'local' is selected, enforcement of `runtime_image` could be skipped.

Please note that because a pipeline's validation also occurs when the pipeline is opened, we should persist the runtime name (and schema name) and pipeline name on initial save/execution.  This will also enable the ability to default persisted names and runtimes on subsequent runs.

kevin-bates:
Since pipelines are also validated on 'save', I'm not sure what should happen in this case since it doesn't make sense to prompt for a runtime and the pipeline may have not been run before (so there's no persistent information from which to determine the runtime).

I'd be inclined to say we should not validate on save.

lresende:
We had some UX suggestions around this, where:

* We could have something similar to `classic notebook kernel selection dropbox` to select/associate a runtime to the pipeline
* We could use the current `run pipeline` button to run the pipeline with the associated runtime
* We could create a new `run as` which will either drop a list of runtimes or re-use the current UI to select the runtime

A side note, we could get rid of asking for pipeline name and reuse the pipeline file as we are using a timestamp to bypass the KFP limitation of unique pipeline names.

PR #1080, opened 2020-11-17
fix bugs in circular reference detection

bourdakos1:
There were a few cases with the old cycle detection algorithm that would cause it to infinitely recurse. This PR instead uses a loop to make it easier to timeout on long checks or an unforeseen infinite loop. I also added some tests to check for some common cycle patterns and tests for where the previous algorithm was failing. 
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -253,8 +253,6 @@ export class PipelineEditor extends React.Component<
     this.beforeEditActionHandler = this.beforeEditActionHandler.bind(this);
     this.tipHandler = this.tipHandler.bind(this);
 
-    this.nodesConnected = this.nodesConnected.bind(this);
-
     this.state = {
       showPropertiesDialog: false,
       propertiesInfo: {},
@@ -625,43 +623,23 @@ export class PipelineEditor extends React.Component<
     }
   }
 
-  /*
-   * Checks if there is a path from sourceNode to targetNode in the graph.
-   */
-  nodesConnected(
-    sourceNode: string,
-    targetNode: string,
-    links: any[]
-  ): boolean {
-    if (
-      links.find((value: any, index: number) => {
-        return value.srcNodeId == sourceNode && value.trgNodeId == targetNode;
-      })
-    ) {
-      return true;
-    } else {
-      for (const link of links) {
-        if (
-          link.srcNodeId == sourceNode &&
-          this.nodesConnected(link.trgNodeId, targetNode, links)
-        ) {
-          return true;
-        }
-      }
-    }
-    return false;
-  }
-
   beforeEditActionHandler(data: any): any {
+    if (data.editType !== 'linkNodes') {
+      return data;
+    }
+
     // Checks validity of links before adding
-    if (
-      data.editType == 'linkNodes' &&
-      this.nodesConnected(
-        data.targetNodes[0].id,
-        data.nodes[0].id,
-        this.canvasController.getLinks()
-      )
-    ) {
+    const proposedLink = {
+      id: 'proposed-link',
+      trgNodeId: data.targetNodes[0].id,
+      srcNodeId: data.nodes[0].id,
+      type: 'nodeLink'
+    };
+    const links = this.canvasController.getLinks();
+
+    const taintedLinks = checkCircularReferences([proposedLink, ...links]);
+
+    if (taintedLinks.length > 0) {
       this.setState({
         validationError: {
           errorMessage: 'Invalid operation: circular references in pipeline.',
@@ -671,9 +649,9 @@ export class PipelineEditor extends React.Component<
       });
       // Don't proceed with adding the link if invalid.
       return null;
-    } else {
-      return data;
     }
+
+    return data;
   }
 
   /*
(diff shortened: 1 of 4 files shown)


lresende:
@bourdakos1 while you are at it, does it make sense to start decoupling the validation code from the widget and start thinking how it would change its behavior when different runtimes are selected (see #870). And these are definitely different/new prs.

packages/pipeline-editor/src/validation.spec.ts
```diff
@@ -0,0 +1,343 @@
+/*
```
lresende, line 1: For consistency, should tests be placed under the test directory?

bourdakos1:
@lresende I've already decoupled validation from the widget in my large PR so that will come eventually. The behavior in the large PR is that it always performs cycle checks no matter the runtime, but then all other checks come from the node schema. So for the most part the validation is decided by the node, not the runtime. However, the runtime plugin decides what nodes to serve and could modify the schema to have different checks.

I think "running locally" is a special case and could be part of the node schema. for example instead of:
```
- id: runtime_image
  type: string
  required: true
  ...
```

We could have something like:
```
- id: runtime_image
  type: string
  required:
    runtime: true
    local: false
  ...
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1125,32 +1104,30 @@ export class PipelineEditor extends React.Component<
    * @returns null if pipeline is valid, error message if not.
    */
   validateAllLinks(): string {
```
lresende, line 1106: Thank you for starting this cleanup, at some point I really would like to see one call in the widget, which is PipelineService.validatePipeline(....) and everything else is driven from there (with calling other utility classes).

packages/pipeline-editor/src/validation.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// 3 seconds should be ample time.
+// should only protect against any infinite loops.
+const TIMEOUT = 3 * 1000;
```
lresende, line 19: When do you expect to get here? If we are trying to identify infinite loops, should we stop when we find the first occurrence of trying to visit a vertice that has been already processed as that identifies a circular dependency?

packages/pipeline-editor/src/validation.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface ILink {
+  id: string;
+  trgNodeId: string;
+  srcNodeId: string;
+  type: string;
+}
+
+export const checkCircularReferences = (links: ILink[]): string[] => {
```
lresende, line 28: Could you add a quick "function doc" on what it does?

packages/pipeline-editor/src/validation.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+  // │ a │──l0─►│ b │──l1─►│ c │──l5─►│ f │──l6─►│ g │
+  // ╰───╯      ╰───╯      ╰───╯      ╰───╯      ╰───╯
+  //                         │          ▲
+  //                        l2          l4
+  //                         ▼          │
+  //                       ╭───╮      ╭───╮
+  //                       │ d │──l3─►│ e │
+  //                       ╰───╯      ╰───╯
+  {
+    it: 'should not have a bug that a previous implementation had',
```
lresende, line 290: what was the bug? That might be a better description for someone seeing this for the first time...

packages/pipeline-editor/src/validation.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+  //                         │
+  //                        l4
+  //                         ▼
+  //                       ╭───╮      ╭───╮      ╭───╮      ╭───╮      ╭───╮
+  //                       │ f │──l5─►│ g │──l6─►│ h │──l7─►│ i │──l8─►│ j │
+  //                       ╰───╯      ╰───╯      ╰───╯      ╰───╯      ╰───╯
+  //                                                          ▲          │
+  //                                                          ╰────l9────╯
+  {
+    it: 'should handle long forks',
```
lresende, line 236: for the algorithm, what is the difference between this scenario and the one described on the test below?

packages/pipeline-editor/src/PipelineEditorWidget.tsx
bourdakos1: 👍 yea that's how it is in the large PR

packages/pipeline-editor/src/validation.ts
bourdakos1: There should never never be an infinite loop (unless there is a bug my the algorithm) this is just a safety net for that case

packages/pipeline-editor/src/validation.spec.ts
bourdakos1: When the algorithm gets to node `C` it either traverses down `l2` or `l4` first depending on the order of the links, this just checks that it doesn't matter which it traverses first

packages/pipeline-editor/src/validation.spec.ts
bourdakos1: I actually don't know haha... I just had this graph sketched out on a piece of paper of other graphs that had issues and couldn't remember what the issue was when writing tests so I gave it a vague name 😬

It either incorrectly indicated a cycle or had caused an infinite loop, but also could have not been and bug and wrote it down for another reason 🤷‍♂️

marthacryan:
Good catch! +1 on all of Luciano's comments (especially placement of spec file) but otherwise this is a good addition.

marthacryan (approved):

lresende (approved):

PR #1081, opened 2020-11-18
Support namespace configuration when using dex with kfp

### LINKED ISSUE #1053: Invalid input error: Invalid resource references for experiment. Expect one namespace type with owner relationship
I deployed kubeflow from using [this kfdef](https://gist.github.com/reevejd/7116780665e871f043005df6781f6ab7) (uses manifests from [v1.1-branch](https://github.com/kubeflow/kubeflow/tree/v1.1-branch))

When trying to run a pipeline on an external, multi-user auth cluster I’m getting a modal pop-up with this error:
Screen Shot 2020-11-09 at 5 20 14 PM

The relevant parts of the JupyterLab logs seem to be:
```
Validate experiment request failed.: Invalid input error: Invalid resource references for experiment. Expect one namespace type with owner relationship. Got:[]","message":"Validate experiment request failed.: Invalid input error: Invalid resource references for experiment.
```
[The full trace is in this gist](https://gist.github.com/reevejd/01cdacb743215969c279908fd68723c0).

Elyra successfully creates the pipeline, and if I then log in to the kubeflow GUI and manually create an experiment from the pipeline Elyra created, it works as expected.

I notice the same error message in a [kubeflow/pipelines issue here](https://github.com/kubeflow/pipelines/issues/4440) but it seems to only apply only to applications using the in-cluster endpoint, which isn’t the case for me.

akchinSTC:
Add user_namespace field to kfp metadata schema and
update kfp processor to allow submission of pipelines
to a dex enabled kfp cluster

Fixes #1053

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -53,6 +53,7 @@ class KfpPipelineProcessor(PipelineProcessor):
         # TODO: try to encapsulate the info below
         api_username = runtime_configuration.metadata.get('api_username')
         api_password = runtime_configuration.metadata.get('api_password')
+        user_namespace = runtime_configuration.metadata['user_namespace']
 
         self.log_pipeline_info(pipeline_name, "submitting pipeline")
         with tempfile.TemporaryDirectory() as temp_dir:
@@ -99,7 +100,10 @@ class KfpPipelineProcessor(PipelineProcessor):
                 else:
                     raise lve
 
-            run = client.run_pipeline(experiment_id=client.create_experiment(pipeline_name).id,
+            experiment = client.create_experiment(name=pipeline_name,
+                                                  namespace=user_namespace)
+
+            run = client.run_pipeline(experiment_id=experiment.id,
                                       job_name=timestamp,
                                       pipeline_id=kfp_pipeline.id)
 
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -51,7 +51,7 @@ class KfpPipelineProcessor(PipelineProcessor):
         cos_bucket = runtime_configuration.metadata['cos_bucket']
 
         user_namespace = None
-        if runtime_configuration.metadata['user_namespace']:
+        if runtime_configuration.metadata.get('user_namespace'):
             user_namespace = runtime_configuration.metadata['user_namespace']
 
         # TODO: try to encapsulate the info below
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -50,9 +50,7 @@ class KfpPipelineProcessor(PipelineProcessor):
         cos_endpoint = runtime_configuration.metadata['cos_endpoint']
         cos_bucket = runtime_configuration.metadata['cos_bucket']
 
-        user_namespace = None
-        if runtime_configuration.metadata.get('user_namespace'):
-            user_namespace = runtime_configuration.metadata['user_namespace']
+        user_namespace = runtime_configuration.metadata.get('user_namespace')
 
         # TODO: try to encapsulate the info below
         api_username = runtime_configuration.metadata.get('api_username')
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -50,10 +50,13 @@ class KfpPipelineProcessor(PipelineProcessor):
         cos_endpoint = runtime_configuration.metadata['cos_endpoint']
         cos_bucket = runtime_configuration.metadata['cos_bucket']
 
+        user_namespace = None
+        if runtime_configuration.metadata['user_namespace']:
+            user_namespace = runtime_configuration.metadata['user_namespace']
+
         # TODO: try to encapsulate the info below
         api_username = runtime_configuration.metadata.get('api_username')
         api_password = runtime_configuration.metadata.get('api_password')
-        user_namespace = runtime_configuration.metadata['user_namespace']
 
         self.log_pipeline_info(pipeline_name, "submitting pipeline")
         with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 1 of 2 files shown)


lresende:
Could you please add the related issue to the description...

elyra/pipeline/processor_kfp.py
```diff
@@ -99,7 +100,10 @@ def process(self, pipeline):
                 else:
                     raise lve
 
-            run = client.run_pipeline(experiment_id=client.create_experiment(pipeline_name).id,
+            experiment = client.create_experiment(name=pipeline_name,
+                                                  namespace=user_namespace)
```
lresende, line 104: This is not a required name, I believe @kevin-bates mentioned before that there is a need to handle `defaults` as a special case here as an empty string wouldn't sufice?

lresende: Never mind, I looked in the code and we are ok.

lresende (approved):
LGTM

lresende:
@akchinSTC  Should I wait for @reevejd feedback before merging?

akchinSTC:
@lresende - yes lets wait until he responds. He will be testing first thing in the morning.

kevin-bates (approved):
These changes look good Alan - thanks.  
(I'm assuming you confirmed that `runtime_configuration.metadata['user_namespace']` is `None` when a user namespace value is not specified.)

kevin-bates:
Thanks for the last update. I'm afk but will look into the metadata behavior tomorrow. I think we should be able to use something like .get('user_namespace') and should look why that isn't happening. I may also be misremembering things. 🙂

kevin-bates:
I think the issue with `user_namespace` being the empty string when the intention is to not set is the front-end issue that they always set a value, in this case the empty string.  I believe this is because they can't distinguish between null and the empty string in the UI.  As a result, the `user_namespace` field will always have a "value" - and logic like Alan has provided will be necessary since that field is persisted with a value of an empty string.

Were the UI able to distinguish "set" vs. "not set", then they could simply not include `user_namespace` in the metadata when it is created in the first place resulting in the object's persistence w/o `user_namespace`.  

In fact, we should probably add an existence check to see if the key (`user_namespace`) is present just in case this occurs.  I believe the CLI tool could probably create such an instance (i.e., one where no `user_namespace` field is persisted).

elyra/pipeline/processor_kfp.py
```diff
@@ -50,6 +50,10 @@ def process(self, pipeline):
         cos_endpoint = runtime_configuration.metadata['cos_endpoint']
         cos_bucket = runtime_configuration.metadata['cos_bucket']
 
+        user_namespace = None
+        if runtime_configuration.metadata['user_namespace']:
```
kevin-bates, line 54: Per my [last comment](https://github.com/elyra-ai/elyra/pull/1081#issuecomment-729879775) I recommend the following:
```suggestion
        if 'user_namespace' in runtime_configuration.metadata and runtime_configuration.metadata['user_namespace']:
```

lresende: or use `get('user_namespace')`

reevejd:
> @akchinSTC Should I wait for @reevejd feedback before merging?

I can confirm that @akchinSTC 's fix is working. Thanks a bunch!

kevin-bates:
Excellent.  Thank you for all your help with this James.  Alan - great (and simple) fix!  (The best kind there is!)

elyra/pipeline/processor_kfp.py
akchinSTC: @lresende - removing the if statement and using `get('user_namespace')` when using an empty `user_namespace` field in the runtime metadata returns the original error in the issue. Issue with using `get` atm is that it will never return `None` since the `user_namespace` key will always exist in the metadata dict, even when the field is empty. 
@kevin-bates - Since the key will always exist, the extra check for the key in metadata is redundant. I think the current check should be enough where an empty string would be considered "falsy"

kevin-bates: I think once the front-end issue is resolved, the key won't exist when no value is present because it won't be persisted.

kevin-bates: This statement (directly from the docs) should produce an instance in which `user_namespace` is not persisted and is valid metadata:
```
elyra-metadata install runtimes --schema_name=kfp \
       --name=my_kfp \
       --display_name="My Kubeflow Pipeline Runtime" \
       --api_username=username@email.com \
       --api_password=mypassword \
       --cos_endpoint=http://minio-service.kubeflow:9000 \
       --cos_username=minio \
       --cos_password=minio123 \
       --cos_bucket=test-bucket
```

akchinSTC: > I think once the front-end issue is resolved, the key won't exist when no value is present because it won't be persisted.

Ive opened #1087 for this. Should we wait then until that issue is fixed and merged before merging this one?

kevin-bates: There's no need to revisit this if you add the field existence test now.  This does mean a condition will be required before the .get() call.

kevin-bates: Also note that the condition can occur now via the CLI tool.  You should be able to produce a `KeyError` if not setting a user_namespace value (and without adding the field existence test).

elyra/pipeline/processor_kfp.py
```diff
@@ -51,7 +51,7 @@ def process(self, pipeline):
         cos_bucket = runtime_configuration.metadata['cos_bucket']
 
         user_namespace = None
-        if runtime_configuration.metadata['user_namespace']:
+        if runtime_configuration.metadata.get('user_namespace'):
             user_namespace = runtime_configuration.metadata['user_namespace']
```
kevin-bates, line 55: Damn - I'm sorry.  I spaced that using `.get('non-existent-key')` will return `None`!  So all you need (as Luciano pointed out long ago :blush: ) is the assignment statement:
```suggestion
            user_namespace = runtime_configuration.metadata.get('user_namespace')
```
with out the `if` statement.  I apologize.

akchinSTC: removed and updated

Issue #785, opened 2020-07-28
Application should be the root of the elyra build tree


ajbozarth:
After creating a graphviz of the Elyra build tree and a long discussion with @lresende we should refactor the services in `application` to not depend on `ui-components`. How we should handle the refactor is not yet decided, but I will include my opinions based on my conversation with @lresende in a comment below.

Current build tree:
![elyra-buils](https://user-images.githubusercontent.com/13952758/88607031-08563b80-d033-11ea-8f22-e1eee341c639.png)

ajbozarth:
My idea for this would include multiple refactors around how `application` is currently organized. In no particular order:

- move `service.tsx` out of `application` into `metadata-common`, this would also include doing some renaming like changing the class and file name to `MetadataServices` to more accurately match it's actually use and be in line with similar classes such as `CodeSnippetService` and `PipelineService`
- remove `parsing.ts` and relocation it's contents. `NotebookParser` make more sense to be in `utils.ts` in `pipeline-editor` (especially since the merger of pipeline-editor and submit-notebook). `IDictionary` could then also move into `requests.tsx` since it would be used as the return type for all the it's functions after the next refactor bullet.
- the big refactor: In `requests.tsx`, move the error handling functions to `ui-components` and have the functions return actually errors instead. The error handling would then be moved to the callees of these functions (`CodeSnippetService`, `PipelineService`, and the newly refactored `MetadataService`), whose packages already depend on `ui-components`. This would remove `application`'s dependency on `ui-components`
- One other refactor to consider is renaming. After the above work `requests.ts` would be the only remaining file and might warrant a better name, similarly `application` itself might be more accurately named `services`.

As stated these are my ideas on handling this purposed refactor, if they are generally agreed with I would be willing to tackle this refactor after I finish on my next project (so in a sprint or two)

ajbozarth:
Also note that bullet 3's refactor would allow for each service to handle it's errors uniquely, such as PipelineService displaying it's error using the new error banner on the pipeline-editor

ajbozarth:
I've opened the following PRs tackling the points above:

> 1. move `service.tsx` out of `application` into `metadata-common`, this would also include doing some renaming like changing the class and file name to `MetadataServices` to more accurately match it's actually use and be in line with similar classes such as `CodeSnippetService` and `PipelineService`

https://github.com/elyra-ai/elyra/pull/1083

[...]

PR #1083, opened 2020-11-18
Move metadata service out of application into metadata-common

ajbozarth:
Initial work on #785 

This specifically addresses point 1 in my comment detailing my plan https://github.com/elyra-ai/elyra/issues/785#issuecomment-664719618

- Moves `services.ts` from `application` to `metadata-common`. This is where it logically belongs, but `metadata-common` did not exist when `services.ts` was created.
- Rename `FrontendServices` to `MetadataService`, this brings it in line with other service classes and clarifies that it only handles metadata related services.
- Move related tests and sets up tests on metadata-common since it did not previously have tests.
 

Next steps:

Now that all service classes that call `requests.ts` are out of `application` we can move error handling downstream from `requests.ts` into the service classes that call it. Thus removing `application` dependency on `ui-components`

           in the file; or

           it.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -17,5 +17,4 @@
 import '../style/index.css';
 
 export * from './parsing';
-export * from './services';
 export * from './requests';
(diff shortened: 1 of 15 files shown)


ajbozarth:
I am working on this branch iteratively and will be immediately starting work on the next step on top of this.

I have made the move and name change separate commits so if the move is approved but the name change is not I can revert the name change easily

karlaspuldaro:
Looks good to me so far :)
Also tested it locally just in case and it works well.

PR #1084, opened 2020-11-18
Replace common properties dialog with side pane

marthacryan:
Opens a side pane view of the node properties editor instead of a dialog. 
image

- [x] Replace dialog with side pane
- [x] Show currently selected node's properties in the side pane (changes on single-clicking a node)
- [x] Replace broken "title editor" at top of side pane with the "label" of the node **OR** fix "title editor" and allow users to change the label of a given node. 
- [x] **EDIT** No longer adding this in this PR: Add refactored code from #969 
- [x] Clean up css to look closer to the JupyterLab appearance (i.e. save / cancel buttons, etc.) 
- [x] Move descriptions of inputs from inside the text box to a question mark that can be hovered over

           in the file; or

           it.


kevin-bates:
I like this idea - thank you.  

I think Luciano may have touched on this in the scrum regarding real estate, but I'm wondering if there's a way to introduce ellipses at a certain point to prevent the wrapping of things like `Filename` and even our checkbox labels.  I think it would be nice to keep those to a single line but allow the full content to be visible when hovering over the ellipses or something.  I've seen some extremely long (obnoxious really) file hierarchies and one of those would fill up the window if left unattended. :smile:

Can this pane be "un-docked" then resized/widened?

marthacryan:
@kevin-bates There'll be some CSS changes coming soon to address the ellipses comment 👍. In terms of the floating dock, it doesn't seem to be in the scope of this PR to add that, but I added an issue https://github.com/elyra-ai/canvas/issues/379 and we can also look into implementing that ourselves in a future PR.

marthacryan:
Current appearance:
image

Expanded:
image

On hover for long paths (**note**: need feedback on this - any ideas on what I should be doing here?) :
**Before hover:**
image
**While hovering:**
image

Dropdown:
image

ajbozarth:
I'll take a look at this once I return from vacation in 2 weeks (since this is still WIP and no rush)

ptitzler:
We should mark required properties.

ptitzler:
An undesired side effect that this design change introduces is that a user might unintentionally loose changes, as illustrated in this scenario:
- add two nodes to a pipeline
- change the properties of the first node (without clicking `OK`)
- select the second node
- select the first node - the change in the first node is lost

Previously a user was forced to either save or discard changes before another node was selected. If possible the behavior should be replicated:
- if node properties were changed and the user selects another node, a prompt should be displayed, notifying the user

marthacryan:
@ptitzler Thank you for the feedback!

> * We should mark required properties. If the widget doesn't support that (yet) natively, the info text should state if a value is required.

Good idea! Just added it. I don't think we marked any fields as required before this PR, does this look ok to you? 
image

> * Probably there should not be empty default entries ^^ just like there is none for dependencies

I actually added that on purpose (I thought maybe it'd be easier to have that initial empty entry? The user isn't typing with the dependencies list so I removed this behavior for that one). But it's easy enough to remove, I'm not that attached to that behavior. 

> * The button labels should be specific:`Add item` -> `Add environment variable` + ...

Good idea - just added that as well (see above)

ptitzler:
> I don't think we marked any fields as required before this PR, does this look ok to you?

marthacryan:
> Inconsistency:
> 
> * drag a notebook onto the canvas and define the required properties. save the changes. open the properties again.
>   image
> * select a different file in the properties view. save the changes. open the properties again. the file extension is gone
>   image

@ptitzler I found the code that causes this inconsistency - it appears to have been intentional to remove the extension, do you think it's better to remove the extension consistently or to keep the extension consistently? @ajbozarth @lresende I think I remember intentionally removing the file extension, did we stop doing that because of python file support in pipelines? (note: you can still tell which type of file it is based on the icon to the right of the label) I can change it either way just let me know.

ptitzler:
> I found the code that causes this inconsistency

🥳 

> it appears to have been intentional to remove the extension, do you think it's better to remove the extension consistently or to keep the extension consistently?

From a usability point of view I believe it'd be better to keep the file extension by default. If desired, a user can always change the node name to anything they'd like, such as an arbitrary string or just the file's base name.

marthacryan:
Alternate idea for showing the longer filepaths:
image

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -58,6 +58,7 @@ import { Color } from '@material-ui/lab/Alert';
 
 import 'carbon-components/css/carbon-components.min.css';
 import '@elyra/canvas/dist/styles/common-canvas.min.css';
+import '../style/index.css';
```
karlaspuldaro, line 61: Maybe this import isn't necessary since it's also in `index.ts`

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1270,6 +1298,27 @@ export class PipelineEditor extends React.Component<
       this.handleOpenPipeline();
     });
   }
+
+  componentDidUpdate(): void {
+    const inputFields = document.querySelectorAll('.properties-readonly');
```
karlaspuldaro, line 1303: I wonder if there is a way to access how uihints are rendered from `properties.json` and use it instead of reading info from the DOM to create a tooltip from scratch?

karlaspuldaro (commented):
Really nice work! Just had a couple minor comments to add so far.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
marthacryan: I agree - I tried looking into it and this was my last option :/ And it seemed better than having a long text field

I might be missing something, but from investigating I found that you could have descriptions based on the value of fields: https://github.com/elyra-ai/canvas/wiki/3.2-Common-Properties--UI-Hints#dynamic-text-expressions, but I wasn't able to get it working and it seems like it might be only for creating a sum or percent? 

I also looked into the `readonly` [code](https://github.com/elyra-ai/canvas/blob/04cc96417a35eb9aff5e409ab3398ef2ebbf77a9/canvas_modules/common-canvas/src/common-properties/controls/readonly/readonly.jsx#L98) and it seems like there's only a tooltip if the control is within a table? I don't know much about it and it doesn't seem to be documented anywhere.

marthacryan:
@ptitzler The editor size being persisted is an issue on the canvas side but I opened an issue there: https://github.com/elyra-ai/canvas/issues/436

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: We can always clarify that with the canvas team later, I was just curious if we could reuse their tool, good to know you've already tried something out :) I'm fine with keeping it as it for now.

marthacryan:
@ptitzler You should be good to check it out again!

marthacryan:
Updated to include auto-updating the properties instead of pressing an "OK" button. 

Here's the tooltip that shows the full file path:
image

Here's the appearance in general:
image

ptitzler:
Super. I'll test drive it tomorrow!

packages/pipeline-editor/src/properties.json
```diff
@@ -47,48 +53,72 @@
...
+        "description": {
+          "default": "Docker image used as execution environment"
         }
       },
       {
         "parameter_ref": "dependencies",
         "label": {
           "default": "File Dependencies"
         },
-        "place_holder_text": {
```
lresende, line 58: In general, the long texts were to help self-document what users can expect/do with these fields. Are we removing/trimming them due to lack of space?

marthacryan: I moved this text into the description because I changed the UI and there is no longer placeholder text for dependencies. I kept the descriptions the same in general, but in this case, I removed two lines because "One filename or expression (e.g. *.py) per line. Supported patterns: ? and *." is unnecessary - now you add entries by pressing the "Add dependencies" button. (see below)
image

lresende: With this, are we not allowing the user to just type (*.py)? What happens if the user tries to load a pipeline that already has (*.py) on it? I believe we should still enable users to type (*.py) or (*.csv) as there might be a lot of files? What do others think?

ptitzler: I agree. It should still be possible to type in an expression, like it was possible in previous releases.

marthacryan: Oh ok - how do you think that should be structured? I can definitely make the fields editable. Should I add a separate button to create a blank field? I could add it next to the "Add dependencies" button, or it could be in the file browser dialog?

marthacryan: Something like this? (Add dependency creates a blank entry, Browse files opens the filebrowser) 
image

ptitzler: This seems to one of the cases where a special kind of widget would work best - one that allows for the selection of a file or specification of an "arbitrary" string. In essence an amalgamation of what we use today select the input file and to type in output file specifications.

marthacryan:
image
Updated UI for file dependencies.

lresende (approved):
LGTM

lresende:
I will leave it open for the day in case others want to review it before I go ahead and merge.

karlaspuldaro (approved):
Looks great!
Maybe just check the warnings on the `renderControl()` function, otherwise LGTM

PR #1088, opened 2020-11-19
Remove ui-components dependency from application

ajbozarth:
This addresses point 3 in https://github.com/elyra-ai/elyra/issues/785#issuecomment-664719618

This moves the error handling code in `requests.tsx` that depends on `ui-components` into `ui-components` and replaces them with a Promise reject

Then it wraps all the calls to `requests.tsx` in try catch blocks that call the error handling that was moved to `ui-components`

This PR is built on top of https://github.com/elyra-ai/elyra/issues/785#issuecomment-664719618 and thus also includes all it's changes

           in the file; or

           it.

### DIFF
--- a/packages/application/package.json
+++ b/packages/application/package.json
@@ -32,11 +32,9 @@
     "watch": "tsc -w"
   },
   "dependencies": {
-    "@elyra/ui-components": "^1.5.0-dev",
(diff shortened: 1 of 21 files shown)


ajbozarth:
For reference this is what the dependency tree looks like after these changes:
graphviz

Issue #1091, opened 2020-11-20
"Making server request..." dialog content should be customizable


ptitzler:
Various tasks in the Elyra UI display a generic message dialog 

![image](https://user-images.githubusercontent.com/13068832/99827972-2dede000-2b0f-11eb-947e-101561f6ddc2.png)

It would be nice if both, the heading and the text would be optionally customizable to provide the user with relevant information. 

For example, a pipeline export could take a while and it's not necessarily obvious to the user why. Context information might provide insights why:

```
Processing pipeline ...

Packaging and uploading pipeline artifacts to cloud storage. This may take some time.

[OK]
```

ptitzler:
As far as I can tell the following features should benefit from this:
 - local pipeline execution
 - remote pipeline execution
 - export pipeline

ajbozarth:
I have this partial implemented and will be continuing work on it today

Issue #917, opened 2020-09-09
Document naming convention to use in Elyra components, file names, etc


lresende:

kevin-bates:
Could you please provide more details and/or examples of what you're talking about?  Is this purely for developers or are you referring to artifact naming conventions - pipeline/notebook files, outputs/dependencies, etc?  Thanks.

lresende:
This is for developers/contributors, and mostly I want to know how to name new files (see example below)

![image](https://user-images.githubusercontent.com/382917/93790430-dfdc5f00-fbe7-11ea-8d9b-b235aa228e93.png)

PR #1116, opened 2020-11-26
Remove ui-component dependency from application

### LINKED ISSUE #785: Application should be the root of the elyra build tree

ajbozarth:
Fixes #785 

- Moves the error handling ui to `ui-components`
- Updates RequestHandler to return rejected Promises instead of the error dialog
- Updates Service classes to pass error handling onto the callee (not use `await`)
- Renames `FrontendServices` to `MetadataService`
 
           in the file; or

           it.


ajbozarth:
This is an alternate implementation to #1088 based on discussions with @lresende 

I've marked this WIP because I was unable to test the error case where the Elyra api endpoint can't be found due to running into a new bug.

I'll take another look at said bug when I get back on Tuesday, but @kevin-bates if you are able to look at it a bit, when I ran `jupyter serverextension disable elyra` to disable the server extension for testing it is correctly listed as disabled, but when I run lab the elyra api is still available (ie the server extension is running). This seems like it may be similar to the same issue with front end extensions I found in https://github.com/jupyterlab/jupyterlab/issues/9234

ajbozarth:
graphviz
build dep chart for ref

ajbozarth:
@kevin-bates pointed out that for server extensions the command for disabling is different, so I was able to test the 404 case and this is now ready for general review. 

I also opened https://github.com/elyra-ai/elyra/pull/1127 to display server extensions properly to avoid future confusion

karlaspuldaro (approved):
Tested it locally, LGTM.

packages/application/src/MetadataService.tsx
```diff
@@ -26,7 +26,7 @@ const ELYRA_METADATA_API_ENDPOINT = 'elyra/metadata/';
 /**
  * A service class for accessing the elyra api.
  */
-export class FrontendServices {
+export class MetadataService {
```
lresende, line 29: Should this be renamed to .ts?

packages/application/src/requests.tsx
```diff
@@ -25,40 +24,6 @@ import * as React from 'react';
  * A service class for making requests to the jupyter lab server.
  */
 export class RequestHandler {
```
lresende, line 26: This file should also be renamed to .ts and you will most likely see that the inline React notation for HTML will break. I also noticed we are returning showDialog on 404 handler which should not be on this package if we are removing UI pieces altogether.

lresende (changes requested):
See inline comments

packages/application/src/MetadataService.tsx
ajbozarth: yes we should change this to a .ts ext

ajbozarth:
@lresende I made the changes you recommended, there's still some Dialog related code there, but that will be removed in my work for https://github.com/elyra-ai/elyra/issues/1091 once I finish that (I have a local wip that I'm working on)

packages/application/src/MetadataService.ts
```diff
@@ -14,35 +14,16 @@
  * limitations under the License.
  */
 
-import { Dialog, showDialog } from '@jupyterlab/apputils';
-import * as React from 'react';
-
 import { IDictionary } from './parsing';
-import { RequestHandler } from './requests';
+import { RequestHandler } from './RequestHandler';
```
lresende, line 18: I believe this should be `./requests`

packages/application/src/MetadataService.ts
```diff
@@ -14,35 +14,16 @@
...
-   */
-  static noMetadataError(namespace: string): Promise<Dialog.IResult<any>> {
-    return showDialog({
-      title: 'Error retrieving metadata',
-      body: <p>No {namespace} metadata has been configured.</p>,
-      buttons: [Dialog.okButton()]
-    });
-  }
-
+export class MetadataService {
```
lresende, line 26: this file should be `metadata.ts`

ajbozarth: I though we had decided to use the standard react naming convention of naming a file after the class it contains, even for utils. (We discussed in one of Nick's PRs last month I believe and decided to update them as we work on them rather than doing a refactor) I took the chance to update the naming convention for these since I was already changing the signature anyway. I'm ok changing them back to `requests` and `metadata` if you still think we should, I don't really have strong opinions on this.

lresende: Where is that documented? I still see this https://github.com/elyra-ai/elyra/issues/917 open. I am really looking for consistency, either we document and make sure we are using what is documented or not, but I really don't want to change one by one and have a mess.

ajbozarth: understandable, and it's possible I'm remembering a verbal conversation at a scrum rather than one on a PR or Issue. At least for now I'll update this to leave the naming as is and we can revisit the naming issue after we return from the holidays

ajbozarth:
@lresende I believe I've address all your previous review comments

lresende (approved):

PR #1127, opened 2020-12-01
List server extension on clean script make install

ajbozarth:
Currently we only list notebook server extensions, we should also list jupyter server extensions since elyra is one now
 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ install-external-extensions:
 install: install-server install-ui install-external-extensions ## Build and install
 	jupyter lab build
 	jupyter serverextension list
+	jupyter server extension list
 	jupyter labextension list
 
 watch: ## Watch packages. For use alongside jupyter lab --watch
--- a/etc/scripts/clean-jupyterlab.sh
+++ b/etc/scripts/clean-jupyterlab.sh
@@ -64,5 +64,7 @@ jupyter --version
 echo " "
 jupyter serverextension list
 echo " "
+jupyter server extension list
+echo " "
 jupyter labextension list
 echo " "


Makefile
```diff
@@ -107,6 +107,7 @@ install-external-extensions:
 install: install-server install-ui install-external-extensions ## Build and install
 	jupyter lab build
 	jupyter serverextension list
+	jupyter server extension list
```
lresende, line 110: I believe that on master, which has migrated to "jupyter server" we can remove the live above using `serverextension`.

etc/scripts/clean-jupyterlab.sh
```diff
@@ -64,5 +64,7 @@ jupyter --version
 echo " "
 jupyter serverextension list
 echo " "
+jupyter server extension list
+echo " "
 jupyter labextension list
```
lresende, line 69: same as above

Makefile
ajbozarth: we could remove it since we, elyra, doesn't use notebook server apps on master, but they are still supported by lab and "could" be installed still so I think leaving it in is worthwhile for clarity during development

lresende (approved):

ptitzler:
Does the command need to be also added to https://elyra.readthedocs.io/en/latest/getting_started/installation.html#verify-installation  and https://github.com/elyra-ai/elyra#verify-installation?

ajbozarth:
@ptitzler yes, I forgot to check for the line in doc files. Do you mind making that update, my branch is in the middle of other work

PR #1141, opened 2020-12-07
update container-image targets in Makefile

ptitzler:
This PR:
 - renames the existing [Elyra] `docker-image` target to `container-image` (so we start using more general terminology)
 - adds a `publish-container-image` target, which  pushes the Elyra container image to Docker Hub and quay.io. This is a privileged task and only users with the appropriate access to the target project on hub.docker.com and quay.io can perform it by default, after they've logged in to the registries. 

The main goal of this PR is to assure that each task in the Elyra lifecycle is "documented", which is not yet the case for the container image publication process is not, as far as I can tell scanning at the documentation. 
 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,11 +146,16 @@ dist-ui: build-ui
 
 release: dist-ui build-server ## Build wheel file for release
 
-docker-image: ## Build docker image
+container-image: ## Build container image
 	@mkdir -p build/docker
 	cp etc/docker/elyra/Dockerfile build/docker/Dockerfile
 	cp etc/docker/elyra/start-elyra.sh build/docker/start-elyra.sh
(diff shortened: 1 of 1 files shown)


ptitzler:
Please let me know if such is change is okay. If it is'll add one more commit to reflect the change in the appropriate documentation topic https://elyra.readthedocs.io/en/latest/developer_guide/development-workflow.html#building-the-elyra-container-image

akchinSTC (commented):
Should we update the directory names as well or does it still make sense to keep them as is?

ptitzler:
For the sake of consistency I would be in favor to do that too, but the main gap I was trying to close is the lack of "documentation" of the publication process. Even though there are just a few folks actually performing the container image publication, anybody should be able to easily find out how it is done. FWIW I stumbled across this when I was looking for clues about how `latest` images are formally built and released and checked the `Makefile`)

lresende (approved):

PR #1146, opened 2020-12-09
Use pytest-plugins array to declare fixtures since entry points have been removed

kevin-bates:
The `pytest-jupyter` package has removed its use of entry points to register its fixtures since this approach would require that `jupyter_server` be installed for **any** packages being tested in environments in which `pytest-jupyter` has been installed, even though those tests have nothing to do with Jupyter Server.  As a result, it is now necessary to use the `pytest-plugins=['plugin-name']` approach to letting pytest know which fixtures are needed. 

This PR adds that declaration and addresses a test failure introduced by another recent fixture change relative to the base URL fixture `jp_base_url`.

**Note: All metadata-handler tests will fail until this PR is merged.**
 
           in the file; or

           it.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,19 @@
+#
+# Copyright 2018-2020 Elyra Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


kevin-bates:
Just FYI - it sounds like the pytest-jupyter package will be going away soon (the repo will remain as a place to document various jupyter pytest fixtures).  Instead, they'll be moving the jupyter_server and jupyter_core fixture definitions back to their respective repositories since having them separated was just too difficult to manage during release publications.  Until then, this PR will be necessary to complete CI testing for Elyra 2.0.  I will update our references to the new locations once the change is in place.

lresende (approved):

PR #1155, opened 2020-12-11
Enables removal of optional metadata fields

### LINKED ISSUE #1087: Metadata editor should not persist empty values for optional fields unless explicitly set
Feel free to rename issue ^^

When retrieving values from the `runtime_configuration.metadata` dictionary for use in the processor, empty fields in the UI should not produce empty values for keys in the dictionary. Ideally the key would not be present in the dictionary at all. 

See #1053 and PR #1081

ajbozarth:
Currently once you add value for an optional field in the
MetadataEditor you are unable to remove a value entirely.

This only enables removing an optional value for text fields

Fixes #1087

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -245,6 +245,11 @@ export class MetadataEditor extends ReactWidget {
     // Special case because all metadata has a display name
     if (schemaField === 'display_name') {
       this.displayName = event.nativeEvent.target.value;
+    } else if (
+      !event.nativeEvent.target.value &&
+      !this.requiredFields.includes(schemaField)
+    ) {
+      delete this.metadata[schemaField];
     } else {
       this.metadata[schemaField] = event.nativeEvent.target.value;
     }


kevin-bates (approved):
The code changes look good (thank you) and I tested using the `user_namespace` and `description` fields of kfp schema and the `description` field of the code-snippet schema.

@ajbozarth - you mention this only addresses text fields.  Array-based fields (like tags) appear fine, but what about other optional _scalar_ fields like integer, number (float), boolean, etc.?  Although we don't have schemas (other than [test schemas](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/metadata-test.json)) that define fields of those types yet, are there plans to add such handling only once there are real (non-test) uses or prior?

((FWIW, if the server is started with env `METADATA_TESTING=1`, you will be able to see (and use) namespace `metadata-tests` and schemas `metadata-test` and `metadata-test2` in case you wanted to check out other data types ahead of time.))

ajbozarth:
text field is the default catch all on the front end for all those fields at the moment, so if the front end adds customized handling for them in the future we would want to add a case like this accordingly

kevin-bates:
ok - thanks.

lresende (approved):

PR #1204, opened 2021-01-16
Re-add resuse support

### LINKED ISSUE #1161: Add support for jupyter-resource-usage
jupyter-resource-usage, previously nbresuse, has moved it's jupyterlab UI out of core and into a separate extension as of lab 3.0. Previously we enabled the UI by adding a dependency on the notebook server extension via our setup.py

Once a new version of jupyter-resource-usage including the labextension is released we should add the extension to elyra in the same way as we do the git extension

https://github.com/jupyter-server/jupyter-resource-usage

ajbozarth:
Fixes #1161 

nbresuse UI code was moved from core lab into an extension within the api, which was renamed and moved to the juoyter-server org. In order to keep the UI in Elyra we need to add it manually like we do git

Also fix some miswritten build lines for lsp support

 
           in the file; or

           it.


lresende:
How easy would be to enable CPU usage tracking as well?

lresende:
Also, do we need to enable the extension as we do with the others at https://github.com/elyra-ai/elyra/tree/master/etc/config/jupyter_server_config.d. It does seem that it gets enabled by the pip install, but just in case...

ajbozarth:
> Also, do we need to enable the extension as we do with the others at https://github.com/elyra-ai/elyra/tree/master/etc/config/jupyter_server_config.d. It does seem that it gets enabled by the pip install, but just in case...

I'll add it

> How easy would be to enable CPU usage tracking as well?

I'll take a quick look and see how trivial or non-trivial this is, iirc it was non-trivial when it was in core

ajbozarth:
@lresende added the config json file, but as for CPU tracking, it's not even in the lab extension currently. So it's non-trivial, but we could contribute it to the lab extension ourselves if we want to. It wouldn't be very difficult add, but It would still take a non-trivial amount of time

lresende (approved):
LGTM

PR #1209, opened 2021-01-19
Schedule GitHub actions to run daily

lresende:

           in the file; or

           it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,9 @@ on:
     branches: '*'
   pull_request:
     branches: '*'
+  schedule:
+    # once a day
+    - cron:  '0 3 * * *'
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ on:
   pull_request:
     branches: '*'
   schedule:
-    # once a day
+    # once a day at 3 am (UTC) (7 pm (PST))
     - cron:  '0 3 * * *'
 
 jobs:


.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ on:
     branches: '*'
   pull_request:
     branches: '*'
+  schedule:
+    # once a day
+    - cron:  '* */3 * * *'
```
kevin-bates, line 24: I don't think this is correct for "once a day".  I believe this is saying to run the job every minute of every third hour of every day.  I think we need to pin a _time_ to run, let's say 8:42 am (UTC), so that would be:
```suggestion
    # every day at 8:42 am (UTC)
    - cron:  '42 8 * * *'
```
It might be good, practically speaking, to pick a time > 8 am so that the job runs on the same day, regardless of TZ (except for folks in Hawaii).

lresende: From the [documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule), my understanding was that it was going to run 3 AM, I can set it to run on the 0 minutes.

.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ on:
     branches: '*'
   pull_request:
     branches: '*'
+  schedule:
+    # once a day
```
kevin-bates, line 23:
```suggestion
    # once a day at 3 am (UTC) (7 pm (PST))
```

kevin-bates (approved):
Thanks.  I suggested an optional comment update that might prevent having to look up the cron scheduling syntax.

PR #1221, opened 2021-01-21
Switch to using the new pipeline editor package

bourdakos1:
Todo:
- [x] Export pipeline is using the wrong path
- [x] Use css variables from jupyterlab
- [x] Fix icons on string array fields of properties editor
- [x] Run pipeline throws error
- [x] Not re-running validation when saving properties
- [x] Fix CI
- [x] Rebase off master
- [x] Fix icon on checkbox for properties editor
- [x] Double click opens file represented by pipeline node
- [x] Switch to using LabIcon
- [ ] Polish css
- [x] Move errors from dialog to banner
- [ ] Add jest tests
- [x] Filter file types based on node types in file browser dialog
- [x] Add new fields to properties editor
- [x] Fix clear button
- [x] Add dynamic runtime images in properties
- [x] Fix console error on runtime image load
- [x] Fix dropdown chevron icon
- [x] Add file to pipeline sends signal many times
- [x] open file from context menu
- [x] Add file to pipeline adds to all open pipelines (might need help on this one)
- [x] Resource usage doesn't persist
- [x] Filepaths when submitting don't convert back to correct filepath

           in the file; or

           it.


marthacryan:
Todo:
- [ ] Export pipeline using the wrong path
- [ ] Use css variables from jupyterlab
- [ ] Fix icons on string array fields of properties editor
- [ ] Run pipeline throws error
- [ ] Not re-running validation when saving properties

.eslintrc.json
```diff
@@ -11,9 +11,16 @@
   "rules": {
     "react/prop-types": "off",
     "func-style": "error",
-    "@typescript-eslint/interface-name-prefix": [
+    "@typescript-eslint/naming-convention": [
```
ajbozarth, line 14: is there a reason we're changing this so it no longer matches jupyterlab's lint rules?

packages/ui-components/style/icons/clear-pipeline.svg
```diff
@@ -1,4 +1,4 @@
-<svg focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16" viewBox="0 0 32 32" aria-hidden="true">
+<svg id="icon" focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16" viewBox="0 0 32 32" aria-hidden="true">
```
ajbozarth, line 1: svg files shouldn't have id's in them like this since multiple instances of it can be added to the dom, this is what was causing the R icons issues we were dealing with last sprint. if this was added for css reasons instead use class

.eslintrc.json
bourdakos1: `@typescript-eslint/interface-name-prefix` has been deprecated in favor of `@typescript-eslint/naming-convention`

bourdakos1: this probably shouldn't be a part of this PR though

lresende (approved):
LGTM, various other contributors also reviewed and tested.

PR #1224, opened 2021-01-21
Change kernel name to display name

### LINKED ISSUE #1223: Use kernel display name on python editor toolbar
The current Python editor kernel dropdown displays kernel names as options.
Using the kernel display names instead would avoid the scenarios:

1. Remote kernel names as python variable names
image

2. Conda kernel names are long and hard to read
image

karlaspuldaro:
Fixes #1223 

Using kernel display names:
image

           in the file; or

           it.

### DIFF
--- a/packages/python-editor/src/KernelDropdown.tsx
+++ b/packages/python-editor/src/KernelDropdown.tsx
@@ -16,7 +16,6 @@
 
 import { ReactWidget } from '@jupyterlab/apputils';
 import { KernelSpec } from '@jupyterlab/services';
-import { ISpecModel } from '@jupyterlab/services/lib/kernelspec/kernelspec';
 import { HTMLSelect } from '@jupyterlab/ui-components';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


ptitzler:
Would there ever be a case where no display name is defined for a kernel?

kevin-bates:
Thanks @karlaspuldaro.  I agree with this, although I believe this is a departure from what is displayed as the kernel name in the notebook - so I suppose one could have a "consistency argument".  Also, could you please include an applicable screenshot for the conda kernel?  I believe they use a bracketed name of the conda env somewhere (prefix or suffix).  I had no idea the kernel name was so obnoxious for conda kernels.

To @ptitzler's point, would it be possible to fall back to the kernel name in cases where there is no display name?  Since there's no enforcement of this, it's possible to have a kernelspec w/o a display name.  I wonder what the launcher shows in such cases, since the launcher shows display names. (confirmed: it doesn't display any text, just the registered icon.)

karlaspuldaro:
@kevin-bates This is how the conda kernels look like now:
image

And good catch @ptitzler , will add Kevin's suggestion for that case.

packages/python-editor/src/KernelDropdown.tsx
```diff
@@ -83,12 +84,13 @@ class DropDown extends React.Component<DropDownProps, DropDownState> {
...
-    kernelNames.forEach((specName: string, i: number) => {
+    const kernelSpecs: ISpecModel[] = Object.values(specs.kernelspecs);
+    kernelSpecs.forEach((spec: ISpecModel, i: number) => {
+      const displayName = spec.display_name;
       const elem = React.createElement(
         'option',
-        { key: i, value: specName },
-        specName
+        { key: i, value: displayName },
+        displayName
```
lresende, line 93: Just to be clear, we should `display` the `displayName`, but continue using the `specName` as the value retrieved and associated during kernel creation.

lresende (approved):

PR #1232, opened 2021-01-22
Handle null/none values in pipeline operation ENVs

### LINKED ISSUE #1225: Pipeline Run/Export fails after deleting environment variable
**Describe the issue**
Create a pipeline and add a node with multiple environment variables
Delete one of the environment variables
Try to run:

```
 HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/elyra/pipeline/schedule?1611290432757', version='HTTP/1.1', remote_ip='::1')
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_local.py", line 74, in process
        operation_processor.process(operation)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_local.py", line 184, in process
        envs.update(operation.env_vars_as_dict())
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/pipeline.py", line 115, in env_vars_as_dict
        if len(nv) > 0:
    TypeError: object of type 'NoneType' has no len()
```

Try to export:
```
Uncaught exception POST /elyra/pipeline/export?1611290528686 (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/elyra/pipeline/export?1611290528686', version='HTTP/1.1', remote_ip='::1')
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_kfp.py", line 150, in export
        kfp.compiler.Compiler().compile(pipeline_function, absolute_pipeline_export_path)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/kfp/compiler/compiler.py", line 923, in compile
        package_path=package_path)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/kfp/compiler/compiler.py", line 977, in _create_and_write_workflow
        pipeline_conf)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/kfp/compiler/compiler.py", line 813, in _create_workflow
        pipeline_func(*args_list)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_kfp.py", line 149, in <lambda>
        pipeline_function = lambda: self._cc_pipeline(pipeline, pipeline_name)  # nopep8
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_kfp.py", line 257, in _cc_pipeline
        result = [x.strip(' \'\"') for x in env_var.split('=', 1)]
    AttributeError: 'NoneType' object has no attribute 'split'
    
    The above exception was the direct cause of the following exception:
    
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 54, in post
        pipeline_overwrite
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor.py", line 81, in export
        None, processor.export, pipeline, pipeline_export_format, pipeline_export_path, overwrite)
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/concurrent/futures/thread.py", line 57, in run
      File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/pipeline/processor_kfp.py", line 153, in export
        format(pipeline_name, absolute_pipeline_export_path), str(ex)) from ex
    RuntimeError: ('Error compiling pipeline demo-heterogeneous for export at /Users/lresende/opensource/jupyter-notebooks/elyra-pipelines/demo-heterogeneous.yaml', "'NoneType' object has no attribute 'split'")
```

We should fix this in the front end, and tolerate this on the backend, and possibly add tests.

akchinSTC:
Fixes #1225

 
           in the file; or

           it.


akchinSTC:
Please don't merge until ive pushed the backend tests for this

kevin-bates:
Hmm.  Since it looks like `None` instances are in the arrays sent from the client - all of which reside on the `Operation` instance, I wonder if it might be better to have a `scrub_list()` method that is used for the various arrays residing on `Operation`.  It might look something like:

```python
@staticmethod
def scrub_list(dirty: List) -> List
    if dirty is None:
        return []
    return [clean for clean in dirty if clean]
```
then you'd have something like...
```python
    def __init__(...)
        ...
        self._env_vars = Operation.scrub_list(env_vars)
        self._inputs = Operation.scrub_list(inputs)
        self._outputs = Operation.scrub_list(outputs)
```
This way we nip all the lists in the bud.

akchinSTC:
> ```python
> return [clean for clean in dirty if clean]
> ```

this was the exact list comprehension that I was about to suggest as well. Probably better nipping it earlier. Will make the change and edit local processor tests with the changes

lresende:
I moved the scrub logic to the parser which is the one that does all the work to produce the `value objects` which should be mostly just data objects.

lresende (approved):

lresende (approved):

elyra/pipeline/parser.py
```diff
@@ -198,3 +206,14 @@ def _get_parent_operation_links(node: Dict, embedded_node_id: Optional[str] = No
...
+    @staticmethod
+    def _scrub_list(dirty: Optional[List]) -> List:
+        """
+        Clean an existing list by filtering out None and empty string values
+        :param dirty: a List of values
+        :return: a clean list without None or empty string values
+        """
+        if not dirty:
+            return []
+        return [clean for clean in dirty if clean]
```
kevin-bates, line 219: The comments reminded me that this will remove `None` and _falsey_ values (e.g., empty strings).  Do we want to remove empty strings?  Since this method was meant for generic lists, it will also remove '0' and 'False' values from integer and boolean lists, which we definitely do not want to do.

Since empty string values weren't problematic before this change, this is a departure from what worked yesterday (in cases where empty strings may have been included). 
```suggestion
        return [clean for clean in dirty if clean is not None]
```
Or we should change the signature (and method name) to indicate it only deals with string lists.  If so, the typing would be: `dirty: Optional[List[Optional[str]]]) -> List[str]`

elyra/pipeline/pipeline.py
```diff
@@ -61,11 +61,11 @@ def __init__(self, id, type, classifier, filename, runtime_image, dependencies=N
...
-        self._dependencies = dependencies or []
+        self._dependencies = Operation.scrub_list(dependencies)
         self._include_subdirectories = include_subdirectories
-        self._env_vars = env_vars or []
-        self._inputs = inputs or []
-        self._outputs = outputs or []
+        self._env_vars = Operation.scrub_list(env_vars)
+        self._inputs = Operation.scrub_list(inputs)
+        self._outputs = Operation.scrub_list(outputs)
         self._parent_operations = parent_operations or []
```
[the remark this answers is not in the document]
kevin-bates: Since `parent_operations` are collected differently and we already guarantee that no operations result in the empty list, we can remove the `or []` clause here.

elyra/pipeline/parser.py
```diff
@@ -185,8 +192,9 @@ def _get_input_node_ids(node_input: Dict) -> List[str]:
 
     @staticmethod
     def _get_parent_operation_links(node: Dict, embedded_node_id: Optional[str] = None) -> List[str]:
```
kevin-bates, line 194: The result type hint here isn't correct since it looks like this returns a list of lists of strings: `List[List[str]]`

kevin-bates (commented):
Thanks Alan and Luciano - had a couple of comments.

elyra/pipeline/pipeline.py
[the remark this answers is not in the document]
lresende: Going back to the assumption that these are optional parameters, I have restored the same implementation that we have in the master branch today,  where if not passed, a list parameter is initialized as an empty list

elyra/pipeline/parser.py
lresende: I think the string signature seems better...

elyra/pipeline/parser.py
lresende: Good catch.... What a fine eye...

lresende:
I will merge this to enable an easier release preparation/pre-tests. If there are additional feedback, let's address them on new prs. Thanks.

PR #1235, opened 2021-01-25
Handle json syntax error when opening a pipeline

### LINKED ISSUE #1226: Trying to load invalid pipeline can cause data loss
**Describe the issue**
Create a pipeline and add a node with multiple environment variables
Delete one of the environment variables
Then, right-click the pipeline file, and open it with the editor (not the pipeline editor)
Remove the extra env var that has the value `null` but doesn't remove the coma from the previous env var. 

Now save and try to open on the pipeline editor

No message is displayed for the user, and an empty editor is shown

Two main things:
- The user should be notified that something went wrong when loading the pipeline
- We should not open an empty editor but cancel/close the open to avoid the user being able to save an empty pipeline and lose all previous values from the file.

karlaspuldaro:
Fixes #1226 

When the pipeline JSON file is broken, when trying to open the file on a pipeline editor, the user now sees the dialog:
image
The pipeline tab is closed after pressing the OK button  (avoiding the previous behaviour of displaying a blank pipeline and risking user to save that state and losing all data).

           in the file; or

           it.


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -920,8 +920,19 @@ export class PipelineEditor extends React.Component<
...
   async handleOpenPipeline(): Promise<void> {
     this.widgetContext.ready.then(async () => {
-      let pipelineJson: any = this.widgetContext.model.toJSON();
-      if (pipelineJson == null) {
+      let pipelineJson: any = null;
+
+      try {
+        pipelineJson = this.widgetContext.model.toJSON();
+      } catch (error) {
+        if (error instanceof SyntaxError) {
```
kevin-bates, line 928: Is this a wide-enough net?  Could any other error could get us into the same data loss situation or is the fact that such errors would now be ignored prevent that from happening?

lresende, line 928: It's ok if we want to have specific handling for SyntaxError, but we should not restrict to only SyntaxError? If there is an error that is not recoverable, don't open the pipeline and let the user know?

packages/ui-components/src/RequestErrors.tsx
```diff
@@ -94,4 +94,23 @@ export class RequestErrors {
...
+
+  /**
+   * Displays an error dialog containing a syntax error.
+   *
+   * @param fileType - the type of file that was being accessed when
+   * the error occurred
+   *
+   * @returns A promise that resolves with whether the dialog was accepted.
+   */
+  static syntaxError(
```
lresende, line 106: this is not a request error... maybe just make this a utility function on PipelineEditorWidget? or make a more generic error class and make this part of a new class like RuntimeError or something like that...

packages/pipeline-editor/src/PipelineEditorWidget.tsx
karlaspuldaro: That's a good point, I'll remove the condition

karlaspuldaro: I agree. 

> don't open the pipeline and let the user know

In that case we open the pipeline editor only to immediately display the dialog to let the user know what's wrong, then clicking OK will close the tab. Is this a valid behaviour?

ajbozarth (approved):

karlaspuldaro:
The dialog message:
image

lresende (approved):

PR #1239, opened 2021-01-26
Add support for running pipelines on Argo and Tekton

### LINKED ISSUE #1222: Compilation of Elyra-Pipelines to Tekton based Kubeflow fails
**Describe the issue**
The elyra pipeline editor can't be used to compile and push pipelines to tekton based kubeflow pipeline services

Please see discussion on [Stackoverflow](https://stackoverflow.com/questions/65832799/compilation-of-elyra-pipelines-to-tekton-based-kubeflow-fails)

lresende:
Red Hat ODH release 1.0 will be using Kubeflow Pipelines
  with Tekton support, so we are enabling Elyra to compile
  pipelines for both Argo (default) or Tekton based on
  runtime configuration.

Fixes #1222 
 
           in the file; or

           it.


elyra/metadata/schemas/kfp.json
```diff
@@ -67,6 +68,15 @@
             "category": "Kubeflow Pipelines"
           }
         },
+        "engine": {
+          "title": "Kubeflow Pipelines engine",
+          "description": "The Kubeflow Pipelines engine in use (Argo or Tekton)",
+          "type": "string",
```
kevin-bates, line 74: We should leverage schema constructs for this.  With the following, if `engine` does not specify a value, a value of `"Argo"` will be used.  It will also automatically enforce only `"Argo"` and `"Tekton"` can be set as values.
```suggestion
          "type": "string",
          "enum": ["Argo", "Tekton"],
          "default": "Argo",
```

lresende:
KFP servers based on Argo and Tekton are now accepting and running the pipelines, but Tekton deployment is having issues displaying the results due to some metrics related issues (FYI @ptitzler)

elyra/pipeline/processor_kfp.py
```diff
@@ -68,10 +69,15 @@ def process(self, pipeline):
 
             self.log.debug("Creating temp directory %s", temp_dir)
 
+            engine = runtime_configuration.metadata.get('engine')
+            print(f'>>> {engine}')
             # Compile the new pipeline
             try:
                 pipeline_function = lambda: self._cc_pipeline(pipeline, pipeline_name)  # nopep8 E731
-                kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
+                if ('argo' == engine or 'Argo' == engine):
```
kevin-bates, line 77: JSON enum values are case-sensitive, so there should never be a **validated** instance of runtime metadata that contains the value `"argo"`.  Are you seeing otherwise?

lresende:
On Argo:
image

On Tekton:
image

elyra/pipeline/processor_kfp.py
lresende: No :)

elyra/pipeline/processor_kfp.py
```diff
@@ -68,10 +69,14 @@ def process(self, pipeline):
...
             # Compile the new pipeline
             try:
                 pipeline_function = lambda: self._cc_pipeline(pipeline, pipeline_name)  # nopep8 E731
-                kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
+                if ('Argo' == engine):
+                    kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
+                else:
+                    kfp_tekton.compiler.TektonCompiler().compile(pipeline_function, pipeline_path)
             except Exception as ex:
                 raise RuntimeError('Error compiling pipeline {} at {}'.
```
kevin-bates, line 81: Since the `else` statement is non-specific, I would suggest updating the error message to include the `engine` value, just in case some bogus value (like 'argo' :smile: ) sneaks in.
```suggestion
                raise RuntimeError('Error compiling pipeline {} for engine '{}' at {}'.
```
Might as well convert this to an f-string if you decide to update it.

elyra/pipeline/processor_kfp.py
```diff
@@ -68,10 +69,14 @@ def process(self, pipeline):
 
             self.log.debug("Creating temp directory %s", temp_dir)
 
+            engine = runtime_configuration.metadata.get('engine')
             # Compile the new pipeline
             try:
                 pipeline_function = lambda: self._cc_pipeline(pipeline, pipeline_name)  # nopep8 E731
-                kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
+                if ('Argo' == engine):
```
ptitzler: In a way this is perhaps not consistent with schema's default, which is `Argo`. If, for whatever reason, the runtime configuration does contain an invalid value `XYZ`, the  logic would run the Tekton compiler.

elyra/pipeline/processor_kfp.py
```diff
@@ -68,10 +69,14 @@ def process(self, pipeline):
 
             self.log.debug("Creating temp directory %s", temp_dir)
 
+            engine = runtime_configuration.metadata.get('engine')
             # Compile the new pipeline
             try:
                 pipeline_function = lambda: self._cc_pipeline(pipeline, pipeline_name)  # nopep8 E731
-                kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
+                if 'Argo' == engine:
+                    kfp.compiler.Compiler().compile(pipeline_function, pipeline_path)
```
ptitzler, line 77: Would be great if we could log which compiler was used. This way there's no ambiguity. (Same comment below)

elyra/pipeline/processor_kfp.py
ptitzler: This is also inconsistent with the logic in https://github.com/elyra-ai/elyra/blob/bde3b4f3d284407dc0f42a3c80f8983e1066b749/elyra/pipeline/processor_kfp.py#L157, where we test for `Tekton`.

ptitzler:
It appears that export does not yet take into account the engine type if Python DSL is generated. The Jinja template https://github.com/elyra-ai/elyra/blob/master/elyra/templates/kfp_template.jinja2#L43 always uses the Argo compiler (unless there is some other PR that contains additional changes.

elyra/pipeline/processor_kfp.py
lresende: other values should not really come in as the metadata is validated and invalid ones will not be shown in the user interface....

lresende:
The export has already been updated by @akchinSTC

ptitzler:
Need a separate doc PR to reflect the changes in https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html

docs/source/user_guide/runtime-conf.md
```diff
@@ -178,10 +183,10 @@ Password used to access your KubeFlow Pipelines API endpoint. This setting is re
 
 Example: `mypassword`
 
-##### user_namespace
-The namespace used to run your pipeline in Kubeflow Pipelines. This setting is required if the Kubeflow Pipelines deployment is multi-user, auth enabled. SEE NOTE.
+##### engine
+The engine being used by Kubeflow Pipelines: Argo or Tekton (default is Argo).
```
ptitzler, line 187:
```suggestion
The engine being used by Kubeflow Pipelines: `Argo` or `Tekton` (default is `Argo`).
```

akchinSTC (approved):
I am good with this up to the doc changes

ptitzler:
The doc update looks good. I also ran a couple tests without issues.

Issue #822, opened 2020-08-05
Make Elyra pipelines more flexible to enable better support for sharing via GitHub or other packaging


ajbozarth:
Umbrella task for handling pipeline editor enhancements focused around enabling "moving" pipelines

- [x] Validate pipeline which includes validation of file location #815
   - [x] Consider developing lifecycle, some types of validation only applies to when executing versus opening/saving
- [x] Enhance the pipeline editor to enable updates to the location of the notebook (including changing the notebook) #732
- [x] Support relative location of the notebook (relative to pipeline file) #823 
   - [x] When opening
   - [x] When drag-and-drop
   - [x] When submitting the pipeline
- [x] Pipeline migration #825
- [x] Documentation
   - [x] Best practices (Elyra 1.0) #820
   - [x] Validation #824

This umbrella should address the following issues:
#608 #659 #815

ajbozarth:
Note: all issues under this umbrella now have open or merged PRs

PR #1246, opened 2021-01-27
Add elyra-pipeline cli tool

bourdakos1:
The purpose of this CLI is to be able to run pipelines locally and to submit them to kfp or other runtimes from outside the context of jupyter lab.

The proposed user experience for the initial version is to introduce two commands `run` and `submit`. 

The `run` command would run the pipeline locally, in the context of CLI and there will be no support for `runtime` and `runtime-config` parameters.

Example of running a pipeline locally:
```
elyra-pipeline run this/is/an/example.pipeline
```

The `submit` would submit the pipeline to a supported remote runtime such as KFP or Airflow. In this context, the `runtime-config` parameters will point to the necessary configuration to submit the pipeline to the remote runtime. 

Example of submitting a pipeline to be run in the context of the Jupyter server:
```
elyra-pipeline submit this/is/an/example.pipeline --runtime-config kfp-shared-tekton
```

~~Eventually, I would like to see the `elyra-metadata` CLI pulled into the `elyra` CLI for simplicity and for the default stora~ge location to be somewhere like `~/.elyra` so it's not so tied to jupyter. Additionally, I would like to see a feature for setting a default config to alleviate the need for passing config information to the `submit` command.~~

           in the file; or

           it.


kevin-bates:
Since this is a POC, I'm not sure it's open for comments/questions, but I have a few of questions...

1. I would have thought the main logic of this would be nearly identical to the [handler's logic](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/handlers.py#L87-L90).  I understand the preparation step, which I assume applies the necessary expansions of the file paths (from relative to absolute) and any other items the front-end submits the pipeline with, but it seems preferable, and more maintainable, to have the core logic the same here.  This POC is replicating fair amounts of the parser.  We should fix the existing object structure rather than change it just for CLI operations.
2. The CLI click update appears to allow for the entire set of runtime metadata fields to be individually updated rather than take the name of a runtime metadata instance.  Do we want to go there?
3. These comments suggest this will be replacing the metadata-driven tool relative to runtime metadata, is that correct?
```
# # elyra add runtime
# # elyra get/list runtimes
# # elyra remove/delete runtime <runtime>
# # elyra replace runtime <runtime>
```

lresende:
I agree with what Kevin is saying above, we really need to have a better understanding of what is the goal here:
- Is this something similar to `elyra-metadata` where we do `elyra-pipeline run foo.pipeline --runtime-config=kfp` otherwise, what are the requirements?

bourdakos1:
@kevin-bates oh yea I see what you are saying, I pulled out parse to remove the check for runtime, but I should just add the runtime to the pipeline 🤦‍♂️ don't know what I was thinking haha... 

@lresende I didn't realize `elyra-metadata` existed I thought all the config stuff went through the ui. That could probably work then, I'll have to take a closer look.

I definitely open to suggestions, but my thinking was that `elyra add runtime` would create a runtime config, which looks like what `elyra-metadata install runtime` does. `list` would list, `delete` would delete, `replace` would let you update a config. 

I think the main divergence is `elyra use runtime <runtime-name>` which would set the current runtime, kinda like kubectx I guess? That way you wouldn't need to pass runtime info with `elyra submit foo.pipeline` 

I was also thinking that we should have 2 distinct commands, one for running locally and the other for actually submitting a pipeline

bourdakos1:
oh also, it seems like the runtime config stuff is all stored in a very jupyter specific place, don't know if it makes sense to move it to somewhere like `~/.elyra`?

kevin-bates:
> I definitely open to suggestions, but my thinking was that elyra add runtime would create a runtime config, which looks like what elyra-metadata install runtime does. list would list, delete would delete, replace would let you update a config.

Yeah, `elyra-metadata` supports 'install', 'list', and 'remove' options that map to 'add', 'list' and 'delete'.  Within 'install' is a sub-option `--replace` that would then satisfy 'update'.  It is completely metadata-driven, working off the specific schema - so every aspect of our metadata (code-snippets, runtime-images, runtimes and test-schemas) use the same tooling.

> I think the main divergence is elyra use runtime <runtime-name> which would set the current runtime, kinda like kubectx I guess? That way you wouldn't need to pass runtime info with elyra submit foo.pipeline

This is an interesting idea.

> I was also thinking that we should have 2 distinct commands, one for running locally and the other for actually submitting a pipeline

Hmm. I would have thought an option, mutually exclusive from `'--runtime xxx'`, like `'--local'` would be all that is necessary, since those are processed identically relative to what the handlers do.  You would need to convey the ["runtime"](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor.py#L64) (which is either `"local"` or the schema name of the referenced runtime instance) in the pipeline definition as part of the preparation step based on the CLI options and the post-process messaging may need to be conditional on whether `--local` was used or not.

> oh also, it seems like the runtime config stuff is all stored in a very jupyter specific place, don't know if it makes sense to move it to somewhere like ~/.elyra?

Yeah, we leverage Jupyter's data directory hierarchy for metadata persistence.  I personally wouldn't mind us taking complete ownership and scrapping the Jupyter hierarchy - although the downside is that hierarchy is well-known to Jupyter users.  Run `jupyter --paths` to get an idea of what's all involved.  It's a bit overwhelming (IMHO).

bourdakos1:
My thinking behind having two separate commands was more from a user perspective than an implementation perspective. I had a CLI that would submit training jobs to WML that could also run them locally and I went with a `--local` flag, it worked, but there was always a bit of a mental load for double checking if I had the `--local` flag sent or not. I have since used other CLIs that have a completely separate command for running things locally and it just feels really nice ❤️

but that's just me, I'm open to using `--local` or something similar

kevin-bates:
No, you're right, my focus was on the core implementation.  I haven't confirmed behavior but given that [`console-scripts` appear to be mapped to a class](https://github.com/elyra-ai/elyra/blob/master/setup.py#L96), it seems we could easily have two CLI commands associated with the same implementation - just two different subclasses.

Since runtimes are really _submission_ actions, while local processing is _immediate execution_, perhaps having different actions/verbs could be the distinction.  For example, if `elyra submit [runtime] foo.pipeline` was the example for submission to a runtime, then we could have `elyra execute foo.pipeline` (where [runtime] is not permitted) would perform the local execution. 

At any rate, I believe there are lots of options and I don't want to narrow the scope too much here.  Thanks for working on this!

bourdakos1:
I updated the PR to remove the custom parse function in favor of just updating the pipeline with the runtime info. It seems to be working pretty well :) Out of curiosity, is there a reason why we do the path expansion on the frontend and not here?

I think we need to have a discussion about the metadata/runtime stuff before I can go any further with this

elyra/cli.py
```diff
@@ -154,42 +95,23 @@ def submit(pipeline_path, name):
...
 
-    click.echo(Fore.CYAN + "❯ Runtime" + Style.RESET_ALL)
-    click.echo("  type: " + pipeline.runtime)
-    click.echo("  config: " + pipeline.runtime_config)
-    click.echo()
+    print_info("Info", [("name", pipeline.name)])
+    print_info("Runtime", [("type", pipeline.runtime), ("config", pipeline.runtime_config)])
     
     with yaspin(text="Submitting Pipeline..."):
         os.environ["ELYRA_METADATA_PATH"] = os.path.join(os.path.expanduser("~"), ".elyra")
```
kevin-bates, line 107: I don't think this should be here.  If users which to alter where metadata instances are stored, this should be set in their environment already and not something we impose in the tool.

kevin-bates:
> Out of curiosity, is there a reason why we do the path expansion on the frontend and not here?

I believe the paths persisted in the pipeline file are relative to the server's notebook (root) dir to facilitate sharing.  This way the pipeline file (and its related files, etc.) can be copied (or downloaded from github) into another user's environment.

> I think we need to have a discussion about the metadata/runtime stuff before I can go any further with this

Sounds good.  What is the issue?

elyra/cli.py
bourdakos1: No yea I agree, I just have my config sitting there right now (I also have the runtime config stuff hardcoded here too)

bourdakos1:
Ohhh I was under the assumption that the paths were relative to the pipeline... Does this mean that if the pipeline is stored at `/xxx/xxx/notebook-root/some/deep/path/foo.pipeline` and a notebook was sitting next to it, it's path in the pipeline would be `some/deep/path/foo.ipynb`?

elyra/cli.py
```diff
@@ -0,0 +1,135 @@
...
+
+    pipeline_definition = prepare_pipeline(pipeline_path, name=name, runtime='kfp', runtime_config='fun')
+    pipeline = PipelineParser().parse(pipeline_definition)
+
+    print_info("Info", [("name", pipeline.name)])
+    print_info("Runtime", [("type", pipeline.runtime), ("config", pipeline.runtime_config)])
+
+    with yaspin(text="Submitting Pipeline..."):
+        os.environ["ELYRA_METADATA_PATH"] = os.path.join(os.path.expanduser("~"), ".elyra")
+        msg = asyncio.get_event_loop().run_until_complete(submit_pipeline(pipeline))
```
bourdakos1, line 108: @kevin-bates I'm very new to python, is this how to properly run an async function? how do I handle errors? `try`/`except` doesnt seem to work right

kevin-bates:
> Ohhh I was under the assumption that the paths were relative to the pipeline

Actually, you're right.  The paths referenced in pipelines are relative to the pipeline file's location (to facilitate sharing) and then the server adjusts all remaining relative paths by prefixing the notebook (root) dir during the pipelines submission/run.  (Sorry, I forgot about that.). This was part of https://github.com/elyra-ai/elyra/issues/822.

elyra/cli.py
bourdakos1: nvm I was listening for the wrong error

kevin-bates: Hmm - that looks right and I have a small sample I updated to model this who's try/except is properly behaved.  If you wrap the `run_until_complete` in a try/except, what happens on exceptions?

kevin-bates: Sorry - I hadn't seen your response - glad things are moving forward.

lresende:
Could we step back and define (and update the description) with:
- What is the user experience (e.g. how the user runs CLI)
- Also, what are the assumptions/requirements regarding the environment

Regarding local vs remote, today that is defined by the "runtime-config" and I would expect it to continue that way, but local really means in the context of the jupyter server, not in the context of the CLI which to me is only a middle man between the user and the REST APIs provided by elyra and hosted by the jupyter server. 

Once we are on the same page for these items, the CLI should be a small and simple rapper very similar to `elyra-metadata`

bourdakos1:
I pushed a few changes removing the hardcoded runtime stuff and added help/docs info. Everything is working properly with the current `elyra-metadata` CLI so I don't think we need any changes there. I think this is ready to be more thoroughly reviewed

lresende:
> Everything is working properly with the current `elyra-metadata` CLI so I don't think we need any changes there. I think this is ready to be more thoroughly reviewed

I want to say it should be implemented following the same pattern:
https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/metadata_app.py

And should be collocated in the pipeline package/folder

And registered in:
https://github.com/elyra-ai/elyra/blob/master/setup.py

bourdakos1:
@lresende I updated the description

> I want to say it should be implemented following the same pattern:
> https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/metadata_app.py

Are you saying we shouldn't use `click` for handling the CLI stuff?

> And should be collocated in the pipeline package/folder

👍

> And registered in:
> https://github.com/elyra-ai/elyra/blob/master/setup.py

What do you mean by this?

lresende:
@bourdakos1 I have both `run` and `submit` working and following the same implementation of the `elyra-metadata` cli tool we have. The changes are a PR to this pr, please review and once you merge we can remove the current PR from [WIP] and merge.

kevin-bates:
@lresende - please note that every test (except for the single 'bad argument' test) for the metadata app tests is failing.  The bad argument test just happens to pass because it encounters the expected error (that all the other tests are encountering).  I would suggest we get those fixed before merging into this PR.

bourdakos1:
@lresende a few questions/issues so far
- `elyra pipeline` doesn't work, will that be coming in a later PR?
- when running the CLI it always dumps this out:
```
[I 2021-02-17 11:37:13,278.278] Generating grammar tables from /opt/miniconda3/envs/luciano/lib/python3.7/site-packages/blib2to3/Grammar.txt
[I 2021-02-17 11:37:13,301.301] Writing grammar tables to /Users/niko/Library/Caches/black/20.8b1/Grammar3.7.9.final.0.pickle
[I 2021-02-17 11:37:13,302.302] Writing failed: [Errno 2] No such file or directory: '/Users/niko/Library/Caches/black/20.8b1/tmpi5k7u2gx'
[I 2021-02-17 11:37:13,302.302] Generating grammar tables from /opt/miniconda3/envs/luciano/lib/python3.7/site-packages/blib2to3/PatternGrammar.txt
[I 2021-02-17 11:37:13,303.303] Writing grammar tables to /Users/niko/Library/Caches/black/20.8b1/PatternGrammar3.7.9.final.0.pickle
[I 2021-02-17 11:37:13,304.304] Writing failed: [Errno 2] No such file or directory: '/Users/niko/Library/Caches/black/20.8b1/tmpefm20g58'
```
- there doesn't seem to be help messages set up for `run`
```
AttributeError: 'super' object has no attribute 'print_help'
```
- the help message for `submit` doesn't provide any info for what options are needed
- it seems that you need a `=` for options, for example `--pipeline the/path/example.pipeline` doesn't work, but `--pipeline=the/path/example.pipeline` does

I haven't tried actually running or submitting anything yet

bourdakos1:
Also I think I would prefer `elyra pipeline run the/path/example.pipeline` vs `elyra pipeline run --pipeline=the/path/example.pipeline`

but I'm fine if you want to allow both

ptitzler:
How does one invoke the cli after install? 

```
$ make install
 ...
$ elyra
zsh: command not found: elyra

```

lresende:
This brings support for `elyra-pipeline`, the wrapper to enable `elyra pipeline/metadata` will come in the future, as needs a bunch of refactoring and does not block functionality.

ptitzler (changes requested):
```
$ elyra-pipeline run -h

  File "/opt/anaconda3/envs/elyra_dev/bin/elyra-pipeline", line 8, in <module>
    sys.exit(PipelineApp.main())
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 175, in main
    elyra_pipeline.start()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 186, in start
    return subinstance.start()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 98, in start
    self.process_cli_options(self.options)  # process options
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/util/app_utils.py", line 237, in process_cli_options
    self.log_and_exit(display_help=True)
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/util/app_utils.py", line 162, in log_and_exit
    self.print_help()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 117, in print_help
    super(AppBase, self).print_help()
AttributeError: 'super' object has no attribute 'print_help'
```

ptitzler:
Shouldn't dump a stacktrace
```
$ elyra-pipeline run
'--pipeline' is a required parameter.

  File "/opt/anaconda3/envs/elyra_dev/bin/elyra-pipeline", line 8, in <module>
    sys.exit(PipelineApp.main())
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 175, in main
    elyra_pipeline.start()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 186, in start
    return subinstance.start()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 98, in start
    self.process_cli_options(self.options)  # process options
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/util/app_utils.py", line 240, in process_cli_options
    self.process_cli_option(option)
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/util/app_utils.py", line 222, in process_cli_option
    format(cli_option.cli_option), display_help=True)
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/util/app_utils.py", line 162, in log_and_exit
    self.print_help()
  File "/opt/anaconda3/envs/elyra_dev/lib/python3.7/site-packages/elyra/pipeline/pipeline_app.py", line 117, in print_help
    super(AppBase, self).print_help()
AttributeError: 'super' object has no attribute 'print_help'
```

ptitzler:
No runtimes defined in my env.

```
$ elyra-metadata list runtimes
No metadata instances found for runtimes
```

Not specifying a runtime.

```
$ elyra-pipeline submit --pipeline=analyze_NOAA_weather_data.pipeline
```
Nothing happens for 40 seconds ...
```

Success?
```
$ echo $?
0
```

Please add error checking. Thanks.

File: elyra/cli/pipeline_app.py [hunk 3]
```diff
@@ -0,0 +1,172 @@
...
+    pipeline_dir = os.path.dirname(pipeline_abs_path)
+    pipeline_name = os.path.splitext(os.path.basename(pipeline_abs_path))[0]
+
+    with open(pipeline_abs_path) as f:
+        pipeline_definition = json.load(f)
+
+    for pipeline in pipeline_definition["pipelines"]:
+        for node in pipeline["nodes"]:
+            if node["app_data"]["filename"]:
+                abs_path = os.path.join(os.getcwd(), pipeline_dir, node["app_data"]["filename"])
```
kevin-bates, line 43: Is getcwd necessary here? Seems like pipeline_dir is already absolute and reflects working_dir already.

kevin-bates (commented):
Reviewed from my phone and lacking markdown sugar.

lresende:
@bourdakos1 could you check this out and help test

bourdakos1:
@lresende why is the `work-dir` flag necessary? Aren’t the dependencies guaranteed to be relative to the pipeline file?

setup.py
```diff
@@ -80,6 +80,7 @@
         'traitlets>=4.3.2',
         'urllib3>=1.24.2',
         'websocket-client',
+        'yaspin',
```
bourdakos1, line 83: Should we add colorama and click here as well?

elyra/cli/pipeline_app.py
```diff
@@ -0,0 +1,172 @@
...
+def _preprocess_pipeline(pipeline_path, runtime, runtime_config, work_dir):
+    pipeline_path = os.path.expanduser(pipeline_path)
+    working_dir = os.path.expanduser(work_dir)
+    pipeline_abs_path = os.path.join(working_dir, pipeline_path)
+    pipeline_dir = os.path.dirname(pipeline_abs_path)
+    pipeline_name = os.path.splitext(os.path.basename(pipeline_abs_path))[0]
+
+    with open(pipeline_abs_path) as f:
+        pipeline_definition = json.load(f)
+
```
bourdakos1, line 39: We do pipeline validation in `PipelineParser().parse(...)` but that happens after this preprocess function. If the pipeline is invalid I would assume this would crash before making it to the nicer error messages in `parse`. 

We should also check the pipeline version, I don’t think I saw logic for that in `parse`

bourdakos1:
So far testing the run command:

Invalid json (for example an blank pipeline file) dumps a stacktrace
valid json but not a pipeline file (for example a jupyter notebook) dumps a stacktrace
missing the filename field in the pipeline json dumps a stacktrace

bourdakos1:
So far testing the `submit` command:

(leaving out the same errors from `run`)
running the command without jupyter running dumps a stacktrace

bourdakos1:
We don't re-validate the pipeline file, so if you use the cli on a pipeline with a circular reference, it dumps a stacktrace for `maximum recursion depth exceeded`

bourdakos1:
Overall this feels pretty good to me. We just need to rethink how we handle pipeline validation a bit and have a better error message for when Jupyter isn't running (it still feels a bit weird that you need to have Jupyter running in order to submit a pipeline, but at least you don't need to for the local run)

Also a bit confused about `--work-dir`

lresende:
I added validation for the running server, but we still only discover local running servers. Should we also add a "server url" parameter to enable submitting to remote servers? 

For the validations, could you please create an issue with the options we should validate on the backend?

bourdakos1:
@lresende oh actually what if instead of searching for a running server, we submit the pipeline from the context of the terminal, but if you want to submit through Jupyter you can pass a flag for "server url" that could be a local server like localhost:8888 or a remote server?

bourdakos1:
opened issue #1370 to track backend validation

elyra/cli/pipeline_app.py
```diff
@@ -0,0 +1,188 @@
...
+from colorama import Fore, Style
+
+from elyra import __version__
+from elyra.pipeline import PipelineParser, PipelineProcessorManager
+from jupyter_server.serverapp import list_running_servers
+
+
+def _validate_pipeline_file(pipeline_file):
+    extension = os.path.splitext(pipeline_file)[1]
+    if extension != '.pipeline':
```
bourdakos1, line 32: Do we need to verify that the file has `.pipeline` as an extension?

lresende: Mostly to avoid other file extensions like you mentioned above (e.g. ipynb, py, etc)

kevin-bates:
> it still feels a bit weird that you need to have Jupyter running in order to submit a pipeline, but at least you don't need to for the local run

I agree and didn't realize the inconsistency introduced with local mode.  Seems like the "modes" should work the same (so local mode should go through the web API).

> oh actually what if instead of searching for a running server, we submit the pipeline from the context of the terminal, but if you want to submit through Jupyter you can pass a flag for "server url" that could be a local server like localhost:8888 or a remote server?

I was just about to type the same thing - only use the web API if `--server-url` is provided, others go direct.

Is`elyra-metadata` going to require the server be running?  (And updating a remote server's metadata would also be useful.)

bourdakos1:
I added some basic validation to the preprocessing step in the cli, I think a lot of it is a duplicate of what happens again in pipeline parser. I think we should move the preprocessing into the parser if possible. This doesn't include checks for circular references or properties validation yet

lresende:
> > it still feels a bit weird that you need to have Jupyter running in order to submit a pipeline, but at least you don't need to for the local run
> 
> I agree and didn't realize the inconsistency introduced with local mode. Seems like the "modes" should work the same (so local mode should go through the web API).
> 
> > oh actually what if instead of searching for a running server, we submit the pipeline from the context of the terminal, but if you want to submit through Jupyter you can pass a flag for "server url" that could be a local server like localhost:8888 or a remote server?
> 
> I was just about to type the same thing - only use the web API if `--server-url` is provided, otherwise go direct.
> 
> Is`elyra-metadata` going to require the server be running? (And updating a remote server's metadata would also be useful.)

I think there are cases, and there are cases. To me, if we think in our day-to-day development workflow we always have the pipelines, the notebook server, elyra, dependencies, etc installed and available locally, so being able to just 'run foo.pipeline' makes complete sense and is very doable. 

Now, consider the scenario where to run this pipeline, you need access to GPUs from a shared server (e.g. a cloud-hosted elyra deployment) or you need access to data that is not available locally and cannot be accessible locally (e.g. a protected cluster where PII data is shared), in this case, running the pipeline locally will not provide you the expected results. 

I added the notion of `run` vs `submit` to try to solve these type of issues, where at one stage or for a sub-set of users, you will always be ok to just execute the pipeline locally, but for others, for one reason or another, we would need to execute them in the context of a Notebook Server deployment (where you were testing the pipeline execution).

Now, if we use ['run' vs 'submit'] or ['run' vs 'run --server-url`] ... or if we use `run for synchronized runs (local modes in the machine or in the context of the server)` vs `submit for  async submissions (kfp/airflow)`. Then it's a different discussion, and I believe I would be ok with both. 

The other aspect is that we should be consistent across UI and CLIs, and it seems that in our UI we are only using `Run` (see examples below), so that should help us on the naming discussion as well. 

image

image

Having said all that, I also ok stepping back and focusing on the day-to-day development workflow scenario, and only provide basic support for running pipelines in the local dev environment, and expand the use cases when users start adopting these and aks for further capabilities, this might actually help us get this in quick and without controversy.

bourdakos1:
> Although I believe sometimes we introduce basic validation on the client-side to enable us to fail fast, in this case, particularly when we are calling direct backend APIs like parse and execute, then we shouldn't code this on the client-side code but share this validation on the backend APIs that is then reused to validate all pipelines, independent from who is integrating/submitting it.
> 
> Are these validations cases just not available on the server-side APIs?
> 
> Or is there some limitation or scenario I might have overseen?

@lresende I think the issue is that we need to do some validation of the pipeline before we can preprocess (expand relative paths to absolute, add runtime info etc...). I think it would be best if the server could handle all of this, if possible.

bourdakos1:
@lresende A thought related to this run vs submit discussion... Would it make sense to have 2 buttons in the UI, to match the CLI?

- A "run" button that runs using the "local" runtime, since most of the options in the dialog seem to be redundant for local (there's no config needed to be chosen and I don't think the run name is actually used?). You could simply click "run" to run without having to configure anything
- A "submit" button which would prompt for a run name, runtime platform and runtime config

I would assume most people would only go back and forth between local runs and a specific runtime config for a given project. With 2 buttons they wouldn't have to reconfigure every time they go between local vs an actual submission.

@ptitzler I feel like you would have a good opinion on this, thoughts?

bourdakos1:
@lresende should we set up a call to discuss this further?

lresende:
After a quick discussion, I have updated the description with the direction we agreed to go, basically supporting run locally and submitting to kfp/airflow.

elyra/cli/pipeline_app.py
```diff
@@ -128,6 +117,21 @@ def _preprocess_pipeline(pipeline_path, runtime, runtime_config, work_dir):
     return pipeline_definition
 
 
+def _execute_pipeline(pipeline_definition):
+    try:
+        with yaspin(text="Processing pipeline ..."):
+            # parse pipeline
+            pipeline_object = PipelineParser().parse(pipeline_definition)
+            # process pipeline
+            asyncio.get_event_loop().run_until_complete(PipelineProcessorManager.instance().process(pipeline_object))
```
bourdakos1, line 126: I haven't tested this yet, but `local` mode has a progress bar when executing nodes, so we might get weird rending issues if we use a spinner on top of it.

elyra/cli/pipeline_app.py
```diff
@@ -162,103 +166,52 @@ def pipeline():
...
-              required=False,
-              default=None,
-              help='Jupyter Notebook server url including access token')
+@click.argument('pipeline_path')
 @click.option('--runtime',
-              required=False,
-              type=click.Choice(['local', 'kfp', 'airflow'], case_sensitive=False),
-              default='local',
+              required=True,
+              type=click.Choice(['kfp', 'airflow'], case_sensitive=False),
```
bourdakos1, line 172: Should we restrict choices here? I think the processor raises an error if the runtime type isn't supported. If we rely on that instead, we won't have to remember to keep updating this list when we add runtimes. Although, we were talking about detecting the runtime type from the config at some point, so this might not really matter for long

bourdakos1 (commented):
This is looking good, I'll do some testing in a bit

elyra/cli/pipeline_app.py
bourdakos1: tested, `Processing pipeline ...` and the progress bar fight over each other. I think we should only render the spinner for `submit` and not for `run`. I also think "Submitting pipeline..." would make more sense

docs/source/user_guide/pipelines.md
```diff
@@ -87,6 +87,38 @@ To run a pipeline on Kubeflow Pipelines:
 
 Refer to the [Kubeflow Pipelines tutorial](/getting_started/tutorials.md) for details.
 
+#### Running a pipeline using the command line
+
```
bourdakos1, line 91: just a reminder to update docs

elyra/cli/pipeline_app.py
lresende: I thought about this, but you will see sort of a "hang" for a long time if you are executing non-papermill things such as long python or R script.

elyra/cli/pipeline_app.py
```diff
@@ -0,0 +1,222 @@
...
+
+# TODO: Is there a place to get this version number already?
+CURRENT_PIPELINE_VERSION = 3
+
+
+def _validate_pipeline_file(pipeline_file):
+    extension = os.path.splitext(pipeline_file)[1]
+    if extension != '.pipeline':
+        click.echo('Pipeline file should be a [.pipeline] file.\n')
+        raise click.Abort()
```
bourdakos1, line 37: I think we should raise a `ClickException` instead, that will print a the error message and exit. The `Aborted!` message feels a bit harsh

elyra/cli/pipeline_app.py
```diff
@@ -0,0 +1,208 @@
...
+def _execute_pipeline(pipeline_definition):
+    try:
+        with yaspin(text="Processing pipeline ..."):
+            # parse pipeline
+            pipeline_object = PipelineParser().parse(pipeline_definition)
+            # process pipeline
+            asyncio.get_event_loop().run_until_complete(PipelineProcessorManager.instance().process(pipeline_object))
+    except ValueError as ve:
+        click.echo(f'Error parsing pipeline: \n {ve}')
+        raise click.ClickException(f'Error parsing pipeline: \n {ve}')
```
bourdakos1, line 116: double message

elyra/cli/pipeline_app.py
bourdakos1: ah, I see, should we add some sort of spinner/progressbar internally for non-papermill executions?

bourdakos1: maybe it's better to just drop the spinner and use a simple `click.echo()`, that way it won't obstruct the papermill progress bars, but we still get the indication that it's submitting/running something

elyra/cli/pipeline_app.py
```diff
@@ -36,41 +36,6 @@ def _validate_pipeline_file(pipeline_file):
...
-
-    if found_primary_pipeline["app_data"]["version"] != CURRENT_PIPELINE_VERSION:
-        raise click.ClickException("Pipeline file is invalid: \n Primary pipeline version is incompatible")
-
-    if 'nodes' not in found_primary_pipeline:
-        raise click.ClickException("Pipeline file is invalid: \n Primary pipeline is missing field 'nodes'")
-
-    if not isinstance(found_primary_pipeline["nodes"], list):
-        raise click.ClickException("Pipeline file is invalid: \n Primary pipeline field 'nodes' should be a list")
-
```
bourdakos1: These checks need to happen before `_preprocess_pipeline`

elyra/cli/pipeline_app.py
```diff
@@ -0,0 +1,165 @@
...
+    such as Kubeflow Pipelines or Apache Airflow.
+
+    Find more information at: https://elyra.readthedocs.io/en/latest/
+    """
+    pass
+
+
+@click.command()
+@click.argument('pipeline_path')
+@click.option('--runtime',
```
kevin-bates, line 114: We talked about inferring runtime from `runtime-config`.  Since this would essentially be an API change if we did this _after_ initially supporting `--runtime`, and since its trivial code to infer the runtime from the config, and since you need to validate that the runtime-config exists anyway, I think we should remove `--runtime` in the first draft.

bourdakos1: Sounds good to me

elyra/cli/pipeline_app.py
lresende: I have added a minimal set of validation back and also surround it with some try/except, but left the rest on the parser so it is not duplicated.

lresende:
One thing that I am not sure how to handle is the warnings or outputs from the execution.
An example is:
```
.../Users/lresende/opt/anaconda/envs/dev/lib/python3.8/site-packages/kfp/dsl/_container_op.py:1032: FutureWarning: Please create reusable components instead of constructing ContainerOp instances directly. Reusable components are shareable, portable and have compatibility and support guarantees. Please see the documentation: https://www.kubeflow.org/docs/pipelines/sdk/component-development/#writing-your-component-definition-file The components can be created manually (or, in case of python, using kfp.components.create_component_from_func or func_to_container_op) and then loaded using kfp.components.load_component_from_file, load_component_from_uri or load_component_from_text: https://kubeflow-pipelines.readthedocs.io/en/stable/source/kfp.components.html#kfp.components.load_component_from_file
  warnings.warn(
```

Or, a script that print results:
```
⠴ Processing pipeline ...[1] "The factorial of 7 is 5040"
```

Thoughts? Is this something we want to try to actually hide from the user?

elyra/cli/pipeline_app.py
lresende: click.echo() will not help on long-running scripts.

bourdakos1:
@lresende I had logic in here at some point that hid the warnings

bourdakos1:
```python
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
```

elyra/cli/pipeline_app.py
bourdakos1: @lresende I agree that a spinner would be preferable, but as you also mentioned in another comment the scripts can print things and it gobbles the progress bar that is much more useful than a simple spinner. I think a static echo is an okay compromise until we have a better way to handle spinners and outputs together

lresende: @bourdakos1 Please checkout the covid-notebook repo and try running/submitting `covid-notebooks/pipelines/us_data.pipeline` and see if the static waiting is acceptable. The spinner at least gives some sense of something going on while the user wait a long task.

lresende:
Current submit output:

image

elyra/cli/pipeline_app.py
bourdakos1: yea, it's not ideal, but I think I'm okay with this for now. When using `run` it shows nice progress bars for each notebook being run (I know it won't show anything for regular scripts). When using `submit` it doesn't take incredibly long to complete, an animation is nice, but `...` also lets me know it might take a minute. We could also add more details saying it might take a moment to finish.

Eventually we can use the blanket spinner for `submit` and rely on the papermill logs for executing notebooks and add our own logging for scripts, unless that is something you want to add to this PR

bourdakos1:
@lresende currently, when submitting a pipeline in the UI we get the following dialog (with no loading animations):
```
Packaging and submitting pipeline ...
This may take some time
```
I'll make a video of the difference between the 2 options

bourdakos1:
With spinner:
https://user-images.githubusercontent.com/4212769/112345109-21cf8600-8c9b-11eb-9fe3-2ff80f2410d2.mov

Without spinner:
https://user-images.githubusercontent.com/4212769/112345135-26943a00-8c9b-11eb-86bb-f49a21d9939e.mov

kevin-bates:
Of the two, I prefer 'with spinner' as it's less ambiguous as to what happened. I really don't like how the progress bar gets to 27%, then everything disappears.  It would be nice to see a "success" message of some sort (in either case).  (I'm assuming the user will be presented with some indication of failures.)

bourdakos1:
Oh yea the progress bar doesn’t actually disappear, I just stopped recording haha

ptitzler:
Not sure I understand how to properly interpret the progress bar. In the last screen recording, what does 1/44 refer to? Is that the cell number within a notebook that's being processed? 

If it is ...
- What would this look like for a Python/R script where there's no such concept of a cell? 
- What if there are multiple notebooks/script in the pipeline? Would there be one progress bar for each? 
- How would a user know how many nodes/files are still left to be processed?

bourdakos1:
Here is the full video at 5000% speed 
progress-bars

bourdakos1:
@ptitzler These progress bars are what we currently output as logs when running pipelines from the UI. I believe scripts would log nothing, but I think we should add some sort of indication that a script is running. I think we should also put some sort of indicator of how many nodes will be run. However, I think we can wait for another PR to add these features

bourdakos1:
The main thing that concerns me with out current implementation is that we are logging over the paper mill logs

ptitzler:
>  I believe scripts would log nothing, but I think we should add some sort of indication that a script is running. 

Yes. Script processing might take as long or longer than notebook processing. Some sort of indication that it is still being processed would be helpful for long running jobs.

> I think we should also put some sort of indicator of how many nodes will be run. 

Yes.

> However, I think we can wait for another PR to add these features

your call.

lresende:
The 'run' scenario you showed, where there are only notebooks and only using the run command is not a good example. In that case that you showed, there is no discussion that the run with only notebooks, the spinner is not so desirable. 

I am really more concerned with:
- Processing scripts (a real one that takes a long time to execute)
- Submitting scripts, where the dependencies are archived and uploaded to cos (try submitting the covid scenario)

lresende:
> The main thing that concerns me without current implementation is that we are logging over the paper mill logs

Would you feel better removing the mill logs?

bourdakos1:
@lresende Yea, at least for now. I’m fine if we silence it’s output and use the spinner

lresende (approved):

bourdakos1:
@lresende did you remove the papermill logs for cli runs?

bourdakos1:
✅ @bourdakos1 approved these changes

kevin-bates (approved):

lresende (approved):

ptitzler:
- The help text for the `runtime-config` option is ambiguous. It's not clear whether one needs to specify the  ` instance name` (`my_runtime`) or `resource name` (`path/to/myruntime.json`). 
- Is there a way to replace `TEXT` with something more meaningful?

```
$ elyra-pipeline submit --help 
Usage: elyra-pipeline submit [OPTIONS] PIPELINE_PATH

  Submit a pipeline to be executed on the server

Options:
  --runtime-config TEXT  Runtime config where the pipeline should be processed
                         [required]

  --help                 Show this message and exit.
```

ptitzler:
The help commands should be consistent (either both support `--help` or both support `-h` or both support both):

```
$ elyra-pipeline -h
Usage: elyra-pipeline [OPTIONS] COMMAND [ARGS]...
Try 'elyra-pipeline --help' for help.

Error: no such option: -h

$ elyra-metadata -h        

Manage Elyra metadata.


```

bourdakos1:
@ptitzler `help` is handled by `click` so metadata will be consistent once @lresende finishes that transition

bourdakos1:
We can add `-h`, but it's not included by default: https://click.palletsprojects.com/en/7.x/documentation/#help-parameter-customization

ptitzler:
Usability: Help should also display help for arguments, not just options. For example, as is the `PIPELINE_PATH` argument name  communicates that the parameter value is supposed to be a path but not necessarily a file name.

```
$ elyra-pipeline run --help
Usage: elyra-pipeline run [OPTIONS] PIPELINE_PATH

  Run a pipeline in your local environment

Options:
  --help  Show this message and exit.
```

ptitzler (approved):
lgtm, assuming that missing functionality is added later, such as  - for local execution -  displaying (prior to execution) the number of nodes in the pipeline and (during execution) the node name + filename being processed.

PR #1279, opened 2021-02-09
Add ability to include image pull policy with runtime image

### LINKED ISSUE #1233: Add support to set imagePullPolicy attribute on runtime images
**Describe the issue**

KFP has introduced the ability to set imagePullPolicy attribute for a given pipeline and/or node. By default, the Image pull policy can be one of Always, Never, IfNotPresent. Defaults to Always if :latest tag is specified, or IfNotPresent otherwise. Cannot be updated. More info: https://kubernetes.io/docs/concepts/containers/images#updating-images

This is useful due to the recent rate limitations on the docker.io repository.

akchinSTC:
Fixes #1233

 
           in the file; or

           it.

### DIFF
--- a/elyra/metadata/schemas/runtime-image.json
+++ b/elyra/metadata/schemas/runtime-image.json
@@ -37,6 +37,15 @@
           "type": "string",
           "minLength": 1
         },
+        "pull_policy": {
+          "title": "Image Pull Policy",
+          "description": "The pull policy to use when selecting this image",
+          "type": "string",
+          "enum": ["Always", "IfNotPresent", "Never"],
+          "uihints": {
+            "field_type": "dropdown"
+          }
+        },
         "tags": {
           "title": "Tags",
           "description": "Tags for categorizing runtime images",
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -23,8 +23,8 @@ class Operation(object):
     """
 
     def __init__(self, id, type, classifier, filename, runtime_image, memory=None, cpu=None, gpu=None,
-                 dependencies=None, include_subdirectories: bool = False, env_vars=None, inputs=None, outputs=None,
-                 parent_operations=None):
+                 dependencies=None, include_subdirectories: bool = False, env_vars=None, inputs=None,
+                 outputs=None, parent_operations=None):
         """
         :param id: Generated UUID, 128 bit number used as a unique identifier
                    e.g. 123e4567-e89b-12d3-a456-426614174000
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -237,14 +237,16 @@ class RuntimePipelineProcess(PipelineProcessor):
                            format(operation.name), exc_info=True)
             raise ex from ex
 
-    def _get_runtime_configuration(self, name):
+    def _get_metadata_configuration(self, namespace, name=None):
         """
-        Retrieve associated runtime configuration based on processor type
+        Retrieve associated metadata configuration based on namespace provided and optional instance name
         :return: metadata in json format
         """
         try:
-            runtime_configuration = MetadataManager(namespace=MetadataManager.NAMESPACE_RUNTIMES).get(name)
-            return runtime_configuration
+            if not name:
+                return MetadataManager(namespace=namespace).get_all()
+            else:
+                return MetadataManager(namespace=namespace).get(name)
         except BaseException as err:
-            self.log.error('Error retrieving runtime configuration for {}'.format(name), exc_info=True)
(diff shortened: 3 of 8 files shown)


elyra/pipeline/pipeline.py
```diff
@@ -141,6 +141,14 @@ def env_vars_as_dict(self, logger: Optional[object] = None) -> Dict:
                         print(f"Could not process environment variable entry `{nv}`, skipping...")
         return envs
 
+    @property
+    def image_pull_policy(self):
+        image_tag = self._runtime_image.split(":")[1]
+        if image_tag == "latest":
```
kevin-bates, line 147: Since 'latest' is also implied when a tag does not exist, I believe that should be handled as well.

kevin-bates (commented):
I understood the issue to be that we need a way for a user to specify the pull policy of a runtime image.  This seems to only retain the k8s default behavior.  

Use case:  I'm developing a new image, tagged with 'dev', but I want that image to use a pull policy of 'Always', how can I do that?

I figured this would require an enum added to the metadata of ['Always', 'Never' and 'IfNotPresent'] or something like that.  Then, the runtime would consult the instance and set the desired pull policy.

Hmm - thinking about that implies that we must also convey the runtime-image metadata name (or have a way to get it), but I believe we only set the image name in the pipeline props, so I suppose that would be an issue.  I suppose pull policy could come from the client (front end) as well.

elyra/pipeline/pipeline.py
akchinSTC: Updated

akchinSTC:
Based on the initial comments in the issue, it seemed like we wanted a behavior set based on image name rather than adding more configuration options in the metadata at the image level or common properties at the node level. 

thoughts @lresende ?

lresende:
My understanding is:
- Runtime-image schema needs to be updated with an optional image_policy enum field
- When a user is creating/editing a runtime-image, he can select the associated image policy
   - As @kevin-bates mentioned (and assuming version tags are immutable), the pull policy only makes sense with non-version tags such as none, latest, dev, etc. Having said that, I don't think we need to make any attempt to add any logic to the selection, either the user selects one or not independent of the type of tag associated with the image. 
- As we did with resources, if the user does not select an image_policy value, we should not send anything and the K8s default behavior should apply

Does this clarify a bit more?

kevin-bates:
Yes. Thanks. Will we just copy the pull policy and store the value, if present, in the node properties as we do today with the image name, or change to using a by-reference approach for the runtime instance? If the former, updates to the image name should unconditionally reset the pull policy value.

lresende:
I would say similar to what we do today, where we would probably add an additional field ... something like `runtime-image-pull-policy`

akchinSTC:
I think I opted for the latter approach from Kevin. The schema has been updated and I do a lookup of the image metadata when we generate the pipeline to find the pull policy and set it if it exists. I modified the get_runtime_configuration function to do double duty and get both runtime and image metadata depending on input, but it might be a little messy since I think they return different types. I may just create another function for either.

kevin-bates:
I think this approach may be better than by-value (where the pull-policy is persisted) for the following reasons:
1. The pull policy isn't persisted in the pipeline file beside the image.
2. If the server doesn't find a matching image name in its set of runtime-images, the Kubernetes behavior kicks in, and that seems sufficient.  Users can create a matching runtime-image instance associated with the referenced image name if they want pull behavior other than the defaults.

Since pull policy is really an attribute of the processing server and not so much an attribute of the image, this approach satisfies that.  It also prevents the "stickiness" of existing pipelines and folks having to potentially visit each node to change the pull policy.

kevin-bates (approved):
Thanks Alan.

elyra/metadata/schemas/runtime-image.json
```diff
@@ -37,6 +37,12 @@
           "type": "string",
           "minLength": 1
         },
+        "pull_policy": {
+          "title": "Image Pull Policy",
+          "description": "The pull policy to use when selecting this image",
+          "type": "string",
+          "enum": ["Always", "IfNotPresent", "Never"]
```
lresende, line 44:
```suggestion
          "enum": ["Always", "IfNotPresent", "Never"],
          "uihints": {
            "field_type": "dropdown"
          }          
```

lresende (commented):
LGTM, I added a suggestion to make the field a dropdown. We also need to resolve conflicts

akchinSTC:
> LGTM, I added a suggestion to make the field a dropdown. We also need to resolve conflicts

Updated and rebased on master after airflow feature merge. 
Airflow template and processor have been updated with the pull policy logic.

lresende (approved):

PR #1284, opened 2021-02-10
Add support for kubernetes pod labels and annotations for KFP pipelines

### LINKED ISSUE #679: Add support for versioning instead of timestamp when submitting pipelines to KFP
KFP didn't have the notion of pipeline version, and due to this Elyra submits a new pipeline with a timestamp appended to it's name to work around this issue.

It looks like in KFP 1.0.0 they now support pipeline versions, and we should leverage that in Elyra which wich will help remove the workarounds introduced around this area.

image

### LINKED ISSUE #1277: Add labels and annotations to container to support debugging and operational tasks
When a pipeline node is executed, KFP launches a pod, which executes the node's Python script or notebook. The pod name, labels, and annotations are auto-generated and don't contain any information that associates the pod with the node. Therefore currently the only way to associate a node with a pod is to inspect the KFP UI and/or? CLI , which doesn't allow for easy reverse lookup ("determine which pipeline and node is associated with pod XYZ that is [insert something that has a negative impact on the system]") 

image

Example list of pods in one of our Kubernetes test environments:

```
$ kubectl get pods --namespace anonymous
NAME                                              READY   STATUS      RESTARTS   AGE
aaa-c5x7s-1198940926                              0/2     Completed   0          20h
lambda-2rvgx-3787107965                           0/2     Completed   0          21h
lambda-4wbg9-3142036415                           0/2     Completed   0          5h38m
lambda-qz7l8-3348423942                           0/2     Completed   0          23h
lambda-qz7l8-3574047449                           0/2     Completed   0          23h
lambda-s4cfs-777187791                            0/2     Completed   0          20h
ml-pipeline-ui-artifact-c8fbcb8f9-km6x6           2/2     Running     0          2d21h
ml-pipeline-visualizationserver-6b78c9646-p6lkm   2/2     Running     0          2d21h
untitled-48cqr-1098260914                         0/2     Completed   0          3h9m
```
Inspecting `lambda-2rvgx-3787107965 `, which embodies a pipeline node

```
$ kubectl describe pod lambda-2rvgx-3787107965 --namespace anonymous
Name:         lambda-2rvgx-3787107965
Namespace:    anonymous
Priority:     0
Node:         devises1.fyre.ibm.com/9.30.211.16
Start Time:   Thu, 04 Feb 2021 14:41:06 -0800
Labels:       pipelines.kubeflow.org/cache_enabled=true
              pipelines.kubeflow.org/cache_id=
              pipelines.kubeflow.org/metadata_context_id=7
              pipelines.kubeflow.org/metadata_execution_id=12
              pipelines.kubeflow.org/metadata_written=true
              workflows.argoproj.io/completed=true
              workflows.argoproj.io/workflow=lambda-2rvgx
Annotations:  pipelines.kubeflow.org/execution_cache_key: 8f6d7dfa82faa95956d0b182db12021712e01161a06711cbfef332a755017b0e
              pipelines.kubeflow.org/metadata_input_artifact_ids: []
              pipelines.kubeflow.org/metadata_output_artifact_ids:
                [{"id": 11, "name": "mlpipeline-ui-metadata", "uri": "minio://mlpipeline/artifacts/lambda-2rvgx/lambda-2rvgx-3787107965/mlpipeline-ui-meta...
              sidecar.istio.io/inject: false
              workflows.argoproj.io/node-name: lambda-2rvgx.untitled
              workflows.argoproj.io/outputs:
                {"artifacts":[{"name":"mlpipeline-metrics","path":"/tmp/mlpipeline-metrics.json","optional":true},{"name":"mlpipeline-ui-metadata","path":...
              workflows.argoproj.io/template:
                {"name":"untitled","inputs":{},"outputs":{"artifacts":[{"name":"mlpipeline-metrics","path":"/tmp/mlpipeline-metrics.json","optional":true}...
Status:       Succeeded
IP:           172.17.0.61
IPs:
  IP:           172.17.0.61
Controlled By:  Workflow/lambda-2rvgx
Containers:
  wait:
    Container ID:  docker://51edfc93aff09c08fc2056a456c75a112bef17057c736bf00bf88a2d8c3af51c
    Image:         argoproj/argoexec:v2.3.0
    Image ID:      docker-pullable://argoproj/argoexec@sha256:85132fc2c8bc373fca885df17637d5d35682a23de8d1390668a5e1c149f2f187
    Port:          <none>
    Host Port:     <none>
    Command:
      argoexec
      wait
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Thu, 04 Feb 2021 14:41:09 -0800
      Finished:     Thu, 04 Feb 2021 14:41:44 -0800
    Ready:          False
    Restart Count:  0
    Environment:
      ARGO_POD_NAME:  lambda-2rvgx-3787107965 (v1:metadata.name)
    Mounts:
      /argo/podmetadata from podmetadata (rw)
      /argo/secret/mlpipeline-minio-artifact from mlpipeline-minio-artifact (ro)
      /var/run/docker.sock from docker-sock (ro)
      /var/run/secrets/kubernetes.io/serviceaccount from default-editor-token-6j9xd (ro)
  main:
    Container ID:  docker://9253f7abda499c6b59d86f74870134c21490069e4aa0882800dac16f8644a83a
    Image:         quay.io/ptitzler/pandas:1.1.4
    Image ID:      docker-pullable://quay.io/ptitzler/pandas@sha256:b55a9e1a15250aa86f5fd6ddb6025a12b295dae6e919f31479be269622d9d78c
    Port:          <none>
    Host Port:     <none>
    Command:
      sh
      -c
    Args:
      mkdir -p ./jupyter-work-dir/ && cd ./jupyter-work-dir/ && curl -H "Cache-Control: no-cache" -L https://raw.githubusercontent.com/ptitzler/kfp-notebook/label/etc/docker-scripts/bootstrapper.py --output bootstrapper.py && curl -H "Cache-Control: no-cache" -L https://raw.githubusercontent.com/ptitzler/kfp-notebook/label/etc/requirements-elyra.txt --output requirements-elyra.txt && python3 -m pip install  packaging && python3 -m pip freeze > requirements-current.txt && python3 bootstrapper.py --cos-endpoint http://devises1.fyre.ibm.com:30927 --cos-bucket pipeline-artifacts --cos-directory "untitled-0204144104" --cos-dependencies-archive "Untitled-eae61c44-e3df-46c8-ac41-60a2c6e93c7a.tar.gz" --file "Untitled.ipynb" 
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Thu, 04 Feb 2021 14:41:10 -0800
      Finished:     Thu, 04 Feb 2021 14:41:43 -0800
    Ready:          False
    Restart Count:  0
    Environment:
      AWS_ACCESS_KEY_ID:             REDACTED
      AWS_SECRET_ACCESS_KEY:         REDACTED
      ELYRA_ENABLE_PIPELINE_INFO:    True
      ELYRA_WRITABLE_CONTAINER_DIR:  /tmp
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-editor-token-6j9xd (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  podmetadata:
    Type:  DownwardAPI (a volume populated by information about the pod)
    Items:
      metadata.annotations -> annotations
  docker-sock:
    Type:          HostPath (bare host directory volume)
    Path:          /var/run/docker.sock
    HostPathType:  Socket
  mlpipeline-minio-artifact:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  mlpipeline-minio-artifact
    Optional:    false
  default-editor-token-6j9xd:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-editor-token-6j9xd
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                 node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:          <none>
```

there appears to be not much identifying information.
By adding labels and annotations, that for example identify in a unique way the pipeline and the node the problem could be resolved. With labels, for example, one could easily query the system to only return pods that belong to a specific pipeline.

### LINKED ISSUE #1289: Pipeline runs launched from the Pipeline Editor always use the Tekton client
**Describe the issue**

When a user runs a pipeline from the Visual Pipeline Editor, Elyra always uses the Tekton client:

https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_kfp.py#L96
```
client = kfp_tekton.TektonClient(host=api_endpoint, cookies=session_cookie)
```

Could this be causing any issues if there are incompatibilities between the two clients?

**Deployment information**
 - Elyra version: 2.0.x

 - Kubeflow Pipelines

ptitzler:
This PR 
 - adds support for pipeline versioning
 - enables the pod labels and annotations support that is delivered by https://github.com/elyra-ai/kfp-notebook/pull/83
 - attaches label/annotation metadata to containers before running a pipeline
 - attaches label/annotation metadata to containers when exporting a pipeline
 - fixes #679 
 - fixes #1277 
 - fixes #1289 

Example of metadata that is attached to each KFP pipeline node:

```
metadata:
  ...
  labels:
    elyra/experiment-name: my-labeled-pipeline-g-0210142652
    elyra/node-name: notebook
    elyra/node-type: notebook-script
    elyra/pipeline-name: my-labeled-pipeline-g
    elyra/pipeline-version: my-labeled-pipeline-g-0210142652
    ...
  annotations:
    elyra/node-file-name: notebook.ipynb
    ...
```

> Values that are derived from user provided input (`experiment-name`,  `node-name`, `pipeline-name`, `pipeline-version` / future, and `node-file-name`) don't necessarily match the user provided input, due to restrictions imposed by Kubernetes. For example, if a pipeline was named by a user `%off-analysis`, the name cannot be used as-is as a label value, because `%` is not a supported character. Elyra takes care of converting this string into a format that meets the restrictions. However, that means that the listed values might deviate significantly from the one a user might be familiar with.
 
User workflow scenarios that are covered:
 - **S1**: Run pipeline from Pipeline editor
 - **S2**: Export pipeline to YAML > upload YAML in KFP UI > create an experiment > run pipeline
 - **S3**: Export pipeline to Python DSL > run generated Python DSL in terminal window to create an experiment and run the pipeline

More in-depth discussion about issues in the comments below.

           in the file; or

           it.


ptitzler:
#### `elyra/node-name`:

Identifies the pipeline node that the Kubernetes pod is processing.
 
```
elyra/node-name: notebook
```
Issue: The node name is currently the notebook's | script's basename (e.g. `notebook` for `notebook.ipynb`) and not the user-provided value. @lresende is working on a fix that should pass through the correct information.

ptitzler:
#### `elyra/pipeline-version`:

Pipeline versioning is currently not supported. Therefore the value is '' for every pipeline. 
 
```
elyra/pipeline-version: ''
```

ptitzler:
#### `elyra/experiment-name`:

In KFP an experiment identifies a "workspace" in which a pipeline is executed. Currently there is a 1-to-1 relationship between pipelines and experiments (and their names) if a pipeline is executed through the Elyra UI.

```
elyra/experiment-name: my-labeled-pipeline-g-0210142652
```

ptitzler:
#### `elyra/pipeline-name`:

Identifies the pipeline that the Kubernetes pod is associated with.
 
```
elyra/pipeline-name: my-labeled-pipeline-g
```

ptitzler:
#### `elyra/node-type`:

Identifies the pipeline node type, which initially is always `notebook-script` to indicate that the pod processes a Python script or notebook. Additional types can be introduced in the future, once pipelines can be assembled from components.
 
```
elyra/node-type: notebook-script
```

ptitzler:
#### `elyra/node-file-name` (annotation):

This annotation identifies the notebook | Python script file name for pods that are labeled with `elyra/node-type: notebook-script`. This annotation complements the `elyra/node-name` label.

```
elyra/node-file-name: notebook.ipynb
```

ptitzler:
Basic test pipeline:
- Download [i1284.tar.gz](https://github.com/elyra-ai/elyra/files/5968831/i1284.tar.gz)
- Extract the archive in your JupyterLab workspace
- Open `my-labeled-pipeline`

image

ptitzler:
Example screen shots taken during test of scenario 1 (run from Elyra UI)

KFP _Pipelines_ view:

image

KFP _Experiments_ view:

image

KFP _component_ view (for one of the nodes):

image

elyra/pipeline/processor_kfp.py
```diff
@@ -51,6 +51,8 @@ def process(self, pipeline):
         t0_all = time.time()
         timestamp = datetime.now().strftime("%m%d%H%M%S")
         pipeline_name = f'{pipeline.name}-{timestamp}'
```
kevin-bates, line 53: Shouldn't `pipeline_name` be `pipeline.name` (sans timestamp)?

ptitzler:
The tests are failing https://github.com/elyra-ai/elyra/pull/1284/checks?check_run_id=1891031113#step:12:1407 because the outdated `kfp-notebook` package is used. The tests should succeed once elyra-ai/kfp-notebook#83 was merged and the updated package published.

elyra/pipeline/processor_kfp.py
```diff
@@ -276,11 +307,14 @@ def _cc_pipeline(self, pipeline, pipeline_name):
 
             # create pipeline operation
             notebook_ops[operation.id] = NotebookOp(name=operation.name,
+                                                    pipeline_name=pipeline.name,
```
kevin-bates, line 310: Shouldn't the variable `pipeline_name` be used here and not `pipeline.name`?  This seems to support my previous question.

elyra/templates/kfp_template.jinja2
```diff
@@ -58,13 +64,13 @@ if __name__ == "__main__":
 
     client = kfp.Client(host='{{ api_endpoint }}')
 {% endif %}
-    name = '{{ pipeline_name }}'
-    exp = client.create_experiment('{{ pipeline_name }}',
+    run_name = '{{ pipeline_name }}-run'
```
kevin-bates, line 67: Should the 'run' be associated with the `experiment_name` rather than `pipeline_name`?

kevin-bates (commented):
It seems like we're conflating the pipeline name with the experiment name or I don't understand the true difference.  I guess I view the _experiment_ to be like an _instance_ of a pipeline.  A runnable _instance_.

ptitzler:
> I guess I view the _experiment_ to be like an _instance_ of a pipeline. A runnable _instance_.

In KFP an [experiment is a workspace](https://www.kubeflow.org/docs/pipelines/overview/concepts/experiment/) and can contain runs from different pipelines. For example, a user can create an experiment X1 and then run pipeline P1 and pipeline P2 in that experiment. If the user chooses to create another experiment X3, he/she can run P1, P2, or any other pipeline P3 in this experiment.

elyra/templates/kfp_template.jinja2
ptitzler: No, because a run (also sometimes referred to as job) is an instance of a pipeline execution. See my earlier clarification what experiments are in response to your earlier comment.

elyra/pipeline/processor_kfp.py
ptitzler: Yes. I didn't realize that this implementation violates the definition in https://github.com/elyra-ai/elyra/pull/1284#issuecomment-777779353, which states that the value is the pipeline file name or the value the user provided when the run was initiated.

However, the issue is that that value currently does not surface in the KFP UI, as shown in the screen captures for this scenario: https://github.com/elyra-ai/elyra/pull/1284#issuecomment-778510188.

kevin-bates:
> In KFP an experiment is a workspace and can contain runs from different pipelines. For example, a user can create an experiment X1 and then run pipeline P1 and pipeline P2 in that experiment. If the user chooses to create another experiment X3, he/she can run P1, P2, or any other pipeline P3 in this experiment.

Thank you for this clarification.  I had nearly the opposite definition.

Today, Elyra only supports _single-pipeline experiments_.  Are there plans to support multi-pipeline experiments?  I haven't heard us talk about that kind of thing and a quick search doesn't find anything, but let's not rule that out.

Since the "experiment" is really an end-user notion, it seems to me that, by default, the experiment name should be the thing provided by the user and, if not provided, default to the basename of the pipeline filename (if multi-pipeline experiments are supported, this default would need to be revisited).  It seems like when the user runs a pipeline, they should have the option of entering a name (that defaults to the pipeline name).

`pipeline_name` is purely static and is the basename of the pipeline file.  There would really be no need to prompt for the user to change this value, it should be derived directly from the pipeline filename.  They also get to set/change the name when they create or rename the pipeline file.

And the `run_name` would be the timestamped `pipeline_name` and no need for the `-run` suffix.

I also believe that `experiment_name` should be something passed _into_ the pipeline processor (assuming a given processor processes a single pipeline) so that when/if we support true (multi-pipeline) "experiments", that same name is conveyed to each processor.

This could be summarized as:
- _experiment_name_ is provided at runtime and defaults to the pipeline file basename.
- _pipeline_name_ is the basename of the pipeline filename. 
- _run_name_ is purely the timestamped `pipeline_name`.
- And, in the KFP UI, the _pipeline_name_ is not surfaced.

Hmm - since experiments can be run multiple times (and is why we timestamp the pipeline name) I wonder if the experiment name passed to `create_experiment` should be the run-name **prefixed** with `experiment_name`?  That is `experiment_name`-`pipeline_name`-timestamp.

I think it would be good to get a solid definition of these terms in particular and align them with the roadmap.  We should also consider how they might equate to Airflow parlance since we essentially want the same pipeline editor for both/all platforms.  If this means naming "experiments" to be more like a "workspace" (although I would view a workspace to be a container of experiments), then we could define our own terms with the platform-specific terms mapped in the respective platform-specific processors.

ptitzler:
Based on the discussion've we had this morning, I updated processing behavior if a pipeline is run from the pipeline editor.

First time run:
- Pipeline file name: `version-test3.pipeline`
- Pipeline name: `version-test3` (the default when the job was submitted)
- KFP UI pipeline name: `version-test3`
- KFP UI version name: `version-test3`
- KFP UI experiment name: `version-test3`
- KFP UI run name: `version-test3-0215144656`
- COS artifact location: `.../version-test3-0215144656`

Second/n-th time run (property values that have changed are marked in **bold**):

- Pipeline file name: `version-test3.pipeline`
- Pipeline name: `version-test3` (the default when the job was submitted)
- KFP UI pipeline name: `version-test3`
- KFP UI version name: **`version-test3-0215153755`** (Note the appended timestamp to achieve uniqueness)
- KFP UI experiment name: `version-test3`
- KFP UI run name: **`version-test3-0215153755`** (Note the updated timestamp to achieve uniqueness)
- COS artifact location: **`.../version-test3-0215153755`** (Note the updated timestamp to achieve uniqueness)

From a user's point of view the only thing that changes is that all pipeline runs are now grouped under the same pipeline name and experiment. Previously each pipeline run was treated as a separate (logical) pipeline and experiment.

ptitzler:
With this new behavior, a user can finally compare performance metrics across multiple runs: 

image

elyra/templates/kfp_template.jinja2
kevin-bates: Isn't `run_name` the time-stamped `pipeline_name` now?  Does it require its own parameter since it's not always the `pipeline_version`?

kevin-bates (commented):
This looks really good - thank you for spending the necessary time on this!  Just had a couple comments - one of which from the previous pass.

ptitzler:
Quick update - `process()` - as far as I can tell - currently has no awareness of what the original pipeline file name is. Therefore, if a user overrides the pipeline name when the job is submitted, the implementation currently does not have awareness that this has happened.

elyra/templates/kfp_template.jinja2
ptitzler: Right, I still need to fix export!

kevin-bates:
> Quick update - process() - as far as I can tell - currently has no awareness of what the original pipeline file name is. Therefore, if a user overrides the pipeline name when the job is submitted, the implementation currently does not have awareness that this has happened.

I guess this makes sense. Does the server ever need to know the pipeline's filename since the pipeline is actually sent in the request?  The file is never read on the server (other than the new CLI tool) - to the best of my knowledge.

Has all the `basename` talk been conflated with node and notebook file names? :smile: 
If so, that's probably "on me" and I apologize!

ptitzler:
Export as YAML update. I tested two scenarios:
 - upload exported YAML as new version
 - upload exported YAML as new pipeline
 
The results surface an issue that, as is, cannot be resolved properly without loosing information.

Pod metadata for upload-as-new-version scenario:
```
    elyra/experiment-name: ''     # this is correct because at compile time we don't know which experiment the user will choose eventually
    elyra/node-name: notebook
    elyra/node-type: notebook-script
    elyra/pipeline-name: version-test3 # this is correct because the pipeline was manually uploaded as a new version
    elyra/pipeline-version: '' # this is correct because at compile time we don't know which experiment the user will choose eventually
    elyra/node-file-name: ...
```

Pod metadata for upload-as-new-pipeline scenario:
```
    elyra/experiment-name: '' # this is correct because at compile time we don't know which experiment the user will choose eventually
    elyra/node-name: notebook
    elyra/node-type: notebook-script
    elyra/pipeline-name: version-test3 # this is **not correct** because the pipeline was manually uploaded as new pipeline `version-test4`
    elyra/pipeline-version: ''
    elyra/node-file-name: notebook.ipynb
```

From a conceptual point of view it'd probably be better to set `elyra/pipeline-name` to the empty string, which is at least not wrong. However, with pipeline name and version not set, there is no good way to associate the pod(s) with a pipeline at all ...
```
    elyra/experiment-name: '' 
    elyra/node-name: notebook
    elyra/node-type: notebook-script
    elyra/pipeline-name: ''
    elyra/pipeline-version: ''
    elyra/node-file-name: notebook.ipynb
```

ptitzler:
> Does the server ever need to know the pipeline's filename since the pipeline is actually sent in the request? 

Not the server, but it's the operator/monitor that is potentially loosing context information. The results of the export-to-YAML test highlight that nicely. In that scenario having another piece of metadata would help. Maybe add one more _annotation_ that contains the original pipeline file?  Annotations have the advantage that their values are a lot less restricted.

kevin-bates:
Yes, an annotation makes sense for this, but where is that value going to come from?  The processor doesn't "see" the filename since the body of the request _is_ the file content.

ptitzler:
> Yes, an annotation makes sense for this, but where is that value going to come from? The processor doesn't "see" the filename since the body of the request _is_ the file content.

Is this an HTTP request? If it is, could the information be enclosed in the header?

kevin-bates:
> Is this an HTTP request? If it is, could the information be enclosed in the header?

Yes.  I suppose it could, but I don't know enough about web-programming to answer definitively.   Anyone else?

ptitzler:
To resolve the "export as YAML" issue I'm changing the value of  `elyra/pipeline-name` to be empty because I believe no information is better than (potentially) incorrect information.

elyra/templates/kfp_template.jinja2
ptitzler: Should be okay now

elyra/pipeline/processor_kfp.py
ptitzler: fixed!

kevin-bates (approved):
Looks really good Patrick - thank you!

elyra/pipeline/processor_kfp.py
ptitzler: Yep. Fixed

akchinSTC (approved):
LGTM, testing on kfp with tekton+argo engines

lresende (approved):

Issue #1282, opened 2021-02-10
Code Snippets Languages


jingsta:
How does one add more code snippets languages to the selectable list.

Example is I have powershell and Rust I want to add as well to the list.

lresende:
The list is dynamic with the most used examples pre-populated. When you create a new code-snippet, please type the new entry (e.g. Rust) and it should be available after that.

jingsta:
Lol.... man i feel dumb. lol. Thanks man!

ptitzler:
I do believe we should update https://elyra.readthedocs.io/en/latest/user_guide/code-snippets.html, because unlike https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html#configuration-settings (for example) it doesn't provide any help that describes the properties.  I do agree that the current selection widget is ambiguous and the input field can be interpreted as a _search/filter_ condition for the default language list.

karlaspuldaro:
@ptitzler I created #1286 to track your suggestion

ptitzler:
Related issues, which should make it easier for users to use the metadata widgets: 
 - https://github.com/elyra-ai/elyra/issues/1291 (add tool tips)
 - https://github.com/elyra-ai/elyra/issues/838 (add link to documentation)

ptitzler:
@jingsta If you would like to permanently add language options to the default list, feel free to update https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/code-snippet.json#L40 and open a pull request.

PR #1293, opened 2021-02-16
Switch to using Material UI in metadata editor

marthacryan:
After discussion on #1282 indicating that the UX of the metadata editor dropdowns wasn't very intuitive, there were several requests to change the dropdown (i.e. change the placeholder text in the dropdown text input, add a "clear button" to enable having no choice for the dropdown). We use blueprint.js to create the components used in the metadata editor because we wanted to be consistent with the JupyterLab UI components, but JupyterLab is currently planning on moving away from blueprint.js toward material UI components. To align with the JupyterLab UI, we should switch to material UI, and by switching, the UX of the dropdowns also becomes more user-friendly. It slightly changes the appearance of the metadata editor, but overall the experience is mostly the same. A couple of changes:
* There is an asterisk next to the label of required fields (if preferred, it could be changed to instead include "(required)" after required fields)

New code snippet:
image

Partially filled in code snippet:
image

Dropdown:
image

Dark mode:
image

Validation:
image

Todo:
- [x] Remove "ResizeSensor" component from the metadata editor
- [x] Bug in loading values
- [x] Bug in dark mode for save button
- [x] Add description
- [x] Move description into tooltip
- [x] Fix alignment of password labels
- [x] Hard-code width of API endpoint field for runtimes editor
- [x] Add asterisk instead of required, with "* = required" at the top
- [x] Adjust wording for asterisk = required

           in the file; or

           it.


ptitzler:
As part of the general due diligence, could you please add a list of widget types we are currently using in our UI? There are probably special considerations we need to take into account. For example, the KFP runtime configuration has a two-value logic -- `argo` and `tekton` -- for engine type. To support this, the select widget we use must support disabling the "create" feature because custom values would be invalid.

ptitzler:
> There is an asterisk next to the label of required fields (if preferred, it could be changed to instead include "(required)" after required fields)

From an accessibility point of view displaying the text `required` instead of `*` would probably be better because there's no ambiguity about its meaning.

marthacryan:
Going to work on fixing the broken CI tests

marthacryan:
After receiving feedback that the descriptions were taking up too much space and should be moved into a tooltip, I looked into adding a small info icon that revealed the description of a field when hovered over. However, due to the dynamic nature of the labels (the label is a "placeholder" until the field is in focus or populated with a value), the normal placement for that type of icon seemed awkward, so I made the tooltip appear when the *field itself* is hovered over:
image

Thoughts? @ptitzler

ptitzler:
Hard to say without trying it:
 - How likely is it that a user will trigger the hover and thus _notice_ that there are tool tips? That's one advantage an i(nformation) icon has over a "hidden" trigger.
 - How likely is it that the user will accidentally trigger the tool tips and get annoyed because they interfere with the task of entering information?

marthacryan:
> * How likely is it that a user will trigger the hover and thus _notice_ that there are tool tips? That's one advantage an i(nformation) icon has over a "hidden" trigger.

It appears quickly when hovering over a field as well as when you're entering a value in a field, so I think it's likely that users would notice, but it's hard to say for sure. 

> * How likely is it that the user will accidentally trigger the tool tips and get annoyed because they interfere with the task of entering information?

The tooltip is present only while the user is hovering over a field or when the field is in focus (i.e. the user is entering a value in the field). It wouldn't be in the way of them entering information because it is below the actual field. 

@ptitzler My other idea for the info icon was to have it be at the right end of the input, basically where the eye icon is placed for password fields. Does that sound better to you?

ptitzler:
I just tested it. The user experience is actually pretty good and doesn't seem annoying. The layout of the fields is also much better now. Good work!

tool tip

One problem is that tool tips don't seem to work for the `display name`, perhaps because it's not a metadata property. 

Some good info about `*` and wording: https://ux.stackexchange.com/questions/372/required-field-validation. `All fields marked with an asterisk are required`

ptitzler:
Have you tried what this looks like in the pipelines properties view? The default size of that view is rather small, which might cause issues

marthacryan:
> Have you tried what this looks like in the pipelines properties view? The default size of that view is rather small, which might cause issues

@ptitzler Oh oops I totally forgot to mention - I was thinking that adding this to the properties editor actually should be in a separate PR / may not be necessary because we'll be changing over to the new pipeline editor relatively soon. Does that sounds okay?

ptitzler:
Sure!

marthacryan:
@ptitzler Just added the ability to specify placeholders for text input after playing around with it on my branch!

ptitzler:
Very nice 👍 

image

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -14,25 +14,30 @@
...
+  createMuiTheme,
+  ThemeProvider
+} from '@material-ui/core';
 
 import * as React from 'react';
 
 import { MetadataEditorTags } from './MetadataEditorTags';
 
 const ELYRA_METADATA_EDITOR_CLASS = 'elyra-metadataEditor';
+const ELYRA_METADATA_EDITOR_DARK = 'elyra-metadataEditor-dark';
```
lresende, line 40: Should we append `class` for consistency on class constants? If so, please update the usages...
```suggestion
const ELYRA_METADATA_EDITOR_DARK_CLASS = 'elyra-metadataEditor-dark';
```

packages/ui-components/src/DropDown.tsx
```diff
@@ -26,93 +31,61 @@ export interface IDropDownProps {
...
-        >
-          <Button
-            rightIcon="caret-down"
-            text={this.props.choice ? this.props.choice : '(No selection)'}
-          />
-        </Select>
-      </FormGroup>
-    );
-  }
+          renderInput={params => (
```
lresende, line 76: are these warnings something simple we could fix?

lresende (commented):
Overall, looks good to me, with some small comments/suggestions below. 

Another thing I would like to discuss, outside of the scope of this pr, is how to better centralize and manage styles, particularly when related to light/dark mode. It seems we have some in the pipeline editor, others in the metadata-common and maybe even on ui-components? Then, for example, how do we enforce that all components are using the same background? Today it seems that the Metadata editor is using a different one from the rest of the UI.

packages/metadata-common/style/index.css
```diff
@@ -92,20 +92,23 @@
...
+  margin-right: 70px;
 }
 
-.bp3-select-popover {
-  max-height: 150px;
-  overflow-y: auto;
+.elyra-metadataEditor-dark.elyra-metadataEditor .MuiButtonBase-root {
+  color: #fff;
+  border-color: #fff;
+}
```
bourdakos1, line 102: Should this be part of the theme instead of raw css?

karlaspuldaro:
I really like how this looks! ❤️

I just have one comment.
Hovering over input text fields highlights the input border, as well as adds this little thing at the bottom:
image
Is this intentional as a second visual indicator to track what's in focus?
If so, maybe we also need this behaviour for the name field, it's the only text field that only gets the border highlight (although  in my opinion it looks better without it)

marthacryan:
@karlaspuldaro Oh that's a bug for when that's undefined - just pushed a fix

packages/ui-components/src/DropDown.tsx
lresende: Made some updates, hopefully, will resolve the warnings

lresende (approved):

ptitzler:
> @karlaspuldaro Oh that's a bug for when the description is undefined - just pushed a fix

Hmmmm, every property in our metadata schemas should include a description. If one of them doesn't, that'd be a metadata bug. I checked the schema for AA and KFP and it doesn't look like neither one is missing a description for the COS password. I am puzzled

marthacryan:
@ptitzler This PR is using the `uihints` field `description`, is that where you're looking? It could be switched to look at the actual field's description

marthacryan:
Oh wait nevermind on that comment ^^ I can investigate that inconsistency, not really sure what's going on?

marthacryan:
I think the tests are failing because they use the selector `.elyra-metadataEditor-form-github_api_endpoint > input`, but with material UI there are 2 wrappers on the `input` field, so we'll want to use `.elyra-metadataEditor-form-github_api_endpoint >> input`. I'll push the fix now.

marthacryan:
Ok looks like we were setting it inconsistently, all of the fields have descriptions now.

karlaspuldaro (approved):
Nice work! 
Beautiful green tests 💚

Issue #1298, opened 2021-02-17
Manage runtime configuration commands provide semantically inconsistent information


ptitzler:
**Describe the issue**
See next section

**To Reproduce**
1. Create a KFP runtime configuration.
2. Create an Apache Airflow runtime configuration.
3. Open the command palette <cmd><shift><c>
4. Select "Manage Apache Airflow runtimes" and check the runtime configuration list. All runtime configurations are displayed.
5. Open the command palette <cmd><shift><c>
6. Select "Manage Kubeflow Pipelines runtimes" and check the runtime configuration list. All runtime configurations are displayed.

**Expected behavior**
Several options
 - only expose one command "Manage runtimes" in the palette
 - the commands open the same metadata view but have a filter pre-selected that temporarily hides runtimes that are not of the expected type
 - ...

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 2.x
 - Operating system: macos
 - Installation source: from source / master branch
 - Deployment type: local installation

ptitzler:
Should be fixed by https://github.com/elyra-ai/elyra/pull/1327

PR #1327, opened 2021-02-23
Add support for multiple schemas in Metadata Editor

ajbozarth:

           in the file; or

           it.


ptitzler:
For the runtime configuration type specific display in

image

where you are going to replace  "Kubeflow Pipelines UI" with the appropriate text for the runtime type, would it also be possible to display additional information based on the type?

For Airflow, a user would want to have access to the configured github repository where the DAGs are stored. The URL can be assembled from the information provided in the runtime configuration. 

image


Is something like this already possible purely based on the existing metadata or is the metadata editor too limited?

ajbozarth:
I am going to update that text based on the schema. As for having different info depending on the schema, that would be a "new feature" since the displayed info is currently hard-coded, but it would be easy to add a uihint value that could determine which properties are shown instead of hard-coding it. I will look into that when I get back to that part of the code

ajbozarth:
Some To Do's:
- Add drop down to `+` button
- Remove hard-coded schema references
- update UI to display correct content in sidebar

ptitzler:
User scenarios we need to verify:

- [ ] User can choose which type of runtime configuration to create
- [ ] User can create/edit/delete a KFP runtime configuration
- [ ] User can create/edit/delete an Apache Airflow runtime configuration

ajbozarth:
Still To Do:
- [ ] Replace double `+` icons with a drop down
- [ ] Fix command pallets (could be done in followup PR) #1298
- [ ] Customize Runtimes UI content based on schema (prob in a followup PR) https://github.com/elyra-ai/elyra/pull/1327#issuecomment-785325718
- [ ] Update tests

ptitzler:
Turns out I forgot to list a user workflow. The "Open Runtimes" button in the Pipeline Editor doesn't work and needs to be added to the TODO list.

Tested:
- User can choose which type of runtime configuration to create
- User can create/edit/delete a KFP runtime configuration
- User can create/edit/delete an Apache Airflow runtime configuration

With the (for now) unified API Endpoint label it's impossible to determine the runtime type without opening the configuration if the user chooses a generic name and description:
  
image

Is there a way to expose the type information to solve the dilemma?

ajbozarth:
I pushed an update addressing Patricks previous comments. 

I've also decided on using Material-UI's ButtonGroup library to handle the dropdown and will be implementing that tomorrow

ptitzler:
Thanks Alex! I was now able to complete all scenarios.

ptitzler:
We should probably revisit the label text in the future.

image

ajbozarth:
I've pushed a working implementation of the drop down, there's still a few things to clean up, specifically it opens a drop down when there's only one schema (it should just open the new editor instead), the text for the tooltips and dropdown are a bit verbose, and lastly I want to clean up and streamline the new code itself. Also I still need to deal with the tests

bourdakos1:
actually, that might be easier said then done... it looks like this widget is a lot more complicated than I thought at first glance haha

bourdakos1:
I think the bigger issue here is that we do a lot of inheritance, when react places nicer with composition

packages/metadata-common/src/MetadataWidget.tsx
```diff
@@ -308,14 +319,83 @@ export class MetadataDisplay<
   }
 }
 
+interface IButtonProps {
+  schemas: IDictionary<any>[];
+  addMetadata: (schema: string) => void;
+}
+
+// eslint-disable-next-line func-style
+function AddMetadataButton(props: IButtonProps): React.ReactElement {
```
lresende, line 328: Could we use the arrow function style?

ptitzler:
If no runtime schema is defined, the following error is raised:

image

Note that this is an artificial scenario. I removed the schema files in `elyra/metadata/schemas/` to simulate the use case where a customer wants to prevent users from creating/using any runtime configuration.

ptitzler:
OK to defer to later:
 - manually add a third runtime configuration schema
 - build elyra
 - create a runtime configuration using the new type
 - in the list of runtime configurations the new runtime configuration is displayed as a kubeflow  configuration

packages/metadata-common/src/MetadataWidget.tsx
ajbozarth: done

ptitzler:
> If no runtime schema is defined, the following error is raised:
...

Opened https://github.com/elyra-ai/elyra/issues/1360.

karlaspuldaro:
Just built the latest changes and I see some react warnings in the chrome dev console related to the `AddMetadataButton` component
image

ajbozarth:
I just pushed an update that fixes current tests (does not add new ones though)

ptitzler (approved):
LGTM. Excluding issues that were deferred.

ajbozarth:
I'll look into the chrome errors, I dont believe I was seeing them in safari.

Also I still need to get the dropdown to support dark mode

lresende (approved):

Issue #1314, opened 2021-02-19
Metadata [editor] allow for optional specification of placeholder value for text inputs


ptitzler:
If specified in the schema, the text in bold is displayed as placeholder in the input text field:

For example,

```
 "github_repo": {
          "title": "Github DAG Repository",
          "description": "The git repository for Airflow DAGs",
          "type": "string",
          "uihints": {
            "category": "Airflow Pipelines",
           "placeholder": "git-id/repository",
          }
 }
```

might be displayed as

![image](https://user-images.githubusercontent.com/13068832/109359520-76b5e300-783a-11eb-80d4-ff1a144193e7.png)

lresende:
Fixed by #1239

PR #1345, opened 2021-02-26
Add UI placeholder metadata to KFP and Apache Airflow configurations

ptitzler:
This PR adds placeholder UI hints to the Apache Airflow and Kubeflow Pipelines metadata. The `placeholder` property is optional. If present, it is rendered by the metadata editor if the input type is "string".

See https://github.com/elyra-ai/elyra/issues/1314 for background information
 
           in the file; or

           it.

### DIFF
--- a/elyra/metadata/schemas/airflow.json
+++ b/elyra/metadata/schemas/airflow.json
@@ -38,7 +38,7 @@
           "format": "uri",
           "uihints": {
             "category": "Apache Airflow",
-            "placeholder": "http://your-apache-webserver:port"
+            "placeholder": "https://your-airflow-webserver:port"
           }
         },
         "github_api_endpoint": {
@@ -49,7 +49,7 @@
           "default": "https://api.github.com",
           "uihints": {
             "category": "Apache Airflow",
-            "placeholder": "https://api.github.com"
+            "placeholder": "https://your-github-endpoint"
           }
         },
         "github_repo": {
--- a/elyra/metadata/schemas/airflow.json
+++ b/elyra/metadata/schemas/airflow.json
@@ -37,7 +37,8 @@
           "type": "string",
           "format": "uri",
           "uihints": {
-            "category": "Apache Airflow"
+            "category": "Apache Airflow",
+            "placeholder": "http://your-apache-webserver:port"
           }
         },
         "github_api_endpoint": {
@@ -47,7 +48,8 @@
           "format": "uri",
           "default": "https://api.github.com",
           "uihints": {
-            "category": "Apache Airflow"
+            "category": "Apache Airflow",
+            "placeholder": "https://api.github.com"
           }
         },
         "github_repo": {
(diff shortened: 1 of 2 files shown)


ptitzler:
@marthacryan fyi

elyra/metadata/schemas/airflow.json
```diff
@@ -37,7 +37,8 @@
           "type": "string",
           "format": "uri",
           "uihints": {
-            "category": "Apache Airflow"
+            "category": "Apache Airflow",
+            "placeholder": "http://your-apache-webserver:port"
```
kevin-bates, line 41: I think "your-airflow-webserver:port" seems more specific since "apache" _is_ a webserver and, in this case, could be ambiguous.

elyra/metadata/schemas/airflow.json
```diff
@@ -47,15 +48,17 @@
           "format": "uri",
           "default": "https://api.github.com",
           "uihints": {
-            "category": "Apache Airflow"
+            "category": "Apache Airflow",
+            "placeholder": "https://api.github.com"
```
kevin-bates, line 52: Although the placeholder value will never be visible when a default value also exists, it seems like the placeholder should be like the others nevertheless  (e.g., `"https://your-github-endpoint"`).

ptitzler: I was thinking that by using the same value the user would be reminded what the default was that was intentionally removed

kevin-bates: I  guess this depends on how 'defaulting' works in the UI and how careful it is to distinguish between not-set and set to nothing.  If not-set, I would expect the UI to fill with a default (when a default exists, and the placeholder otherwise).  If set to nothing, I would expect the UI to show nothing (including not showing the placeholder). 

Is placeholder a UI thing in that what is displayed as the placeholder isn't considered an inputted value in any way, whereas default is considered an inputted value?  This is probably a splitting-hairs thing I guess and see that you already committed the update - so moving on.  Just to add, an advantage of the updated approach is that we can more easily determine which of the two values is being used - which might help testability - and perhaps we can then come to a conclusion on what would be best.

lresende (approved):

elyra/metadata/schemas/airflow.json
ptitzler: > Is placeholder a UI thing in that what is displayed as the placeholder isn't considered an inputted value in any way, whereas default is considered an inputted value? 

Yes to both. A placeholder is supposed to provide guidance. Take, for example, an input field for a social security number. `xxx-xxx-xxxx`. It's not a valid entry but makes it clear that the expectation is that the input is supposed to be entered in the specified format. (Placeholders are not used though to validate or enforce anything.) Generally speaking laceholders are a front-end thing only, whereas default applies now to the front-end (is displayed there) and back-end (stored/passed through).

Issue #1334, opened 2021-02-24
Standardize display location of tags in editor windows


ptitzler:
When defining metadata, a user can optionally specify tags. Currently there's no standard as to where the tags are displayed:

### Runtime configuration (top)

![image](https://user-images.githubusercontent.com/13068832/109078998-206a6800-76b3-11eb-9fed-c4e0bb8f83cd.png)

### Runtime image (bottom)

![image](https://user-images.githubusercontent.com/13068832/109079329-a4bceb00-76b3-11eb-8372-e630f4b1cb32.png)

### Code snippet (bottom)

![image](https://user-images.githubusercontent.com/13068832/109079104-4b54bc00-76b3-11eb-8c96-915ea037b808.png)

For the sake of consistency the tags should be displayed on top, right after _display name_ and _description_ (like it is done for the runtime configuration).

PR #1347, opened 2021-02-27
Group tags with name and description

ptitzler:
This PR
 - Enforces that tags are displayed after the display name and description but before all other properties for code snippets and runtime images. See next comment for motivation why a new category `source` was introduced.
 - Closes https://github.com/elyra-ai/elyra/issues/1334
 - Adds a placeholder for the runtime image `image_name` property. For the placeholder to be rendered, https://github.com/elyra-ai/elyra/pull/1293 is required.
 
New arrangements:

image

image

           in the file; or

           it.

### DIFF
--- a/elyra/metadata/schemas/code-snippet.json
+++ b/elyra/metadata/schemas/code-snippet.json
@@ -31,13 +31,22 @@
           "description": "Code snippet description",
           "type": "string"
         },
+        "tags": {
+          "title": "Tags",
+          "description": "Tags for categorizing snippets",
+          "type": "array",
+          "uihints": {
+            "field_type": "tags"
+          }
+        },
         "language": {
           "title": "Language",
           "description": "Code snippet implementation language",
           "type": "string",
           "uihints": {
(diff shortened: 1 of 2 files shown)


ptitzler:
This is what the metadata would be rendered like without the introduction of the `source` category. Note the placement, which doesn't cleanly separate metadata metadata (display name, description, tags) and metadata:

image

image

kevin-bates:
Interesting.  In thinking about ordinals for this, it strikes me that if we do add a `uihint` to address this, it should probably be two-dimensional, so that you can express that something should be on a row by itself.  More like a _coordinate_ (e.g., (1,1) = first row, first item; (3,2) = third row, second item; etc.)

Also, this implies a category named "metadata" that perhaps could be treated specially in that it doesn't trigger a group label and consists entirely of `name`, `description`, and `tags`.  That way, those three items would always precede the instance's _application_ properties.

lresende (approved):

PR #1364, opened 2021-03-03
Add additional property for pipeline response

### LINKED ISSUE #1362: Allow for custom properties in elyra.pipeline.processor.PipelineProcessorResponse
Currently a [PipelineProcessorResponse](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor.py#L91) is limited to three pieces of information:
- URL where a pipeline run can be monitored
- URL linking to the COS GUI
- Path within the the COS instance where the pipeline artifacts are stored

For Apache Airflow pipelines there is a fourth piece of information that's of interest to the user: the GitHub URL where the DAG was uploaded to.

akchinSTC:
Adds an additional `git_url` property to the submission dialog box
to provide users a link to the location of the DAG in the GitHub
repository

Fixes #1362

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -18,13 +18,12 @@ import entrypoints
 import os
 import time
 
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from elyra.metadata import MetadataManager
 from elyra.util.cos import CosClient
 from elyra.util.archive import create_temp_archive
 from elyra.util.path import get_expanded_path
 from traitlets.config import SingletonConfigurable, LoggingConfigurable, Unicode, Bool
-from typing import Optional
 
 
 elyra_log_pipeline_info = os.getenv("ELYRA_LOG_PIPELINE_INFO", True)
@@ -88,17 +87,21 @@ class PipelineProcessorManager(SingletonConfigurable):
         return res
 
 
-class PipelineProcessorResponse(object):
-    def __init__(self,
-                 run_url: str = '',
-                 object_storage_url: str = '',
-                 object_storage_path: str = '',
-                 git_url: Optional[str] = ''):
+class PipelineProcessorResponse(ABC):
 
+    _type = None
+
+    def __init__(self, run_url, object_storage_url, object_storage_path, **kwargs):
+
+        super(PipelineProcessorResponse, self).__init__(**kwargs)
         self._run_url = run_url
-        self._git_url = git_url
         self._object_storage_url = object_storage_url
         self._object_storage_path = object_storage_path
+
+    @property
+    @abstractmethod
+    def type(self):
+        raise NotImplementedError()
 
     @property
     def run_url(self):
@@ -123,19 +126,9 @@ class PipelineProcessorResponse(object):
         """
         return self._object_storage_path
 
-    @property
-    def git_url(self):
-        """
-        :return: The GitHub URL where the DAG was uploaded to
-        """
-        return self._git_url
-
+    @abstractmethod
     def to_json(self):
-        return {"run_url": self.run_url,
-                "git_url": self.git_url,
-                "object_storage_url": self.object_storage_url,
-                "object_storage_path": self.object_storage_path
-                }
+        raise NotImplementedError()
 
 
 class PipelineProcessor(LoggingConfigurable):  # ABC
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -91,9 +91,7 @@ class PipelineProcessorResponse(ABC):
 
     _type = None
 
-    def __init__(self, run_url, object_storage_url, object_storage_path, **kwargs):
-
-        super(PipelineProcessorResponse, self).__init__(**kwargs)
+    def __init__(self, run_url, object_storage_url, object_storage_path):
         self._run_url = run_url
         self._object_storage_url = object_storage_url
         self._object_storage_path = object_storage_path
@@ -126,9 +124,12 @@ class PipelineProcessorResponse(ABC):
         """
         return self._object_storage_path
 
-    @abstractmethod
     def to_json(self):
-        raise NotImplementedError()
+        return {"platform": self.type,
+                "run_url": self.run_url,
+                "object_storage_url": self.object_storage_url,
+                "object_storage_path": self.object_storage_path
+                }
 
 
 class PipelineProcessor(LoggingConfigurable):  # ABC
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -24,6 +24,7 @@ from elyra.util.cos import CosClient
 from elyra.util.archive import create_temp_archive
 from elyra.util.path import get_expanded_path
 from traitlets.config import SingletonConfigurable, LoggingConfigurable, Unicode, Bool
+from typing import Optional
 
 
 elyra_log_pipeline_info = os.getenv("ELYRA_LOG_PIPELINE_INFO", True)
@@ -88,8 +89,14 @@ class PipelineProcessorManager(SingletonConfigurable):
 
 
 class PipelineProcessorResponse(object):
-    def __init__(self, run_url='', object_storage_url='', object_storage_path=''):
+    def __init__(self,
+                 run_url: str = '',
+                 object_storage_url: str = '',
+                 object_storage_path: str = '',
+                 git_url: Optional[str] = ''):
+
         self._run_url = run_url
+        self._git_url = git_url
         self._object_storage_url = object_storage_url
         self._object_storage_path = object_storage_path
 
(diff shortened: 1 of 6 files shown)


akchinSTC:
image

lresende:
Clicking the GitHub link:
image

Note the `double slash` on the URL, even after #1363, also, because we use the token to access the repo, there is no guarantee that the user submitting the pipeline will have access to the shared git repository used to store the DAGs. 

Anyway, just my $0.02 :)

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -159,7 +159,44 @@ export class PipelineService {
     ).then(response => {
       let dialogTitle;
       let dialogBody;
-      if (response['run_url']) {
+      if (response['git_url']) {
```
kevin-bates, line 162: This seems to introduce a lot of duplicated code.  SInce the presence of `git_url` is a subset of the presence of `run_url`, it seems like the original block (`if response['run_url']`) could still exist, with a sub-condition that checks for `git_url` to produce the git-related destination.

Also, is there something else in `response` (or elsewhere) that indicates the platform (e.g., 'airflow', 'kfp')?  If not, I think we should probably add some indicator so that we can remove the assumption that `git_url` is Airflow.

akchinSTC:
@lresende - not able to access the repo cause you guys dont have permissions via UI I think. 
Ill strip out the extra slash in the url context

elyra/pipeline/processor.py
```diff
@@ -88,8 +89,14 @@ async def export(self, pipeline, pipeline_export_format, pipeline_export_path, o
 
 
 class PipelineProcessorResponse(object):
-    def __init__(self, run_url='', object_storage_url='', object_storage_path=''):
+    def __init__(self,
```
kevin-bates, line 92: Along the lines of my previous comment - in which I asked whether the platform indicator is present in the response - I think it would make sense to make this an abstract base class and let each platform processor define a concrete subclass.

In that case, `AirflowProcessorResponse` would _add_ a _required_ `git_url` to the constructor, whereas `KfpProcessorResponse` would simply subclass the ABC `PipelineProcessorReponse`.  The ABC would then define an "abstract attribute" named, say, `platform`, and each of the concrete classes would set `platform` accordingly.

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth: I agree, we should nest the relevant new code inside the existing if block, I'll detail below

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -159,7 +159,44 @@ export class PipelineService {
...
+          <p>
+            Apache Airflow DAG has been pushed to the{' '}
+            <a
+              href={response['git_url']}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              GitHub Repository.
+            </a>
+            <br />
```
ajbozarth, line 175: You should just add this chunk of code in the current `dialogBody` wrapped in the `if (response['git_url'])`

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -159,7 +159,44 @@ export class PipelineService {
...
+            >
+              object storage
+            </a>
+            .
+          </p>
+        );
+      } else if (response['run_url']) {
         // pipeline executed remotely in a runtime of choice
         dialogTitle = 'Job submission to ' + runtimeName + ' succeeded';
         dialogBody = (
```
ajbozarth, line 202: eg for suggestion above, adding something like this (untested):
```
        dialogBody = (
          <p>
            {
              response['git_url'] ?
                (
                  <p>
                  Apache Airflow DAG has been pushed to the{' '}
                   <a
                     href={response['git_url']}
                     target="_blank"
                     rel="noopener noreferrer"
                   >
                     GitHub Repository.
                   </a>
                   <br />
                  </p>
                ) : null
            }
            Check the status of your job at{' '}
```

ajbozarth (changes requested):
some frontend review

packages/pipeline-editor/src/PipelineService.tsx
akchinSTC: updated

packages/pipeline-editor/src/PipelineService.tsx
akchinSTC: updated

packages/pipeline-editor/src/PipelineService.tsx
akchinSTC: Updated with `response`json  with `platform` field

elyra/pipeline/processor.py
akchinSTC: created new base class with suggested parameters and respective subclasses in each runtime processor

ajbozarth (approved):
The front end code look good, note that I haven't reviewed the backend code or tested this locally though

elyra/pipeline/processor.py
```diff
@@ -116,11 +126,9 @@ def object_storage_path(self):
         """
         return self._object_storage_path
 
+    @abstractmethod
     def to_json(self):
-        return {"run_url": self.run_url,
-                "object_storage_url": self.object_storage_url,
-                "object_storage_path": self.object_storage_path
-                }
+        raise NotImplementedError()
```
kevin-bates, line 131: I would leave this method as-is, not make it abstract, but add `"platform": self.type,` to the result.  The KFP subclass would then not need to implement `to_json()`, while the Airflow subclass would call super().to_json(), then add in its `git_url` value.

elyra/pipeline/processor_airflow.py
```diff
@@ -96,8 +96,8 @@ def process(self, pipeline):
                                    f"pipeline pushed to git: {github_url}",
                                    duration=(time.time() - t0_all))
 
-            return PipelineProcessorResponse(
-                # TODO - Add another field to return the url of the job in Airflow UI
+            return self.AirflowPipelineProcessorResponse(
```
kevin-bates, line 99: You shouldn't need `self.`
```suggestion
            return AirflowPipelineProcessorResponse(
```

elyra/pipeline/processor_airflow.py
```diff
@@ -275,3 +275,23 @@ def create_pipeline_file(self, pipeline, pipeline_export_format, pipeline_export
...
+        def type(self):
+            return self._type
+
+        def to_json(self):
+            return {"platform": self.type,
+                    "run_url": self.run_url,
+                    "git_url": self.git_url,
+                    "object_storage_url": self.object_storage_url,
+                    "object_storage_path": self.object_storage_path
+                    }
```
kevin-bates, line 297:
```suggestion
        def to_json(self):
            response = super().to_json()
            response['git_url'] = self.git_url
            return response
```

elyra/pipeline/processor_kfp.py
```diff
@@ -192,7 +192,7 @@ def process(self, pipeline):
                                    f"pipeline submitted: {api_endpoint}/#/runs/details/{run.id}",
                                    duration=(time.time() - t0_all))
 
-            return PipelineProcessorResponse(
+            return self.KfpPipelineProcessorResponse(
```
kevin-bates, line 195: Shouldn't need `self`

elyra/pipeline/processor.py
```diff
@@ -87,12 +87,22 @@ async def export(self, pipeline, pipeline_export_format, pipeline_export_path, o
...
 
-class PipelineProcessorResponse(object):
-    def __init__(self, run_url='', object_storage_url='', object_storage_path=''):
+class PipelineProcessorResponse(ABC):
+
+    _type = None
+
+    def __init__(self, run_url, object_storage_url, object_storage_path, **kwargs):
+
+        super(PipelineProcessorResponse, self).__init__(**kwargs)
```
kevin-bates, line 96: I recently learned the `super` calls can be simplified.
```suggestion
        super().__init__(**kwargs)
```
Hmm, do we need the keyword args entry?  Also, I'm not sure we need the call to super for the ABC actually.

Did you find the **kwargs was necessary?  I see you define the known parameters are regular args (*args), yet call the constructor using keyword-args, so I wonder if it's confused.

kevin-bates (commented):
Looking good, just had some suggestions and a question about what your findings were with kwargs.

elyra/pipeline/processor.py
akchinSTC: simplified, removed call to super and kwargs

akchinSTC:
@kevin-bates, thanks kevin, yeah didnt need the kwargs or super call. cleaned it up a bit.

kevin-bates (approved):
Thanks Alan.  Full disclosure: I didn't test this other than by inspection.

ptitzler:
Not introduced by this PR but also not resolved yet is the path issue: 
The following runtime config settings
```
https://api.github.com
akchinstc/test-repo
main
```
yield the following URL:

```
https://github.com//akchinstc/test-repo/tree/main
```

ptitzler (approved):
Working as expected for KFP and AA, aside from the URL issue I mentioned.

akchinSTC:
@ptitzler - added a fix for the extra slash in the generated URL

ptitzler:
Can you please add automated tests for `get_github_url`? I am sure we'll hit issues in the future once we see more URLs in the wild that'll break the current logic. This way we can just keep adding more tests when that happens.

PR #1369, opened 2021-03-04
Fix dark mode for metadata UI

ajbozarth:
This is a follow up to #1327 fixing various bugs as well as adding
dark mode support.

Dark Mode support for material-ui has now been abstracted to a new
ui-component and MetadataEditor has been updated to use it.

 
           in the file; or

           it.

### DIFF
--- a/packages/ui-components/src/JpThemeProvider.tsx
+++ b/packages/ui-components/src/JpThemeProvider.tsx
@@ -18,7 +18,7 @@ import { IThemeManager } from '@jupyterlab/apputils';
 
 import { createMuiTheme, ThemeProvider } from '@material-ui/core';
 
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 export interface IJpThemeProviderProps {
   children: React.ReactElement;
@@ -26,47 +26,47 @@ export interface IJpThemeProviderProps {
   updateWidget?: () => void;
 }
 
-export const JpThemeProvider = (
-  props: IJpThemeProviderProps
-): React.ReactElement => {
-  let isLight =
-    props.themeManager.theme &&
-    props.themeManager.isLight(props.themeManager.theme);
-
-  const updateTheme = (): void => {
-    const isLightUpdate =
-      props.themeManager.theme &&
-      props.themeManager.isLight(props.themeManager.theme);
-
-    if (isLightUpdate != isLight) {
-      isLight = isLightUpdate;
-      if (props.updateWidget) {
-        props.updateWidget();
-      }
-    }
-  };
-
-  updateTheme();
-
-  const lightTheme = createMuiTheme({
-    palette: {
-      type: 'light'
-    }
-  });
-
-  const darkTheme = createMuiTheme({
-    palette: {
-      type: 'dark'
-    }
-  });
-
-  if (props.themeManager) {
-    props.themeManager.themeChanged.connect(updateTheme);
+const lightTheme = createMuiTheme({
+  palette: {
+    type: 'light'
   }
+});
+
+const darkTheme = createMuiTheme({
+  palette: {
+    type: 'dark'
+  }
+});
+
+const isLightTheme = (themeManager: IThemeManager): boolean => {
+  return themeManager.theme && themeManager.isLight(themeManager.theme);
+};
+
+export const JpThemeProvider = ({
+  children,
+  themeManager,
+  updateWidget
+}: IJpThemeProviderProps): React.ReactElement => {
+  const [isLight, setIsLight] = useState(isLightTheme(themeManager));
+
+  useEffect(() => {
+    const updateTheme = (): void => {
+      setIsLight(isLightTheme(themeManager));
+      if (updateWidget) {
+        updateWidget();
+      }
+    };
+
+    updateTheme();
+
+    if (themeManager) {
+      themeManager.themeChanged.connect(updateTheme);
+    }
+  }, [themeManager, updateWidget]);
 
   return (
     <ThemeProvider theme={isLight ? lightTheme : darkTheme}>
-      {props.children}
+      {children}
     </ThemeProvider>
   );
 };
--- a/packages/ui-components/src/JpThemeProvider.tsx
+++ b/packages/ui-components/src/JpThemeProvider.tsx
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018-2021 Elyra Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


ajbozarth:
Note those those I tagged for review, this PR touches a lot of code from #1293 if you could make sure I haven't broken anything from that PR (I spent all afternoon making sure I didn't but odd are I missed something)

ajbozarth:
For reasons I am still unsure of, this PR makes the code editor in the MetadataEditor (used by code snippets) not load. This is (obviously) causing the tests to fail. I will look into why this is happening tomorrow

packages/ui-components/src/JpThemeProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  const darkTheme = createMuiTheme({
+    palette: {
+      type: 'dark'
+    }
+  });
+
+  if (props.themeManager) {
+    props.themeManager.themeChanged.connect(updateTheme);
```
bourdakos1, line 64: This should go into an effect hook

bourdakos1 (changes requested):

packages/ui-components/src/JpThemeProvider.tsx
ajbozarth: like this? 
```
  useEffect(() => {
    if (props.themeManager) {
      props.themeManager.themeChanged.connect(updateTheme);
    }
  });
```
I just googled it and it seems to the the function equivalent of componentDidMount, why should this be in an effect hook?

bourdakos1: you also need a dependency array otherwise this will run every render

bourdakos1:
```
useEffect(() => {
    if (themeManager) {
     themeManager.themeChanged.connect(updateTheme);
    }
  }, [themeManager]);
```

bourdakos1: this will make it so it only connects whenever themeManager or updateTheme change, does this have a disconnect function? because we should do that here as well

bourdakos1: I'll write up a quick demo with notes on how this works with hooks

bourdakos1:
```js
// The themes should be created outside of the component, otherwise they will be re-created
// every render which could be expensive.
const lightTheme = createMuiTheme({
  palette: {
    type: 'light'
  }
});

const darkTheme = createMuiTheme({
  palette: {
    type: 'dark'
  }
});

function isLightTheme(themeManager) {
  return themeManager.theme && themeManager.isLight(themeManager.theme);
}

function JpThemeProvider({ themeManager, children, updateWidget }) {
  // variables in a component should not be directly modified by side effects. If we want to 
  // modify a variable from an effect/callbacks it should be a state hook.
  // Note: It is normally frowned upon to derive state from props, because the can become 
  // stale. In this case it is okay because we update this value through an effect.
  const [isLight, setIsLight] = useState(isLightTheme(themeManager))

  useEffect(() => {
    // Any type of listeners are side effects and should be set up and torn down inside
    // an effect hook. Otherwise, you will register a listener every render and end up
    // a big leak

    // `updateTheme` is small and isn't called anywhere else so we can define it inside
    // of the effect so we don't need to add it to the dependency array.
    // It's important to note that if this is defined inside of the component and listed in
    // the dependency array the effect would be rerun on every render because a new 
    // function is created every render. You can avoid this by using a `callback` hook
    // which will memoize the function when possible 
    function updateTheme() {
      setIsLight(isLightTheme(themeManager))
      updateWidget()
    }

    if (themeManager) {
      themeManager.themeChanged.connect(updateTheme);
    }
    return () => {
      // disconnect when the when the component is unmounted/the dependency array has 
      // changed and a new connection has been made.
      themeManager.themeChanged.disconnect(updateTheme);
    }
  // This is the dependency array, anytime one of these change this effect will be ran. It will
  // also run when the component is first mounted.
  // Notice, that `updateWidget` is listed here, but not `isLightTheme` or `setIsLight`. This
  // is because any `setState` functions are guaranteed not to change as well as the function
  // defined outside of this component. However, `updateWidget` is a prop and could change 
  // from render to render. (There is an eslint rule that will automatically generate this array for you)
  }, [themeManager, updateWidget])

  return (
    <ThemeProvider theme={isLight ? lightTheme : darkTheme}>
      {children}
    </ThemeProvider>
  )
}
```

bourdakos1: Here is an example of `updateTheme` with the callback hook:
```
const updateTheme = useCallback(() => {
  setIsLight(isLightTheme(themeManager))
  updateWidget()
}, [themeManager, updateWidget])
```

bourdakos1: there's also the option to memoize the theme and use it inside of the component:
```
const theme = useMemo(() => {
  return createMuiTheme({
    palette: {
      type: isLight ? 'light' : 'dark'
    }
  });
}, [isLight])
```

ajbozarth:
I have made some changes to match @bourdakos1 comments above but I was dealing with rendering race conditions all afternoon. @bourdakos1 I'll ping you tomorrow and see if we can hop on a Webex and you can help me try to figure it out.

If I cant figure it out by early afternoon (pacific time) I’m going to partially roll back my work so that the part of the PR that fixes dark mode for the drop down can at least be merged since that part already works fine. That will unblock the 2.1 release, and then we can deal with moving the MetadataEditor to use the JpThemeProvider component in a follow up PR since that’s where the race condition is happening.

bourdakos1 (approved):
This looks good to me

ajbozarth:
@lresende once tests pass this should be good to merge (unless you want more reviews first)

packages/ui-components/src/JpThemeProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+export interface IProps {
+  themeManager: IThemeManager;
+}
+
+const isLightTheme = (themeManager: IThemeManager): boolean => {
+  // Default to light theme
+  return themeManager?.theme ? themeManager.isLight(themeManager.theme) : true;
+};
+
+export const JpThemeProvider: React.FC<IProps> = ({
```
lresende, line 32: I would like to avoid introducing `Jp`, 'Lab`, or `Elyra`prefixes to class names or similar suffixes.

packages/ui-components/src/JpThemeProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  themeManager,
+  children
+}) => {
+  const [isLight, setIsLight] = useState(isLightTheme(themeManager));
+
+  const theme = useMemo(() => {
```
lresende, line 38: to promote usage, could we add some inline docs on what these do?

ajbozarth: @bourdakos1 you understand this better, could you add an inline comment?

ajbozarth: done

packages/ui-components/src/JpThemeProvider.tsx
ajbozarth: done

lresende (approved):

PR #1385, opened 2021-03-08
Add warning before submitting notebook with unsaved changes

### LINKED ISSUE #1356: Prompt user for saving a modified file when submitting script/notebook
**Describe the issue**
Prompt user for saving a modified file when submitting script/notebook

**To Reproduce**
1. Create a bogus script/notebook and submit
2. The run will fail
3. Fix the file, and resubmit
4. It will still fail unless you explicitly save the file

**Expected behavior**
I would expect the application to warn me to save or pick up the `in-memory` contents as we do with the `pipeline` file. 

This is probably also an issue when submitting a pipeline, that has artifacts open on the editor, and pending modifications, but I am ok not considering that in the scope of this issue.

marthacryan:
Fixes #1356. Adds a warning if there are unsaved changes in a notebook before submitting. If the user presses "OK", the changes are saved and the user moves to the submission dialog. If the user presses "Cancel", the changes will not be saved, but the user will still move to the submission dialog, and the notebook will be submitted without the in-memory changes. 
image

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -891,9 +891,9 @@ export class PipelineEditor extends React.Component<
 
     if (this.widgetContext.model.dirty) {
       const dialogResult = await showDialog({
-        title: 'This pipeline has unsaved changes. Save before submitting?',
+        title: 'This pipeline contains unsaved changes. To submit the pipeline the changes need to be saved.',
         buttons: [
-          Dialog.cancelButton({ label: "Don't submit" }),
+          Dialog.cancelButton(),
           Dialog.okButton({ label: 'Save and Submit' })
         ]
       });
@@ -1298,9 +1298,9 @@ export class PipelineEditor extends React.Component<
 
     if (this.widgetContext.model.dirty) {
       const dialogResult = await showDialog({
-        title: 'This pipeline has unsaved changes. Save before submitting?',
+        title: 'This pipeline contains unsaved changes. To submit the pipeline the changes need to be saved.',
         buttons: [
-          Dialog.cancelButton({ label: "Don't submit" }),
+          Dialog.cancelButton(),
           Dialog.okButton({ label: 'Save and Submit' })
         ]
       });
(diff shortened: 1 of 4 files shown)


lresende:
I would say that, if the user hits cancel, it should stop and not move to the submission dialog.
Also, we need the same thing for the `submit script` path

ptitzler:
> I would say that, if the user hits cancel, it should stop and not move to the submission dialog. 

I agree, even if this might not be consistent with the behavior of other editors in JupyterLab. If a user was allowed to run a notebook/script/pipeline without saving the source first, one would not necessarily know which exact version was used.

> Also, we need the same thing for the submit script path

And the run and export pipeline workflow to maintain consistency. In neither scenario is the user currently required to save changes first.

... and would lobby to change the button labels to remove ambiguity. `OK` is typically used to acknowledge something, not to make a decision. Compare, for example "OK"/"Cancel" with "Save and submit" / "Don't submit".

marthacryan:
Ok sounds good - so to be clear, I should add this dialog to the submit script and the run and export pipeline dialogs, and change the buttons to say "Save and submit" / "Don't submit"?

marthacryan:
New dialog text:
image

ptitzler:
I believe that works even though the `Don't submit` button label isn't really answering  directly the "Save before submitting?" question

lresende:
> I believe that works even though the `Don't submit` button label isn't really answering directly the "Save before submitting?" question

That's one reason I was thinking of leaving it as `Cancel` which sort of means, cancel the operation (e.g. what is described on the location of the ok button).

ptitzler:
Yes, but `Cancel` does not convey what is canceled. If a user needs to click a button to find out what it does we have a usability issue. 

Option 1: Choose a more descriptive label, e.g.  `Cancel Submission`?
Option 2: Change the narrative by changing the info text. "This XXX contains unsaved changes. To submit the XXX the changes need to be saved." `<cancel> <save and submit>`

ajbozarth:
I like @ptitzler Option 2, I think it has the clearest text and buttons

marthacryan:
@lresende If you confirm I can go ahead and change it to option 2

lresende (approved):

PR #1394, opened 2021-03-09
Validate resource requests for zero or negative values

### LINKED ISSUE #1390: Resource values (cpu, memory, gpu) should be validated
**Describe the issue**
When editing a node's properties, I am able to enter negative values for CPU and GPU. Note that I can enter a negative value for memory, but it is caught _on the server_.  Also, I can enter invalid values (32 GPU) when there aren't any, and the pipeline's execution will succeed w/o error.

**To Reproduce**
1. Open a pipeline's node properties and enter -8 for the CPU.  Although there is a warning, the value is permitted.  Also enter a value of some large number of GPUs (e.g., 37)
2. Execute the pipeline successfully.

I tried this on KFP and local.  Did not try on Airflow, but I suspect the same issue.

**Expected behavior**
At a minimum, invalid values like negative numbers should be caught on the client.  I understand it might be harder (or impossible) to validate actual numbers of CPUs and GPUs since a runtime may be involved.

akchinSTC:
Fixes #1390

 
           in the file; or

           it.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -15,5 +15,5 @@
 #
 import pytest
 
-#pytest_plugins = ["jupyter_server.pytest_plugin"]
+pytest_plugins = ["jupyter_server.pytest_plugin"]
 
--- a/conftest.py
+++ b/conftest.py
@@ -15,5 +15,5 @@
 #
 import pytest
 
-pytest_plugins = ["jupyter_server.pytest_plugin"]
+#pytest_plugins = ["jupyter_server.pytest_plugin"]
 
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -70,9 +70,9 @@ class Operation(object):
         self._inputs = inputs or []
         self._outputs = outputs or []
         self._parent_operations = parent_operations or []
-        self._cpu = cpu
-        self._gpu = gpu
-        self._memory = memory
+        self._cpu = self.validate_resource_value("CPU", cpu)
+        self._gpu = self.validate_resource_value("GPU", gpu)
+        self._memory = self.validate_resource_value("MEMORY", memory)
 
     @property
     def id(self):
@@ -141,6 +141,13 @@ class Operation(object):
                     else:
                         print(f"Could not process environment variable entry `{nv}`, skipping...")
         return envs
+
+    def validate_resource_value(self, name: str, value):
+        if value is None or int(value) > 0:
+            return value
+        else:
+            raise ValueError(f"Zero or negative values are not permitted for {name} "
+                             f"for operation with {os.path.basename(self.filename)}")
 
     @property
     def inputs(self):
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 import os
+import sys
+
 from typing import Dict, Optional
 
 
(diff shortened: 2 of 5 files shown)


elyra/pipeline/pipeline.py
```diff
@@ -142,6 +142,13 @@ def env_vars_as_dict(self, logger: Optional[object] = None) -> Dict:
                         print(f"Could not process environment variable entry `{nv}`, skipping...")
         return envs
 
+    def validate_resource_value(self, name: str, value):
+        if value is None or int(value) > 0:
```
lresende, line 146:
```suggestion
        if value and int(value) > 0:
```

bourdakos1:
It should not let you run the pipeline unless you have positive resources now

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -1172,6 +1172,15 @@ export class PipelineEditor extends React.Component<
     ) {
       validationErrors.push('no runtime image');
     }
+    if (node.app_data.cpu <= 0) {
+      validationErrors.push('CPU must be greater than 0');
+    }
+    if (node.app_data.gpu <= 0) {
```
kevin-bates, line 1178: Seems like 0 GPU should be allowed.  The same might go for CPU, although I don't really know if something that uses GPUs can exclude the use of CPUs (I kinda doubt it).

bourdakos1: oh yea that makes sense, I was just going off of what the ui controls were validating for, I'll update those too

so, like this?
cpu > 0
memory > 0 
gpu >= 0

kevin-bates: Those make sense to me - thanks.

elyra/pipeline/pipeline.py
```diff
@@ -70,9 +70,9 @@ def __init__(self, id, type, classifier, filename, runtime_image, memory=None, c
         self._inputs = inputs or []
         self._outputs = outputs or []
         self._parent_operations = parent_operations or []
-        self._cpu = cpu
-        self._gpu = gpu
-        self._memory = memory
+        self._cpu = self.validate_resource_value("CPU", cpu)
+        self._gpu = self.validate_resource_value("GPU", gpu)
```
kevin-bates, line 74: Regarding my earlier comment relative to the front-end, it seems like we should tolerate a zero value for GPU.  Sorry, this complicates the check and messaging.

conftest.py
```diff
@@ -15,5 +15,5 @@
 #
 import pytest
 
-pytest_plugins = ["jupyter_server.pytest_plugin"]
+#pytest_plugins = ["jupyter_server.pytest_plugin"]
```
kevin-bates, line 18: Is this causing issues?  I _think_ it's necessary for tests that use fixtures from jupyter_server.  The metadata tests use a number of these.  That said, I think I've found this to be unnecessary, but that never made sense to me - so I'm curious if you found some adverse to defining this.  (The pytest fixture stuff is still a bit of magic to me.)

akchinSTC: My mistake, I left this in there. I've had to comment this out when testing locally due to an error being thrown when using pytest. Let me re-enable it. 
I'm kind of in the same boat you are, this file has been unnecessary for me but needed for others.

packages/pipeline-editor/src/properties.json
```diff
@@ -298,7 +298,7 @@
               "condition": {
                 "parameter_ref": "gpu",
                 "op": "greaterThan",
-                "value": 0
+                "value": -1
```
kevin-bates, line 301: The `validation.fail_message.message.default` field above should read to the effect of "GPU must be greater than or equal to 0."

bourdakos1: oh yep, good catch

kevin-bates (approved):
Thanks @akchinSTC and @bourdakos1!

lresende (approved):

PR #1402, opened 2021-03-09
Remove error label when user fixes invalid field

### LINKED ISSUE #1388: Metadata editor leaves field in error state following valid entry
**Describe the issue**
When triggering a required value error in the pipeline runtime editor, the field remains in an error state even after entering data.  This makes it confusing as to which fields still must be addressed.

**To Reproduce**
1. Go to the runtimes editor and enter a name.
2. Click on save and should be presented with required value errors on several fields that require a value
3. Enter "valid" information that should suppress the error state.
4. Can't easily determine which of the remaining error fields still need to be addressed.

**Screenshots or log output**
image

**Expected behavior**
I had expected the field error indicator to be removed once a "valid" entry has been entered.

**Deployment information**
 - Elyra version: 2.1.0rc2
 - Operating system: macos
 - Installation source: PyPi

marthacryan:
Fixes #1388. Removes the error label when user fixes invalid field. 
 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -59,7 +59,9 @@ interface ICodeBlockProps {
   initialValue: string;
   language: string;
   onChange?: (value: string) => any;
-  error: boolean;
+  initialError: boolean;
+  label: string;
+  required: boolean;
 }
 
 const CodeBlock: React.FC<ICodeBlockProps> = ({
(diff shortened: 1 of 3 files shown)


lresende (approved):

lresende:
Tested the metadata editor by entering invalid/valid values and all seem to be showing/removing the error messages based on the correct values.

ajbozarth:
Code looks good to me, I haven't had a chance to test locally though since I'm in the middle of testing 1404

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -59,14 +59,26 @@ interface ICodeBlockProps {
...
+  onChange,
+  initialError,
+  label,
+  required
 }) => {
+  const [error, setError] = React.useState(initialError);
+
+  React.useEffect(() => {
+    setError(initialError);
+  }, [initialError]);
```
bourdakos1, line 80: this shouldn't be needed

bourdakos1 (approved):

bourdakos1:
I updated `DropDown`, `TextInput` and `CodeBlock` to be uncontrolled components, because we aren't using react state in the parent component so they can't be real controlled components.

As an uncontrolled components each form control handles its own state, and notifies the parent of any changes via "onChange". The parent component has no control of the form controls state and can only give it an initial value when the control is first mounted.

The `CodeBlock` component relies on `language` from one of the `DropDown` instances, so the editor must be refreshed when `language` changes. I updated `setFormFocus` to check if any of the form controls had focus before giving focus to the `display_name` input. This stops `display_name` from stealing focus as you type a language into `DropDown`

lresende:
@bourdakos1 looks like your change might have caused some component ids/types to change and is affecting the `code-snippet` tests:

image

PR #1410, opened 2021-03-10
Add GitHub link to Runtimes UI

### LINKED ISSUE #1335: Metadata Editor / Apache Airflow configuration: display custom property
Currently the runtime configuration metadata editor displays two links, based on the property values of the metadata instance:
 - a link to the runtime's GUI (`api_endpoint` property value)
 - a link to cloud object storage (`cos_endpoint` property value)

image

For Apache Airflow runtimes there is a third GUI that we need to expose: a link to the Github location where the DAGs for the runtime instance are stored. The URL is not directly stored but can be assembled from property values as follows:
`modified_github_api_endpoint` + `/` + `github_repo` +  `/tree/` + `github_branch`  

Note that  `modified_github_api_endpoint` is the value of `github_api_endpoint` with the host name removed, e.g. `https://api.github.com` => `https://github.com`

Example: `https://github.com/ptitzler/airflow-dags/tree/main`

ajbozarth:
Fixes #1335

This adds a `GitHub DAG Repository` link to the Runtimes UI for airflow runtimes
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/RuntimesWidget.tsx
+++ b/packages/pipeline-editor/src/RuntimesWidget.tsx
@@ -35,13 +35,11 @@ const addTrailingSlash = (url: string): string => {
 };
 
 const getGithubURLFromAPI = (apiEndpoint: string): string => {
-  // For enterprise instances the api is located at <hostname>/api/
+  // For Enterprise Server the api is located at <hostname>/api/
   let baseURL = new URL(apiEndpoint).origin;
 
-  // For github.com the api endpoint is located at api.github.com
-  if (baseURL.includes('api.github.com')) {
-    baseURL = baseURL.replace('api.', '');
-  }
+  // For Github.com and Github AE the api is located at api.<hostname>
+  baseURL = baseURL.replace('api.', '');
 
   return addTrailingSlash(baseURL);
 };
@@ -81,15 +79,16 @@ class RuntimesDisplay extends MetadataDisplay<
         getGithubURLFromAPI(metadata.metadata.github_api_endpoint) +
         metadata.metadata.github_repo +
         '/tree/' +
-        metadata.metadata.github_branch;
+        metadata.metadata.github_branch +
+        '/';
       githubRepoElement = (
         <span>
-          <br />
-          <br />
           <h6>{metadata_props.github_repo.title}</h6>
           <a href={githubRepoUrl} target="_blank" rel="noreferrer noopener">
             {githubRepoUrl}
           </a>
+          <br />
+          <br />
         </span>
       );
     }
@@ -104,6 +103,7 @@ class RuntimesDisplay extends MetadataDisplay<
         </a>
         <br />
         <br />
+        {githubRepoElement}
         <h6>
           {metadata_props
             ? metadata_props.cos_endpoint.title
@@ -112,7 +112,6 @@ class RuntimesDisplay extends MetadataDisplay<
         <a href={cosEndpoint} target="_blank" rel="noreferrer noopener">
           {cosEndpoint}
         </a>
-        {githubRepoElement}
       </div>
     );
   }
--- a/packages/pipeline-editor/src/RuntimesWidget.tsx
+++ b/packages/pipeline-editor/src/RuntimesWidget.tsx
@@ -28,12 +28,31 @@ import React from 'react';
 
 import { PipelineService, RUNTIMES_NAMESPACE } from './PipelineService';
 
+const RUNTIMES_METADATA_CLASS = 'elyra-metadata-runtimes';
+
+const addTrailingSlash = (url: string): string => {
+  return url.endsWith('/') ? url : url + '/';
+};
+
+const getGithubURLFromAPI = (apiEndpoint: string): string => {
+  // For enterprise instances the api is located at <hostname>/api/
+  let baseURL = new URL(apiEndpoint).origin;
+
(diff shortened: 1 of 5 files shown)


ajbozarth:
Note I'm not 100% sure `addTrailingSlash` and `getGithubURLFromAPI` belong where they are or should be moved to some util file elsewhere.

ajbozarth:
Screenshot:
Screen Shot 2021-03-10 at 3 16 58 PM

packages/pipeline-editor/src/RuntimesWidget.tsx
```diff
@@ -28,12 +28,31 @@ import React from 'react';
 
 import { PipelineService, RUNTIMES_NAMESPACE } from './PipelineService';
 
+const RUNTIMES_METADATA_CLASS = 'elyra-metadata-runtimes';
+
+const addTrailingSlash = (url: string): string => {
+  return url.endsWith('/') ? url : url + '/';
+};
+
+const getGithubURLFromAPI = (apiEndpoint: string): string => {
```
ptitzler, line 37: The  logic to derive the GUI URL needs to mirror https://github.com/elyra-ai/elyra/blob/master/elyra/util/git.py#L89 to handle GitHub Enterprise URL patterns that we are aware of.

packages/pipeline-editor/src/RuntimesWidget.tsx
```diff
@@ -28,12 +28,31 @@ import React from 'react';
...
+  // For enterprise instances the api is located at <hostname>/api/
+  let baseURL = new URL(apiEndpoint).origin;
+
+  // For github.com the api endpoint is located at api.github.com
+  if (baseURL.includes('api.github.com')) {
```
ptitzler, line 42: See above comment. An example for a GHE URL is `https://api.github.ibm.com`, which wouldn't be correctly derived by the logic I initially outlined in the issue.

ptitzler:
If possible, I believe it would be better if the GH DAG repository entry is displayed before the COS URL. This way 
 - the sequence would mimic the property order in the configuration (AA, GH, COS)
 - the AA and GH URLs are semantically more closely related - the two URLs are something an AA user would be familiar with

Tested the code and it generally appears to work fine.

packages/pipeline-editor/src/RuntimesWidget.tsx
ajbozarth: I was following the docs (https://docs.github.com/en/rest/reference/enterprise-admin), but I missed that there is another format than just the GitHub.com and Enterprise Server formats I covered. It seems IBM uses GitHib AE, I'll update this and remove the if statement and always remove `api.`

ajbozarth:
> If possible, I believe it would be better if the GH DAG repository entry is displayed before the COS URL.

Moved it

> An example for a GHE URL is https://api.github.ibm.com

Updated to handle GitHub AE instances, there is now an edge case where this will behave oddly (if a Enterprise Server puts `api.` in their GHE hostname), but that edge case exists in our backend as well.

packages/pipeline-editor/src/RuntimesWidget.tsx
```diff
@@ -28,12 +28,29 @@ import React from 'react';
...
+};
+
+const getGithubURLFromAPI = (apiEndpoint: string): string => {
+  // For Enterprise Server the api is located at <hostname>/api/
+  let baseURL = new URL(apiEndpoint).origin;
+
+  // For Github.com and Github AE the api is located at api.<hostname>
+  baseURL = baseURL.replace('api.', '');
+
+  return addTrailingSlash(baseURL);
```
lresende, line 44: URL (from nodejs) has the option to handle join. Could we use that instead of implementing our own code for that?

ajbozarth: We could but it wouldn't really change the code, `addTrailingSlash` conditionally adds `/` if it's not already there. So if we used `URL.join` we would need to strip the `/` off first. In the end the amount of code wouldn't change, it would just be different. I'm ok either way, I did consider it, but didn't bother when I realized it wouldn't think the code footprint

lresende: The [docs](https://nodejs.org/api/url.html#url_the_whatwg_url_api) seem to imply no striping is needed.

const myURL = new URL('/foo', 'https://example.org/');
// https://example.org/foo

ptitzler:
> Updated to handle GitHub AE instances, there is now an edge case where this will behave oddly (if a Enterprise Server puts `api.` in their GHE hostname), but that edge case exists in our backend as well.

Understood. We'll fix it as we learn more about valid patterns that we need to handle based on what customers might report in the future.

ptitzler (approved):
LGTM. Tested
 - create KFP runtime conf (no change in behavior)
 - edit KFP runtime  conf (no change in behavior)
 - create AA runtime conf (OK)
 - edit AA runtime conf (OK)

packages/pipeline-editor/src/RuntimesWidget.tsx
ajbozarth: Copying what I mentioned in a Webex here: in general trailing slash is not an issue, but we intentionally added it in our code to fix an issue related to kfp instances. I've just updated our code to add it for all our links here

lresende (approved):

Issue #1412, opened 2021-03-11
Resource values that require values > 0 treat null as 0 and produce error


kevin-bates:
**Describe the issue**
When _removing_ a value for CPU or Memory, the node border will show it in an error state even though no value should be considered valid (and was valid when the node was originally added).

**To Reproduce**
1. Bring up a node's properties and set a valid value for CPU or Memory (e.g., 4) and close the properties
2. The node border is blue and no errors
3. Bring up the node's properties and _delete_ the previously entered value, leaving nothing in that portion of the dialog, and close the properties.
4. Note the error state of the node and, when hovering, the message indicating the value must be greater than 0.
5. Since GPU allows >= 0 as valid, this is not an issue there.

**Screenshots or log output**
![image](https://user-images.githubusercontent.com/22599560/110716132-664b2400-81bb-11eb-8bf6-5d8d3db4cd81.png)

**Expected behavior**
Null or empty values should be treated as _no_ value and not as zero (which is invalid).

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 2.1.0rc3
 - Operating system: macos
 - Installation source: PyPI
 - Deployment type: local installation

bourdakos1:
fun fact, in javascript:
`null == 0` is `false`
`null < 0` is `false`
`null > 0` is `false`
but:
`null <= 0` is `true`
`null >= 0` is `true`

PR #1413, opened 2021-03-11
Fix resource values that require values > 0 from treating null as <= 0

bourdakos1:
fixes: #1412 

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -1197,13 +1197,13 @@ export class PipelineEditor extends React.Component<
     ) {
       validationErrors.push('no runtime image');
     }
-    if (node.app_data.cpu <= 0) {
+    if (node.app_data.cpu != null && node.app_data.cpu <= 0) {
       validationErrors.push('CPU must be greater than 0');
     }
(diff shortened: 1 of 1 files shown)


bourdakos1:
woops, how did this happen

akchinSTC:
Built and tested on master. 
- setting saving cpu and memory at 0 or negative value, results in error 
- setting GPU at negative value, results in error, zero value allowed 
- Also removal of the value (setting to null in file) does not throw error. 
image
image

lresende:
@bourdakos1 could you also take a look at [cleanNullProperties()](https://github.com/elyra-ai/elyra/blob/de5ffbf9f476127c45fd7bcaadab6bf11f44ac7b/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L862) and if we need to do something similar to #1406. 

I know we are checking for nulls now, but I think nulls should never reach a saved and/or submitted pipeline.

bourdakos1:
It looks like we only strip nulls when submitting/exporting not saving/validating

lresende:
> It looks like we only strip nulls when submitting/exporting not saving/validating

Could we fix that as well?

bourdakos1:
Is that an issue?

PR #1427, opened 2021-03-13
Propagate operation input/output in sorted way

### LINKED ISSUE #1422: Pipeline Editor should inherit assets (files) output from upstream stages without re-defining them
I'm developing a pipeline, some stages acting on a csv, some on image data, so if stage x defines metadata.csv as output, stage x+1 doesn't stage x+2 can't read metadata.csv because it is not explicitly declared so in stage x+1

Let's continue the example with metadata.csv: If stage x defines it as output, stage x+1 not, it should be available for stage x+2

Otherwise on has to keep track of each and every file any upstream stage might define as output

lresende:
Fixes #1422
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -20,6 +20,7 @@ import time
 
 from abc import ABC, abstractmethod
 from elyra.metadata import MetadataManager
+from elyra.pipeline.parser import Pipeline, Operation
 from elyra.util.cos import CosClient
 from elyra.util.archive import create_temp_archive
 from elyra.util.path import get_expanded_path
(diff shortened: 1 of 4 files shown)


elyra/pipeline/processor.py
```diff
@@ -187,6 +188,57 @@ def log_pipeline_info(self, pipeline_name: str, action_clause: str, **kwargs):
 
             self.log.info(f"{self._type} '{pipeline_name}'{op_clause} - {action_clause} {duration_clause}")
 
+    @staticmethod
+    def _propagate_operation_inputs_outputs(pipeline: Pipeline, sorted_operations) -> None:
```
kevin-bates, line 192: This could use a type hint.  Perhaps `sorted_operations: List[Operation]`?  You'd need `from typing import List`.

Or you could probably just go with `sort_operations: list`.

kevin-bates: Actually, whatever is used should match the return value of `_sort_operations()`. :smile:

kevin-bates (commented):
Very nice cleanup!  Just the typing comment.

elyra/pipeline/processor.py
```diff
@@ -187,6 +188,57 @@ def log_pipeline_info(self, pipeline_name: str, action_clause: str, **kwargs):
...
+            for parent_operation_id in operation.parent_operations:
+                parent_operation = pipeline.operations[parent_operation_id]
+                if parent_operation.inputs:
+                    parent_io.update(parent_operation.inputs)
+                if parent_operation.outputs:
+                    parent_io.update(parent_operation.outputs)
+
+            if parent_io:
+                parent_io.update(operation.inputs)
+                operation.inputs = parent_io
```
kevin-bates, line 210: Should this get "cast" back as a list - `list(parent_io)`?  This might the cause of the test failure.

kevin-bates (approved):
Nice fix.  Thanks Luciano.

Issue #1428, opened 2021-03-13
Material UI style specificity issues


bourdakos1:
Since we have multiple instances of `material-ui` each one of them injects styles into the head without awareness of what the others are injecting.

Each instance only injects the styles that are needed at that moment, so you can get into weird states like this:
One instance  of `material-ui` renders a `MenuItem` which uses both `Button` and `MenuItem` styles, so it injects `Button` styles first and `MenuItem` second so that the `MenuItem` styles override `Button`. 
but then the other instance of `material-ui` is only rendering a `Button` so it injects styles for `Button` and if that happens after `MenuItem` then you end up with  `Button` > `MenuItem` > `Button` with `Button` incorrectly overriding `MenuItem` styles.

There’s a few possible fixes:
- Wrap all of JupyterLab inside of a `StylesProvider` component (the best solution, if possible)
- have a classname generator with `disableGlobal` set to `true` (This will make all the Mui style names non-deterministic, so we won’t be able to override styles via css, but I believe we can override them via a theme)
- create our own classname generator that allows us to create deterministic Mui classes, but with each mui instance prefixed with something custom

bourdakos1:
relevant documentation on generating class names: https://material-ui.com/styles/api/#creategenerateclassname-options-class-name-generator

kevin-bates:
@bourdakos1 - having sat in the initial debug session for this, this analysis is really fantastic!  Thank for you the clear description (especially for noobs like me!).

> Wrap all of JupyterLab inside of a StylesProvider component (the best solution, if possible)

Would the _wrapping_ be something we (elyra) do - as specific consumers of the Lab front-end or is there something that can (and should) be done for Lab extensions in general (and perhaps provided externally)?  Just trying to understand if this would help the Lab community.  (Note: I may not understand your response. :smile: )

bourdakos1:
After reflecting a bit more on this, I don't think _wrapping_ would be a good solution. First, I don't think it's possible because I assume the react widgets in jupyter are all their own react roots so a provider wouldn't be able to pass context to them. Also, wrapping everything in a provider could cause issues for other react widget extensions that are attempting to use material-ui. 

However, I believe we can get the intended effect of wrapping without actually wrapping. If I understand correctly, as long as there is only a single instance of a classname generator across all of our extensions, things should work as expected. 

[...]

PR #1434, opened 2021-03-15
Fix Material UI style specificity issues in Metadata Editor

bourdakos1:
TODO:
- [x] remove hardcoded mui css
- [x] update tests that target `Mui-xxx` to `elyra-Mui-xxx` (we should avoid using classnames to target, but that's another issue)

fixes: #1428 
 
           in the file; or

           it.


bourdakos1:
cypress just bumped from 6.6.0 to 6.7.0 like 8 minutes ago and the package isn't actually available yet, so that's why tests are failing 🙃

ajbozarth (approved):
I fought hard with @lresende about not naming it `ThemeProvider`, I greatly dislike overloading classnames and feel that it can cause confusion and mixups. But if you believe it is better react practice to overload the name then I'll relent on this.

Otherwise I tested this locally and it looks good. I'll leave it to @marthacryan to confirm this doesn't revert #1425

bourdakos1:
Yea, I normally prefer not to overload classnames, but I think in this case there's not really a great alternative

ajbozarth:
> Yea, I normally prefer not to overload classnames, but I think in this case there's not really a great alternative

@lresende and I went back and forth on names offline for a long while before finally setting on `ThemeComponent` as a decent middle ground given the best name would be to overload `ThemeProvider`.

I would rather leave it as `ThemeComponent` than overloading `ThemeProvider` if you don't have strong opinions on the name.

bourdakos1:
I think it definitely needs to have `Provider` at the end and we shouldn't use `Component` as part of a react component's name, because everything is a component

bourdakos1:
I think `ThemeProvider` is kind of the expected name for theme providers

PR #1435, opened 2021-03-15
Add R Editor

### LINKED ISSUE #1263: Add Elyra support for R editor
New feature

Make `Python Editor` a `Script Editor` and add support for R scripts

- Rename `Python Editor` to `Script Editor`
- Refactor editor code to separate UI from controller (ScriptEditorController, REditor, PythonEditor)
- Kernel dropdown displays R kernels
- Add R language server installation (make sure LSP extension features  work on R files)
- Add command to create new R file from Launcher and Main menu
- Update `SubmitScriptButtonExtension` for any python specific references (R support for pipelines reference: #1418)
- Update documentation
- Update integration tests

image

lresende:
Fixes #1263 

To install r kernel:
`conda install r-irkernel `

Tasks:
- [x]  Rename `Python Editor` extension to `Script Editor` extension
- [x]  Rename `PythonRunner` to `ScriptRunner`
- [x]  Refactor editor code to separate UI from controller (create `ScriptController` class)
- [x]  Implement REditor class
- [x]  Add command to create new R file from Launcher and Main menu
- [x]  Kernel dropdown displays R kernels (refactor `KernelDropDown`)
- [x] Update widget on kernel dropdown value change
- [x] Handle no kernel case
- [x]  Implement a generic `ScriptEditor` class
- [x] Update R logo (merge from master - fix icon bug when r editor is open)
- [X]  Update SubmitScriptButtonExtension (merge pipeline support for R from master)
- [x]  Update integration tests
- [x] Update package name to `script-editor-extension` and all its references
- [x]  Document R language server installation - LSP extension features for R files (may go in elyra conda-forge recipe in the future)
- [x]  Update documentation

May be done on a follow up PR:
- Adjust Code Snippet insert method & drag and drop warning
- Update conda recipe for `script-editor-extension`
- Fix linter disable line on` KernelDropDown`
- Make python/r editor children classes of `ScriptEditor`

Screenshots:
image
image

           in the file; or

           it.


packages/python-editor/src/KernelDropdown.tsx
```diff
@@ -16,128 +16,73 @@
...
+  // the change events there as well, but I know this isn't always possible
+  // alongside jupyter.
+  useImperativeHandle(select, () => ({
+    getSelection: (): string => selection
+  }));
+
+  return (
+    <select
+      className={KERNEL_SELECT_CLASS}
+      onChange={(e): void => setSelection(e.target.value)}
```
karlaspuldaro, line 55: We need to update the editor widget with a change in kernel selection, so the runner can start the correct session. For now it always uses the `kernelName` set on init state.

ptitzler:
```
$ jupyter kernelspec list
Available kernels:
  python3    /opt/anaconda3/envs/elyra_dev/share/jupyter/kernels/python3
```

A message should be displayed if no R kernel is defined and the user should  be prevented from running anything. Currently one can run an "R script" (which actually seems to be treated as a Python script):

```
print("Hello")
```

yields
```
[ ] Hello
```

karlaspuldaro:
@ptitzler Good catch. Looks like JupyterLab Session (which is what creates a kernel and manages its lifecycle) is using the default kernel (python) in case the kernel name is falsy in the `startNew` step.
We can add a handler for that case to the dropdown display and the run button to be disabled.

ptitzler:
> We can add a handler for that case to the dropdown display and the run button to be disabled.

Is there a way to issue a warning (e.g. "This script cannot be run because no supporting kernel is defined." or whatever the current message text is if there's no Python kernel) if the user clicks on the run button instead of disabling the button? The reason why I'm asking is that a user wouldn't necessarily know why the button is disabled and might assume it is a bug.

karlaspuldaro:
@ptitzler Yes, I am adding a dialog with a msg for the run button click (I can edit the msg as your suggestion above)
image
Also in that case the console output shouldn't open, I'm fixing that now.

ajbozarth:
Now that #1452 is merged this can be rebased and `rKernelIcon` references can be replaced with `rIcon` references

packages/python-editor/src/KernelDropdown.tsx
```diff
@@ -16,128 +16,86 @@
...
-  private async getKernelSPecs(): Promise<void> {
-    const specs: KernelSpec.ISpecModels = await this.props.runner.getKernelSpecs();
-    this.filterPythonKernels(specs);
-
-    // Set kernel to default
-    this.updateKernel(specs.default);
-
-    this.createOptionElems(specs);
-    this.setState({ kernelSpecs: specs });
+// eslint-disable-next-line react/display-name
```
ajbozarth, line 40: afaik you shouldn't need this since you're using an arrow function

packages/python-editor/src/KernelDropdown.tsx
```diff
@@ -16,128 +16,86 @@
...
+  ) {
     super();
-    this.runner = runner;
-    this.updateKernel = updateKernel;
   }
 
   render(): React.ReactElement {
     return (
-      <DropDown {...{ runner: this.runner, updateKernel: this.updateKernel }} />
+      <div>
```
ajbozarth, line 96: any particular reason you're wrapping this in a `div`?

ajbozarth (commented):
This is looking good code-wise, I've also done some locally testing while working on the icon bug, but nothing intensive.

My biggest hangup would be the handful of `if python else r` lines in the code, I feel like having the edit default to r isn't the best, if possible I think we should try to make a `if python else if r else other` scenario where we handle the other case properly, either as an error or otherwise. This is how I wrote my icon fix (at least as it is now, still working on it)

ajbozarth:
I've pushed a fix for the R icon display issue. My fix uses deprecated functionality in lumino to bypass a lumino bug that I've filed here: https://github.com/jupyterlab/lumino/issues/160

packages/python-editor/src/KernelDropdown.tsx
karlaspuldaro: You're right, that can be removed

packages/python-editor/src/KernelDropdown.tsx
karlaspuldaro: deleting that line I get this error:
```
  40:18  error  Component definition is missing display name  react/display-name

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
```

ajbozarth: hmm that's not the error I thought you were getting, seems like a react thing, @bourdakos1 what is the correct way to write this so we don't have to disable the linter?

bourdakos1: Umm I think its because we are passing an anonymous function to forwardRef. I think it’s fine to ignore this, but you could also try setting the component to a variable and pass that to forwardRef and if it’s still complaining you can do .displayName = xxx

kevin-bates:
From a functional level, this appears to be working well. I do see some differences between the same code running in an R notebook vs. the R script.  For example, the results of `system("hostname", intern=TRUE)` display in the notebook, but not in the script.  For those results to display in the script I must wrap the command in `print()` and then I get the square-bracketed counter prefixing the output (which is not present when not using `print()`).  This square-bracketing behavior _does_ match that of notebook, but I don't see output of `system` until I wrap in `print()`  in the R editor.

I also notice that the _stop_ button is actually the _interrupt_ button in the notebook.  This is fine, but there is a behavioral difference here as well.  In the notebook, when I select the interrupt button, the processing is halted immediately.  In the editor, when selecting the stop button, the processing continues about 5 seconds before it stops - due to the fact that a shutdown kernel request takes some time.  This is also true of Python scripts, and I wonder if we should (under the covers) submit an interrupt request before stopping the kernel from the editor.

karlaspuldaro:
Thank you @kevin-bates 
Our runner picks certain msg output types coming from the kernel session, this is how it looks
```
future.onIOPub = (msg: any): void => {
        const msgOutput: any = {};

        if (msg.msg_type === 'error') {
          msgOutput.error = {
            type: msg.content.ename,
            output: msg.content.evalue
          };
        } else if (msg.msg_type === 'execute_result') {
          if ('text/plain' in msg.content.data) {
            msgOutput.output = msg.content.data['text/plain'];
          } else {
            // ignore
            console.log('Ignoring received message ' + msg);
          }
        } else if (msg.msg_type === 'stream') {
          msgOutput.output = msg.content.text;
        } else if (msg.msg_type === 'status') {
          msgOutput.status = msg.content.execution_state;
        } else {
          // ignore other message types
        }
```
This portion of code hasn't changed, but maybe for R the `system()` output returns a different msg type we are not handling? I can debug it and find out.

For the stop button, we call `await this.sessionConnection.shutdown()` which kills the kernel and shuts down the session.
I believe to have a faster response as the kernel interrupt button, we should access the kernel directly somehow, instead of doing so through the session.

lresende:
> I do see some differences between the same code running in an R notebook vs. the R script. For example, the results of `system("hostname", intern=TRUE)` display in the notebook, but not in the script. For those results to display in the script I must wrap the command in `print()` and then I get the square-bracketed counter prefixing the output (which is not present when not using `print()`). This square-bracketing behavior _does_ match that of notebook, but I don't see output of `system` until I wrap in `print()` in the R editor.

If you have a`one cell notebook` does it behave similarly to the `R script` then?

karlaspuldaro:
@lresende this is what Kevin described
image
In the R Editor, the output is only displayed when wrapped in `print()`

lresende:
Yes, that's what I wanted to understand, if this R specific or Script specific (e.g. also in Python script as well)

kevin-bates:
> I believe to have a faster response as the kernel interrupt button, we should access the kernel directly somehow, instead of doing so through the session.

I see.  This isn't a big deal.  It might come up with folks wondering why the "little square button" isn't more responsive in scripts, but if interrupts can't be performed via the session (which makes sense that they can't) then let's not worry about. 
Thanks for looking into it!

lresende:
We added session to fix a bug related to working directory. Adding session also helped with imports and later on will enable better integration with debugger.

kevin-bates (commented):
I'm sorry. I just realized that this review was never submitted sometime yesterday and, by now, some of these comments may not apply, but I thought I'd send what I had.

packages/python-editor/src/ScriptEditor.tsx
```diff
@@ -85,15 +92,20 @@ export class PythonFileEditor extends DocumentWidget<
...
     this.runDisabled = false;
+    this.controller = new ScriptEditorController();
+    this.editorLanguage = this.kernelName.toLowerCase().includes(PYTHON)
+      ? PYTHON
+      : R;
 
-    // Add python icon to main tab
-    this.title.icon = pyIcon;
+    // Add icon to main tab
+    this.title.icon = this.editorLanguage === PYTHON ? pyIcon : rIcon;
```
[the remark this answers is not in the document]
karlaspuldaro: I agree, making subclasses would make these language-specific properties better contained.
This was the very first implementation idea, but we decided in handling it all in one class for now. 
The editor language is defined by the file type specified for the script editor factory, and since we are only creating script editors for Python or R files, the Script editor language assumptions can be considered safe, only not the most elegant. 😅  
We can revisit the implementation on a follow up PR.

packages/python-editor/src/ScriptEditor.tsx
```diff
@@ -122,14 +132,39 @@ export class PythonFileEditor extends DocumentWidget<
...
+    this.initializeKernelSpecs();
   }
 
+  initializeKernelSpecs = async (): Promise<void> => {
+    const kernelSpecs = await this.controller.getKernelSpecsByLanguage(
+      this.editorLanguage
+    );
+
+    if (this.editorLanguage === PYTHON) {
+      this.kernelName = kernelSpecs.default;
```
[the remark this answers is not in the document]
karlaspuldaro: Good catch, I assumed python was always jupyterlab default kernel, we should handle init kernel name properly then.

> Also, does getKernelSpecsByLanguage() preserve the "default" entry? I thought it just processed the kernelspecs themselves.

Yes

lresende:
Overall it looks good, one question I still have is related to the coupling of the two extensions, in the sense that either you install both Python and R editor or none. Should we leave all the shared code in the script editor module and have a Python editor and R editor with the minimal code to install/register the editors independently?

@skukhtichev any thoughts here?

skukhtichev:
@lresende 
> Should we leave all the shared code in the script editor module and have a Python editor and R editor with the minimal code to install/register the editors independently?

I think yes, because separating the Python and R script editors will allow to activate Script extensions depending on supported kernels. E.g. If there is only python kernel installed then users could activate only Python script editor, and default text editor will be opened for all other file types

kevin-bates (approved):
Thanks @karlaspuldaro - good stuff!

ptitzler (changes requested):
After installing from the source there are two Python editor icons:
image

ptitzler:
General comment:  for the sake of consistency the label text probably be `<X> file` or `<X> editor` because all three tasks do the same thing: create a new file of type X and open it in the editor. The current naming scheme does not convey this. 
 
image

ptitzler:
Would an R user specify _Python_ `file dependencies`? 

image

ptitzler:
General (?) submit dialog bug. A scroll bar is displayed when there is less content

submit

karlaspuldaro:
Thanks @ptitzler for the feedback.
1. I cannot reproduce the duplicated python button in the launcher :/ 
2. That's a good idea to make them all consistent to `X Editor`
3. Yes the file dependency type should be fixed
4. We can tweak the dialog css on this PR to remove the scrollbar. Looks like the last HTML element is slightly bigger than the display, so the bar shows up to scroll just a little.

lresende:
> After installing from the source there are two Python editor icons:
> image

This might be if you had a previously installed python editor and have not run the clean-jupyterlab.sh script because the branch still has the extension folder renamed. . This should not happen when we merge this with the "two extensions packages"  as the Python editor will be updated and the others (r script and script-editor added).

docs/source/getting_started/overview.md
```diff
@@ -116,6 +116,18 @@ The [JupyterLab Language Server Protocol (LSP) integration](https://github.com/k
 
 By default, Elyra installs the [Python Language Server package](https://pypi.org/project/python-language-server/), enabling the productivity features in the notebook editor and the Python editor. Support for other languages can be added by manually installing additional Language Server packages and their prerequisites.
 
+##### R-Editor Language Server Protocol
```
ptitzler, line 119: The R editor section in the new script editor topic might be a better location for this content because
 - it is too low level for an overview
 - it separates related content into multiple documents, forcing the user to look for related information in several locations (R LSP content is in the overview document but the R enterprise gateway content is in the editor document)

docs/source/user_guide/enhanced-script-support.md
```diff
@@ -0,0 +1,80 @@
...
+
+![Enhanced Python Support](../images/python-editor.gif)
+
+The execution of these scripts leverages the available Python and R based Kernels. This enables users to run their scripts in different configurations and environments.
+
+Elyra also allows submitting a Python and R scripts as a single node pipeline for execution in a Kubeflow Pipelines or Apache Airflow environment in the cloud. This feature is accessible when the Elyra [AI Pipelines](../user_guide/pipelines.md) extension is also enabled.
+
+## Python script execution support
+
+In the Jupyter Lab Launcher, click the `Python File` icon to create a new Python Script.
```
ptitzler, line 32:
```suggestion
In the Jupyter Lab Launcher, click the `Python File` icon to create a new Python script.
```

docs/source/user_guide/enhanced-script-support.md
```diff
@@ -0,0 +1,80 @@
...
+![Create new Python file](../images/launcher-python-file.png)
+
+When used in conjunction with `Jupyter Enterprise Gateway`, the dropdown will be populated with more kernel options,
+allowing users to run their scripts with remote kernels with more specialized resources.
+
+To run your script locally, select the `Python 3` option in the dropdown menu, and click the `Run` icon.
+
+## R script execution support
+
+In the Jupyter Lab Launcher, click the `R File` icon to create a new R Script.
```
ptitzler, line 43:
```suggestion
In the Jupyter Lab Launcher, click the `R File` icon to create a new R script.
```

docs/source/user_guide/enhanced-script-support.md
```diff
@@ -0,0 +1,80 @@
...
+
+![Create new R file](../images/launcher-r-file.png)
+
+To run your R script locally you will need to install an R kernel.
+1. If you are using a [conda](https://docs.conda.io/en/latest/miniconda.html) environment:
```
ptitzler, line 48: Incorrect indentation breaks formatting, leading to ambiguity. Is the last command supposed to be only run when the user performs (2)?

image

docs/source/user_guide/enhanced-script-support.md
```diff
@@ -0,0 +1,80 @@
...
+2. Alternatively, you can install it via [CRAN](https://cran.r-project.org/) on an R console:
+```
+install.packages('IRkernel')
+```
+then enable it on Jupyter:
+```
+IRkernel::installspec()
+```
+
+To run the script, from the Script editor toolbar, select the `R` option in the dropdown menu, and click the `Run` icon.
```
ptitzler, line 62:
```suggestion
To run the script, from the Script editor toolbar, select the `R` option in the kernel selection drop-down, and click the `Run` icon.
```

docs/source/user_guide/enhanced-script-support.md
```diff
@@ -0,0 +1,80 @@
...
+conda install -y r-irkernel
+```
+
+2. Alternatively, you can install it via [CRAN](https://cran.r-project.org/) on an R console:
+```
+install.packages('IRkernel')
+```
+then enable it on Jupyter:
+```
+IRkernel::installspec()
```
ptitzler, line 59: Where does the user run this? Since the Elyra documentation does not visually distinguish between commands entered at the prompt
```
$ ls -al
```

and somewhere else
```
print('hello world')
```

the user would already need to know. But beginners necessarily do.

ptitzler: The other problem is, if I understand the instructions in https://irkernel.github.io/installation/ correctly, that one needs to   run the command from R. However, I don't have R installed ... and I can't run R commands because the editor doesn't have a kernel (:

ptitzler:
> This might be if you had a previously installed python editor and have not run the clean-jupyterlab.sh script because the branch still has the extension folder renamed.

Thanks! Confirmed. It's gone after a clean clean install.

docs/source/user_guide/enhanced-script-support.md
karlaspuldaro: It should be run from an R console as described in line 53, but you're right, the way it is worded can be ambiguous.

docs/source/getting_started/overview.md
akchinSTC: moved to the script editor topic

docs/source/user_guide/enhanced-script-support.md
akchinSTC: removed the numbering, argue that its clearer without it given there is really only one step with one given alternative.

docs/source/user_guide/enhanced-script-support.md
akchinSTC: added some language to inform users that R must available prior to installation and updated code snippets

ajbozarth (changes requested):
A handful of comments, and all my comments on the python editor index file apply to the r editor index file as well since it's essentially a copy. Also note I only looked at the front file, not the backend, docs or tests.

ajbozarth:
Pushed a commit resolving my review comments, in general I tweaked the index and package files so as to minimize changes, most of which were caused by the dev process of renaming the project, then splitting it into 3 packages.

ajbozarth (approved):
lgtm

lresende:
I am not sure what's going on with the build, took a quick look but could not reproduce it locally, played with some dependency versions without much luck, will need to resolve this tomorrow.

PR #1443, opened 2021-03-16
Add default / placeholder to dropdown field in metadata editor

### LINKED ISSUE #1368: Metadata - select box uses hard-wired placeholder
**Describe the issue**

The select box in the KFP runtime configuration appears to be using a hard-wired placeholder text `Create or select  ...` https://github.com/elyra-ai/elyra/blob/master/packages/ui-components/src/DropDown.tsx#L88

image

**To Reproduce**
1. Create a Kubeflow Pipelines runtime configuration
2. Inspect the `Kubeflow Pipelines engine` select box

**Expected behavior**
Scenario 1: The select box input area displays the `default` value (if one is defined) if the user can only choose from a set of predefined options but cannot add an option ("read-only" drop down) https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/kfp.json#L76

Scenario 2: The select box input area displays the `uihints.placeholder` value, if one is defined:
https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/kfp.json#L71 (none is defined yet in this case) and it's an "editable drop down")

**Deployment information**
 - Elyra version: 2.1/master

marthacryan:
Fixes #1368.

* If `uihints.placeholder` is defined, the field will display that as a placeholder. Otherwise, it will use "Create or select <field name>" as a placeholder for dropdowns
* If the schema defines a `default` for a field, an empty field will automatically use that for the value of the field. Otherwise, the field will be empty.

The code snippet language field remains the same:
image

The KFP runtimes editor automatically shows this when creating a new runtime config: 
image

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditor.tsx
+++ b/packages/metadata-common/src/MetadataEditor.tsx
@@ -479,7 +479,9 @@ export class MetadataEditor extends ReactWidget {
           description={this.schema[fieldName].description}
           required={required}
           defaultError={uihints.error}
-          defaultValue={this.metadata[fieldName]}
+          placeholder={uihints.placeholder}
+          defaultValue={this.schema[fieldName].default}
+          initialValue={this.metadata[fieldName]}
           options={this.getDefaultChoices(fieldName)}
           onChange={(value): void => {
             this.handleDropdownChange(fieldName, value);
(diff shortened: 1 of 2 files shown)


ptitzler:
In the KFP scenario, a user should not be able to edit the input field - only choose one. Do we need to introduce another property to indicate that the drop down is supposed to be r/o? 

engine_type

ptitzler (changes requested):
See earlier comment about read only behavior

ptitzler:
Not sure why we are not using a basic select list. There's really no need for text input/clear for read-only properties.

simple_select_list

marthacryan:
@ptitzler I was actually just working on adding that for readonly dropdowns! Here's what it looks like now:
select

marthacryan:
Here's the dropdown in runtime images (no default value):
image
image

ptitzler (approved):
LGTM. Simple and a lot more user friendly.

PR #1446, opened 2021-03-16
Add placeholder property to uihints

ptitzler:
The placeholder is displayed instead of the static default, which is "Create or select ...". PR https://github.com/elyra-ai/elyra/pull/1443 takes advantage of this new property.

placeholder-property

 
           in the file; or

           it.

### DIFF
--- a/elyra/metadata/schemas/kfp.json
+++ b/elyra/metadata/schemas/kfp.json
@@ -77,7 +77,8 @@
           "default": "Argo",
           "uihints": {
             "field_type": "dropdown",
-            "category": "Kubeflow Pipelines"
+            "category": "Kubeflow Pipelines",
+            "placeholder": "Select engine type"
           }
         },
         "cos_endpoint": {


elyra/metadata/schemas/kfp.json
```diff
@@ -77,7 +77,8 @@
           "default": "Argo",
           "uihints": {
             "field_type": "dropdown",
-            "category": "Kubeflow Pipelines"
+            "category": "Kubeflow Pipelines",
+            "placeholder": "Select engine type"
```
lresende, line 81: But I believe this is one of those places that most people will be using the default engine that is deployed with KFP which is Argo. Should we actually have a `default value` or Argo here? Which is what we do if this is not selected?

ptitzler: See https://github.com/elyra-ai/elyra/pull/1443#issuecomment-801223295, which proposes a simpler solution that eliminates the need for this placeholder altogether for this specific property.

ptitzler:
Withdrawing this PR because there is no longer a need for it.

PR #1447, opened 2021-03-16
Can't remove string array properties

### LINKED ISSUE #1426: Pipeline editor: cannot remove empty node properties
**Describe the issue**

image

**To Reproduce**
1. Create pipeline with one node
2. Add a notebook/script
3. open properties
4. Add a bunch of emoty env variables (for example)
5. close properties and save pipeline
6. open properties
7. remove some of the empty properties
8. close the pipeline
9. open the pipeline
9. open node properties again. they are back!

**Screenshots or log output**
<pre>
Paste the log output here.
</pre>
</details>

**Expected behavior**

**Deployment information**
 - Elyra version: [e.g. 1.5.3]
 - Operating system: [e.g. macos, linux]
 - Installation source: [e.g. PyPI, conda, from source, official container image, custom container image]
 - Deployment type: [e.g. local installation, Docker, Kubernetes, Kubeflow [notebook server] , Open Data Hub]

ajbozarth:
When `StringArrayInput` deletes a value in string array properties
it does not actually delete it, instead leaving an `undefined` value.

This needs to be addressed by removing the undefined values when
saving property changes. This is due to js allowing sparse arrays.

Fixes #1426

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -545,9 +545,13 @@ export class PipelineEditor extends React.Component<
     }
 
     app_data.runtime_image = propertySet.runtime_image;
-    app_data.outputs = propertySet.outputs;
-    app_data.env_vars = propertySet.env_vars;
-    app_data.dependencies = propertySet.dependencies;
+    app_data.outputs = propertySet.outputs.filter((x: any) => x !== undefined);
+    app_data.env_vars = propertySet.env_vars.filter(
+      (x: any) => x !== undefined
+    );
+    app_data.dependencies = propertySet.dependencies.filter(
+      (x: any) => x !== undefined
+    );
     app_data.include_subdirectories = propertySet.include_subdirectories;
     app_data.cpu = propertySet.cpu;
     app_data.memory = propertySet.memory;


ajbozarth:
Note: I decided to deal with the issue at save and not delete because `StringArrayInput` is rendered using index in the keys, so by removing values, rather than deleting and leaving it sparse, will cause the "hard-coded" indexes to be wrong

bourdakos1 (approved):
Just an fyi, this is not an issue in the new pipeline-editor component. We also had a discussion around using an index as the key vs using stable IDs. We decided to keep using the index + work around instead of moving to stable IDs

ptitzler (approved):
I was able to confirm that removed empty values are no longer persisted.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -545,9 +545,13 @@ export class PipelineEditor extends React.Component<
...
-    app_data.dependencies = propertySet.dependencies;
+    app_data.outputs = propertySet.outputs.filter((x: any) => x !== undefined);
+    app_data.env_vars = propertySet.env_vars.filter(
+      (x: any) => x !== undefined
+    );
+    app_data.dependencies = propertySet.dependencies.filter(
+      (x: any) => x !== undefined
+    );
     app_data.include_subdirectories = propertySet.include_subdirectories;
     app_data.cpu = propertySet.cpu;
```
lresende, line 556: Would the same apply here? We are seeing null values being saved when you initially set these to a value, and then try to delete the value from the node property (cpu, gpu, mem)

ajbozarth: I'm unsure, in the case of string array we wrote our own component to handle them (`StringArrayInput`) and that component caused this bug. Whereas iiuc we leverage a canvas component for the dropdown so I would argue that that is a separate issue than this.

bourdakos1: Yea, I would say that is a separate issue. I would consider storing `dependencies: ["one", null, null, "two", null, "three"]` a bug, but storing `cpu: null` more of an opinion. I think the same discussion could be had for when the array is empty should we store `dependencies: []` or just delete dependencies

PR #1418, opened 2021-03-11
Enable support for adding R Script to Pipeline

lresende:
Fixes #1424 

Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #1411, opened 2021-03-10
Expose error due to failure of local python pipeline node execution

kiersten-stokes:
Resolves #1382 by adding an except block to trap `CalledProcessError`'s that occur specifically as a result of python subprocess execution failure.

### New look
new look

 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

Issue #1467, opened 2021-03-23
Trap and format namespace-related errors during pipeline processing in a more user-friendly way


kiersten-stokes:
**Describe the issue**
When submitting a pipeline to a KFP runtime with an incorrect or missing namespace (if one is required), the error message produced is too verbose and is not user-friendly.

**To Reproduce**
1. Enter a KFP runtime with an incorrect or empty namespace (for an endpoint that requires a namespace)
2. Submit a pipeline to this runtime
3. See below error

**Screenshots/log output**
When namespace entered is incorrect:
<img width="1019" alt="Screen Shot 2021-03-23 at 10 58 26 AM" src="https://user-images.githubusercontent.com/31816267/112187429-17948580-8bd0-11eb-8c0c-9c7ec91bb531.png">

<details><summary>Log Output</summary>
<pre>
Traceback (most recent call last):
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/tornado/web.py", line 1705, in _execute
    result = await result
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/elyra/pipeline/handlers.py", line 89, in post
    response = await PipelineProcessorManager.instance().process(pipeline)
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/elyra/pipeline/processor.py", line 78, in process
    res = await asyncio.get_event_loop().run_in_executor(None, processor.process, pipeline)
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/elyra/pipeline/processor_kfp.py", line 178, in process
    experiment = client.create_experiment(name=experiment_name,
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/kfp/_client.py", line 345, in create_experiment
    experiment = self.get_experiment(experiment_name=name, namespace=namespace)
  File "/Users/kierstenstokes/miniconda3/envs/dev1/lib/python3.9/site-packages/kfp/_client.py", line 449, in get_experiment
    list_experiments_response = self.list_experiments(page_size=100, page_token=next_page_token, namespace=namespace)
[...]

PR #1485, opened 2021-03-25
Refactor script processors, include brief detail on generic errors

### LINKED ISSUE #1337: Missing kernelspec on new notebooks created in VSCode
**Describe the issue**
Notebooks created in VSCode don't have a kernel spec until you run them, which causes issues when trying to run them with Elyra.

**To Reproduce**
1. Create and open an `.ipynb` file in VSCode
2. Add some content to it
3. Save the notebook file
4. Open the notebook Jupyter Lab
5. Add the notebook to an Elyra Pipeline
6. Set a runtime image for the node
7. Submit notebook for local execution
8. Get error `KeyError: 'kernelspec'`

**Screenshots or log output**
Screen Shot 2021-02-25 at 10 38 17 AM

**Expected behavior**
Either a better error message with steps to resolve the issue, or just magically run somehow.
(This can be fixed by opening the notebook in Jupyter Lab and re-saving the file or by running the notebook at least once in VSCode)

**Deployment information**
 - Elyra version: 1.5.3
 - Operating system: macos

 - Local execution

**Notebook source:**
```
{
 "metadata": {
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": 3
  },
  "orig_nbformat": 2
 },
 "nbformat": 4,
 "nbformat_minor": 2,
 "cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {},
   "outputs": [],
   "source": [
    "print(\"hello world\")"
   ]
  }
 ]
}
```

kevin-bates:
Although this change was instigated by #1337, it primarily focuses on the aftermath of the merges for #1411 and #1418 - which essentially crossed paths, leaving the R script processor behind the Python script processor in terms of its error handling.  

While looking into #1337, it was found that Papermill release 2.3.3 handles the missing kernelspec much more gracefully than raising a `KeyError`.  Instead, it raises a `ValueError` indicating the kernel could not be determined nor was one provided.  However, the general error processing for notebook and both forms of script operations drop any level of details due to the unknown nature of error messages.  

This PR does the following:
- It introduces a new base class named `ScriptOperationProcessor` from which the _existing_`PythonScriptOperationProcessor` and `RScriptOperationProcessor` classes now derive.  This base class contains 90% of the applicable code with the subclasses providing their name and argument vectors.
- It introduces a `log_and_raise()` method on the base `FIleOperationProcessor` class that is available to all file-based operations.  Building on the work done in #1411, this method checks the length of the error message and truncates it to around the max (80), replacing overflow with an ellipses (...).
- Adds a test that removes the kernel metadata from a notebook node and ensures the appropriate error is raised.

When the kernelspec information is missing, the following message will be produced:
Screen Shot 2021-03-25 at 2 35 42 PM

Resolves #1337
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_local.py
+++ b/elyra/pipeline/processor_local.py
@@ -42,7 +42,7 @@ class LocalPipelineProcessor(PipelineProcessor):
     _type = 'local'
 
     def __init__(self, root_dir, **kwargs):
-        super(LocalPipelineProcessor, self).__init__(root_dir, **kwargs)
+        super().__init__(root_dir, **kwargs)
         notebook_op_processor = NotebookOperationProcessor(self.root_dir)
         python_op_processor = PythonScriptOperationProcessor(self.root_dir)
         r_op_processor = RScriptOperationProcessor(self.root_dir)
@@ -121,7 +121,7 @@ class FileOperationProcessor(OperationProcessor):
     MAX_ERROR_LEN: int = 80
 
     def __init__(self, root_dir: str):
-        super(FileOperationProcessor, self).__init__()
+        super().__init__()
         self._root_dir = root_dir
 
     @property
@@ -171,9 +171,6 @@ class FileOperationProcessor(OperationProcessor):
 
 class NotebookOperationProcessor(FileOperationProcessor):
     _operation_name = 'execute-notebook-node'
-
-    def __init__(self, root_dir: str):
-        super(NotebookOperationProcessor, self).__init__(root_dir)
 
     def process(self, operation: Operation):
         filepath = self.get_valid_filepath(operation.filename)
@@ -263,9 +260,6 @@ class PythonScriptOperationProcessor(ScriptOperationProcessor):
     _operation_name = 'execute-python-node'
     _script_type = 'Python'
 
-    def __init__(self, root_dir):
-        super(PythonScriptOperationProcessor, self).__init__(root_dir)
-
     def get_argv(self, file_path) -> List[str]:
         return ['python3', file_path, '--PYTHONHOME', os.path.dirname(file_path)]
 
@@ -273,9 +267,6 @@ class PythonScriptOperationProcessor(ScriptOperationProcessor):
 class RScriptOperationProcessor(ScriptOperationProcessor):
     _operation_name = 'execute-r-node'
     _script_type = 'R'
-
-    def __init__(self, root_dir):
-        super(RScriptOperationProcessor, self).__init__(root_dir)
 
     def get_argv(self, file_path) -> List[str]:
         return ['Rscript', file_path]
--- a/elyra/pipeline/processor_local.py
+++ b/elyra/pipeline/processor_local.py
@@ -23,7 +23,7 @@ from elyra.util.path import get_absolute_path
 from jupyter_server.gateway.managers import GatewayClient
 from subprocess import run, CalledProcessError, PIPE
 from traitlets import log
-from typing import Dict
+from typing import Dict, List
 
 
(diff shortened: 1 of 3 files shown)


lresende (approved):

lresende:
Should we really truncate messages? I would say that the current issues we see on the dialog box limitation might not be present on different UIs? and if we truncate the message it would not be available for UIs that have the proper way to display them? Also, if it's only in the logs, what happens with managed environments such as Hub/Binder that does not expose the logs to the user?

lresende (changes requested):
Changing to requested changes to discuss the direction of truncating error messages on the server-side versus on the UI.

kevin-bates:
We can open up the message, it wasn't there prior to this, so I thought I'd add a length to it.  My concern is that we can easily get back to situations described [here](https://github.com/elyra-ai/elyra/issues/1467#issue-838948082) where some exception decides to be more verbose about things. 

> Also, if it's only in the logs, what happens with managed environments such as Hub/Binder that does not expose the logs to the user?

The details twisty thing essentially contains the portion of the "logs" for a user to send to an admin.

kevin-bates:
Drilling down in this comment:
> what happens with managed environments such as Hub/Binder that does not expose the logs to the user?

I'm curious what you're getting at here?  In my opinion, exposing access to a server's log - whether that be directly or via some propagation mechanism using third-party software like Grafana - is the responsibility of the administrator/installation.  We don't even instruct users to redirect the server's console output to a file where the "logs" can be persisted and "exposed".  

Even if the logs were exposed to the users, are we expecting users to decipher what's going on - especially in error situations?  

Are you proposing we capture things like status information during a given request so that that information can be returned to the user, or that we expose logs via some kind of service to work around the scenario where logs are intentionally not made available to the end-user?

I'm just trying to understand the end-goal here because we'll need to design/conform our logging differently (and perhaps add state management altogether).

One thought would be to assign every request a, well, _request ID_, that is randomly generated at the start of each request.  The request ID would be logged on entry and exit from the handlers and potentially at _interesting data points_ and certainly be included in [`log_pipeline_info()`](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor.py#L166) along with logged exceptions.  It would also be included _on_ the error dialog.  Then the user would simply reference the _request id_ as the "handle" that an administrator could use to decipher that particular request.  This is tantamount to a _session id_ associated with each request.

kevin-bates:
After a sidebar (scrum), it was determined that having the traceback in the error details is sufficient _log information_ and inclusion of state or progress is not necessary.

We also discussed possibly moving the truncation to the client since other client interfaces may be able to handle the more unwieldy messages better.  We decided, for now, to continue with the truncation on the server after confirming that the full (non-truncated) message is available in the traceback.  Here's a screenshot, in which I temporarily shorted the max message length from 80 to 40 characters, showing that the full message is indeed in the traceback:
Screen Shot 2021-03-26 at 9 36 52 AM

Again, note that this truncation only happens for _general_  exceptions. We explicitly handle `PapermillExceutionError` (for notebooks) and `CalledProcessError` (for scripts) exceptions and they do not apply to this discussion.

SInce it sounds like the discussion is complete, I'm removing this from DRAFT.

kiersten-stokes (approved):
Looks great!

lresende (approved):

PR #1492, opened 2021-03-29
Fix common properties bug

### LINKED ISSUE #1493: Properties copy over from one node to another when switching properties editor
**Describe the issue**
Properties copy over from one node to another when switching properties editor (i.e. properties editor is open and new node is selected in pipeline editor).

**To Reproduce**
1. Drag file onto pipeline editor
2. Drag a different file onto pipeline editor
3. Open properties editor
4. Select one of the nodes
5. Select the other node
6. Select the first node again
7. The filename for the first node is now the filename of the second node.

marthacryan:
Fixes #1493. The properties editor has a bug that causes properties to copy over to a different node when a new node is selected in the pipeline editor. This PR adds logic to apply the properties changes as they happen instead of when switching nodes to prevent that from happening. 
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -278,7 +278,6 @@ export class PipelineEditor extends React.Component<
       selectedNodes: []
     };
 
-    this.applyPropertyChanges = this.applyPropertyChanges.bind(this);
     this.propertiesActionHandler = this.propertiesActionHandler.bind(this);
     this.propertyListener = this.propertyListener.bind(this);
     this.propertiesControllerHandler = this.propertiesControllerHandler.bind(
@@ -410,8 +409,7 @@ export class PipelineEditor extends React.Component<
     const propertiesCallbacks = {
       actionHandler: this.propertiesActionHandler,
       controllerHandler: this.propertiesControllerHandler,
-      propertyListener: this.propertyListener,
-      applyPropertyChanges: this.applyPropertyChanges
+      propertyListener: this.propertyListener
     };
 
     let commProps = null;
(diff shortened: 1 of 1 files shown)


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -568,67 +565,49 @@ export class PipelineEditor extends React.Component<
   }
 
   propertyListener(data: any): void {
+    console.log(data);
```
bourdakos1, line 568: log

bourdakos1:
The original issue is now fixed. However, there is still a similar issue with the array items. I'm guessing the custom component isn't fully controlled and doesn't trigger `UPDATE_PROPERTY` for every change

bourdakos1:
I think this needs to be `value` not `defaultValue`, but changing this will probably break other things: https://github.com/marthacryan/elyra/blob/c3cecbd7580aae637e0c27b65aa2b0c9ec546ed4/packages/pipeline-editor/src/StringArrayInput.tsx#L88

bourdakos1:
I think we can fix this issue separately. This doesn't effect the underlying data, just a rendering bug

bourdakos1 (approved):

marthacryan:
> The original issue is now fixed. However, there is still a similar issue with the array items. I'm guessing the custom component isn't fully controlled and doesn't trigger `UPDATE_PROPERTY` for every change

Could you add the steps to reproduce? I tried and couldn't figure out how to reproduce

bourdakos1:
oh yea sorry, It's similar to the previous issue:
- open properties
- add an item to a list (for example `1`)
- click on another node and add an item to a list (for example `2`)
- click anywhere on the canvas
- reopen the node properties where you entered `2` and it should incorrectly say `1`

PR #1508, opened 2021-03-31
Fix default return value for node when using submit button

akchinSTC:

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/parser.py
+++ b/elyra/pipeline/parser.py
@@ -143,7 +143,8 @@ class PipelineParser(LoggingConfigurable):
             id=node_id,
             type=node.get('type'),
             classifier=node.get('op'),
-            name=PipelineParser._get_ui_data_field(node, 'label'),
+            name=PipelineParser._get_ui_data_field(node, 'label',
+                                                   default_value=PipelineParser._get_app_data_field(node, 'filename')),
             cpu=PipelineParser._get_app_data_field(node, 'cpu'),
             gpu=PipelineParser._get_app_data_field(node, 'gpu'),
             memory=PipelineParser._get_app_data_field(node, 'memory'),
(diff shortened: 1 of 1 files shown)


elyra/pipeline/parser.py
```diff
@@ -143,7 +143,8 @@ def _create_pipeline_operation(node: Dict, super_node: Optional[Dict] = None):
             id=node_id,
             type=node.get('type'),
             classifier=node.get('op'),
-            name=PipelineParser._get_ui_data_field(node, 'label'),
+            name=PipelineParser._get_ui_data_field(node, 'label',
+                                                   default_value=PipelineParser._get_app_data_field(node, 'filename')),
```
lresende, line 147: you have filename a few lines below, don't want to move this down below and use that as the default value?

ptitzler (approved):
Submit now seems to work for Python scripts, R scripts and notebooks.

There's still the issue with the bogus file extension for notebooks but it doesn't seem to have an impact on this issue.

image

elyra/pipeline/parser.py
kevin-bates: I agree.  I guess I was surprised to see the default value added here, but then I noticed that the Operation constructor requires a value for `name` and that the defaulting performed within the `name` property accessor is really all about dropping the suffix.  So, this default value seems fine.

Regarding leveraging the `filename` keyword value as suggested, I'm not sure if that would work.  Can a keyword reference another keyword in a call? No - just checked.  You'll get `NameError: name 'filename' is not defined`.  You'd need to move it into a regular variable prior to the instance construction.

kevin-bates (approved):
Since the issue is resolved, I'm cool with dealing with the default value detail until later.

lresende (approved):

PR #1536, opened 2021-04-08
Build elyra-image from released artifacts when building non dev tags

### LINKED ISSUE #1525: Support building Elyra container images from released artifacts if building release container tags (e.g. not dev)
It should be possible to create the (`kf-notebook` container image](https://elyra.readthedocs.io/en/latest/recipes/using-elyra-with-kubeflow-notebook-server.html) from PyPI ("use published packages") or from a local github branch ("use locally built packages").  

See https://github.com/elyra-ai/elyra/issues/1522#issuecomment-813208070

The idea would be to check the container tag being built:
- dev tags require building from source and installing
- x.y.z tags, which corresponds to a release, can skip the build, and be installed from published pypi artifacts

This should speed up the process of building and help reduce the size of the built image.

lresende:
Fixes #1525 

 
           in the file; or

           it.


ptitzler:
This works fine for `make` (with `2.2.1` as hard-wired default) but fails for `make` (with `dev` as hard-wired default) and `make TAG=2.0.1` (to override the hard-wired default to build an older version).

lresende:
There is also an issue with releases (e.g 2.2.1) where if you want to start the docker image as in the readme, you get two work dirs, one with some extra char at the name and it is messing things up. I didn't have much time to get to the bottom of it.

ptitzler:
- `make elyra-image TAG=2.2.1` OK (sample pipeline runs fine locally)
- `make elyra-image TAG=2.1.0` OK (sample pipeline runs fine locally)
- `make elyra-image TAG=2.0.1` FAIL (Build errors with `The Jupyter Server requires tornado >=6.1.0`)

Looks like there are limits to how far back we can go version wise with a generic build approach. Sooner or later the referenced base image does require a package version that conflicts with what Elyra needs.

ptitzler:
` make elyra-image TAG=dev` fails because `make` isn't available during build

etc/docker/elyra/Dockerfile
```diff
@@ -18,29 +18,31 @@
...
 
 ADD start-elyra.sh /usr/local/bin/start-elyra.sh
 
-RUN chmod ugo+x /usr/local/bin/start-elyra.sh && \
-    apt-get update && apt-get install -y build-essential curl && \
-    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add - && \
-    echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update && apt-get install -y yarn && \
-    rm -rf /var/lib/apt/lists/*
+RUN chmod ugo+x /usr/local/bin/start-elyra.sh
```
ptitzler, line 27:
```suggestion
RUN chmod ugo+x /usr/local/bin/start-elyra.sh
RUN if [ "$TAG" == "dev" ] ; then \
            apt-get update && apt-get install -y build-essential && \
            rm -rf /var/lib/apt/lists/*; \
         fi
```

ptitzler (changes requested):
See above

lresende:
This is ready for review now, there seems to be an existing issue that has not been introduced by these changes and I created https://github.com/elyra-ai/elyra/issues/1550 to track it.

lresende:
Also, this seems to have reduced the image name in about 1GB when building it from released artifacts

```
REPOSITORY                  TAG       IMAGE ID       CREATED             SIZE
elyra/elyra                 dev       758d3bbe9bd3   31 minutes ago      3.95GB
elyra/elyra                 2.2.1     0e162c701944   2 days ago          2.83GB
```

ptitzler:
One thing I didn't realize is that `make elyra-image TAG=dev` builds the container image from the official dev repository's master branch, not the user's _local_ file system. So as is, this can't handle these types of builds without requiring changes by the user.

kevin-bates:
>  make elyra-image TAG=dev builds the container image from the official dev repository's master branch, not the user's local file system.

Hmm - that seems like an issue.  The dev build should install whatever wheel file is in the local `dist` or `build` directories.

lresende:
> > make elyra-image TAG=dev builds the container image from the official dev repository's master branch, not the user's local file system.
> 
> Hmm - that seems like an issue. The dev build should install whatever wheel file is in the local `dist` or `build` directories.

We tried that in the past, and there were a few issues related to how lab handles dependencies around extensions and no extensions. One way it would be to copy the user-cloned contents to the docker build directory and "COPY/ADD" that... but this is one thing that should be changed across all images and I would take that as another issue/pr.

ptitzler:
> but this is one thing that should be changed across all images 

The other images pull from PyPI. This is currently the only Dockerfile with support to build from the local file system.

kevin-bates:
> One way it would be to copy the user-cloned contents to the docker build directory and "COPY/ADD" that... but this is one thing that should be changed across all images and I would take that as another issue/pr.

Seems like the Developer Guide should include the necessary steps to build a custom image (that can include changes to both front and back ends) since that's probably a typical use case these days. (And just asked about on gitter. :smile:  )

ptitzler:
>  Seems like the Developer Guide should include the necessary steps

The topic is right here: https://elyra.readthedocs.io/en/latest/developer_guide/development-workflow.html#building-the-elyra-container-image. Ideally the instructions are as simple as `make elyra-image TAG=dev` to build from local source and `make elyra-image TAG=X.Y.Z` to create an image from a published version on PyPI.

lresende:
> > but this is one thing that should be changed across all images
> 
> The other images pull from PyPI. This is currently the only Dockerfile with support to build from the local file system.

Which I don't believe is the proper way, I would expect that xxx:1.2.0 builds from a release but xxx:dev build from the tip of master otherwise you can't really validate something on that image before a release.

lresende:
Updated docs

akchinSTC (approved):
This PR works as advertised for me, that is to be able to build an elyra image from released artifacts (pypi). If we want to be able to build a image from local environment, we should open a separate issue.

lresende:
@ptitzler see comment above https://github.com/elyra-ai/elyra/pull/1536#pullrequestreview-635095369

ptitzler:
Want to include support for https://github.com/elyra-ai/elyra/issues/1557 or handle it in a separate PR?

ptitzler (approved):
LGTM

PR #1547, opened 2021-04-09
[POC] Add npm scripts

bourdakos1:
This PR updates the GitHub actions to ensure that our `yarn.lock` isn't taken into consideration when running our integration tests. This should help make us aware of any runtime bugs hidden during development or that arise post-release due to version bumps of dependencies.

These changes required a bit of reworking of our script runner, but instead of modifying the current `Makefile`, I decided to test out a workflow only using `npm scripts`. I'm not very experience with Python so I would especially appreciate any feedback in that area. If we like using `npm scripts` I will update the documentation, if we don't like the `npm scripts` I will turn them into `make` commands.

To install dependencies run:
```
yarn install
pip install ".[test]"
```

At this point, you can either run any linting/formatting:
```
yarn format
yarn lint
```

To do anything else, you will need to build Elyra:
```
yarn build
```

After Elyra is built, you can run unit tests:
```
yarn test
```

To try out the extension in JupyterLab or run integration tests, you need to actually install the extension:
```
yarn lab:install
```

Here, you have the option of doing a watch mode or full build.

For a full build you can run:
```
yarn lab:build 
jupyter lab
```

For watch mode:
```
// in one terminal
yarn lab:watch

// in another terminal
yarn watch
```

## TL;DR
To install/build everything and start jupyter:
```
yarn install
pip install .
yarn start
```


.github/workflows/build.yml
```diff
@@ -42,46 +42,59 @@ jobs:
...
+          packages/code-snippet/dist.tgz \
+          packages/metadata/dist.tgz \
+          packages/metadata-common/dist.tgz \
+          packages/pipeline-editor/dist.tgz \
+          packages/python-editor/dist.tgz \
+          packages/r-editor/dist.tgz \
+          packages/script-editor/dist.tgz \
+          packages/services/dist.tgz \
+          packages/theme/dist.tgz \
+          packages/ui-components/dist.tgz
```
lresende: Multiple things here, most of these should be part of make, `rm -rf node_modules` should be part of clean, the `lab extensions` some are to be linked others to be installed. Having in make as we have today avoid having this duplications that can introduce errors.

.github/workflows/build.yml
```diff
@@ -42,46 +42,59 @@ jobs:
...
+        yarn test:integration
```
lresende, line 97: Again, these depend on other things and should be just part of the make

Makefile
```diff
@@ -40,16 +40,18 @@ help:
 
 purge:
 	rm -rf build *.egg-info yarn-error.log
-	rm -rf node_modules lib dist
-	rm -rf $$(find packages -name node_modules -type d -maxdepth 2)
+	rm -rf lib dist
```
lresende, line 43: this is already available as the next couple lines

Makefile
```diff
@@ -75,6 +77,8 @@ uninstall:
 
 clean: purge uninstall ## Make a clean source tree and uninstall extensions
 
+full-clean: purge purge-yarn uninstall ## Should be unnecessary to run this
```
lresende, line 80: clean should be the `full clean`. Why creating a new target? Just add purge-yarn to clean.

lresende (changes requested):
Please don't duplicate `make targets` commands in `build.yml`. We should keep enhancing the Makefile as needed so that build and local developers can execute similar steps.

bourdakos1:
@lresende this is still a WIP, just trying to get the actions working right now. I think some of the make targets will need to be reworked in order to get some of this working

bourdakos1:
@lresende should the yarn dependency caching for the github actions we talked about be a part of this PR or should that be separate?

lresende:
Ideally i would have one pr for enabling lock and another for refactoring/enhancing the ci script

.github/workflows/build.yml
lresende: what about `make install`?

bourdakos1: @lresende what is this referring to?

bourdakos1: Oh, I see, this comment shows up on outdated code in the conversation view. `make install` wouldn't be ideal here, because we don't want to rebuild any of our code. We only want to install our prebuilt packages and build JupyterLab. This will minimize build errors since we have deleted the yarn.lock at this point, but alert us to any runtime errors

lresende:
It looks like a few new commands have been introduced into `package.json` but not in use by the makefile. IMHO, the ci should mimic the developers' workflow via make and avoid any duplications. Once that is resolved I will take a deeper look at the pr.

lresende:
Also, as suggested in https://github.com/elyra-ai/elyra/pull/1547#issuecomment-817079872 this to me looks more like two prs:
- Enabling lock (uncontroversial and small)
- Updates to scripts/ci (needs discussion)

bourdakos1:
@lresende I left the ci enhancements out of this PR, but this does still update ci to make sure that we are running integration tests on unlocked dependencies. Should that be part of a separate PR as well?

PR #1551, opened 2021-04-12
Serve pipeline configuration from server

### LINKED ISSUE #454: Implement more dynamic pipeline editor configuration

lresende:
This is a bring-up of #650 without any front-end updates and updates of the json files. 

Fixes #454 
 
           in the file; or

           it.


elyra/pipeline/handlers.py
```diff
@@ -93,3 +95,26 @@ async def post(self, *args, **kwargs):
...
+
+    valid_resources = ["palette", "properties"]
+
+    @web.authenticated
+    def get(self, resource):
+        if(resource in self.valid_resources):
+            msg_json = self._read_config(resource)
+        else:
+            # invalid resource, throw an error
+            raise ValueError("Invalid configuration name {}".format(resource))
```
kevin-bates, line 111: Should raise HTTPError with a 400 status here.

elyra/pipeline/handlers.py
```diff
@@ -93,3 +95,26 @@ async def post(self, *args, **kwargs):
...
+        self.write(msg_json)
```
kevin-bates, line 112: I think we want to call `self.finish(msg_json)` here.  Should probably set the content-type header as well.

lresende: `finish` already sets the content-type, but explicitly doing that to follow the same pattern used in other elyra handlers

lresende:
@marthacryan is this easy to integrate with your pr for the new pipeline-editor component?

kevin-bates (changes requested):
The openAPI doc needs to be updated and tests should be implemented.

lresende:
Do we have a simple way to test handlers by bootstrapping a server that is exposing these handlers? Could you send a reference link, please?

lresende:
> The openAPI doc needs to be updated

Once it is documented, does it become an official/supported API? If so, I would prefer not to do so as this will evolve to a completely different API once we add the notion of "component registry", and only document the final API.

kevin-bates:
> Once it is documented, does it become an official/supported API?

Only in the sense that everything else is "official".   (Elyra should really be using version indicators in its endpoints actually.)

> this will evolve to a completely different API once we add the notion of "component registry"

The "component registry" will change how envs are extracted from notebooks and scripts?  Interesting.

kevin-bates:
> Do we have a simple way to test handlers by bootstrapping a server that is exposing these handlers? 

Not that I'm aware of.  I currently use the [RESTClient extension](http://restclient.net/) on Firefox if no tests exist. 

> Could you send a reference link, please?

The metadata service includes handler testing: https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/tests/test_handlers.py

@kiersten-stokes and I discussed adding this in a separate PR since there was going to be a release today.  If that isn't the case, it seems like this could be included here as well.

kevin-bates:
I'm sorry, but my last two comments are really regarding #1460 - although I suppose they do have some application here - since both tests and open API updates do not exist.

lresende:
@kevin-bates basic tests have been added

elyra/pipeline/handlers.py
```diff
@@ -92,4 +94,28 @@ async def post(self, *args, **kwargs):
...
+
+
+class PipelineConfigHandler(HttpErrorMixin, APIHandler):
+    """Handler to expose method calls to retrieve pipelines editor configuration"""
+
+    valid_resources = ["palette", "properties"]
+
+    @web.authenticated
+    def get(self, resource):
+        print(f'>>> retrieving {resource}')
```
kevin-bates, line 106: Please remove or convert to an appropriate log statement.

elyra/pipeline/handlers.py
```diff
@@ -92,4 +94,28 @@ async def post(self, *args, **kwargs):
...
+        if(resource in self.valid_resources):
```
kevin-bates, line 107: Python doesn't really use parenthesis by default.  I suspect some linters/IDEs will flag these as redundant.

elyra/pipeline/resources/palette.json
```diff
@@ -0,0 +1,158 @@
+{
+  "version": "3.0",
+  "categories": [
+    {
+      "label": "Notebook",
+      "image": "",
+      "id": "notebooks",
+      "description": "Notebook",
```
kevin-bates, line 8: After seeing the demo for Pipelines 2.0 in today's community meeting, should the icons for each of the palette entries be represented here?

elyra/pipeline/resources/properties.json
```diff
@@ -0,0 +1,372 @@
...
+        "label": {
+          "default": "Output Files"
+        },
+        "control": "custom",
+        "custom_control_id": "elyra-string-array-input",
+        "description": {
+          "default": "Files generated during execution that will become available to all subsequent pipeline steps.\n One filename or expression (e.g. *.csv) per line. Supported patterns: ? and *."
+        },
+        "data": {
+          "placeholder": "foo.csv",
```
kevin-bates, line 164: This doesn't look right.  Perhaps a name like `outputs.csv` or `results.csv` would be more appropriate?

elyra/pipeline/tests/test_handlers.py
```diff
@@ -0,0 +1,88 @@
...
+except ImportError:
+    # Try backported to PY<37 `importlib_resources`.
+    import importlib_resources as pkg_resources
+
+from .. import resources
+
+
+# Set Elyra server extension as enabled (overriding server_config fixture from jupyter_server)
+@pytest.fixture
+def jp_server_config():
```
kevin-bates, line 34: We should move this definition (as well as the one in the [metadata tests](https://github.com/elyra-ai/elyra/blob/7f7a1cf006232f27a6f41f252d97113d7b8b39dd/elyra/metadata/tests/conftest.py#L89-L98)) to the [parent conftest.py](https://github.com/elyra-ai/elyra/blob/master/conftest.py) since that will become the location for fixture definitions that span all subsystems.  The contents stuff will need that as well.

Note: I'm not certain if this location should be more in the elyra location, rather than the root of the repo. Will have to see how CI goes, etc.

elyra/pipeline/tests/test_handlers.py
```diff
@@ -0,0 +1,88 @@
...
+    assert payload == properties
+
+
+def expected_http_error(error, expected_code, expected_message=None):
+    """Check that the error matches the expected output error."""
+    e = error.value
+    if isinstance(e, tornado.web.HTTPError):
+        if expected_code != e.status_code:
+            return False
+        if expected_message is not None and expected_message != str(e):
```
kevin-bates, line 75: I think a strict equality test will be difficult (and see there are not calls that actually validate the message text).  I would suggest using `expected_message in str(e)` instead.  Then the dev only needs to provide the applicable segment of the message while allowing for variations (like temp filenames for example) introduced by the test framework.

elyra/pipeline/tests/test_handlers.py
```diff
@@ -0,0 +1,88 @@
...
+        return True
+    elif any([
+        isinstance(e, tornado.httpclient.HTTPClientError),
+        isinstance(e, tornado.httpclient.HTTPError)
+    ]):
+        if expected_code != e.code:
+            return False
+        if expected_message:
+            message = json.loads(e.response.body.decode())['message']
+            if expected_message != message:
```
kevin-bates, line 86: Same as previous (use `in` clause).

elyra/__init__.py
```diff
@@ -49,8 +49,11 @@ def _load_jupyter_server_extension(nb_server_app):
         (url_path_join(web_app.settings['base_url'], r'/elyra/namespace'), NamespaceHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/schedule'), PipelineSchedulerHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/export'), PipelineExportHandler),
+        (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/config/%s' % (resource_regex)),
+         PipelineConfigHandler),
```
kevin-bates, line 53: This endpoint should be added to the [elyra.yaml](https://github.com/elyra-ai/elyra/blob/master/elyra/api/elyra.yaml) open api doc.

kevin-bates (commented):
Good stuff.  Just had a few (mostly minor) comments.  Thanks.

elyra/pipeline/resources/palette.json
lresende: this is a copy of what is in master today, when we merge the new pipeline 2.0 there is going to be needed another integration pr, which can handle any additional requirements that comes from that pr.

elyra/pipeline/tests/test_handlers.py
lresende: so, this utility function is a copy from metadata, which is a copy from jupyter_server. Can we get rid of these local copies and use the one from jupyter_server? should we expose those from there?

elyra/__init__.py
lresende: As I mentioned, this part is the one I would like to hold off until the "component registry" is done.

elyra/pipeline/tests/test_handlers.py
lresende: Same as above :)

elyra/pipeline/handlers.py
lresende: My bad, left over of troubleshooting cis

elyra/pipeline/tests/test_handlers.py
lresende: had the same thought, but not sure where the location should be.

kevin-bates: Yeah.  I don't know pytest well enough to know if added to the root of the repo, then would all tests need to run from there or not.  Also, we should probably decide if we want users to be able to run all tests after installation - I think its a good idea - at least server unit tests.  It might make more sense to move the top-most file into elyra.

Regardless, we should remove the duplication with this PR, whether that means moving the top-most file down or creating another in elyra.  The contents/properties handler tests will need this as well.

elyra/pipeline/tests/test_handlers.py
kevin-bates: Yeah, this rings a bell now.   This function was used directly in `metadata/tests/test_utils.py` because, at that time, the server tests were external to the package.  However, since then, they've been moved back into the package so that users could run tests after `pip install`.  So, I agree, just use the one from server now that it's available.  Do you mind removing the function from the metadata/tests/test_utils and updating the import statement in metadata/tests/test_handlers?

I see none of the metadata tests use the message parameter, but there are some jupyter_server tests that use specific substitutions for the message parameter to allow equality to work - which I'm sure requires debugging to get that right, but that's fine (especially since none of our tests use it).

lresende: removed the duplication and we are now importing it from jupyter-server

elyra/pipeline/tests/test_handlers.py
lresende: Created #1578 for agreement on how to best refactor backend tests aiming for better code reuse.

lresende:
@kevin-bates We are thinking of merging the new pipeline editor soon. Any other changes on this one before we can merge so we can start working on integrating this to the new front-end?

PR #1553, opened 2021-04-12
Update list of documented Makefile targets

lresende:
Some Makefile targets are composed of more fine-grained
targets (e.g. build is composed by build-ui and build-server)

This will cleanup the documented targets to only display
the corase grained targets when make is invoked without commands.

 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,7 @@ container-images: elyra-image kf-notebook-image airflow-image ## Build all conta
 
 publish-container-images: publish-elyra-image publish-airflow-image publish-kf-notebook-image ## Publish all container images
 
-validate-runtime-images: ## Validates delivered runtime-images meet minimum criteria
+validate-runtime-images: # Validates delivered runtime-images meet minimum criteria
 	@required_commands=$(REQUIRED_RUNTIME_IMAGE_COMMANDS) ; \
 	pip install jq ; \
 	for file in `find etc/config/metadata/runtime-images -name "*.json"` ; do \
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,8 @@
 #
 
 .PHONY: help purge install uninstall clean test-dependencies lint-server lint-ui lint yarn-install
-.PHONY: build-ui build-server install-server
-.PHONY: watch test-server test-ui test-ui-integration-debug test docs-dependencies docs dist-ui release
+.PHONY: build-ui build-server install-server watch
+.PHONY: test-server test-ui test-integration test-integration-debug test docs-dependencies docs dist-ui release
 .PHONY: validate-runtime-images elyra-image publish-elyra-image kf-notebook-image
 .PHONY: publish-kf-notebook-image airflow-image publish-airflow-image container-images publish-container-images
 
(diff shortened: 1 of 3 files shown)


lresende:
Looks like the double hashtag comments `##` identify the documented targets. So I moved the new undocumented ones to use only one hashtag `#` which will still enable the comment to describe the target.

bourdakos1 (approved):

Makefile
```diff
@@ -123,21 +123,21 @@ install: install-server install-ui ## Build and install
...
 
-test-ui-integration: ## Run frontend cypress integration tests
+test-ui-integration: # Run frontend cypress integration tests
 	npm run test:integration
 
-test-ui-unit: ## Run frontend jest unit tests
+test-ui-unit: # Run frontend jest unit tests
 	npm run test:unit
 
 test-ui-debug: ## Open cypress integration test debugger
```
kevin-bates, line 137: This target seems out of place in this new context of only displaying top-level targets.

lresende: I thought about that, but thought this one was good for users to be aware of as it would launch the cypress debugger to help identify issues. But I agree that it feels a little out of place. If you prefer I can hide it.

kevin-bates: If we're going to be consistent, it needs to be hidden.  There are plenty of targets that would be good for users performing specific actions to be aware of and they'll just need to peruse the Makefile to discover them.

Target `validate-runtime-images` should also probably be removed from the display since its performed automatically by CI and would only produce different results when the factory runtime-images are updated (which is rare).

bourdakos1: The part of this that bothers me a bit is that the target is called `test-ui-debug`, but it only runs the integration tests. Shouldn't we call this `test-integration-debug`?

akchinSTC (approved):
+1 to hiding the debug target

Makefile
kevin-bates: `test-integration-debug` should be hidden and can be documented elsewhere, just like all the other useful (but hidden) targets.

Makefile
```diff
@@ -123,21 +123,21 @@ install: install-server install-ui ## Build and install
...
-test-ui-unit: ## Run frontend jest unit tests
+test-ui-unit: # Run frontend jest unit tests
 	npm run test:unit
 
-test-ui-debug: ## Open cypress integration test debugger
+test-ui-integration: # Run frontend cypress integration tests
+	npm run test:integration
+
+test-ui-integration-debug: # Open cypress integration test debugger
 	npm run test:integration:debug
```
bourdakos1, line 138: Shouldn't it be `test-integration` / `test-integration-debug`, because it tests the system as a whole, not just the UI?

lresende: Maybe, but as there is a possibility for us to have backend integration tests as well, and there is already a pattern of using UI for UI-related targets, I would prefer to leave it this way for now.

kevin-bates: hmm - I figured _integration_ meant front and back already.  What does integration mean within the front-end?

bourdakos1: I think the what is considered an `integration` test can be a bit of a blurred line, so I tend to not use the term. I normally opt for calling cypress tests that cover the entire system `end-to-end` tests. Then all of my ui tests are a mix `unit` and what could be considered `integration` tests (excluding the E2E tests), but I don't really make any kind of distinction between them and just call them "tests"/ "ui tests" / "unit tests"

If we want to be able to use "integration tests" for things that are exclusively frontend/backend, maybe we should also opt for a different word for tests that test the system as a whole?

ptitzler (commented):
Since you are already updating the dev doc, can you please also fix `make container-image`. It should be `make elyra-image`.

lresende:
> Since you are already updating the dev doc, can you please also fix `make container-image`. It should be `make elyra-image`.

You already have a pr for that, so we would just need to rebase on top of this one which is cleaning the makefile

Issue #1544, opened 2021-04-09
Proposal to track yarn.lock


bourdakos1:
Currently, we do not commit `yarn.lock` files. This is a highly recommended practice by `yarn` itself, see [Lockfiles should be committed on all projects](https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/). From time to time we experience build issues stemming from bugs in small version changes of dependencies. This results in wasted dev hours trying to track down exactly which packages are causing the issue, for example (https://github.com/elyra-ai/elyra/pull/1541).

I believe that the majority of these issues arise from temporary bugs in large build tools (babel, webpack, etc..), completely halting our development until they are resolved. Additionally, new contributors to Elyra could get frustrated when trying to make a contribution, but met with a failing build. Keeping a `yarn.lock` should solve this problem and guarantee a stable build for that version of the `yarn.lock`. Without a `yarn.lock` our builds can start failing, unexpectedly, at any point in time (even without any code changes from us).

A common concern for keeping a `yarn.lock` is that it will hide these bugs, showing themselves when an end user tries to install Elyra. This is very true, but only true for **runtime** dependencies (which I believe is a much rarer occurrence). Additionally, runtime dependencies normally don't result in complete system failures, but smaller UI bugs. It is also important to note that with and without the `yarn.lock`, a problematic version bump in a runtime dependency could be released the second after we release Elyra, causing issues for end users, but unseen by us when releasing. This can be combated with stricter pining of any problematic dependencies and having a cron job GitHub action that runs our integration tests on a version of Elyra that was built with the lockfile, but uses the latest allowed versions of the runtime dependencies.

lresende:
Could you please also document best practices to bring it up to date, in particular during the release process.

bourdakos1:
Normally, it will only ever be updated when you add/upgrade/remove dependencies. I don't think it is a best practice to force it to update.

lresende:
So, let's say you have jupyterlab-xxxx:^3.0.0 and there is a new lab release which updates a transient dependency that has a security fix. 

When we release Elyra, how do we make sure the lock file is pointing to the latest lab release that has the security fix?

lresende:
Btw, lab has a script to update all the dependencies (or a subset of them) and its used before a release.

bourdakos1:
[...]

PR #1555, opened 2021-04-12
Use yarn lock

bourdakos1:
This PR removes the yarn.lock file from the gitignore in order to reduce the occurrence of build-time errors.

closes: #1544
 
           in the file; or

           it.


lresende:
Should you include an initial yarn lock with the pr? And document what to do and when to update it.

lresende:
You were quick, looks like yarn.lock has already beein added.

bourdakos1:
>  And document what to do and when to update it.

@lresende nothing should really change from a development perspective

bourdakos1:
Actually, maybe we should document merge conflicts? Where should that go?

If there is a merge conflict with the `yarn.lock` you shouldn't manually resolve it, but instead just run `yarn` or `yarn install` and it will resolve the conflict for you

lresende:
> Actually, maybe we should document merge conflicts? Where should that go?
> 
> If there is a merge conflict with the `yarn.lock` you shouldn't manually resolve it, but instead just run `yarn` or `yarn install` and it will resolve the conflict for you

Maybe on the developer workflow doc?

lresende:
Also, the release script build, should it actually use the yarn lock file or be a fresh build from scratch (e.g. remove the yarn lock file first)?

bourdakos1:
> Also, the release script build, should it actually use the yarn lock file or be a fresh build from scratch (e.g. remove the yarn lock file first)?

We should keep the `yarn.lock` in the release script. We should never remove the `yarn.lock` for anything (The only exception is the possibility of removing it for the integration tests).

lresende:
Overall LGTM, but I would like to get an approval from others on the frontend side (@karlaspuldaro, @marthacryan, @ajbozarth)

bourdakos1:
@lresende I just added a section to the docs "Contributing back to Elyra" with a note about merge conflicts with the `yarn.lock`. Does that all look okay?

marthacryan (approved):
Haven't tried locally but LGTM

.github/workflows/build.yml
```diff
@@ -75,6 +75,8 @@ jobs:
     - name: Build
       run: |
         export PATH=~/.npm-global/bin:$PATH
+        # make sure nothing was manually added to the package.json
+        yarn install --frozen-lockfile
```
lresende, line 79: `make install` already calls yarn install... could you please update the makefile instead

lresende:
> @lresende I just added a section to the docs "Contributing back to Elyra" with a note about merge conflicts with the `yarn.lock`. Does that all look okay?

We have a generic [contributor guidelines](https://github.com/elyra-ai/community/blob/master/contributing.md#before-creating-a-pull-request) that is linked from every `CONTRIBUTING.MD` in all repositories. Should this actually go there, together with the section that talks about resolving conflicts?

@ptitzler Thoughts here on documentation location?

.github/workflows/build.yml
bourdakos1: `yarn install --frozen-lockfile` is something that should only need to be run in ci. It is basically just a check to make sure that a contributor didn't use `npm install` to add a package, or manually add something to the package.json and never ran `yarn install` to install things

bourdakos1: Adding `yarn install --frozen-lockfile` to `make install` would throw an error if you tried to add something to a package.json and follow up with a `make install` (something that should be okay to do). Without `yarn install --frozen-lockfile` in `make install`, adding something to the package.json and then running `make install`, would just install the packages and update the `yarn.lock`, which is normally the expected behavior

bourdakos1: The npm equivalent of this is `npm ci`: https://docs.npmjs.com/cli/v7/commands/npm-ci

bourdakos1:
Oh yes, this looks like a better place to put it, I'll remove my doc changes and open a PR there instead

ajbozarth (approved):
like Martha I didn't test this locally, but afaik it lgtm

lresende (approved):

PR #1579, opened 2021-04-16
Update OpenAPI spec to cover new contents/properties endpoint

### LINKED ISSUE #1572: Define Open API entry for contents/properties endpoint
**Describe the issue**
#1460 introduced a new endpoint (`elyra/contents/properties`) that should be documented in the Open API spec: https://github.com/elyra-ai/elyra/blob/master/elyra/api/elyra.yaml

~Also absent in that PR were unit tests to test the handler (endpoint) interactions that should be included as well.~ (Edit: as pointed out below, this was already written up in #1542)

kiersten-stokes:
Resolves #1572. 
 
           in the file; or

           it.

### DIFF
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -342,6 +342,35 @@ paths:
                     type: string
                     description: The pipeline export path.
 
+  /elyra/contents/properties/{path}:
+    get:
+        tags:
+        - properties
+        summary: Get a dictionary of properties for a file.
+        parameters:
+        - name: path
+          in: path
+          description: The path to the file to be parsed, relative to the JupyterLab server root directory.
+          required: true
+          schema:
+            type: string
+        responses:
+          200:
+            description: The dictionary of properties found while parsing the file.
+            content:
+              application/json:
+                schema:
+                  type: object
+                  description: A dictionary of property dictionaries. Each property dictionary consists of a key for each instance found and its value (if applicable), or, if parsing fails, is empty.
+                  oneOf:
+                    - $ref: '#/components/schemas/ResourceBody'
+                    - $ref: '#/components/schemas/ResourceBodyEmpty'
+          400:
+            description: File is a directory.
+          404:
+            description: File not found.
+
+  
 components:
   schemas:
     MetadataResourceBody:
@@ -490,6 +519,41 @@ components:
         overwrite:
           type: boolean
           description: Determines if the existing export should be overwritten.
+          
+    ResourceBody:
+      description: The set of resources comprising the request body.
+      required:
+        - env_vars
+        - inputs
+        - outputs
+      type: object
+      properties:
+        env_vars:
+          type: object
+          properties: 
+            ENV_VAR_NAME1:
+              type: string
+            ENV_VAR_NAME2:
+              type: string
+        inputs:
+          type: object
+        outputs:
+          type: object
+          
+    ResourceBodyEmpty:
+      description: The set of resources comprising the request body for files where no resources could be found.
+      required:
+        - env_vars
+        - inputs
+        - outputs
+      type: object
+      properties:
+        env_vars:
+          type: object
+        inputs:
+          type: object
+        outputs:
+          type: object
 
   parameters:
     namespace:
@@ -503,6 +567,13 @@ components:
       name: resource
       in: path
       description: The name of the resource in a given namespace.
+      required: true
+      schema:
+        type: string
+    path:
+      name: path
+      in: path
+      description: The path to the file to be parsed, relative to the JupyterLab server root directory.
       required: true
       schema:
         type: string
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -361,7 +361,7 @@ paths:
               application/json:
                 schema:
                   type: object
-                  description: A dictionary of property dictionaries. Each property dictionary consists of a key for each instance found and its value (if applicable), or, if parsing fails, is empty.
+                  description: A dictionary of properties.
                   $ref: '#/components/schemas/ContentsPropertiesResource'
           400:
             description: File is a directory.
@@ -528,13 +528,16 @@ components:
       properties:
         env_vars:
           type: object
+          description: A dictionary with a key for each environment variable name found and a value corresponding to the variable value or null.
           additionalProperties:
             type: string
         inputs:
           type: array
+          description: A list of input file names.
           items:
             type: string
         outputs:
           type: array
+          description: A list of output file names.
           items:
             type: string
--- a/elyra/api/elyra.yaml
+++ b/elyra/api/elyra.yaml
@@ -342,13 +342,13 @@ paths:
                     type: string
                     description: The pipeline export path.
 
-  /elyra/contents/properties/{path}:
+  /elyra/contents/properties/{filepath}:
     get:
         tags:
         - properties
         summary: Get a dictionary of properties for a file.
         parameters:
-        - name: path
+        - name: filepath
           in: path
           description: The path to the file to be parsed, relative to the JupyterLab server root directory.
           required: true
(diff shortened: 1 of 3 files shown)


elyra/api/elyra.yaml
```diff
@@ -342,6 +342,35 @@ paths:
...
+        responses:
+          200:
+            description: The dictionary of properties found while parsing the file.
+            content:
+              application/json:
+                schema:
+                  type: object
+                  description: A dictionary of property dictionaries. Each property dictionary consists of a key for each instance found and its value (if applicable), or, if parsing fails, is empty.
+                  oneOf:
+                    - $ref: '#/components/schemas/ResourceBody'
```
kevin-bates, line 366: The convention we've been using is that this name reflects the resource relative to the endpoint.  For example, we should probably use `ContentsPropertiesResource` or `ContentsPropertiesResponse`.

Looks like metadata chose to use _Resource_, while pipeline chose _Response_.  I think _Resource_ is more conformant to the open API spec terminology, but it probably doesn't make a difference.

The _Body_ term applies more to _input_ to POST and PUT calls.

elyra/api/elyra.yaml
```diff
@@ -342,6 +342,35 @@ paths:
...
+                    - $ref: '#/components/schemas/ResourceBodyEmpty'
```
kevin-bates, line 367: Since the empty response is literally an empty response (and normal), I don't think we need an explicit entity for that.

elyra/api/elyra.yaml
```diff
@@ -490,6 +519,41 @@ components:
...
+        - outputs
+      type: object
+      properties:
+        env_vars:
+          type: object
+          properties: 
+            ENV_VAR_NAME1:
+              type: string
+            ENV_VAR_NAME2:
+              type: string
```
kevin-bates, line 537: Each of the properties (`env_vars`, `inputs`, `outputs`) is a list, correct?  This should equate to type `array`.

Since `env_vars` are a list of strings, `env_vars:` should be expressed as something like:
```
env_vars:
    type: array
    items: string
```
I suspect the others (`inputs` and `outputs`) could be similarly defined as well.

elyra/api/elyra.yaml
```diff
@@ -506,3 +570,10 @@ components:
       required: true
       schema:
         type: string
+    path:
+      name: path
+      in: path
+      description: The path to the file to be parsed, relative to the JupyterLab server root directory.
```
kevin-bates, line 576: Hmm.  I'm wondering if we should make this name slightly more specific (e.g., `filepath`) and the description, less specific (eg., _The path to the file relative to the JupyterLab server root directory._) since I would imagine we'll have other endpoints that take a file.

The other option would be to name this parameter specifically for this endpoint (e.g., `contents_properties_path`) and leave the description alone, but I think making it more generic might be best at this point.

kevin-bates (commented):
This looks good Kiersten.  Just a few comments.

I would also recommend, if you haven't already, starting the server, determining the token, then enter the URL into the swagger editor at: https://editor.swagger.io/

 image

This will allow you to poke at this particular endpoint and interact with it.

elyra/api/elyra.yaml
kiersten-stokes: Ah that does make more sense, especially considering a GET request wouldn't have a body. `ContentsPropertiesResource` it is!

elyra/api/elyra.yaml
kiersten-stokes: Sounds good. I'll change the parameter description to "The path to the file". Does this also apply to the description in line 353 or should we leave that one more specific?

elyra/api/elyra.yaml
kiersten-stokes: Each property is actually it's own dictionary, since we also return the value if application. Here's an example from a real request I just made:

```
{
    "env_vars": {
        "DATASET_URL": "https://dax-cdn.cdn.appdomain.cloud/dax-noaa-weather-data-jfk-airport/1.1.4/noaa-weather-data-jfk-airport.tar.gz"
    },
    "inputs": {},
    "outputs": {}
}
```

kiersten-stokes: I was thinking about `inputs` and `outputs` while I was putting this together actually. Since they won't really be key:value pairs like environment variables can be, I think we'll eventually want to change them to arrays rather than dictionaries. But for now our implementation is an empty dictionary for each.

kevin-bates: Ok - thanks for the information.  I think, then, we want to convey these are dictionaries of strings and it appears that is done [this way](https://swagger.io/docs/specification/data-models/dictionaries/):

```
env_vars:
    type: object
    additionalProperties:
         type: string
```

Regarding the `inputs` and `outputs` entries, if we know they'll be lists of strings, I think it might be a good idea to update the implementation to return an empty list, otherwise, we could be looking at an incompatibility - both with "auto-processors" of this document but also with our frontend devs since they may want to populate results - albeit empty - and should use the correct data type.

elyra/api/elyra.yaml
kevin-bates: > Does this also apply to the description in line 353 or should we leave that one more specific?

Good question.  Currently, all parameter descriptions match their use in the endpoints section and their definition, but I would think the description in the endpoints section can specific to _that_ endpoint.  Let's leave the description alone in the endpoints sections and then see how the swager.io editor handles these - in terms of which description is shown and, if both, where those different values appear.  (More out of curiosity than anything else IMO).

kiersten-stokes: Yeah, I was curious about that too! In swagger.io, the description it uses is the one in its endpoint section. I can't see anywhere that the other description appears. I tried deleting the description altogether in the endpoint section thinking that it would default to the parameter description, but it doesn't. And in fact now that I'm testing it, deleting the whole parameters section has no affect at all as far as I can see. 

I think this is because we don't reference the parameters using `$ref` and instead are sort of re-defining them for each endpoint. In order to pull in the parameter description, we would have to have the following in the endpoint:

```
parameters:
    - $ref: '#/components/parameters/filepath'
```

elyra/api/elyra.yaml
kiersten-stokes: Sounds good! I'm happy to update the implementation as well -- would we want that in a separate PR?

kevin-bates: I think updating the implementation in this PR is fine, especially since there aren't any consumers yet.

kiersten-stokes: Makes sense!

elyra/api/elyra.yaml
kevin-bates: Ah, right.  As best as I can determine, the ability to use `$ref` for path parameters appears to be relatively new (OpenAPI 3.0).  Kinda seems like we should be using `$ref` for path parameters or remove their definitions from the bottom parameters section unless they also appear in responses. 

Hmm, since we may want richer descriptions per use, I'm leaning toward scrapping the bottom parameters section and not using references.  What do you think?

kiersten-stokes: Yeah I'm inclined to agree. Especially because we don't have so many endpoints that maintaining the descriptions per endpoint is too involved. I'll just go ahead and remove the parameters section in that case!

elyra/api/elyra.yaml
```diff
@@ -362,9 +362,7 @@ paths:
                 schema:
                   type: object
                   description: A dictionary of property dictionaries. Each property dictionary consists of a key for each instance found and its value (if applicable), or, if parsing fails, is empty.
```
kevin-bates, line 364: I think we should shorten this description to something more like `A dictionary of properties.`, then add descriptions on each of the properties in the reference itself since their types now vary (dictionary vs. list).

The current description is obsolete as we raise exceptions on failures and empty sets are only returned when the type of file is not recognized - which I'm not sure needs to mentioned here.

kiersten-stokes: Yeah, I just caught that too! Thoughts on the below descriptions? The `env_vars` one is still perhaps a bit too wordy.

```
env_vars:
          type: object
          description: A dictionary with a key for each environment variable name found and a value corresponding to the variable value or null.
          additionalProperties:
            type: string
inputs:
          type: array
          description: A list of input file names.
          items:
            type: string
outputs:
          type: array
          description: A list of output file names.
          items:
            type: string
```

kevin-bates: That does seem wordy, but it's difficult.  I think what you have is fine.

kevin-bates (approved):
Looks good - thanks @kiersten-stokes.

PR #1588, opened 2021-04-20
Pin documentation links to stable/specific Elyra version

### LINKED ISSUE #1566: Documentation and container image tag `latest`  have different semantic meanings
In the context of Elyra container images the `latest` tag refers to the more current official release, e.g. as of 4/14 it should resolve to `2.2.1`. 

However, the documentation `https://elyra.readthedocs.io/` resolves to https://elyra.readthedocs.io/en/latest/?badge=latest, which is the documentation associated with the master branch in https://github.com/elyra/ai/elyra. The equivalent of `latest` (2.2.1 as of today) appears to be `stable` https://elyra.readthedocs.io/en/stable/

While this may seem like a nit, it does have a profound impact because in the `latest` document versions the tag/string `:dev` is displayed as is, whereas it is replaced with the actual version number during the release process. 

Example:
- User wants to install/use the latest Elyra release (2.2.1)
- User opens https://elyra.readthedocs.io/, which resolves to `latest` by default
- User browses documentation, which  references the `dev` tag. If the user follows the instructions, the most current dev build is installed, not the most current official release. Most users should not be installing `:dev`, unless they need access to code that has not been formally released.

ptitzler:
This PR
 - updates the documentation in link the repository README (now pointing to the stable documentation version instead of the development version)
 - updates the release script to pin the Launcher's Elyra documentation link to the installed Elyra version
 - adds a warning message if a user accesses the documentation development version 
    **Rendered page on `latest` version:**
   
    image

    **Rendered page on other versions:**
    image

 
The PR does not_close #1566 because there is additional work that cannot be delivered in a PR.
 
           in the file; or

           it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ limitations under the License.
 [![PyPI version](https://badge.fury.io/py/elyra.svg)](https://badge.fury.io/py/elyra)
 [![Anaconda-Server Badge](https://anaconda.org/conda-forge/elyra/badges/version.svg)](https://anaconda.org/conda-forge/elyra)
 [![Downloads](https://pepy.tech/badge/elyra)](https://pepy.tech/project/elyra)
-[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=latest)](https://elyra.readthedocs.io/en/latest/?badge=latest)
+[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=stable)](https://elyra.readthedocs.io/en/stable/?badge=stable)
 [![GitHub](https://img.shields.io/badge/issue_tracking-github-blue.svg)](https://github.com/elyra-ai/elyra/issues)
 [![Gitter](https://badges.gitter.im/elyra-ai/community.svg)](https://gitter.im/elyra-ai/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ limitations under the License.
 [![PyPI version](https://badge.fury.io/py/elyra.svg)](https://badge.fury.io/py/elyra)
 [![Anaconda-Server Badge](https://anaconda.org/conda-forge/elyra/badges/version.svg)](https://anaconda.org/conda-forge/elyra)
 [![Downloads](https://pepy.tech/badge/elyra)](https://pepy.tech/project/elyra)
-[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=stable)](https://elyra.readthedocs.io/en/stable/?badge=stable)
+[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=latest)](https://elyra.readthedocs.io/en/latest/?badge=latest)
 [![GitHub](https://img.shields.io/badge/issue_tracking-github-blue.svg)](https://github.com/elyra-ai/elyra/issues)
 [![Gitter](https://badges.gitter.im/elyra-ai/community.svg)](https://gitter.im/elyra-ai/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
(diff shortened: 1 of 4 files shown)


lresende:
I really think that the main place to change is in the launcher so that when in a released version, link to the documentation of a specific version. This would be changed on the [theme package](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/packages/theme/src/index.ts#L124)

README.md
```diff
@@ -20,7 +20,7 @@ limitations under the License.
 [![PyPI version](https://badge.fury.io/py/elyra.svg)](https://badge.fury.io/py/elyra)
 [![Anaconda-Server Badge](https://anaconda.org/conda-forge/elyra/badges/version.svg)](https://anaconda.org/conda-forge/elyra)
 [![Downloads](https://pepy.tech/badge/elyra)](https://pepy.tech/project/elyra)
-[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=latest)](https://elyra.readthedocs.io/en/latest/?badge=latest)
+[![Documentation Status](https://readthedocs.org/projects/elyra/badge/?version=stable)](https://elyra.readthedocs.io/en/stable/?badge=stable)
```
lresende, line 23: This should not be changed permanently. It would be ok to change in the release script.

create-release.py
```diff
@@ -239,7 +239,7 @@ def update_version_to_dev() -> None:
 
         sed(_source('packages/theme/src/index.ts'),
             r"https://elyra.readthedocs.io/en/stable/",
-            f"https://elyra.readthedocs.io/en/latest/")
+            f"https://elyra.readthedocs.io/en/v{new_version}/")
```
lresende, line 242: You need to consider dev->release and then release->dev and this is only handling on side of it.

README.md
ptitzler: Wy not? [New] users are likely going to access the documentation from the readme if they don't have Elyra installed yet and they should see by default  the documentation for code that is published, not being worked on.

ptitzler:
> I really think that the main place to change is in the launcher so that when in a released version, link to the documentation of a specific version. This would be changed on the [theme package](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/packages/theme/src/index.ts#L124)



> I really think that the main place to change is in the launcher so that when in a released version, link to the documentation of a specific version. This would be changed on the [theme package](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/packages/theme/src/index.ts#L124)

How is that different from the proposed release script change which updates that file? Ideally `index.ts` would just retrieve the current version number (and embed it in the URL at runtime) but I don't see an API to obtain it.

README.md
lresende: The badge describes the build status for documentation. When looking into master README it should describe the status of the master documentation build. 

I believe what we agreed on was to change the "right side documentation link" to point to stable. 

image

lresende:
> > I really think that the main place to change is in the launcher so that when in a released version, link to the documentation of a specific version. This would be changed on the [theme package](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/packages/theme/src/index.ts#L124)
> 
> How is that different from the proposed release script change which updates that file? Ideally `index.ts` would just retrieve the current version number (and embed it in the URL at runtime) but I don't see an API to obtain it.

When you run the build script, all the version info is collected in the [generate config method](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/create-release.py#L456). Then, there is another function that performs changes from [dev_to_release](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/create-release.py#L97) which is how the code will be on the release tag, and after that, it[ brings back to dev](https://github.com/elyra-ai/elyra/blob/c30928925c9b3b7d593940e697b748d23428d7df/create-release.py#L175) which is how the master will show up. 

Your change is only handling one side of it.

README.md
ptitzler: Ah, I missed that part. Do note that the JupyterLab badge in their repo points to the stable version!

lresende (approved):

PR #1591, opened 2021-04-20
Enable creation of pipelines specific to a runtime

### LINKED ISSUE #1003: Enable creation of pipelines specific to a runtime
Users should be able to create a generic pipeline that can run on any support runtime:
   - Only a subset of components are supported
      - Notebook and Python Scripts?
   - Local mode is enabled

Users can create pipelines for specific runtimes which then will leverage specific features from the selected runtime
 - Leverage the full set of components from the specific runtime component registry
 - Moving from one runtime to another will not be allowed
 - Local mode is not supported

image

marthacryan:
Fixes #1003. 

Todo:
- [x] Add buttons to launcher
- [x] Update dialog logic to take runtime into account
- [x] Fix order of launcher buttons
- [x] Change airflow icon? 
 
           in the file; or

           it.


marthacryan:
Current appearance changes:
image
image
image

marthacryan:
@lresende Thoughts on the airflow logo I used / the order of the cards on the launcher?

ptitzler:
image

Shouldn't the labels be "Apache Airflow Pipeline Editor" and "Kubeflow ... Editor"? As is the label text seems to imply that by clicking the buttons one creates a runtime [configuration].

ptitzler:
To maintain visual consistency, should the runtime platform entry be rendered as a select list with a single locked entry?

image


Side note (since this is a WIP), as shown the selection is not correct (KFP/AA is mixed)

marthacryan:
@ptitzler Looks like the full "Kubeflow Pipelines runtime Editor" is too long for the card to show: 
image

Would "KFP Editor" work? Also after giving this a second look we probably shouldn't have the word "runtime" in those titles either

ptitzler:
Re order it'd vote for gen, kfp, aa (aa is least likely to be used)

ptitzler:
> @ptitzler Looks like the full "Kubeflow Pipelines runtime Editor" is too long for the card to show:
> image
> 
> Would "KFP Editor" work? Also after giving this a second look we probably shouldn't have the word "runtime" in those titles either

It's not really an acronym that seems to be commonly used. What if the AA icon includes the text "Apache Airflow" on the bottom and the button text (for both) would be "Pipeline Editor"?

image

marthacryan:
@ptitzler How's this?
image

ptitzler:
> @ptitzler How's this?
> image

Super!

marthacryan:
Updated dialogs:
image
image

@ptitzler thoughts on the wording? Should I use something other than "Run pipeline **in** <runtime>"?

kevin-bates:
Seems like _run_ needs to use **in** or **on**: 
"Run pipeline **in** Kubeflow Pipelines" or 
"Run pipeline **on** Kubeflow Pipelines"

I suppose **using** could work as well:
"Run pipeline **using** Kubeflow Pipelines"


For _export_, I want to say the preposition (yes, I had to look that up) should be something more like **for**: 
"Export pipeline **for** Kubeflow Pipelines"

but, I defer to @ptitzler, et al.

ptitzler:
We've used "on" in the past in our article drafts and the editor's didn't object to that.

ptitzler:
... and "for" for export like Kevin mentioned.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -297,10 +298,20 @@ const PipelineWrapper: React.FC<IProps> = ({
       RequestErrors.serverError(error)
     );
 
+    const pipelineRuntime = pipeline?.pipelines[0]?.app_data?.ui_data?.runtime;
```
bourdakos1, line 301: I think this would need to be:
```ts
pipeline?.pipelines?.[0]?.app_data?.ui_data?.runtime
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -417,12 +428,19 @@ const PipelineWrapper: React.FC<IProps> = ({
     };
     schema.unshift(JSON.parse(JSON.stringify(localSchema)));
 
+    let title = 'Run pipeline';
+    const pipelineRuntime = pipeline?.pipelines[0]?.app_data?.ui_data?.runtime;
```
bourdakos1, line 432: same as above comment

packages/pipeline-editor/src/PipelineExportDialog.tsx
```diff
@@ -48,7 +49,7 @@ export class PipelineExportDialog extends React.Component<IProps, IState> {
   state = {
     displayedRuntimeOptions: new Array<IRuntime>(),
     fileTypes: new Array<Record<string, string>>(),
-    selectedRuntimePlatform: '',
+    selectedRuntimePlatform: this.props.runtime?.name || '',
```
bourdakos1, line 52: I'm a little confused where this state is being use? Seems like it's only ever accessed in componentDidMount?

Also, I would use nullish coalescing if possible: `this.props.runtime?.name ?? ''`

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
```diff
@@ -33,7 +34,9 @@ interface IState {
 export class PipelineSubmissionDialog extends React.Component<IProps, IState> {
   state = {
     displayedRuntimeOptions: new Array<IRuntime>(),
-    selectedRuntimePlatform: this.props.schema[0] && this.props.schema[0].name,
+    selectedRuntimePlatform:
+      this.props.runtime?.name ||
+      (this.props.schema[0] && this.props.schema[0].name),
```
bourdakos1, line 39: If avoidable, we should try to not use props inside state

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
```diff
@@ -88,22 +91,26 @@ export class PipelineSubmissionDialog extends React.Component<IProps, IState> {
...
+          <div>
+            <label htmlFor="runtime_platform">Runtime Platform:</label>
+            <br />
+            <select
+              id="runtime_platform"
+              name="runtime_platform"
+              className="elyra-form-runtime-platform"
+              data-form-required
+              defaultValue={selectedRuntimePlatform}
+              onChange={this.handleUpdate}
```
bourdakos1, line 104: if we use `onChange` to update the `selectedRuntimePlatform`, then `defaultValue` should just be `value`

packages/pipeline-editor/src/index.ts
```diff
@@ -144,21 +150,68 @@ const extension: JupyterFrontEndPlugin<void> = {
...
+          ? 'New Pipeline Editor'
+          : args['runtime'] && args['runtime']['display_name']
+          ? 'Pipeline Editor'
+          : 'Generic Pipeline Editor';
+      },
+      icon: (args: any) => {
+        if (args['isPalette']) {
+          return undefined;
+        } else {
+          if (args['runtime'] && args['runtime']['display_name']) {
```
bourdakos1, line 164: this could be simplified to `args.runtime?.display_name`

packages/pipeline-editor/src/index.ts
```diff
@@ -170,19 +223,42 @@ const extension: JupyterFrontEndPlugin<void> = {
       category: 'Elyra'
     });
 
+    const schema = await PipelineService.getRuntimesSchema().catch(error =>
+      RequestErrors.serverError(error)
+    );
```
bourdakos1, line 228: this should probably be a try/catch if we are  using async/await

bourdakos1 (changes requested):

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
marthacryan: What would be the better way to go about setting the state to that value (`this.props.runtime?.name`) by default?

packages/pipeline-editor/src/PipelineExportDialog.tsx
marthacryan: Oh good catch, not sure why exactly but I'm suspecting it's from copying over code from `PipelineSubmissionDialog`. Just removed it and it's working as expected.

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
bourdakos1: I would try to construct it outside of state and leave state for only storing the information that comes from some sort of event (user input / network request / etc...)

packages/pipeline-editor/src/index.ts
bourdakos1: I meant you should be able to do something like this: 
```
let schema 
try {
  schema = await PipelineService.getRuntimesSchema()
} catch (e) {
  RequestErrors.serverError(e)
}
```

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
bourdakos1: something like this: https://codesandbox.io/s/sharp-waterfall-iymcv?file=/src/App.tsx

marthacryan:
Since we're removing the display name of each runtime from the launcher cards, the only place the display name shows up is in the header of the export/run dialogs.
image
Is it okay to use the original display name so that the other extensions that use that field (i.e. runtimes editor) don't need to change?

ptitzler:
We might want/need to expose the display name in other places and the appended "runtime" string likely would need to be stripped off. (From a translation point of view it'd be also better.) It's always easier to append something than to remove something. I would therefore vote for making the change even though it's a bit more work atm.

lresende:
@marthacryan it looks like pipelines are now embedding the runtime schema to the pipeline file, is that by design?

karlaspuldaro:
Now that the pipeline types are being split, maybe we should treat the `run` dialog the same way `submit/export` handles empty runtimes
 
instead of displaying an empty runtime configuration dropdown
image
This wasn't previously handled for run pipeline dialog because there would always be the local run option, which now is only the case for the generic pipeline.

ptitzler:
> maybe we should 

+1, with the caveat that the message should include include the runtime display name, e.g. `No runtime configuration for Kubeflow Pipelines is defined.` to make it clear that the user is working with a _typed_ pipeline.

packages/pipeline-editor/src/index.ts
```diff
@@ -170,18 +223,43 @@ const extension: JupyterFrontEndPlugin<void> = {
...
+        );
+        for (const runtime of schema) {
+          menu.fileMenu.newMenu.addGroup(
+            [
+              {
+                command: openPipelineEditorCommand,
+                args: { runtime }
+              }
+            ],
+            runtime.name === 'kfp' ? 2 : runtime.name === 'airflow' ? 3 : 4
```
karlaspuldaro, line 258: For some reason the generic pipeline editor isn't following the specified order
image

ptitzler: Slightly off-topic - in this case we should probably not rely on the icon to identify the typed editor. In this context menu the icons are really small and users might find it hard to distinguish them. Can we use `Kubeflow Pipelines Pipeline Editor` and `Apache Airflow Pipeline Editor` as text instead? We are already doing this for the script editors (:

marthacryan:
Dialog that now shows up when running/exporting a pipeline that has runtime specified if no runtime configs for that runtime exist:
image
image

packages/pipeline-editor/src/index.ts
marthacryan: Ohh ok sounds good I'll add that

marthacryan: So the rest of the code that uses the `schema` variable will need to be in that try as well though, right? Also what is the benefit of using try/catch over the await/catch?

bourdakos1: Schema will just be undefined if it throws an error 

.catch() is just using the promise syntax, there’s nothing wrong with it, but we should stick to async/await or the promise syntax and not mix and match

Also, typescript wise, I think this would be consider schema as defined instead of potentially undefined:
```
const schema = await getSchema().catch()
```

marthacryan: Oh ok thanks for explaining - is the way it's implemented currently (see https://github.com/elyra-ai/elyra/pull/1591/files#diff-2280d00b71af312d9eecec1377484159361d5145193a35ef246a7fa9a37abaf1R234) using a `then` with resolve/reject handlers and all logic that uses the `schema` variable equivalent though?

packages/pipeline-editor/src/PipelineSubmissionDialog.tsx
```diff
@@ -62,20 +58,16 @@ export class PipelineSubmissionDialog extends React.Component<IProps, IState> {
...
 
   render(): React.ReactNode {
     const { name } = this.props;
-    const {
-      displayedRuntimeOptions,
-      selectedRuntimePlatform,
-      validSchemas
-    } = this.state;
+    const { selectedRuntimePlatform, validSchemas } = this.state;
+    const displayedRuntimeOptions = this.updateRuntimeOptions(
```
bourdakos1, line 68: maybe just rename this to `getRuntimeOptions` since this doesn't update anything anymore, but other than that I think this looks good to go

bourdakos1 (approved):

marthacryan:
New file menu:
image

ptitzler:
Come to think of it, in the context of "File" > "New" wouldn't it be semantically more appropriate  to use "X file" instead of "X Editor" as text? Granted, the X task opens an X editor but also creates an untitled file of type X, the latter being the intention of the user.

bourdakos1:
I agree with @ptitzler, but also what is the difference a “python file” and the “python editor”, shown there?

karlaspuldaro:
> what is the difference a “python file” and the “python editor”, shown there?
Not sure why we see both, I just tried master and only see python/r editor options
image

I also agree with @ptitzler on making them all "files". This can go on a separate issue.

ptitzler:
> what is the difference a “python file” and the “python editor”, shown there?

Apparently there's another Python editor out there. I remember seeing an image in a blog post.

bourdakos1:
If someone were to click on either of these, would they both result in the same editor being opened? Or does it respect the editor that put the menu item there? 

Does this menu option:
A) open a new X editor and the editor happens to create the file as a side-effect
or
B) create a new X file and open the default editor for that file type

karlaspuldaro:
> Does this menu option:
> A) open a new X editor and the editor happens to create the file as a side-effect
> or
> B) create a new X file and open the default editor for that file type

I believe it's B) based on [this code](https://github.com/elyra-ai/elyra/blob/213b93ca527b4544c000b98a803fa44f77cbfbb5/packages/python-editor/src/index.ts#L210) where we handle the file menu command for python files

karlaspuldaro (approved):
Looks good. Thanks Martha!

PR #1595, opened 2021-04-20
[WIP] LSP support for pipeline files

bourdakos1:
This PR uses a language server for validating pipeline files. Please note that I am not a Python person so this code might be really yucky 😁


elyra/lsp/client.py
```diff
@@ -0,0 +1,100 @@
...
+            if line.startswith("Content-Length: "):
+                line = line[len("Content-Length: "):]
+                message_size = int(line)
+
+        jsonrpc_res = self.pipe.read(message_size).decode("utf-8")
+        self.res = json.loads(jsonrpc_res)
+        self.done = True
+
+
+class LSPClient:
```
akchinSTC, line 56: Not sure how to phrase this, is the front end e..g VPE going to be communicating with the LSP in real time or just a one-off validation on submission? e.g. triggering validation when something is dragged or changed on the canvas. If so do we need to implement more of the ls protocol in the client? like  `"method": "textDocument/didChange",`

elyra/lsp/yarn.lock
```diff
@@ -0,0 +1,65 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
akchinSTC, line 1: a bit weird to have npm related files in the python dirs, possible we can find another place for these?

akchinSTC (commented):
This is cool. Assuming all the work and `rules` for the validation happen in the vscode-language server, have any good links that check out for a good overview?
Could also provide the info for the `generic_pipeline_core` python package you used, or a description of what that does?

elyra/__init__.py
```diff
@@ -20,7 +20,7 @@
 from .api.handlers import YamlSpecHandler
 from .metadata.handlers import MetadataHandler, MetadataResourceHandler, SchemaHandler, SchemaResourceHandler, \
     NamespaceHandler
-from .pipeline import PipelineExportHandler, PipelineSchedulerHandler, PipelineProcessorManager, PipelineConfigHandler
+from .pipeline import PipelineExportHandler, PipelineSchedulerHandler, PipelineProcessorManager, PipelineConfigHandler, PipelineNodesHandler
```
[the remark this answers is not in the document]
bourdakos1: This handler doesn't do any validation, it just serves a list of nodes/components

elyra/lsp/client.py
bourdakos1: The LSP and the VPE use (this validation package)[https://github.com/bourdakos1/pipeline-editor/tree/create-node/packages/pipeline-services/src/validation] for validation logic. Right now this only uses the LSP for validation at submission time, while the VPE uses the validation package directly in real time. I don't know if we will ever have the VPE use the LSP directly since it has full access to the validation logic without having to make requests to the server

elyra/lsp/yarn.lock
bourdakos1: I agree, it is a bit awkward, is there a place we can put them where they will still be packed up in the python stuff?

bourdakos1:
@akchinSTC 

This is the package with the validation logic:
https://github.com/bourdakos1/pipeline-editor/tree/create-node/packages/pipeline-services/src/validation

This is the actual language server:
https://github.com/elyra-ai/vscode-pipeline-editor/tree/lsp-updates/packages/language-server/

This is the project that generates the python and npm package `generic-pipeline-core` which is just the list of nodes/components. The language server needs this information to validate the node properties so it is bundled as a part of the language server as an npm package. The UI also needs this information so it is served to the UI via a pip package in Jupyter and via an npm package in vscode:
https://github.com/bourdakos1/generic-pipeline-core/tree/simple

bourdakos1:
basically, JSON is generated from [this file](https://github.com/bourdakos1/generic-pipeline-core/blob/simple/src/nodes.ts) and wrapped in python and javascript for easier ingestion and updates

akchinSTC:
In a nutshell, this way is nice but it brings up the issue of maintainability. I myself am not a front-end developer so the idea of needing to write validation rules and supporting in TS is a little daunting...haha. So the implementation of our current set of validation rules, along with any future updates, of which there will be many, will fall on you @bourdakos1  and the front-end devs on the team. Instead of tacking on more work, I'd rather this part live in the backend (in python) also since aligns with the 'jupyter' way of doing things. 

This leads to the next question, can we use a python-based implementation LS instead?  In this WIP,
```
Frontend -> direct access to the validation logic, no need to access LS
Backend -> payload is piped to TS-based LS via subprocesses
```
Using a python LS would mean we would need to update two sets of rules instead of one(the set in the python LS and the set in https://github.com/bourdakos1/pipeline-editor/tree/create-node/packages/pipeline-services/src/validation) 
So we would need to try the ol switcheroo, where front end would need to use a client?  like pipeline submission to check so only need to maintain a single set of rules.

bourdakos1:
@akchinSTC yea we could use a python LS instead, but it might be a lot of work to connect to the LS from the UI. It would also introduce python as a requirement, but that shouldn’t be too much of a problem

bourdakos1:
Would it make sense to go forward with the node LS for now and then work on building a python LS and interfacing in parallel?

akchinSTC:
@bourdakos1 - not just yet, lets put a pin in this. came into some conflicting info and need to reframe my thoughts my previous comments.

akchinSTC:
Pushed this working branch to `contrib-pipeline-language-server`. 
https://github.com/elyra-ai/elyra/tree/contrib-pipeline-language-server

PR #1620, opened 2021-04-26
Add support for pipeline runtime specific components

### LINKED ISSUE #1004: Add support for pipeline runtime specific components
Enable the notion of a registry that can parse and expose the necessary information required to expose runtime specific components on the Elyra pipeline editor. 

image

Examples of KFP Components:
 - https://github.com/kubeflow/pipelines/tree/master/components
Example of Airflow Components:
 - https://github.com/apache/airflow/tree/1.10.15/airflow/operators
 - https://github.com/apache/airflow/tree/1.10.15/airflow/contrib/operators

Things to consider:
- Abstraction of classes to enable the concept of "federated registry" that reads from different sources/locations/types
- Caching on multiple layers: Frontend, Service, Filesystem contents

### LINKED ISSUE #1152: Does pipeline editor support Kubeflow pipeline components directly?
I am trying elyra recently and notice pipeline editor interface ask "start your new pipeline by dragging files from the file browser pane"

Seems it will use python files or notebooks as a step in Kubeflow Pipeline. It's a little bit hard for existing kubeflow pipeline who are familiar with Ops. 

Does pipeline editor supports https://github.com/kubeflow/pipelines/tree/master/components directly? I think it would be great to drag those components in the editor and the experiences will be very close to how we use `kfp` sdk to build the pipeline.

### LINKED ISSUE #1659: Pipeline editor stopped supporting canvas palette format
**Describe the issue**
Pipeline editor stopped supporting canvas palette format. With the new component registry, that has the ability to generate the palette format supported by the canvas, we shouldn't need to perform multiple conversions from Palette json -> node javascript model -> Palette json. 

We should invoke the existing services implemented in #1551 
- /elyra/pipeline/config/palette
- /elyra/pipeline/config/properties

This will enable the easy transition to the new component registry services.

Note, we should also make it simpler to support calls to retrieve a palette by a given runtime, and properties based on the component being used. If it makes it simpler, we can make the fix based on #1620

### LINKED ISSUE #1736: Unable to select a notebook file when adding a notebook component
**Describe the issue**
Unable to select a notebook file when attempting to add a nodebook note component to the pipeline editor, can only select *.R files

**To Reproduce**
1. Go to Palette in the properties panel
2. Drag a new notebook component to the canvas
3. go to properties and click on browse, for filename section

**Screenshots or log output**
image

**Expected behavior**
to be able to select a *.ipynb file from the browser

**Deployment information**
 - Elyra version: master/ component-registry 
 - Operating system: macOS
 - Installation source: source
 - Deployment type:

### LINKED ISSUE #1737: Local pipeline fails to run when using relative script
**Describe the issue**
Local pipeline fails to run when using relative script

**To Reproduce**
1.Create a pipeline on `elyra-pipelines`
2.Add a script `factorial.r` from a sibling directory `sample-script`
3.See error:
image

### LINKED ISSUE #1738: Include subdirectories checkbox should be better formatted
**Describe the issue**
checkbox associated with include subdirectories property is a bit clunky

**To Reproduce**
1. Open node properties for any notebook, r script or python script node
2. See error

**Screenshots or log output**
image
**Expected behavior**
Maybe have the checkbox displayed inline with the text

**Deployment information**
 - Elyra version: 2.3.0/ component registry
 - Operating system: macos
 - Installation source: source

### LINKED ISSUE #1739: Unable to edit node properties of nodes inside of a supernode
**Describe the issue**
Unable to edit node properties of nodes inside of a supernode

**To Reproduce**
1. Create a supernode
2. Expand the supernode flow
3. open the properties of a node

**Screenshots or log output**
Screen Shot 2021-06-09 at 5 12 30 PM

### LINKED ISSUE #1740: Pipeline node property name description defaults to notebook file
**Describe the issue**
Pipeline editor name field description in node properties tab defaults to notebook file independent of the node type

**To Reproduce**
1. Open pipeline editor
2. Add a Python or R script node
3. Select node and open properties tab
4. See the description as below
image

**Expected behavior**
File type description should match node type

**Deployment information**
 - Elyra version: 2.3.0/ component registry
 - Operating system: macOS
 - Installation source: source

kiersten-stokes:
This PR covers the following:

- [x] Read and parse locally-configured component definitions to acquire palette & properties JSON dynamically
  - [x] KFP YAML files
  - [x] Airflow operator python files
developers/users can best understand the use of each input and output parameter
- [x] Filter returned JSON into the UI
- [x] Support execution of new components
  - [x] KFP
  - [x] Airflow
- [ ] Determine which components should be provided locally and add files as needed (see #1698 for KFP) (see #1728 for Airflow)
  - ~~[ ] Massage any pre-configured component specifications to contain as much information as possible such that parsing is as smooth as possible~~

### Current status:

**Must fix for merge**
- [x] Kfp - runtime image validation - Fix in progress-  https://github.com/elyra-ai/pipeline-editor/pull/124
- [x] Airflow - Available operators, and filling up fields on submit - Fix in progress-  https://github.com/elyra-ai/pipeline-editor/pull/124
- [x] Elyra_Airflow_class_name defaulting to empty array instead of string

**Must fix before release**
- [x] Move registry to outside package
- [ ] Validate the list of components
   - [ ] KFP
   - [x] Airflow
- [ ] CI front end test failures

**Nice to have**
- [ ] Change from each component to a location of multiple components
- [ ] Cache
- [ ] make current operators into YAML files

### Potential future directions and changes (separate PRs)
- Add dependency support for non-standard components (currently dependencies are only handled for Notebooks and scripts)
- Change the way components are accessed/queried for their palette and property information (e.g. maybe a folder where all components can be added)
- Implement caching for quicker palette and properties return
- Support ability for the user to define custom components
- Support ability for the user to drag and drop a component YAML/operator directly from the file browser
- Support ability for the user to permanently add components to the the list of locally-configured components
- Determine and document a standard set of component properties that would provide the best UI experience
- Remove custom execution of notebooks and scripts in favor of using runtime-specific components for the same purpose
- Add tests

Fixes #1004
Fixes #1152
Fixes #1659
Fixes #1736
Fixes #1737
Fixes #1738
Fixes #1739
Fixes #1740

 
           in the file; or

           it.


lresende:
@kiersten-stokes I made a few suggestions as a pr to your branch.

elyra/pipeline/handlers.py
```diff
@@ -96,21 +98,47 @@ async def post(self, *args, **kwargs):
         self.finish(json_msg)
 
 
-class PipelineConfigHandler(HttpErrorMixin, APIHandler):
-    """Handler to expose method calls to retrieve pipelines editor configuration"""
+class PipelineComponentHandler(HttpErrorMixin, APIHandler):
+    """Handler to expose method calls to retrieve pipelines editor component configuration"""
 
-    valid_resources = ["palette", "properties"]
+    valid_processors = ["local", "kfp", "airflow"]
```
kevin-bates, line 104: We should expose a `get_processor_types() -> List` method on `PipelineProcessorRegistry` from which these are derived.  This way it's always based on the set of "registered" processors.

The implementation should be as simple as:
```python
def get_processor_types() -> List:
    return self._processors.keys()
```

elyra/pipeline/processor.py
```diff
@@ -69,22 +70,28 @@ def __init__(self, **kwargs):
...
                 self.log.error('Error registering processor "{}" - {}'.format(processor, err))
 
-    async def process(self, pipeline):
-        processor_type = pipeline.runtime
+    def _get_processor_for_runtime(self, processor_type: str):
         processor = self._registry.get_processor(processor_type)
 
         if not processor:
-            raise RuntimeError('Could not find pipeline processor for [{}]'.format(pipeline.runtime))
+            raise RuntimeError('Could not find pipeline processor for [{}]'.format(processor_type))
```
kevin-bates, line 77: Should `get_processor()` raise instead?  It looks like this kind of condition is used elsewhere.  

If we need to _validate_ that a given type has an associated processor, we should look into implementing `__contains__` (similar to what [`MultiKernelManager` does to check if a given `KernelManager` is present](https://github.com/jupyter/jupyter_client/blob/98faaf91d00b4fbf17b2fbf0b08b1fd23adc27bf/jupyter_client/multikernelmanager.py#L164-L165)).

elyra/pipeline/processor.py
```diff
@@ -156,6 +163,12 @@ def __init__(self, root_dir, **kwargs):
     def type(self):
         raise NotImplementedError()
 
+    @abstractmethod
```
kevin-bates, line 166: If this has an implementation, it should not be an abstract method.

elyra/pipeline/processor.py
```diff
@@ -243,6 +256,16 @@ def _sort_operation_dependencies(operations_by_id: dict, ordered_operations: lis
 
 
 class RuntimePipelineProcess(PipelineProcessor):
+    component_registry: ComponentRegistry = ComponentRegistry()
+
+    def get_components(self):
+        # component_registry = ComponentReader.get_instance(registry_type='file', processor_type=self.type)
```
kevin-bates, line 262: I'm confused.  What is the difference between this method and the same method and identical implementation on `PipelineProcessorManager`?

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+    "runtime_image": "",
+    "outputs": [],
+    "env_vars": [],
+    "dependencies": [],
+    "include_subdirectories": False
+}
+
+
+def get_id_from_name(name):
+    return ' '.join(name.lower().replace('-', '').split()).replace(' ', '-').replace('_', '-')
```
kevin-bates, line 61: Please add a docstring indicating what this is doing and why it's necessary.  The removal of `-` in the original, then its re-introduction for space and underscores is a little funky.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+        }
+    }
+
+    return node_type
+
+
+class ComponentParser(SingletonConfigurable):
+    _type = "local"
+    properties: dict() = {}
+    parameters: dict() = {}
```
kevin-bates, line 89: I don't think this syntax is correct since `dict()` is equivalent to `{}`.  I suspect what is meant is:
```python
    properties: Dict = {}
    parameters: Dict = {}
```
where `Dict` comes `from typing import Dict`.

Assuming `self.parameters` is initialized like `self.properties`, these should probably be declared w/o the instance:
```python
    properties: Dict
    parameters: Dict
```

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.properties = self.get_common_config('properties')
+
+    def get_common_config(self, config_name):
+        common_dir = os.path.join(os.path.dirname(__file__), 'resources')
```
kevin-bates, line 96: We should declare `resources_dir` statically and use it here and when constructing `catalog_file` below.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+
+        return common_json
+
+    def _get_component_catalog_json(self):
+        catalog_dir = os.path.join(os.path.dirname(__file__), "resources")
+        catalog_file = os.path.join(catalog_dir, f"{self._type}_component_catalog.json")
+        with open(catalog_file, 'r') as f:
+            catalog_json = json.load(f)
+
+        return catalog_json['components']
```
kevin-bates, line 109: Rather than a List, should this return a Dict indexed by the component's id?  This way, `list_all_components()` becomes: `return self._get_component_catalog_json().values()` and `return_component_if_exists(id)` becomes `return self._get_component_catalog_json().get(id)`.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+        for component in self._get_component_catalog_json():
+            if component['id'] == component_id:
+                return component
+        return None
+
+    def add_component(self, request_body):
+        component_json = {
+            'name': request_body['name'],
+            'id': get_id_from_name(request_body['name']),
+            'path': request_body['path']
```
kevin-bates, line 130: Should we assert `path` has a value here rather than in `list_all_components()`?

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        catalog_json = self._get_component_catalog_json()
+        catalog_json['components'].append(component_json)
+
+        catalog_dir = os.path.join(os.path.dirname(__file__), "resources")
+        catalog_file = os.path.join(catalog_dir, f"{self._type}_component_catalog.json")
+        with open(catalog_file, 'a') as f:
```
kevin-bates, line 138: Since every component is being written to the file, we don't want `append` here. `'w'` mode should be sufficient to rewrite the file.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+            f.write(json.dumps(catalog_json))
+
+    def parse_component_details(self, component, component_name):
+        """Get component name, id, description for palette JSON"""
```
kevin-bates, line 142: Since this class isn't abstract, this should probably raise `NotImplementedError` if this method is required to be overwritten.  If it's optional, `pass` is fine.  

Same comment applies to `parse_component_properties()`

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+    def parse_component_properties(self, component):
+        """Get component properties for properties JSON"""
+        raise NotImplementedError
+
+
+class KfpComponentParser(ComponentParser):
+    _type = "kfp"
+
+    def __init__(self):
+        super().__init__()
```
kevin-bates, line 154: This isn't necessary until it needs to do something.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def parse_component_details(self, component_body, component_name=None):
```
kevin-bates, line 156: Method signature should match the superclass.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+        component_json = {
+            'label': component_body['name'],
+            'image': "",
+            'id': get_id_from_name(component_body['name']),
+            'description': ' '.join(component_body['description'].split()),
```
kevin-bates, line 161: Is `component_body['description']` a required value?  If not, this will result in an exception.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+
+        # Append output group info to parameter details
+        component_parameters['uihints']['group_info'][0]['group_info'].append(output_group_info)
+
+        return component_parameters
+
+    def build_parameter(self, obj, obj_type):
+        new_parameter = {}
+
+        new_parameter['id'] = f"elyra{obj_type}_{obj['name'].replace(' ', '_')}"
```
kevin-bates, line 246: Should we have a separator between `elyra` and the parameter type?  It might read easier.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+            'description': ' '.join(description.split()),
+            'node_types': []
+        }
+
+        node_type = set_node_type_data(get_id_from_name(label), label, description)
+        component_json['node_types'].append(node_type)
+
+        return component_json
+
+    def parse_component_properties(self, component):
```
kevin-bates, line 308: Please remove if this won't be necessary and update superclass to `pass` or `return None` - same difference.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+        else:
+            raise ValueError(f'File type {component_extension} is not supported.')
+
+
+class ComponentRegistry(SingletonConfigurable):
+
+    parsers = {
+        KfpComponentParser._type: KfpComponentParser(),
+        AirflowComponentParser._type: AirflowComponentParser(),
+        ComponentParser._type: ComponentParser()
```
kevin-bates, line 374: This doesn't seem conducive to _bring your own processors_, but something we can deal with when/if that time comes.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def _get_parser(self, processor_type: str):
+        """
+        Find the proper parser based on processor type.
+        """
+
+        try:
+            return self.parsers.get(processor_type)
+        except Exception:
+            raise ValueError(f"Unsupported processor type: {processor_type}")
```
kevin-bates, line 483: Since this is using `get()`, no exception will be thrown and `None` will be returned if `processor_type` is not in the dictionary.  I think you want `self.parsers[processor_type]` which will result in a `KeyError` being raised - which probably should be what is referenced in the `except` clause.

elyra/pipeline/handlers.py
```diff
@@ -96,21 +98,47 @@ async def post(self, *args, **kwargs):
...
-            # invalid resource, throw an error
-            raise web.HTTPError(400, f"Invalid configuration name '{resource}'")
+    async def get(self, processor):
+        print(f'>>> Retrieving pipeline components for {processor}')
+        if processor not in self.valid_processors:
+            raise web.HTTPError(400, f"Invalid processor name '{processor}'")
 
+        components = await PipelineProcessorManager.instance().get_components(processor)
+        print('>>>')
+        print(components)
```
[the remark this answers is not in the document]
lresende: my fault :)

elyra/pipeline/processor.py
```diff
@@ -44,12 +44,18 @@ def add_processor(self, processor):
...
+        if self.is_valid_processor(processor_type):
             return self._processors[processor_type]
         else:
             return None
 
+    def is_valid_processor(self, processor_type: str) -> bool:
+        if processor_type in self._processors.keys():
+            return True
+        else:
+            return False
```
kevin-bates, line 57:
```suggestion
        return processor_type in self._processors.keys()
```

elyra/pipeline/registry.py
kiersten-stokes: That makes a lot of sense!

elyra/pipeline/registry.py
kiersten-stokes: Ah of course! Not sure what my thinking was on that one

elyra/pipeline/registry.py
kiersten-stokes: Great point. Should definitely do some checking around this

elyra/pipeline/registry.py
kiersten-stokes: Working on implementing this this week 😄

elyra/pipeline/registry.py
kiersten-stokes: Yeah this is definitely one of those things where I wasn't sure exactly how it might affect future changes, but I'll keep in for now I guess

elyra/pipeline/registry.py
kiersten-stokes: Ah interesting! Thanks for pointing that out

elyra/pipeline/registry.py
kevin-bates: Oh, yeah, that is odd.  Could `obj_type` be made plural?  e.g., `elyra_inputs_input_data`, `elyra_inputs_packages_to_install`, `elyra_outputs_output_data`, etc.  This might better differentiate the "prefix" from the "parameter".

kiersten-stokes: I like that a lot!

marthacryan:
Just pushed a frontend update that requests the components from the server based on the runtime assigned to the pipeline. 
Todo:
- [ ] Some of the fields are not consistent with the current node definitions we're using (there are additional fields like "label" that provides a label for the node when there is no filename assigned. There are also fields that have changed because the pipeline editor uses custom components, so the `uihints` should be updated for that. 
- [ ] There isn't an API endpoint for a full list of all possible components, so while there are a few new components, I'm only seeing the notebook, python, and R nodes when I send a get request to `elyra/pipeline/components/[runtime]`.

kiersten-stokes:
There's still some fine-tuning to do, but see screenshots below for current status of UI integration.

### Palette:
Screen Shot 2021-05-17 at 11 09 28 AM

### Properties:
Screen Shot 2021-05-17 at 11 08 33 AM

elyra/pipeline/processor.py
kiersten-stokes: I'll move the raise for now and put `__contains__` into my "to-discuss" list!

elyra/pipeline/registry.py
kevin-bates: I think, like runtime pipeline processors, these "component parsers" will need to be registered via entry points as well.  Or perhaps, be a function of the runtime processor implementation.  I.e., the `RuntimePipelineProcessor` subclass provides "its" ComponentParser implementation.  That way third parties just register the runtime processor.  This topic goes with [this discussion](https://github.com/elyra-ai/elyra/discussions/1672).  That said, this is somewhat orthogonal to getting the general support framework in place - so nothing to worry about now.
@lresende - thoughts?

elyra/pipeline/processor.py
kiersten-stokes: Looks like this just a holdover. I'll remove the one in `RuntimePipelineProcess` in favor of the one in `PipelineProcessor`  since the former inherits from the latter.

elyra/pipeline/registry.py
kiersten-stokes: Now that the UI is integrated, I went with a slightly different approach. Turns out the id is only used internally, so I removed the `elyra_inputs` prefix from the inputs altogether, and only added `elyra_outputs` prefix if a prefix of the same name already exists in the parameter list.

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,623 @@
...
+                custom_control_id = "StringControl"
+
+        # Build label name
+        label = obj['name']
+        if obj_type not in obj['name'].lower():
+            label = f"{obj['name']} ({obj_type})"
+
+        # Build parameter info
+        new_parameter = {
+            'parameter_ref': obj['name'].replace(' ', '_'),
```
kevin-bates, line 267: Since the parameter structure is identical across ComponentParsers, it might be better to add a method on the superclass that takes the 5 parameters that are runtime-specific and returns the parameter JSON. 

The same comment goes for determining `custom_control_id`.  That would also prevent the redundant calls to `lower()`.

kiersten-stokes: Yeah, the more `lower()` calls I added, I started to question myself 😄 . Great idea!

elyra/pipeline/registry.py
```diff
@@ -125,19 +130,47 @@ def add_component(self, request_body):
...
+        if parameter_type in ["number", "integer"]:
+            return "NumberControl"
+        elif parameter_type in ["bool", "boolean"]:
+            return "BooleanControl"
+        # elif "array" in parameter_type:
+        #     return "StringArrayControl"
+        else:
+            return "StringControl"
+
+    def return_parameter(self, name, control_id, label, description, data):
```
kevin-bates, line 157: `return_xxx` for a method name is atypical because methods return things.  Might the good 'ole `get_parameter` or `compose_parameter` be more indicative.  I think the latter describes what is taking place in the method and, when called within `build_parameter()` kind of implies the final step.

elyra/pipeline/registry.py
```diff
@@ -125,19 +130,47 @@ def add_component(self, request_body):
...
+    def parse_component_details(self, component, component_name=None):
         """Get component name, id, description for palette JSON"""
-        pass
+        raise NotImplementedError
 
     def parse_component_properties(self, component):
         """Get component properties for properties JSON"""
         raise NotImplementedError
 
+    def get_custom_control_id_and_type(self, parameter_type):
```
kevin-bates, line 146: The method name implies 'type' is going to be returned.  I would suggest just using `get_custom_control_id()`.

elyra/pipeline/registry.py
```diff
@@ -243,39 +280,18 @@ def build_parameter(self, obj, obj_type):
         data_object['format'] = "string"
         custom_control_id = "StringControl"
         if "type" in obj:
-            data_object['format'] = obj['type']
```
kevin-bates: I suspect this was unintentionally removed, although the TODO in the airflow-based class implies it might not be necessary.

elyra/pipeline/registry.py
kiersten-stokes: `compose` it is!

elyra/pipeline/registry.py
kiersten-stokes: Ah right, I meant to remove the `and_type`. Or, at the least, I'll have to look into it further. This is also the reason for the below comment on removing the assignment of type/format. I haven't been able to find a scenario where changing this format information really affects anything. The `custom_control_id` seems to be the key for populating the type information needed on the frontend. I'll do some research!

elyra/pipeline/registry.py
kiersten-stokes: Yeah, I think I will go ahead and add the `data_object['format']` back in. Even though I haven't been able to figure out if it is used anywhere on the frontend, I think it's probably good to have for informational purposes and we may be able to use it in the property descriptions as a type hint to the user. (see https://github.com/elyra-ai/elyra/pull/1620#discussion_r635429036 for more)

elyra/pipeline/handlers.py
```diff
@@ -47,7 +47,7 @@ async def post(self, *args, **kwargs):
         pipeline_export_path = payload['export_path']
         pipeline_overwrite = payload['overwrite']
 
-        pipeline = PipelineParser().parse(pipeline_definition)
+        pipeline = PipelineParser().parse(pipeline_definition, self.settings['server_root_dir'])
```
kevin-bates, line 50: It might be cleaner to make `root_dir` and attribute (member variable) of `PipelineParser` and convert the two static methods that require `root_dir` as a parameter back to instance methods where the latter of the two can simply reference `self.root_dir`.  This way it doesn't need to be "on the surface" down to where it's used.

elyra/pipeline/parser.py
```diff
@@ -128,7 +131,7 @@ def _get_pipeline_definition(pipeline_definitions: Dict, pipeline_id: str) -> [D
         return None
 
     @staticmethod
-    def _create_pipeline_operation(node: Dict, super_node: Optional[Dict] = None):
+    def _create_pipeline_operation(root_dir: str, node: Dict, super_node: Optional[Dict] = None):
```
kevin-bates, line 134: Per previous comment, both this method and `_get_remaining_component_params()` would not be static and `_get_remaining_component_params()` can reference `self.root_dir` directly.

elyra/pipeline/parser.py
```diff
@@ -254,8 +257,14 @@ def _get_remaining_component_params(node: Dict):
                 if key in ["filename", "runtime_image", "cpu", "gpu", "memory", "dependencies",
                            "include_subdirectories", "env_vars", "outputs", "ui_data", "component_source_type"]:
                     continue
-                # Remove parameter id if a unique identifier was added during component properties parsing
+                # Remove unique identifier of parameter id if one was added during component properties parsing
                 if "elyra_outputs_" in key:
                     key = key.replace("elyra_outputs_", "")
+                # For path inputs and outputs, grab the content in order to pass to contructor
+                if "elyra_path_" in key:
```
kevin-bates, line 264: Could this be `key.startswith("elyra_path_")` instead or is "elyra_path_" a substring?  (Same comment for "elyra_outputs_" above).

elyra/pipeline/handlers.py
kiersten-stokes: I was going to ask you about this -- glad you saw! I wanted to get what I had pushed ASAP so that Luciano could test out the functionality, but adding the attribute definitely seems like the way to go

elyra/pipeline/parser.py
kiersten-stokes: I'll have to think through the scenarios here to see if there's a better way to pass variable type info than in the key like this. But if there isn't, I think I'll change this to check "elyra_path_" first (and use `startswith()` to do so) and do the key and value reassignment. And then afterwards check for "elyra_outputs_" with `startswith()` as well, since they'll always be at the start of the string

elyra/pipeline/registry.py
```diff
@@ -0,0 +1,800 @@
...
+        '''
+        # Start with empty properties object
+        component_parameters = copy.deepcopy(empty_properties)
+
+        # Add path details
+        component_parameters['current_parameters']['component_source'] = component_path
+
+        # Add class information as parameter
+        component_parameters['parameters'].append({"id": "elyra_airflow_class_name"})
+        component_parameters['current_parameters']["elyra_airflow_class_name"] = []
```
kevin-bates, line 470: What is the reason for this defaulting? to an empty array?  It seems contrary to the 'For Airflow inputs, remove class name information from key' logic in `PipelineParser._get_remaining_component_params` which expects a string.

Where does the actual value of `elyra_airflow_class_name` come from?

kiersten-stokes: This is part of the way I'm handling an operator file that includes more than one operator class. [This](https://github.com/apache/airflow/blob/1.10.15/airflow/operators/slack_operator.py) is a good example of a file that has 2 possible operators to use. 

So the `elyra_airflow_class_name` var holds an array of classes from which the user can choose in the properties pane (as in the screenshot below). The user will select this class (I suppose I should make it `required` now that I think about it), and the selected class will be propagated as a parameter to the backend where the parser will then use to determine which of the remaining parameters it needs to include as kwargs for the selected class.

Screen Shot 2021-06-07 at 3 06 51 PM

That being said, I'm not sure I love this approach for the long term, but this was a fairly simple way to include that functionality.

elyra/pipeline/parser.py
```diff
@@ -234,3 +242,42 @@ def _scrub_list(dirty: Optional[List[Optional[str]]]) -> List[str]:
...
+                if not value:
+                    continue
+                # Do not include any of the standard set of parameters
+                if key in ["filename", "runtime_image", "cpu", "gpu", "memory", "dependencies", "env_vars", "outputs",
+                           "include_subdirectories", "ui_data", "component_source", "component_source_type",
+                           "elyra_airflow_class_name"]:
+                    continue
+                # For Airflow inputs, remove class name information from key
+                class_name = PipelineParser._get_app_data_field(node, 'elyra_airflow_class_name')
+                if class_name and not key.startswith(class_name.lower().replace(' ', '_')):
```
kevin-bates, line 265: Per the previous comment regarding `elyra_airflow_class_name` being an empty list, I guess I'm still confused why its an empty list here.  Neither of the `class_name.lower()` calls will work.  Here's a screenshot from the debugger.
image
If we're using the empty list as a `None` indicator, we should use the latter since `class_name` should ultimately be a string.

kiersten-stokes: I think it's only an empty list here because a class wasn't selected from the dropdown menu. I'm testing it out now, but ideally what will happen when I make `elyra_airflow_class_name` a required field is that it will force the user to select one of the options from the dropdown. And as such, only the selected class will be passed as a string to the backend

bourdakos1:
I bandaid fixed the CI issue, working on a real fix now

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -413,10 +441,15 @@ export class PipelineService {
...
-      );
+      if (
+        node.op ==
+        ('execute-notebook-node' || 'execute-python-node' || 'execute-r-node')
+      ) {
+        node.app_data.filename = this.getPipelineRelativeNodePath(
+          pipelinePath,
+          node.app_data.filename
+        );
+      }
```
bourdakos1, line 452: This should be
```tsx
if (node.op === 'execute-notebook-node' || node.op === 'execute-python-node' || node.op === 'execute-r-node') {
```

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -426,10 +459,15 @@ export class PipelineService {
...
-      );
+      if (
+        node.op ==
+        ('execute-notebook-node' || 'execute-python-node' || 'execute-r-node')
+      ) {
+        node.app_data.filename = this.getWorkspaceRelativeNodePath(
+          pipelinePath,
+          node.app_data.filename
+        );
+      }
```
bourdakos1, line 470: same as above

bourdakos1 (changes requested):

packages/pipeline-editor/src/PipelineService.tsx
bourdakos1: this should fix #1737

bourdakos1 (approved):

bourdakos1 (changes requested):
I will add support for custom node labels

lresende (approved):

PR #1626, opened 2021-04-27
Fix runtime images enum bug

### LINKED ISSUE #1609: Runtime image selection is showing image tag information
**Describe the issue**
When editing the notebook, the runtime image displayed on the node properties side panel is showing the image tag instead of the title. See image below:

image

**Expected behavior**
- If runtime image is currently available in the metadata store, show the runtime image title
- If runtime image is not available, previous behavior was to use '-----' but in this case, I would be ok to show the runtime image name, but that gives the user the idea that things are ok, where he really should upgrade to one of the available runtime images (e.g. in case of upgrades, etc)

**Deployment information**
 - Elyra master

### LINKED ISSUE #1615: Pipeline editor associates runtime image display name with node instead of runtime image name
**Describe the issue**

**To Reproduce**
1. Create a pipeline
2. Create a notebook
3. Add notebook to pipeline
4. export | run pipeline

Example of an exported pipeline:

```
  - name: hello
    container:
      args: ['mkdir -p ./jupyter-work-dir/ && cd ./jupyter-work-dir/ && curl -H "Cache-Control:
          no-cache" -L https://raw.githubusercontent.com/akchinSTC/kfp-notebook/i-94/etc/docker-scripts/bootstrapper.py
          --output bootstrapper.py && curl -H "Cache-Control: no-cache" -L https://raw.githubusercontent.com/akchinSTC/kfp-notebook/i-94/etc/requirements-elyra.txt
          --output requirements-elyra.txt && python3 -m pip install  packaging &&
          python3 -m pip freeze > requirements-current.txt && python3 bootstrapper.py
          --cos-endpoint https://s3-op.../ --cos-bucket
          opf-datacatalog --cos-directory "world-0423115249" --cos-dependencies-archive
          "hello-fbdba829-00b2-43e4-a79e-1dda85d034c0.tar.gz" --file "hello.ipynb" ']
      command: [sh, -c]
      env:
      - {name: AWS_ACCESS_KEY_ID, value: XYZ}
      - {name: AWS_SECRET_ACCESS_KEY, value: ABC}
      - name: ELYRA_ENABLE_PIPELINE_INFO
        value: "True"
      - {name: ELYRA_WRITABLE_CONTAINER_DIR, value: /tmp}
      image: Pandas on quay.io
```

**Expected behavior**

**Deployment information**
 - Elyra version: 2.3.x/master

marthacryan:
Fixes #1609. 
Fixes #1615.

The pipeline editor was passing the display name rather than the tag of the runtime image selected for each node. This fixes that behavior by swapping out the display name for the tag before propagating changes in the pipeline back to the file. 
 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -134,8 +134,16 @@ const PipelineWrapper: React.FC<IProps> = ({
     const currentContext = contextRef.current;
 
     const changeHandler = (): void => {
-      const pipeline = currentContext.model.toJSON();
-      setPipeline(pipeline);
+      const pipelineJson: any = currentContext.model.toJSON();
+      // Update to display actual value of runtime image
+      for (const node of pipelineJson?.pipelines?.[0]?.nodes) {
+        const app_data = node?.app_data;
+        if (app_data?.runtime_image) {
+          app_data.runtime_image =
+            runtimeImages.current?.[app_data.runtime_image];
+        }
+      }
+      setPipeline(pipelineJson);
       setLoading(false);
     };
 
@@ -150,6 +158,17 @@ const PipelineWrapper: React.FC<IProps> = ({
 
   const onChange = useCallback((pipelineJson: any): void => {
     if (contextRef.current.isReady) {
+      // Update to store tag of runtime image
+      for (const node of pipelineJson?.pipelines?.[0]?.nodes) {
+        const app_data = node?.app_data;
+        if (app_data?.runtime_image) {
+          for (const tag in runtimeImages.current) {
+            if (runtimeImages.current?.[tag] === app_data?.runtime_image) {
+              app_data.runtime_image = tag;
+            }
+          }
+        }
+      }
       contextRef.current.model.fromString(
         JSON.stringify(pipelineJson, null, 2)
       );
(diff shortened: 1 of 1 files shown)


lresende:
So, while testing this with an existing pipeline, I see the following:

image

for a pipeline that has:

image

Also, when I go to node properties

image

Based on the previous implementation, if the field value is there, validation should not flag the node as invalid and execution should work. The node property user experience is questionable, it should allow the user to update to a new image (or not)?

lresende (changes requested):
Some corner cases around validation are still failing.

bourdakos1:
@lresende question related to this, would it be better to just use the actual runtime image name instead of a “display name” so we can show the image name in the dropdown even if the image metadata isn’t available?

lresende:
> @lresende question related to this, would it be better to just use the actual runtime image name instead of a “display name” so we can show the image name in the dropdown even if the image metadata isn’t available?

I would say that a possible solution would be to use the "display name" for items read from metadata, and if the current node value is not there, display the "image name" just in that case. In the past, I believe we were displaying ` - - - ` sort of to indicate that the value is old and needs to be updated, but the main issue is that this scenario does not trigger validation issues.

marthacryan:
@lresende Want to try out the new changes to make sure the edge case is fixed? I could reproduce locally and it's looking better for me after my most recent commit.

lresende (approved):
LGTM, the issue described above seems to be fixed and things seem back to normal. Also tried with images that were not listed as part of metadata and they look ok. Thanks.

lresende:
Waiting for build status before merging.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -134,22 +134,61 @@ const PipelineWrapper: React.FC<IProps> = ({
...
+          const app_data = node?.app_data;
+          if (app_data?.runtime_image) {
+            for (const tag in runtimeImages.current) {
+              if (runtimeImages.current?.[tag] === app_data?.runtime_image) {
+                app_data.runtime_image = tag;
+              }
+            }
+          }
+        }
+      }
```
bourdakos1, line 191: I'm a little hesitant about directly mutating the pipeline json here, but if it's not breaking anything 🤷‍♂️

bourdakos1 (approved):

PR #1653, opened 2021-05-10
[MEGA WIP] Example refactor

bourdakos1:
Think of this PR as my clay for forming my ideas and not something that will ever need to be merged


lresende (changes requested):
It's very hard to review this and compare what should be used as an example of enhancing the code via refactoring and other unrelated issues. For now, I think we should focus on reducing the clutter and providing an example so others can follow the same pattern. 

I would reconsider adding here any lint updates, build-related changes, etc All these additions could still be proposed as separate prs.

bourdakos1:
@lresende I’m just using this as a way to get a better feel for the code base, before drafting up a more concrete strategy. I don’t expect this to ever be something that is merged

Issue #1594, opened 2021-04-20
Document best practice for implementing UI components


lresende:
Some of Elyra UI components are very coupled with related business logic and we are discussing refactor them to promote separation of concerns and sort of use MVC code-pattern (see #1405).

This issue will handle the documentation of the best-practice/patterns to adopt during the refactor and creation of new UI components for Elyra.

As for location, I would say either somewhere in [Elyra docs](https://github.com/elyra-ai/elyra/tree/master/docs/source/developer_guide) or [Community repo](https://github.com/elyra-ai/community). Having said that, I believe we could start with Elyra repo and promoting to the community after a few interactions to make it solid first.

bourdakos1:
# Nick’s guide to a fun frontend

## Component Style
All React components should be functional components (no class components). 
Class components can promote the use of inheritance where React works best with
a composition model. For more information see: https://reactjs.org/docs/composition-vs-inheritance.html
```tsx
interface Props {
    text: string;
}

export const Component: FC<Props> = ({ text }) => {
    return <div>{text}</div>;
}
```

## Hooks
Using functional components will drive the need for the use of hooks. Hooks have
two very important rules:
1. Hooks can only be called in a react functional component or a custom hook
2. Do not call hook conditionally, inside loops or inside nested functions

The two hooks you will most likely come across are `useState` and `useEffect`.

## `useState`
State is used to store a dynamic values produced from a side effect. Examples:
- data fetching
    - fetch("/api/things")
- subscriptions
    - onClick
    - window.addEventLister
    - store.subscribe

The `useState` hook is a function that takes a single argument that acts as the 
initial value of the state. The `useState` function returns an array with only 
two values. This array is normally destructured, the first item being the value
of the state (which can never be manually modified) and the second item being a 
setter for that value.

Example:
```tsx
export const Component: FC = () => {
    const [value, setValue] = useState("Default Value");

    const handleClick = () => {
        setValue("Clicked!")
    }

    return <div onClick={handleClick}>{value}</div>;
}
```

When calling the setter, it will completely replace the value of the state
instead of merging like with `this.setState` in class components. For this reason
it is best to use many smaller states than a large state object.

BAD:
```tsx
export const Component: FC = () => {
    const [state, setState] = useState({ 
        color: "red", 
        text: undefined,
    });

    const handleColorChange = (color) => {
        setState({
            ...state,
            color,
        });
    }

    const handleTextChange = (value) => {
        setState({
            ...state,
            text: value,
[...]

Issue #1405, opened 2021-03-09
Refactor source code to separate UI and control


karlaspuldaro:
**Describe the issue**

When developing Elyra extensions it is recommended to follow the notion of Model-View-Control design pattern whenever applicable.
Some portions of our code could benefit from a code refactor in order to better distinguish the different roles and scopes.
One example would be the [getRuntimes](https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/PipelineService.tsx#L47) function in [PipelinesService.tsx](https://github.com/elyra-ai/elyra/blob/master/packages/pipeline-editor/src/PipelineService.tsx), which currently returns either a list of runtimes, or the result of a UI dialog, where the controller/service has notion of the view component (UI).

Also documenting code practices and conventions would help new contributors to the Elyra source code in general.

PR #1657, opened 2021-05-11
Refactor ui-components

karlaspuldaro:
As part of #1405 and following #1594 documentation notes, this PR refactors React components in `ui-components` package from class to functions, introducing hooks where applicable.

Manual testing of the affected components.

           in the file; or

           it.

### DIFF
--- a/packages/ui-components/src/ExpandableComponent.tsx
+++ b/packages/ui-components/src/ExpandableComponent.tsx
@@ -85,9 +85,7 @@ export const ExpandableComponent: React.FC<IExpandableComponentProps> = props =>
         <button
           className={buttonClasses}
           title={expanded ? 'Hide Details' : 'Show Details'}
-          onClick={(): void => {
-            toggleDetailsDisplay();
-          }}
+          onClick={toggleDetailsDisplay}
         >
           {expanded ? (
             <caretDownIcon.react
@@ -110,9 +108,7 @@ export const ExpandableComponent: React.FC<IExpandableComponentProps> = props =>
               ? DISPLAY_NAME_CLASS
               : DISPLAY_NAME_CLASS + ' ' + DRAGGABLE_CLASS
           }
-          onClick={(): void => {
-            toggleDetailsDisplay();
-          }}
+          onClick={toggleDetailsDisplay}
           onMouseDown={(event): void => {
             props.onMouseDown && props.onMouseDown(event);
           }}
--- a/packages/ui-components/src/ExpandableComponent.tsx
+++ b/packages/ui-components/src/ExpandableComponent.tsx
@@ -60,32 +60,37 @@ export interface IExpandableComponentProps {
 /**
  * A React component for expandable containers.
  */
-export const ExpandableComponent: React.FC<IExpandableComponentProps> = props => {
+export const ExpandableComponent: React.FC<IExpandableComponentProps> = ({
+  displayName,
+  tooltip,
+  actionButtons = [],
+  onExpand,
+  onBeforeExpand,
+  onMouseDown,
+  children
+}) => {
   const [expanded, setExpandedValue] = React.useState(false);
 
-  const toggleDetailsDisplay = (): void => {
+  const handleToggleDetailsDisplay = (): void => {
     // Switch expanded flag
     const newExpandFlag = !expanded;
-    if (props.onBeforeExpand) {
-      props.onBeforeExpand(newExpandFlag);
-    }
+    onBeforeExpand?.(newExpandFlag);
     setExpandedValue(newExpandFlag);
   };
 
   React.useEffect((): void => {
-    props.onExpand && props.onExpand(expanded);
+    onExpand?.(expanded);
   });
 
   const buttonClasses = [ELYRA_BUTTON_CLASS, BUTTON_CLASS].join(' ');
-  const actionButtons = props.actionButtons || [];
 
   return (
     <div>
-      <div key={props.displayName} className={TITLE_CLASS}>
+      <div key={displayName} className={TITLE_CLASS}>
         <button
           className={buttonClasses}
           title={expanded ? 'Hide Details' : 'Show Details'}
-          onClick={toggleDetailsDisplay}
+          onClick={handleToggleDetailsDisplay}
         >
           {expanded ? (
             <caretDownIcon.react
@@ -102,18 +107,18 @@ export const ExpandableComponent: React.FC<IExpandableComponentProps> = props =>
           )}
         </button>
         <span
-          title={props.tooltip}
+          title={tooltip}
           className={
-            props.onMouseDown
+            onMouseDown
               ? DISPLAY_NAME_CLASS
               : DISPLAY_NAME_CLASS + ' ' + DRAGGABLE_CLASS
           }
-          onClick={toggleDetailsDisplay}
+          onClick={handleToggleDetailsDisplay}
           onMouseDown={(event): void => {
-            props.onMouseDown && props.onMouseDown(event);
+            onMouseDown?.(event);
           }}
         >
-          {props.displayName}
+          {displayName}
         </span>
 
         <div className={ACTION_BUTTONS_WRAPPER_CLASS}>
@@ -124,7 +129,7 @@ export const ExpandableComponent: React.FC<IExpandableComponentProps> = props =>
                 title={btn.title}
                 feedback={btn.feedback || ''}
                 className={buttonClasses + ' ' + ACTION_BUTTON_CLASS}
-                onClick={(): void => btn.onClick()}
+                onClick={btn.onClick}
               >
                 <btn.icon.react
                   tag="span"
@@ -137,7 +142,7 @@ export const ExpandableComponent: React.FC<IExpandableComponentProps> = props =>
         </div>
       </div>
       <div className={expanded ? DETAILS_VISIBLE_CLASS : DETAILS_HIDDEN_CLASS}>
-        {props.children || null}
+        {children}
       </div>
     </div>
   );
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
  */
-export class ExpandableComponent extends React.Component<
-  IExpandableComponentProps,
-  IExpandableComponentState
-> {
-  constructor(props: IExpandableComponentProps) {
-    super(props);
-    this.state = { expanded: false };
-  }
+export const ExpandableComponent: React.FC<IExpandableComponentProps> = props => {
```
bourdakos1, line 63: It's normally common practice to destructure props here

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
+  };
 
-  componentDidUpdate(): void {
-    if (this.props.onExpand) {
-      this.props.onExpand(this.state.expanded);
-    }
-  }
+  React.useEffect((): void => {
+    props.onExpand && props.onExpand(expanded);
+  });
```
bourdakos1, line 77: we shouldn't need an effect for this, this should be able to go in `handleToggleDetailsDisplay`

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
-> {
-  constructor(props: IExpandableComponentProps) {
-    super(props);
-    this.state = { expanded: false };
-  }
+export const ExpandableComponent: React.FC<IExpandableComponentProps> = props => {
+  const [expanded, setExpandedValue] = React.useState(false);
 
-  toggleDetailsDisplay(): void {
+  const toggleDetailsDisplay = (): void => {
```
bourdakos1, line 66: it's normally best to start event handler names with "handle" so maybe `handleToggleDetailsDisplay`

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
     // Switch expanded flag
-    const newExpandFlag = !this.state.expanded;
-    if (this.props.onBeforeExpand) {
-      this.props.onBeforeExpand(newExpandFlag);
+    const newExpandFlag = !expanded;
+    if (props.onBeforeExpand) {
+      props.onBeforeExpand(newExpandFlag);
     }
```
bourdakos1, line 71: I try to use optional chaining for event callbacks instead of `if` statements: 
```ts
onBeforeExpand?.(newExpandFlag)
```

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
-  }
+  React.useEffect((): void => {
+    props.onExpand && props.onExpand(expanded);
+  });
 
-  render(): React.ReactElement {
-    const buttonClasses = [ELYRA_BUTTON_CLASS, BUTTON_CLASS].join(' ');
-    const actionButtons = this.props.actionButtons || [];
+  const buttonClasses = [ELYRA_BUTTON_CLASS, BUTTON_CLASS].join(' ');
+  const actionButtons = props.actionButtons || [];
```
bourdakos1, line 80: this should be changed to using nullish coalescing:
```
const actionButtons = props.actionButtons ?? [];
```

but in this case if it's destructured as a prop above we can assign a default like so:
```ts
const Component: React.FC<Props> = ({ actionButtons = [] }) => {
}
```

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,88 @@ export interface IExpandableComponentProps {
...
+              </FeedbackButton>
+            );
+          })}
         </div>
       </div>
-    );
-  }
-}
+      <div className={expanded ? DETAILS_VISIBLE_CLASS : DETAILS_HIDDEN_CLASS}>
+        {props.children || null}
```
bourdakos1, line 140: I don't think `|| null` is needed?

bourdakos1 (changes requested):
I only reviewed `ExpandableComponent` looking good, just a few small things

packages/ui-components/src/ExpandableComponent.tsx
karlaspuldaro: Code snippets needs this callback triggered only after the component state changed to expanded, so that it renders the snippet preview. Calling this inside `handleToggleDetailsDisplay` messes up the rendering order, the first time expanding a code snippet nothing is displayed.

bourdakos1: oh gotcha!

packages/ui-components/src/ExpandableComponent.tsx
```diff
@@ -57,112 +57,93 @@ export interface IExpandableComponentProps {
...
+          {displayName}
+        </span>
+
+        <div className={ACTION_BUTTONS_WRAPPER_CLASS}>
+          {actionButtons.map((btn: IExpandableActionButton) => {
+            return (
+              <FeedbackButton
+                key={btn.title}
+                title={btn.title}
+                feedback={btn.feedback || ''}
```
bourdakos1, line 130:
```suggestion
                feedback={btn.feedback ?? ''}
```

packages/ui-components/src/ExpandableErrorDialog.tsx
```diff
@@ -31,58 +31,53 @@ interface IErrorDialogProps {
...
-  any
-> {
-  dialogNode: HTMLDivElement;
-  collapsedDimensions: number[];
+export const ExpandableErrorDialog: React.FC<IErrorDialogProps> = ({
+  reason,
+  message,
+  timestamp,
+  traceback,
+  default_msg
```
bourdakos1, line 39: this should be `defaultMessage` if possible

packages/ui-components/src/ExpandableErrorDialog.tsx
```diff
@@ -31,58 +31,53 @@ interface IErrorDialogProps {
...
+        (width < ERROR_DIALOG_WIDTH || height < ERROR_DIALOG_HEIGHT)
+      ) {
+        setCollapsedSize({ width, height });
+        dialogNode.style.width = Math.max(width, ERROR_DIALOG_WIDTH) + 'px';
+        dialogNode.style.height = Math.max(height, ERROR_DIALOG_HEIGHT) + 'px';
+      } else if (!expanded && collapsedSize) {
+        dialogNode.style.width = collapsedSize.width + 'px';
+        dialogNode.style.height = collapsedSize.height + 'px';
+      }
+    },
```
bourdakos1, line 62: We are manually mutating the DOM here, which is normally considered a side effect, so this should go into a `useEffect` hook

packages/ui-components/src/ExpandableErrorDialog.tsx
```diff
@@ -31,58 +31,53 @@ interface IErrorDialogProps {
...
-    const height = this.dialogNode.clientHeight;
+  const details = traceback ? (
+    <ExpandableComponent
+      displayName={'Error details: '}
+      tooltip={'Error stack trace'}
+      onBeforeExpand={handleUpdateDialogSize}
+    >
+      <pre>{traceback}</pre>
+    </ExpandableComponent>
+  ) : null;
```
bourdakos1, line 74: I would normally pull this out into a separate component or include it inline in the jsx below.

since these are string literals I would change them to this:
```
<ExpandableComponent
      displayName="Error details: "
      tooltip="Error stack trace"
      onBeforeExpand={handleUpdateDialogSize}
>
```

bourdakos1 (changes requested):
This is looking great, I think the last big this is using an effect hook to handle any dom mutations. The rest is just a couple tiny stylistic changes

packages/ui-components/src/ExpandableErrorDialog.tsx
bourdakos1:
```tsx
const [expanded, setExpanded] = useState(null);

// we can use a ref, because we don't need to rerender the component if its value changes
// this could also be initialized to undefined or null if that is something that needs to be checked
const collapsedSizeRef = useRef({ width: 0, height: 0 });

useEffect(() => {
    const dialogNode = document.querySelector(JP_DIALOG_CONTENT);
    if (expanded) {
        collapsedSizeRef.current = { width: dialogNode.clientWidth,  height: dialogNode.clientHeight}
        dialogNode.style.width = Math.max(width, ERROR_DIALOG_WIDTH) + 'px';
        dialogNode.style.height = Math.max(height, ERROR_DIALOG_HEIGHT) + 'px';
    } else {
        dialogNode.style.width = collapsedSize.current.width + 'px';
        dialogNode.style.height = collapsedSize.current.height + 'px';
    }
}, [expanded])

// this doesn't really need to be a callback hook
const handleSizeChange = (expanded: boolean) => {
    setExpanded(expanded);
}
```

bourdakos1 (approved):
LGTM - nice work!

packages/ui-components/src/ExpandableErrorDialog.tsx
karlaspuldaro: Just to document here the discussions around this topic, the proposed solution changes the DOM tree and therefore the original dialog resize behaviour.
Another solution we've tried was using CSS selectors only, but it didn't work in some cases.
This is a tricky case as we are manipulating and tracking the state of an element external to this component, maybe that's why using effects here doesn't really behave as expected. 
Therefore it was decided to keep it as is.

PR #1664, opened 2021-05-12
Validation service initial implementation

akchinSTC:
Dependent on #1854 being merged first

 Validation service is being served under `{url}/elyra/pipeline/validate`

Validates basic JSON structure
- [x]   contains valid json
- [x] - Contains `primary pipeline’ and is a string
- [x] - Contains at least one `pipeline element` in the json and that is a list
- [x] - Checks pipeline schema version matches or works with the current version being run
- [x] - the primary_pipeline is findable in the pipelines array
- [x] - the found primary_pipeline has an app_data field that is an object containing a version field that is a number
- [x] - the found primary_pipeline has a nodes field that is an array

Validate compatibility 
- [x] - Based on the version of Elyra, determines what operations are supported (a bit moot unless we are planning on retrofitting earlier versions of elyra) for instance python script support was introduced in 1.0.0 and r-script in 2.2.0
- [x] - Construct a support matrix based on runtimes 
- [x] expand to check runtime-specific components at runtime 
- [x] Support validating of pre 3.0 version pipelines e.g. generic components only

Validate node properties
- [x] Check filename for validity
- [x] Check filename for existence 
- [x] - Check image name registry coordinate syntax
- [x] - Check resource values
- [x] - Check dependencies for validity
- [x]  -Check dependencies for existence 
- [x] - Check env vars are properly formatted
- [x] - Check inputs and outputs validity
- [x] - Check ui data label validity against acceptable runtime naming standards (only kfp for now)
~~- [ ] - Update runtime field references when #1710 is completed~~
- [x] - Expand to check runtime specific components properties in pipeline json compared to current component registry

Validate the graph
- [x] - Check for graph cycles
- [x] - Check for singleton nodes to give warning?? Ones without any children
- [x] - Check supernodes

Add tests
- [x] - test resources (test json payloads) will be located under `elyra/pipeline/tests` and `elyra/pipeline/tests/resources`
- [x] Update tests for new component parameters and node label

Front end
- [x]  - Export and submission, what to display in response dialog box when errors in payload found

### Acceptance criteria

Please validate against:
- [x] https://github.com/CODAIT/covid-notebooks
- [x] https://github.com/CODAIT/flight-delay-notebooks
- [x] https://github.com/lresende/elyra-sample/tree/master/consumer-producer/pipelines
- [x] https://github.com/elyra-ai/examples/tree/master/pipelines/dax_noaa_weather_data

Responses will have the following format (example shows a non-numeric string was used) and will differ based on the type of error discovered:
```
{
    "title": "Elyra Pipeline Diagnostics",
    "description": "Issues discovered when parsing the pipeline",
    "issues": [
        {
            "severity": "1",
            "source": "Elyra Pipeline Validation Service",
            "type": "invalidNodeProperty",
            "message": "Property has a non-parsable value",
            "data": {
                "nodeID": "4b1a73b3-3b64-4818-ab98-73965410ea10",
                "propertyName": "cpu"
            }
        },
     ]
}
```
Error Levels:
Currently support 4 error levels on a scale of 1 to 4: `Error, Warning, Info, Hint `
Only `Error`s will cause a fatal exception during pipeline submission and halt the run. 
Error Types:
```
invalidJSON - Invalid JSON detected
data: None

invalidPipeline - Pipeline level payload structure is incomplete, missing or malformed
data: None

invalidRuntime - Incompatible/Unsupported pipeline runtime associated with the pipeline file
data: pipelineRuntime, pipelineID

invalidNodeType - Unsupported node type in the pipeline for the runtime
data: NodeID, PipelineID

invalidNodeProperty - Node level payload structure is incomplete, missing or malformed
data: NodeID, Node Property Name

invalidFilePath - Property has an invalid path to a file/dir or the file/dir does not exist
data: nodeID, Node Property Name, filepath

invalidEnvPair - Invalid env pair in node property
data: nodeID, Node Property Name, env pair

invalidNodeLabel: Label associated with node is invalid for the runtime
data: nodeID, Node Property Name, label name

singletonReference: node is not connected to any part of the pipeline
data: nodeID, pipelineID

circularReference: a cycle was found in the graph
data: cycle number, linkIDList (list of linkIDs associated with the cycle)
```


elyra/validation/handlers.py
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+#
+from elyra.validation import PipelineValidationManager
+from elyra.util.path import get_expanded_path
+from jupyter_server.base.handlers import APIHandler
+from tornado import web
+from ..util.http import HttpErrorMixin
+
+
+class PipelineValidationHandler(HttpErrorMixin, APIHandler):
```
kevin-bates, line 23: Same "location" comment, should this placed into pipeline/handlers.py?

elyra/__init__.py
```diff
@@ -53,6 +54,7 @@ def _load_jupyter_server_extension(nb_server_app):
          PipelineConfigHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/contents/properties/%s' % (path_regex)),
          FileParserHandler),
+        (url_path_join(web_app.settings['base_url'], r'/elyra/validation'), PipelineValidationHandler),
```
kevin-bates, line 57: Seems like this endpoint should be `/elyra/pipeline/validation`.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+from elyra import __version__ as elyra_version
+from elyra.pipeline.parser import PipelineParser
+from packaging import version
+from traitlets.config import SingletonConfigurable
+from typing import Dict, Optional
+
+
+class PipelineValidationManager(SingletonConfigurable):
+
+    def validate(self, root_dir: str, pipeline: Dict) -> 'ValidationResponse':
```
kevin-bates, line 32: I think it might be better to just move the definition of `ValidationResponse` prior to avoid the forward reference "string".

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+
+        try:
+            pipeline = json.loads(json.dumps(pipeline))
+        except ValueError:
+            response.add_message(severity='1',
+                                 message_type="invalidJSON",
+                                 message="Invalid JSON detected, unable to continue.")
+            return response
+
+        self.validate_pipeline_structure(pipeline=pipeline, response=response)
```
kevin-bates, line 49: These sub-methods should be prefixed with `_` since they're not intended to be called publically.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+        :param response: ValidationResponse containing the issue list to be updated
+        """
+
+        current_pipeline_schema_version = '3.0'
+        pipeline_json = json.loads(json.dumps(pipeline))
+
+        # Check for required values.  We require a primary_pipeline, a set of pipelines, and
+        # nodes within the primary pipeline (checked below).
+        if 'primary_pipeline' not in pipeline_json:
+            response.add_message(severity='1', message_type="invalidPipeline",
```
kevin-bates, line 69: See comment about enums or constants for the severity level.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+                                 message="Pipeline 'app_data' is missing from primary pipeline")
+        elif 'version' not in primary_pipeline['app_data']:
+            response.add_message(severity='1', message_type="invalidPipeline",
+                                 message="Pipeline app_data 'version' is missing from primary pipeline")
+        elif not int(primary_pipeline['app_data']['version']) > 0:
+            response.add_message(severity='1', message_type="invalidPipeline",
+                                 message="Primary pipeline version field has an invalid value")
+
+        # TODO: Do we increment the pipeline schema version at all?
+        if pipeline_json['version'] < current_pipeline_schema_version:
```
kevin-bates, line 101: Should the presence of 'version' be confirmed prior to this - like the others are doing?

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+    def validate_compatibility(self, pipeline: Dict, response) -> None:
+        """
+        Checks that the pipeline payload is compatible with this version of elyra (ISSUE #938)
+        as well as verifying all nodes in the pipeline are supported by the runtime
+        :param pipeline: the pipeline definition to be validated
+        :param response: ValidationResponse containing the issue list to be updated
+        """
+        # TODO: Provide option to convert to current format?
+
+        pipeline_json = json.loads(json.dumps(pipeline))
```
kevin-bates, line 114: Do we need to do this on each sub-method since it's already done in `validate()`?

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+
+        # TODO: Get a list of available components from the registry, this should be done dynamically to ensure that
+        # the list of components is current and up to date since the user may have added additional ones since
+        # startup, for now use a hard-coded dict
+        supported_ops = ['execute-notebook-node']
+        if version.parse(elyra_version) >= version.parse("1.0.0"):
+            supported_ops.append('execute-python-node')
+        if version.parse(elyra_version) >= version.parse("2.2.0"):
+            supported_ops.append('execute-r-node')
+        supported_runtimes = ['kfp', 'airflow', 'null']  # need to rethink this null string
```
kevin-bates, line 125: These should probably be derived from the schemas within the runtime namespace.

What is 'null'?  And should 'local' be added to the list of runtimes although it's a special case.  I believe it will be the value of the pipeline's runtime field in that case.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+        support_matrix = {}
+        for runtime in supported_runtimes:
+            support_matrix[runtime] = supported_ops
+
+        # Checks pipeline node types are compatible with the runtime selected
+        pipeline_runtime = pipeline_json['pipelines'][0]['app_data']['runtime']
```
kevin-bates, line 131: Should we add an assertion here?
```suggestion
        pipeline_runtime = pipeline_json['pipelines'][0]['app_data']['runtime']
        assert pipelne_runtime in supported_runtimes, f"Referenced runtime is not in {supported_runtimes}"
```

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+            # Validate the node structure
+            node_field_strings = ['filename', 'runtime_image', 'include_subdirectories', 'cpu', 'gpu', 'memory']
+
+            for field_string in node_field_strings:
+                if field_string not in node_data:
+                    response.add_message(severity='1',
+                                         message_type="invalidNodeProperty",
+                                         message="Node is missing field",
+                                         data={"nodeID": node['id'],
+                                               "propertyName": field_string})
```
kevin-bates, line 168: Given the commonality of this, I think it would make sense to define method that checks existence: 
```python
def validate_field_existence(self, field: str, data: Dict, node_id: str, response: ValidationResponse)
```

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+                elif not isinstance(node_data[field_string], str):
+                    response.add_message(severity='1',
+                                         message_type="invalidNodeProperty",
+                                         message="Node field is incorrect type",
+                                         data={"nodeID": node['id'],
+                                               "propertyName": field_string})
```
kevin-bates, line 174: Same comment, but have the method relative to data-type (validate_field_type).  The two could then be wrapped in another that adds type to the signature (e.g., 'validate_field()`)

These could then be applied below.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+                                            response=response)
+
+    def validate_container_image_name(self, node, response):
+        """
+        Validates the image name exists and is proper in syntax
+        :param node: the node definition to be validated
+        :param response: ValidationResponse containing the issue list to be updated
+        """
+        image_name = node['app_data'].get('runtime_image', '')
+        if not image_name:
```
kevin-bates, line 233: Does this need to tolerate 'local' where images are not required?

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+            response.add_message(severity='1',
+                                 message_type="invalidNodeProperty",
+                                 message="Node is missing image name",
+                                 data={"nodeID": node['id'],
+                                       "propertyName": 'runtime_image'})
+        else:
+            pattern = r"(([A-Za-z0-9.:/]+\/)?[A-Za-z0-9.-]+\/[A-Za-z0-9.-]+:[\w][\w.-]{0,127})"
```
kevin-bates, line 240: hmm - shouldn't the runtime-image name be verified as part of the metadata?  Seems like this pattern should be added to the [runtime-images schema](https://github.com/elyra-ai/elyra/blob/b850476dbcd8a0c85ef26be10d19909ffee63ce9/elyra/metadata/schemas/runtime-image.json#L43).

This seems a bit much to me.  SInce its not validating the image's existence, failure would occur as a non-existent image if malformed.

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+    def validate_environmental_variables(self, node_id: str, env_var: str, response):
+        """
+        Checks the format of the env var to ensure its in the correct form
+        e.g. FOO = 'BAR'
+        :param node_id: the node ID of the node the path is located in
+        :param response: ValidationResponse containing the issue list to be updated
+        :param env_var: the env_var key value pair to check
+        """
+        result = [x.strip(' \'\"') for x in env_var.split('=', 1)]
+        if len(result) < 2:
```
kevin-bates, line 300: Should this be `!= 2`?

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+                        if cycle[0] == link['node_id_ref']:
+                            response.add_message(severity='1',
+                                                 message_type="circularReference",
+                                                 message="A cycle was found within this pipeline",
+                                                 data={"nodeID": node['id'],
+                                                       "pipelineID": pipeline_id})
+            else:
+                for node in node_list:
+                    for link in node['inputs'][0]['links']:
+                        if cycle[0] == link['node_id_ref'] and cycle[1] == node['id']:
```
kevin-bates, line 377: Are there at most two cycles?

elyra/validation/validate.py
```diff
@@ -0,0 +1,424 @@
...
+                    severity: str = '2'):
+        """
+        Helper function to add a diagnostic message to the response to be sent back
+        :param message: A simple message describing the issue
+        :param message_type: The type of message to send back e.g. invalidNodeType, invalidPipeline
+        :param data: granular details regarding the error e.g. the nodeID, pipelineID, linkID etc.
+        :param severity: the severity of the issue on a scale of 1-4
+        :return:
+        """
+        valid_severity_levels = ['1', '2', '3', '4']  # error | warning | information | hint
```
kevin-bates, line 412: I think it would be nice to use an enum for these values so that the enumerated value could be used rather than the stringized integer everywhere.
```python
from enum import Enum
class ValidationSeverity(Enum):
    Error = '1'
    Warning = '2'
    Information = '3'
    Hint = '4'
```
```suggestion
        valid_severity_levels = [ValidationSeverity.Error, ValidationSeverity.Warning, ValidationSeverity.Information, ValidationSeverity.Hint]
```
You would likely need to use `str(severity.value)` below if you truly wanted, say, "1" to be sent as the value.

Is there a reason these are string values and not ints?  If the later, you could derive from `IntEnum` and go with `severity.value`.

At any rate, using enum may not work out, but we should define "constants" for the numeric strings at a minimum.

kevin-bates (commented):
Thanks Alan, this service will be useful.

I kinda wonder if a schema already exists, or could be produced, such that most of this could be accomplished by validating the pipeline against its schema?

elyra/validation/validate.py
akchinSTC: nope there could be more than two. 
further above I iterate thru the cycles found in the graph. 
for each cycle I check to see if its a single node ID, if so, I know its a cycle into itself e.g. node1 -> node1.....
otherwise it should return an edge representing the first edge of the cycle...e.g. node1->node2 

this else statement is just a look-up loop to find the link ID associated with that edge.

elyra/__init__.py
akchinSTC: updated

elyra/validation/validate.py
akchinSTC: updated

elyra/validation/validate.py
akchinSTC: updated

elyra/validation/validate.py
akchinSTC: updated

elyra/validation/validate.py
akchinSTC: you're probably right about moving the pattern verification to the runtime images schema. 
as far as the second part, we had a brief like 30 second talk about this in one of the scrums, and decided that verifying the existence of the images via a call to the registry is out of scope with these changes and shouldnt be attempted. 

with that said, Ill remove this check and leave it as a purely structural check e.g. the parameter is present and is of the correct type.

elyra/validation/validate.py
kevin-bates: Would you mind adding comments as to how cycles are detected?

elyra/validation/validate.py
akchinSTC: updated

elyra/validation/handlers.py
akchinSTC: updated

elyra/validation/validate.py
akchinSTC: sure no problem, its mostly the networkx lib doing the work though

akchinSTC: shoot, welp its back to the drawing board, as the function call to look for cycles does not do a complete np depth first lookup and only returns the first cycle. will tackle this on monday

kevin-bates: Would the `simple_cycles()` method be better for this?  Do we care if there are multiple cycles?  Yes, it would be great to be able to highlight various cycles when multiple are present, but I suspect pointing out one (the first), would alert the user that multiple could be in play and they would either fix the others or find out the next when validating again.

Regarding the code, I have a few more comments:
1. It seems like it would be easier (more clear) to start from an empty graph and, in a single pass through the node_list, add both the node and its edges - avoiding the second loop to add the nodes.
2. I think the edge coordinates might be reversed.  Since these are _inputs_ to the current node, the edge data seems like it should be the `node_id_ref` first, then _this_ node id since the edge is input _to_ this node.
3. I'm not sure we need the specification of the empty string if a given node has no inputs - that doesn't seem correct.  The isolated nodes should be identifiable by not appearing in any edge tuples.

This would yield something like the following...
```python
        graph = nx.DiGraph()
        for node in node_list:
            graph.add_node(node['id'])
            if 'links' in node['inputs'][0]:
                for link in node['inputs'][0]['links']:
                    graph.add_edge((link['node_id_ref'], node['id']))
```

Regarding cycles, I think we do want to use `simple_cycles()` which returns a list of cycles where each cycle is a list of edges (tuples).
Since the graph is built from the _inputs_, the "interesting node" would be the second element of the edge tuple.  It seems like, for cycles, the `data` within the `ValidationResponse` should simply be a list of nodes composing the simple cycle - which can be built directly from the results of the simple cycle call and we shouldn't need to walk the nodes, but just use the second tuple of each cycle element's edge.

```python
        cycles_detected = nx.simple_cycles(graph)

        for cycle in cycles_detected:
            nodes_in_cycle = []
            for edge_u, edge_v in cycle:
                nodes_in_cycle.append(edge_v)
                
            response.add_message(severity=ValidationSeverity.Error,
                                                      message_type="circularReference",
                                                      message="A cycle was found within this pipeline",
                                                      data={"pipelineID": pipeline_id, "nodes_in_cycle": nodes_in_cycle})
```
I don't know what the consumer of these response objects expects in this case, so the array may need to be split up across multiple messages - one for each node in a given cycle - and moved to within the inner loop, eliminating the need for the `nodes_in_cycles` array.

elyra/validation/validate.py
kevin-bates: Never mind - I see this (and the resource validators) are conditional and only called when not 'local'.  Sorry I didn't notice that prevously.

akchinSTC:
the graph validation function has been updated to provide a list of the associated linkIDs in a cycle, for each cycle (numbered as cycleNumber). 
example of a pipeline with a 4-node cycle
```        {
            "severity": 1,
            "source": "Elyra Pipeline Validation Service",
            "type": "circularReference",
            "message": "A cycle was found within this pipeline",
            "data": {
                "cycleNumber": 1,
                "linkIDList": [
                    "fe572234-b61e-40a0-946b-5745471f805c",
                    "83db840c-8a70-43f8-beb4-e1f2aa4bf48c",
                    "c618674d-6134-4fa2-b28b-4b1d929a2140",
                    "d89a0cd3-fc0f-4e42-b67d-b286e160681a"
                ],
                "pipelineID": "primary"
            }
        },
```

elyra/validation/validate.py
```diff
@@ -401,44 +423,57 @@ def _validate_pipeline_graph(self, pipeline: Dict, response):
...
-                        if cycle[0] == link['node_id_ref'] and cycle[1] == node['id']:
-                            response.add_message(severity=ValidationSeverity.Error,
-                                                 message_type="circularReference",
-                                                 message="A cycle was found within this pipeline",
-                                                 data={"nodeID": node['id'],
-                                                       "pipelineID": pipeline_id})
+            size_of_cycle = len(cycle)
+            for i in range(0, size_of_cycle):
+                if cycle_counter not in link_dict_table:
+                    link_dict_table[cycle_counter] = []
```
kevin-bates, line 448: Can this `if` block be moved to prior to this for loop?  Seems like it's only predicated on `cycle_counter`.

elyra/validation/validate.py
```diff
@@ -401,44 +423,57 @@ def _validate_pipeline_graph(self, pipeline: Dict, response):
...
+                if size_of_cycle == 1:
+                    link_dict_table[cycle_counter].append(self._get_link_id(pipeline, cycle[i], cycle[i]))
+                elif i == size_of_cycle - 1:
```
kevin-bates, line 451: These first two clauses are redundant since a cycle of 1 will have `i` be the last node of the cycle.  It might be easier to read if you defined 'the last node' as requiring nodes (`i` and `0`), while all others use (`i` and `i+1`).

elyra/validation/validate.py
akchinSTC: done

bourdakos1:
In order to provide the best user experience, I believe the UI will need to be fairly chatty with the validation service which could cause some performance issues as a normal rest api. Until `pipeline-editor` has implemented language server support I think we will need to settle with duplicated validation logic on the frontend and backend: 
Screen Shot 2021-05-20 at 1 22 25 PM

> **Note:** VS Code requires a language server to get the integrated diagnostics support

Once the UI supports the language server protocol, it shouldn't be to hard wrap our validation service into a language server:
Screen Shot 2021-05-20 at 1 22 33 PM

kevin-bates:
This implies we should not publish the validation service via the REST API.  Not sure how the response should be addressed.  Would the `submit` and `export` endpoints return the `ValidationResponse`, but only when a validation issue is detected, yet return their current response payloads otherwise?

bourdakos1:
> This implies we should not publish the validation service via the REST API.

@kevin-bates I'm not against also having the validation exposed via REST API, but I'm not sure what it would be used for. 

> Would the submit and export endpoints return the ValidationResponse, but only when a validation issue is detected, yet return their current response payloads otherwise?

Yes, I think so. At least, the submission and export APIs should use the validation service as a guard. The submit/export APIs don't necessarily need to conform to the full `ValidationResponse` spec, the response given via the APIs would just be used to show an error message in a dialog in the UI or logging an error message if submitting via CLI. 

The UI will be directly communicating with the language server in order to display all the rich UI hints

elyra/__init__.py
```diff
@@ -53,6 +54,7 @@ def _load_jupyter_server_extension(nb_server_app):
          PipelineConfigHandler),
         (url_path_join(web_app.settings['base_url'], r'/elyra/contents/properties/%s' % (path_regex)),
          FileParserHandler),
+        (url_path_join(web_app.settings['base_url'], r'/elyra/pipeline/validation'), PipelineValidationHandler),
```
lresende, line 57: This should be more like a verb - `/validate`?

elyra/validation/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
```
lresende, line 1: We shouldn't need to add a new folder, this is all in the context of Pipelines so it should be on the pipeline folder.

elyra/validation/validate.py
lresende: I believe we added a method to validate runtimes there with a recent PR or with kersten's prs. Can't remember.

setup.py
```diff
@@ -79,6 +79,7 @@
         'nbconvert>=5.6.1,<6.0',
         'nbdime>=3.0.0.b1',
         'nbformat>=5.1.2',
+        'networkx>=2.5.1',
```
lresende, line 82: How is this new dependency being used?

lresende:
> Until `pipeline-editor` has implemented language server support

I believe the LSP is out of scope for the lab side of things unless lab itself move towards that direction.

setup.py
bourdakos1: I’m guessing this is for the circular references check

bourdakos1:
> I believe the LSP is out of scope for the lab side of things unless lab itself move towards that direction.

@lresende I'm not sure what you mean by this? Either way, I think we will need to use websockets in order to support server <-> client validation. I just think it makes sense to follow the LSP specification, because it should be the same amount of work and then could be used anywhere

elyra/pipeline/handlers.py
```diff
@@ -86,10 +88,18 @@ async def post(self, *args, **kwargs):
...
         self.log.debug("JSON payload: %s", pipeline_definition)
 
-        pipeline = PipelineParser().parse(pipeline_definition)
+        root_dir = self.settings['server_root_dir']
+        expanded_path = get_expanded_path(root_dir)
+        validation_response = PipelineValidationManager().validate(pipeline=pipeline_definition,
+                                                                   root_dir=expanded_path)
+
+        validation_msg = validation_response.to_json()
+        json_msg = json.dumps(validation_response.to_json())
```
kevin-bates, line 97: We shouldn't convert to JSON twice.  Instead, I would recommend changing `validation_msg` to `response`, then perform the `to_json()` call unconditionally _after_ the validation check below (and per the next comment).

elyra/pipeline/handlers.py
```diff
@@ -86,10 +88,18 @@ async def post(self, *args, **kwargs):
...
 
-        response = await PipelineProcessorManager.instance().process(pipeline)
-        json_msg = json.dumps(response.to_json())
+        if "issues" not in validation_msg:
+            pipeline = PipelineParser().parse(pipeline_definition)
+            response = await PipelineProcessorManager.instance().process(pipeline)
+            json_msg = json.dumps(response.to_json())
```
kevin-bates, line 102: As noted above, this indentation could be removed so that `response` either reflects a fatal validation response or the response from the operation.

ptitzler:
Does validation also support detection of non-fatal problems that would only warrant a warning?

akchinSTC:
@ptitzler, yes it def can, if you have a specific warning scenario in mind, please add it here

ptitzler:
Nothing I can communicate yet, but I wanted to make sure that he design already allows for info/warn/error messages. It wasn't clear from the descriptions above.

ptitzler:
Additional requirement for file dependency validation: https://github.com/elyra-ai/elyra/issues/1706 (validation should detect references that cannot be handled by the packaging process)

kevin-bates:
@ptitzler - should this be qualified with "if the node-level property is not already set"?
> if present, a new pipeline-level property overrides node-level property

My understanding was that pipeline-level properties only flow if the same property is not already set on the node (so nodes can override the "default").

ptitzler:
>  should this be qualified with "if the node-level property is not already set"?

Yes, the wording wasn't good. I changed the comment to list the scenarios instead and deferred to the PR instead for details to avoid duplicate discussions

akchinSTC:
Looks like we make the call to validate the pipeline here :
https://github.com/elyra-ai/pipeline-editor/blob/d47d79f9b164350b4d6f57f8f3724482b4c3c6ba/packages/pipeline-services/src/validation/index.ts#L91
which returns a list of similar objects to this validation service called "problems". But im not clear about how these "problems" are used to highlight the issues in the visual pipeline editor. 
@bourdakos1 - if these are used for the highlighting of issues in the UI, is it just a matter of modifying the response from these changes to match your schema and swapping out the validate calls? e.g. Indiana jones and the temple of doom

akchinSTC:
Due to the performance issues, we're only going to hit this service at pipeline run/submission time.

kevin-bates:
Since `parse()` no longer performs basic validation, it seems like we should make sure the CLI option validates the pipeline in (or before) [`_execute_pipeline()`](https://github.com/elyra-ai/elyra/blob/b92e0422147cd54a50e52d62435a0dcf47dcc233/elyra/cli/pipeline_app.py#L88) and conveys the issues list in a reasonable manner.

akchinSTC:
Sample Error output dialog box:
image

akchinSTC:
This PR is dependent on #1854

elyra/pipeline/component_registry.py
```diff
@@ -102,7 +103,7 @@ def get_generic_component(component_id: str) -> Component:
         return ComponentRegistry._generic_components.get(component_id)
 
     @staticmethod
-    def to_canvas_palette(components: List[Component]) -> dict:
+    def to_canvas_palette(components: List[Component]) -> str:
```
kevin-bates, line 106: Just a heads up that #1882 is returning a dict (and calling `json.loads()`) in the method.  Good catch on the invalid hint though.  Same goes for `to_canvas_properties()` below.

elyra/pipeline/handlers.py
```diff
@@ -134,3 +156,31 @@ async def get(self, processor, component_id):
...
+    async def post(self):
+        self.log.debug("Pipeline Validation Handler now executing post request")
+
+        root_dir = self.settings['server_root_dir']
+        expanded_path = get_expanded_path(root_dir)
+
+        pipeline_definition = self.get_json_body()
+        self.log.debug("Pipeline payload: %s", pipeline_definition)
+
+        response = await PipelineValidationManager().validate(pipeline=pipeline_definition,
```
kevin-bates, line 180: The singleton instance should be obtained using `PipelineValidationManager.instance()`.  This comment applies throughout - including tests.

elyra/pipeline/tests/test_validation.py
```diff
@@ -0,0 +1,318 @@
...
+        with open(f'elyra/pipeline/tests/resources/validation_pipelines/{pipeline_filepath}') as f:
+            pipeline = json.loads(f.read())
+            return pipeline, response
+
+    yield _function
+
+
+def test_basic_pipeline_structure(load_pipeline):
+    pipeline, response = load_pipeline('generic_basic_pipeline_only_notebook.pipeline')
+    PipelineValidationManager()._validate_pipeline_structure(pipeline=pipeline,
```
kevin-bates, line 40: Because this is a singleton, it may be best to create a fixture that returns the instance, yields that instance, then clears the instance after the yield so the singleton is destroyed (and re-created) on each test.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,552 @@
...
+            self._response['issues'].append(diagnostic)
+
+        if severity is ValidationSeverity.Error:
+            self._has_fatal = True
+
+    def to_json(self):
+        return self._response
+
+
+class PipelineValidationManager(SingletonConfigurable):
```
kevin-bates, line 89: The initial instance of this should be created in __init__.py with the other singletons.  Please be sure to set the parent parameter.  If that's the only parameter you have, you shouldn't need your own __init__() method definition.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,552 @@
...
+
+        for single_pipeline in pipeline_json['pipelines']:
+            node_list = single_pipeline['nodes']
+            pipeline_runtime = 'local' if pipeline_runtime == 'generic' else pipeline_runtime
+            component_list = await PipelineProcessorManager.instance().get_components(pipeline_runtime)
+            components = json.loads(ComponentRegistry.to_canvas_palette(component_list))
+            for node in node_list:
+                if node['type'] == 'execution_node':
+                    node_data = node['app_data']
+                    if node['op'] in ['execute-r-node', 'execute-python-node', 'execute-notebook-node']:
```
kevin-bates, line 244: We should add a class method on Operator that takes the `node['op']` value and returns true if `isGenericOperator()`.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,552 @@
...
+                                                filename=node_data['filename'], response=response)
+
+                        # If the running locally, we can skip the resource and image name checks
+                        if pipeline_execution != 'local':
+                            self._validate_container_image_name(node, response=response)
+                            for resource_name in ['cpu', 'gpu', 'memory']:
+                                if resource_name in node_data.keys():
+                                    self._validate_resource_value(node, resource_name=resource_name, response=response)
+                        if pipeline_runtime == 'kfp' and node_data['filename'] != node_data['ui_data']['label']:
+                            self._validate_ui_data_label(node=node, label_name=node_data['ui_data']['label'],
```
kevin-bates, line 256: Heads up.  #1882 moves the label field to app_data.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,552 @@
...
+        pipeline_json = json.loads(json.dumps(pipeline))
+
+        for single_pipeline in pipeline_json['pipelines']:
+            node_list = single_pipeline['nodes']
+            pipeline_runtime = 'local' if pipeline_runtime == 'generic' else pipeline_runtime
+            component_list = await PipelineProcessorManager.instance().get_components(pipeline_runtime)
+            components = json.loads(ComponentRegistry.to_canvas_palette(component_list))
+            for node in node_list:
+                if node['type'] == 'execution_node':
+                    node_data = node['app_data']
```
kevin-bates, line 243: Heads up. #1882 moves the _interesting_ node properties under `node['app_data']['component_parameters']`.

elyra/pipeline/validate.py
kevin-bates: Rather than pass `node` to every validation method, it might be cleaner to extract the node-id and node-name information once and pass them individually and `node` only when necessary.

I wonder if it would make since to have a NodeValdation class that takes the node and response in its constructor and has a simple `validate()` method where each of these definitions reside?  This would enable use of instance variables and reduce parameters, etc.  Not a big deal though.

kevin-bates (commented):
Some further comments - mostly in regards to #1882.

elyra/pipeline/component_registry.py
akchinSTC: got it, rebased on 1882 and changing it back

packages/ui-components/src/RequestErrors.tsx
```diff
@@ -36,6 +36,15 @@ export class RequestErrors {
       return this.server404(response.requestPath);
     }
 
+    const issues = JSON.parse(response.message).issues;
```
karlaspuldaro, line 39: This line is causing the code snippets test to fail as it's being shared with other metadata types. 
In that case `response.message` is a string, not a json object. Maybe we should do a try catch here. @ajbozarth  do you agree?

karlaspuldaro: The if statement proposed above solves the server error issue, only not sure it is the best approach, so I'll wait for confirmation before pushing the change to this PR.

ajbozarth: a try catch would be the correct way to handle this, but I believe rather than having this handling on the front end the backend should just return the json as is just like how we do stack traces. I'm writing up a POC for this right now and will share at EOD scrum

ajbozarth:
@akchinSTC I pushed a small fix for the issue @karlaspuldaro found. The `web.HTTPError`s that return the issues have been replaced with a block of code that returns the json as an object rather than a string. That block of code should probably be moved into some utility function somewhere that both locations of it can call instead of having the duplicate code.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,631 @@
...
+            self._has_fatal = True
+
+    def to_json(self):
+        return self._response
+
+
+class PipelineValidationManager(SingletonConfigurable):
+
+    def __init__(self, **kwargs):
+        super().__init__()
```
kevin-bates, line 96: We should remove this or propagate `**kwargs` to the superclass constructor.  I think the former is probably best since it's not doing anything anyway.

elyra/pipeline/handlers.py
```diff
@@ -138,4 +174,33 @@ async def get(self, processor, component_id):
...
+        self.log.debug("Pipeline Validation Handler now executing post request")
+
+        root_dir = self.settings['server_root_dir']
+        expanded_path = get_expanded_path(root_dir)
+
+        pipeline_definition = self.get_json_body()
+        self.log.debug("Pipeline payload: %s", pipeline_definition)
+
+        response = await PipelineValidationManager.instance().validate(pipeline=pipeline_definition,
+                                                                       root_dir=expanded_path)
```
kevin-bates, line 201: After rebasing to master and instantiating the singleton instance in the application startup sequence, you _could_ obtain `root_dir` as a member variable of the singleton where it would be accessible via `self.root_dir` to the validation manager's method.  This might eliminate some potential confusion since other singletons that need the root directory are doing similarly.

elyra/pipeline/validate.py
akchinSTC: just kept it and pushed to the superclass.

elyra/pipeline/handlers.py
akchinSTC: mimic'ed the other singletons in the startup sequence

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,633 @@
...
+                                 message="Invalid JSON detected, unable to continue.")
+            return response
+
+        pipeline_execution = pipeline_json['pipelines'][0]['app_data'].get('runtime')  # local, kfp, airflow
+        pipeline_runtime = self._get_runtime_schema(pipeline, response)
+
+        self._validate_pipeline_structure(pipeline=pipeline, response=response)
+        await self._validate_compatibility(pipeline=pipeline, response=response,
+                                           pipeline_runtime=pipeline_runtime, pipeline_execution=pipeline_execution)
+        await self._validate_node_properties(root_dir=root_dir, pipeline=pipeline,
```
kevin-bates, line 124: `self._validate_node_properties()` should get `root_dir` from `self.root_dir`.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,633 @@
...
+
+    async def validate(self, pipeline: Dict) -> ValidationResponse:
+        """
+        Validates the pipeline JSON payload
+        :param pipeline: the pipeline definition to be validated
+        :return: ValidationResponse containing any and all issues discovered during the validation
+        """
+
+        response = ValidationResponse()
+        root_dir = self.root_dir
```
kevin-bates, line 108: This is not necessary since it's not used (after updating argument list of `self._validate_node_properties()`)

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,633 @@
...
+            response.add_message(severity=ValidationSeverity.Error,
+                                 message_type="invalidNodeProperty",
+                                 message="Property has a non-parsable value",
+                                 data={"nodeID": node_id,
+                                       "nodeName": node_label,
+                                       "propertyName": resource_name})
+
+    def _validate_filepath(self, node_id: str, node_label: str, property_name: str, server_root: str,
+                           filename: str, response: ValidationResponse, root_dir: Optional[str] = None) -> None:
+        """
```
kevin-bates, line 369: the "root" parameters are confusing.  I think think `server_root` should be removed and `self.root_dir` used for _that_ value, and the `root_dir` parameter should be renamed to what appears to be `file_dir` - wherever the file to run resides.

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,633 @@
...
+        Checks the file structure, paths and existence of pipeline dependencies.
+        Note that this does not cross reference with file path references within the notebook or script itself.
+        :param node_id: the unique ID of the node
+        :param node_label: the given node name or user customized name/label of the node
+        :param property_name: name of the node property being validated
+        :param root_dir: the absolute base path of the current elyra workspace
+        :param filename: the name of the file or directory to verify
+        :param response: ValidationResponse containing the issue list to be updated
+        """
+        root_dir = root_dir or server_root
```
kevin-bates, line 379:
```suggestion
        file_dir = fle_dir or self.root_dir
```

elyra/pipeline/validate.py
```diff
@@ -0,0 +1,633 @@
...
+                                                                  response=response)
+
+                        # Check label against kfp naming standards
+                        if pipeline_runtime == 'kfp' and filename != node_label:
+                            self._validate_label(node_id=node['id'], node_label=node_label, response=response)
+                        if dependencies:
+                            notebook_root_relative_path = os.path.dirname(filename)
+                            for dependency in dependencies:
+                                self._validate_filepath(node_id=node['id'], node_label=node_label,
+                                                        root_dir=os.path.join(root_dir, notebook_root_relative_path),
```
kevin-bates, line 290:
```suggestion
                                                        file_dir=os.path.join(self.root_dir, notebook_root_relative_path),
```

kevin-bates (commented):
Some more cleanup to use `self.root_dir` is necessary along with some parameter naming improvements (unless I'm misinterpreting things).

elyra/pipeline/validate.py
akchinSTC: fixed

elyra/pipeline/validate.py
akchinSTC: fixed

elyra/pipeline/validate.py
akchinSTC: fixed

elyra/pipeline/validate.py
akchinSTC: fixed

elyra/pipeline/validate.py
akchinSTC: fixed

akchinSTC:
@kevin-bates - thx for the review (...again) =)

elyra/pipeline/tests/test_validation.py
```diff
@@ -135,10 +135,9 @@ async def test_invalid_node_property_structure(monkeypatch, load_pipeline):
     pvm = PipelineValidationManager.instance()
```
kevin-bates, line 135: Does root_dir need to be conveyed when creating the validation manager instance?

akchinSTC: Ah, this is a hole in the coverage, tests were passing because I was monkeypatching past the filepath checks when using the manager and when when actually testing the filepath, I was calling the filepath test function directly and passing in the directory path. Adding another test case to fill this in.  Weird that the test failing here is passing locally.

elyra/pipeline/tests/test_validation.py
```diff
@@ -0,0 +1,355 @@
...
+        with open(f'elyra/pipeline/tests/resources/validation_pipelines/{pipeline_filepath}') as f:
+            pipeline = json.loads(f.read())
+            return pipeline, response
+
+    yield _function
+
+
+@pytest.fixture
+def validation_manager():
+    yield PipelineValidationManager.instance()
```
kevin-bates, line 40: We should probably follow this `yield` up with an explicit `PipelineValidationManager.clear_instance()` as it may not get cleared for the next test.

elyra/pipeline/tests/test_validation.py
kevin-bates: Yeah, I just added a comment that might help.  Since these are singletons, their behavior can be affected.

akchinSTC:
@kevin-bates - think we are good now, tests are passing although the front end integration tests may be failing due to the new release of JL 3.1

elyra/pipeline/tests/test_validation.py
```diff
@@ -0,0 +1,369 @@
...
+from elyra.pipeline.validate import PipelineValidationManager
+from elyra.pipeline.validate import ValidationResponse
+
+
+@pytest.fixture
+def load_pipeline():
+    def _function(pipeline_filepath):
+        response = ValidationResponse()
+
+        with open(f'elyra/pipeline/tests/resources/validation_pipelines/{pipeline_filepath}') as f:
```
kevin-bates, line 31: This (and the `validation_manager` fixture below) introduces the requirement that testing must occur from the root of the repo.  This will break normal CLI and IDE functionality.

We should be using temporary directories for all tests and emulate installation/deployment relative to those directories.  However, that will take some time to implement and these tests will work fine in fixed environments used in CI - so can be deferred IMO.

kevin-bates (approved):
The tests will be problematic for dev environments, but that can be resolved later.

PR #1668, opened 2021-05-13
Add indication of runtime into operation's environment

### LINKED ISSUE #1663: Set ELYRA_RUNTIME_ENV variable in execution environment for a component to determine it's execution environment
I'm developing the elyra component library and need to know if I'm running inside Kubeflow or in local mode as I need to switch between local data storage and COS data storage for example

Please provide a ELYRA_RUNTIME_ENV env variable set to either
local, kubeflow, airflow, ....

Find some other way (based on env variables already set where I am

**Additional context**
Main purpose now is to support https://github.com/elyra-ai/component-library to change behavior of different components depending on their runtime context

kevin-bates:
This pull request adds an indication of the current runtime into the operation's environment via the variable `ELYRA_RUNTIME_ENV`.  Although the discussion on the issue suggested this value be obtained from the runtime configuration schema, the value is actually a function of the processor _implementation_ corresponding to the runtime and reflected in the abstract `type` property.   I believe we should document that the value of `PipelineProcessor.type` should match the schema name of the runtime (as is the case so far with the exception of the special case "local runtime") and is the suggested value for the  _name_ of the `entrypoint` under which the [processor implementation is registered](https://github.com/elyra-ai/elyra/blob/8f0ead610878e74536cb561242519fde0e1d00b1/setup.py#L121-L124).  This way the three entities (runtime schema, pipeline processor implementation and corresponding `entrypoint` entry) are logically associated.

The current set of valid values are:
- **local**: indicating the operation (or node) is being processed in the local runtime context
- **kfp**: indicating the operation (or node) is being processed in the Kubeflow Pipelines context
- **airflow**:  indicating the operation (or node) is being processed in the Airflow context

These changes also consolidate the collection of envs which helps with testing.

- [x] Extend the [docs](https://github.com/elyra-ai/elyra/blob/master/docs/source/developer_guide/pipelines.md#registering-new-pipeline-processor) for users like Elyra _component_ and _processor_ authors.   This addition should, at a minimum, associate a PipelineProcessor's implementation to its runtime schema name (via the `type` property) and document the "built-in" values which `ELYRA_RUNTIME_ENV` can have.

The approaches for testing the 'local' case are different than the 'kfp' and 'airflow' cases.

For the 'local' test, the generated pipeline files were amended to assert the presence of the environment variable (`ELYRA_RUNTIME_ENV`) has a value of 'local' during the actual execution of the generated pipeline.

For the 'kfp' and 'airflow' scenarios, a `test_collect_envs()` test was added to each set of tests where the new `_collect_envs()` method was called on the respective processor, followed by appropriate assertions for the presence of other runtime-specific environment variables.

Resolves #1663
 
           in the file; or

           it.


elyra/pipeline/processor_kfp.py
```diff
@@ -350,6 +351,17 @@ def export(self, pipeline, pipeline_export_format, pipeline_export_path, overwri
...
 
+    def _collect_envs(self, operation: Operation, **kwargs) -> Dict:
+        """
+        Amends envs collected from superclass with those pertaining to this subclass
+
+        :return: dictionary containing environment name/value pairs
+        """
+        envs = super()._collect_envs(operation, **kwargs)
+        # Only Unix-style path spec is supported.
+        envs['ELYRA_WRITABLE_CONTAINER_DIR'] = self.WCD
```
lresende, line 362: @akchinSTC is this really only used by kfp and not applicable for airflow?

kevin-bates:
Note: we need a better story for bringing your own schema than placing the custom schema into elyra's installation area under `metadata/schemas`.

lresende (approved):
LGTM, there is still a question to @akchinSTC but otherwise good to go.

kevin-bates:
Thanks @lresende.  

Also, note that I did not remove the previous [class-hierarchy image](https://github.com/elyra-ai/elyra/blob/master/docs/source/images/pipeline-class-hierarchy.png) that is now unused.  Since it is incomplete, I think it might be best to remove it - assuming you can generate an updated copy if needed.  If agreed, I'll submit a commit to remove it.

elyra/pipeline/processor_kfp.py
akchinSTC: As far as I can tell yes this is only used for KFP for locating a suitable writable location for the mlpipeline metrics stuff that the kfp UI uses. These metrics do not exist in airflow or the airflow processor.

akchinSTC (approved):

elyra/pipeline/processor_kfp.py
ptitzler: Yes

docs/source/user_guide/pipelines.md
```diff
@@ -104,7 +104,17 @@ elyra-pipeline submit elyra-pipelines/demo-heterogeneous.pipeline \
 
 The `runtime-config`  should be a valid [runtime configuration](/user_guide/runtime-conf.md).
 
+#### Detecting the runtime from a component
```
ptitzler, line 107: We need to introduce the term `component` because afaik we haven't used it prior to 2.3.x in the context of pipeline artifacts. This is also needed because the Pipelines documentation topic starts with "Elyra utilizes its canvas component ", which is semantically completely different and can lead to confusion.

elyra/pipeline/processor_local.py
```diff
@@ -115,6 +115,13 @@ def operation_name(self) -> str:
     def process(self, operation: Operation):
         raise NotImplementedError
 
+    @staticmethod
+    def _collect_envs(operation: Operation) -> Dict:
+        envs = os.environ.copy()  # Make sure this process's env is "available" in the kernel subprocess
+        envs['ELYRA_RUNTIME_ENV'] = "local"  # Special case
```
ptitzler, line 121: Wouldn't it be safer to reference the `_type` class variable https://github.com/elyra-ai/elyra/blob/f17143b3561e0226b3c53de2fcd4a398fa34ff0a/elyra/pipeline/processor_local.py#L42 instead of hard-coding the value?

elyra/pipeline/tests/test_airflow_processor.py
```diff
@@ -296,3 +297,22 @@ def test_pipeline_tree_creation(parsed_ordered_dict, sample_metadata, sample_ima
...
+    pipelines_test_file = 'elyra/pipeline/tests/resources/archive/test.ipynb'
+
+    test_operation = Operation(id='this-is-a-test-id',
+                               type='execution-node',
+                               classifier='airflow',
+                               name='test',
+                               filename=pipelines_test_file,
+                               runtime_image='tensorflow/tensorflow:latest')
+
+    envs = processor._collect_envs(test_operation, cos_secret=None, cos_username='Alice', cos_password='secret')
```
ptitzler, line 312: This test should also cover the `cos_secret` code path, which would not return `envs['AWS_ACCESS_KEY_ID']` and `envs['AWS_SECRET_ACCESS_KEY']` for null/empty.

elyra/pipeline/tests/test_kfp_processor.py
```diff
@@ -112,3 +112,22 @@ def test_get_dependency_archive_name(processor):
...
+    pipelines_test_file = 'elyra/pipeline/tests/resources/archive/test.ipynb'
+
+    test_operation = Op(id='this-is-a-test-id',
+                        type='execution-node',
+                        classifier='kfp',
+                        name='test',
+                        filename=pipelines_test_file,
+                        runtime_image='tensorflow/tensorflow:latest')
+
+    envs = processor._collect_envs(test_operation, cos_secret=None, cos_username='Alice', cos_password='secret')
```
ptitzler, line 127: This test should also cover the `cos_secret` code path, which would not return `envs['AWS_ACCESS_KEY_ID']` and `envs['AWS_SECRET_ACCESS_KEY']` for null/empty.

docs/source/user_guide/pipelines.md
kevin-bates: Since we reference "nodes" I use that.   Thanks.

elyra/pipeline/processor_local.py
kevin-bates: `_type` is not on an `OperationProcessor` - which is where "local" runtimes perform env collection.  I could pass this as a parameter, but it (at least today) will never be used outside of a local "runtime".  Had we treated "local" as a full-fledged runtime (eh um), then we'd be able to use "type" - as I do in `RuntimePipelineProcess._collect_envs()`.

ptitzler:
User test for kfp/run, kfp/export/yaml, kfp/export/dsl, local/run, airflow/export yielded the expected results. Waiting for airflow test env to come back up before airflow/run can be tested.

ptitzler (approved):
LGTM

PR #1670, opened 2021-05-13
update tsconfig and fix errors

bourdakos1:

- [x] Test changes locally  

           in the file; or

           it.


packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -477,9 +489,13 @@ class CodeSnippetDisplay extends MetadataDisplay<
         ].model.value.text = codeSnippet.metadata.code.join('\n');
       } else {
         // Add new snippets
+        const snippetEl = document.getElementById(codeSnippet.name);
```
lresende, line 492:
```suggestion
        const snippetElement = document.getElementById(codeSnippet.name);
```

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -98,19 +98,22 @@ const CodeBlock: React.FC<ICodeBlockProps> = ({
...
     };
 
-    editorRef.current = servicesRef.current.factoryService.newInlineEditor({
-      host: codeBlockRef.current,
-      model: new CodeEditor.Model({
-        value: defaultValue,
-        mimeType: servicesRef.current.mimeTypeService.getMimeTypeByLanguage({
-          name: language,
-          codemirror_mode: language
+    if (codeBlockRef.current !== null) {
```
lresende, line 101: should we use something like isNotEmpty to check for null and empty string?

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -120,7 +123,7 @@ const CodeBlock: React.FC<ICodeBlockProps> = ({
   }, []);
 
   React.useEffect(() => {
-    if (editorRef !== null) {
+    if (editorRef.current !== undefined) {
```
lresende, line 126: On the block above we are checking against null while here we are checking against undefined. What's the differences?

tsconfig.base.json
```diff
@@ -1,20 +1,20 @@
 {
   "compilerOptions": {
+    "target": "es2015",
```
lresende, line 3: Are we missing the schema definition/reference on the top? 
` "$schema": "http://json.schemastore.org/tsconfig",`

lresende:
@bourdakos1 Have you tried the pipeline editor after applying these changes? Just want to make sure that removing some of the files from its tsconfig file does not cause other side effects.

packages/metadata-common/src/MetadataEditor.tsx
bourdakos1: I don't think so, the following would check if the value is `null`, `undefined`, or an empty value for the type (so for string it would check empty string):
```ts
if (codeBlockRef.current) {
}
```

I just prefer to explicitly check for null/undefined/empty string because I personally find it leads to less bugs

packages/metadata-common/src/MetadataEditor.tsx
bourdakos1: `null` and `undefined` are two separate types. I think the original intention was that `undefined` means that a variable hasn't been defined, like `let x;` `x` is `undefined`. Where `null` is a value that represents the absence of an actual value, normally more intensional, like `let x = null;` `x` is `null` here. It's also important to know that `null !== undefined`, but `null == undefined`. `undefined` can also be set, so you can do `let x = undefined;`, for this reason some people try to only use `undefined`, because `undefined` occurs naturally, but `null` doesn't, so if you only use `undefined`  there's normally less corner cases you need to think about. Another thing is that `null` is serializablefor JSON, but `undefined` isn't, so a js object like this: `{ a: "value", b: null }` would serialize to `{ "a": "value", "b": null }`, where `{ a: "value", b: undefined }` would serialize to `{ "a": "value" }`

That was a lot haha...

Now, the reason this is checking for `undefined`, is that `current` might be `undefined`, but can't be `null`. I think the original `null` check on `editorRef` wasn't doing anything

tsconfig.base.json
bourdakos1: I'm not sure, is this normally needed? It seems at least for vscode it is aware of the schema and I get completions, but I can add that if there are issues

bourdakos1:
@lresende it seems that at least the basic functionality is working for me, but this touches a lot of code so we might want more extensive manual tests to be done by others

ajbozarth:
I just pushed some updates to the PRs with:

- Fixed bug where new vars weren't saved on close
- Added (back) support for env var detection on node add
- Updated callback function to be more generic so it can be extended with more properties in the future

The last point was to partially address @bourdakos1 and @lresende discussion above. I believe for now leaving the refresh button inline where it actual applies makes sense, but I updated the code so that when we make it more generic in the future we only have the move the button, the actual callback function is already ready to support a full refresh.

With these changes I'm also moving this out of WIP for review

packages/script-editor/src/ScriptEditor.tsx
```diff
@@ -72,14 +71,14 @@ export class ScriptEditor extends DocumentWidget<
...
-  private scrollingWidget: ScrollingWidget<OutputArea>;
+  private kernelName?: string;
+  private dockPanel?: DockPanelSvg;
+  private outputAreaWidget?: OutputArea;
+  private scrollingWidget?: ScrollingWidget<OutputArea>;
   private model: any;
   private emptyOutput: boolean;
   private runDisabled: boolean;
-  private kernelSelectorRef: RefObject<ISelect>;
+  private kernelSelectorRef: RefObject<ISelect> | null;
```
marthacryan, line 81: Why are you using `... | null` rather than `kernelSelectorRef?: RefObject...`? (I don't have a preference, just wondering the difference)

marthacryan: (seeing this a few other places in the code)

marthacryan:
Code changes LGTM, but haven't tested locally

packages/script-editor/src/ScriptEditor.tsx
bourdakos1: `xxx?: yyy` is equivalent to `xxx: yyy | undefined` which isn't the same as `xxx: yyy | null`. I prefer `xxx?: yyy`, but we use `null` in a lot of places and I didn't want to change that and accidentally break things

packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -146,14 +155,17 @@ class CodeSnippetDisplay extends MetadataDisplay<
...
       ) {
-        notebookCellEditor.replaceSelection(
+        notebookCellEditor.replaceSelection?.(
           this.addMarkdownCodeBlock(snippet.metadata.language, snippetStr)
         );
       } else {
-        notebookCellEditor.replaceSelection(snippetStr);
+        notebookCellEditor.replaceSelection?.(snippetStr);
+      }
+      const cell = notebookWidget.model?.contentFactory.createCodeCell({});
```
ajbozarth, line 164: does moving this into the if statement not cause issues with the usage outside the if statement below?

packages/code-snippet/tsconfig.json
```diff
@@ -1,19 +1,7 @@
 {
-  "extends": "../../tsconfig-base",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "lib",
-    "rootDir": "src"
+    "outDir": "lib"
   },
-  "include": ["src/*"],
-  "references": [
```
ajbozarth, line 8: why are we getting rid of this section?

ajbozarth (commented):
A couple non-blocking question. I don't fully understand how the build still works after removing so much from the tsconfig, but I tested locally and it does work. So I'd say this could just be merged as is.

karlaspuldaro:
Code looks good.
I tested this locally and it works well for code snippets and script editors.

Although I'm seeing this alert element at the bottom of pipeline editor
image
Clicking the x does nothing, it just stays there. Wondering if it's my env or anybody else can reproduce this?

marthacryan:
I'm able to reproduce what Karla commented. I think it's due to this change: https://github.com/elyra-ai/elyra/pull/1670/files#diff-c3ebbc30bb0c07233b81530f86ee864bf04a6eaa2542f1b7f1601ebf595bc513R130
because the alert shows if the alert variable is not equal to *null*, but the default is now an empty string.

ajbozarth (approved):

marthacryan (approved):

lresende (approved):

PR #1671, opened 2021-05-14
CI Enhancements

bourdakos1:
This PR update CI to:
- cache yarn dependencies (I tried caching pip but it didn't really make a difference)
- parallelize validation tasks (linting server, testing server, linting UI, testing UI, and integration tests can all run independently)

Screen Shot 2021-05-21 at 12 46 01 PM

I was going to update this to be one make target per step, but most steps are just a single command.

@lresende should I create make targets for all of these or any other suggestions?

A couple other questions:

- What does "Build Documentation" do? Should we rename it to "Validate Documentation Build" or something?
- Do I need to run `python -m pip install --upgrade pip` before every pip install?
- What is the difference between `python setup.py bdist_wheel sdist && pip install dist/elyra-*-py3-none-any.whl` and `pip install .`
- Why does not using `--use-deprecated=legacy-resolver` take infinity years to install?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install
-        run: |
+        # Note: Tests fail if we don't install `wheel` for some reason.
+        run: | 
           python -m pip install --upgrade pip wheel
           pip install -q -r test_requirements.txt
           pip install --use-deprecated=legacy-resolver .
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,7 +119,7 @@ jobs:
         run: |
           yarn install
           python -m pip install --upgrade pip
-          pip install --use-deprecated=legacy-resolver .
+          pip install .
       - name: Test
         run: yarn test:unit
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,9 +85,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install
         run: |
-          pip install wheel
-          python setup.py bdist_wheel
-          pip install dist/elyra-*-py3-none-any.whl
+          pip install -q --upgrade pip --user
+          pip --version
+          pip install .
       - name: Test
         run: pytest -v elyra
 
@@ -112,9 +112,9 @@ jobs:
       - name: Install
         run: |
           yarn install
-          pip install wheel
-          python setup.py bdist_wheel
-          pip install dist/elyra-*-py3-none-any.whl
+          pip install -q --upgrade pip --user
+          pip --version
+          pip install .
       - name: Test
         run: yarn test:unit
 
@@ -141,9 +141,9 @@ jobs:
         # TODO: `make install` also lints and things, so we should break this down more.
         run: |
           yarn install
-          pip install wheel
-          python setup.py bdist_wheel
-          pip install dist/elyra-*-py3-none-any.whl
+          pip install -q --upgrade pip --user
+          pip --version
+          pip install .
       - name: Build
         run: yarn lerna run build
       - name: Install Elyra into JupyterLab
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,90 +15,131 @@
 #
 name: Elyra Tests
 on:
-  push:
-    branches: '*'
-  pull_request:
-    branches: '*'
+  push: # all branches
+  pull_request: # all branches
   schedule:
     # once a day at 3 am (UTC) (7 pm (PST))
     - cron:  '0 3 * * *'
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest]
-        python-version: [ '3.6', '3.7', '3.8' ]
-    steps:
-    - name: Checkout
-      uses: actions/checkout@v2
-      with:
-        clean: true
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
-      with:
-        python-version: ${{ matrix.python-version }}
-        architecture: 'x64'
-    - name: Set up Node
-      uses: actions/setup-node@v2
-      with:
-         node-version: '*'
-    - name: Create npm configuration for authentication
-      run: |
-        echo "scripts-prepend-node-path=true" >> ~/.npmrc
-        echo "prefix=~/.npm-global" >> ~/.npmrc
-        mkdir -p ~/.npm-global
-    - name: Install Python dependencies
-      run: |
-        python -m pip install --upgrade pip wheel jupyter_server
-        python -m pip install --upgrade jupyterlab
-    - name: Install NPM dependencies
-      run: |
-        sudo apt install yarn
-        npm install -g lerna@3.22.1
-        npm install -g rimraf@3.0.2
-        npm install -g typescript@4.1.3
-        npm install -g yarn@1.22.4
-    - name: Log current build tools version
-      run: |
-        python --version
-        node --version
-        npm --version
-        yarn --version
-        lerna --version
-        pip --version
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -15,90 +15,160 @@
...
+        run: yarn install --frozen-lockfile
+
+  lint-server:
+    name: Lint Server
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install
+        run: |
+          python -m pip install --upgrade pip
```
lresende, line 53: We already have these steps using make, could we avoid duplication here and use the make target.

.github/workflows/build.yml
```diff
@@ -15,90 +15,160 @@
...
+      - uses: actions/cache@v2
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - name: Install
+        run: yarn install
+      - name: Lint
+        run: yarn eslint:check --max-warnings=0
```
lresende, line 76: same as above.

.github/workflows/build.yml
```diff
@@ -15,90 +15,160 @@
...
-          /home/runner/.npm/_logs/*.log
-  image_validation:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install
+        # Note: Tests fail if we don't install `wheel` for some reason.
+        run: | 
+          python -m pip install --upgrade pip wheel
```
lresende, line 94: same as above

.github/workflows/build.yml
```diff
@@ -15,90 +15,160 @@
...
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - name: Install
+        run: |
+          yarn install
+          python -m pip install --upgrade pip
+          pip install --use-deprecated=legacy-resolver .
```
lresende, line 122: same as above

.github/workflows/build.yml
```diff
@@ -15,90 +15,160 @@
...
+      - uses: actions/cache@v2
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+            /home/runner/.cache/Cypress
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - name: Install
+        run: |
+          yarn install
```
lresende, line 147: same as above

lresende (changes requested):
Overall looks ok, but we should avoid duplicating the commands that are already available via make targets.

.github/workflows/build.yml
bourdakos1: We only have a make target `lint-ui` which runs both `eslint` and `prettier` together. I think it would be beneficial to run these in separate steps so it's easier to see what failed. Additionally, `lint-ui` runs both commands in `fix` mode instead of `check` mode. In CI we should also set `max-warnings` to zero so it will fail on any lint warnings/errors. 

Should I make two new make targets intended for checking?

.github/workflows/build.yml
bourdakos1: The `install-server` target will end up running flake as a side effect. Can we rewire the `install-xxx` targets to not do linting? Maybe only lint when running the full `make install`? Or create new make targets that only do installing and `install-xxx` calls those and lint?

kevin-bates: I'd be fine with that.  I would like to make sure we do linting on test-related targets, however.  The key being that linting gets performed prior to PR creation and is something that tends to be forgotten.

bourdakos1: Do you think we should just remove linting from install all together and just have it run on test?

kevin-bates: I guess the confusing thing here is that `make install` is actually _the_ build step and linting does make sense prior to building.  Perhaps we could introduce explicit build-xxx targets that include linting, although I'm not sure that helps from a CI perspective.

I guess I should ask, what is the issue with `flake8` running prior to building the server?

bourdakos1: I'm fine with linting before building the server locally. In CI linting is separated into its own job, so it would end up linting in the linting job twice and then also in every job where the server is built. If linting fails then none of the other tests would run

bourdakos1: The more I'm thinking about this, the more I realize that our make targets try to group everything into one giant command, where the CI is trying to separate everything into much smaller distinct steps

lresende: > The more I'm thinking about this, the more I realize that our make targets try to group everything into one giant command
And why is that a problem? In terms of lint, the only additional thing is "preparing/installing" dependencies that will be used in a future step... In the worst case, we separate in make, but we also don't want contributors to have to issue single commands for every single task.

.github/workflows/build.yml
lresende: But wouldn't you see in the log which task failed (e.g. eslint vs pretier)?

bourdakos1: Yea, you could look in the log to see which task failed, but it’s a bit clearer if you separate the commands into separate steps because then the log output for both commands are separated in the UI.

You can tell much quicker what went wrong

B7319152-5088-45FB-A7AE-4D5A9F196F9A

bourdakos1:
@lresende I tried to diagram the current `make install` flow (left) and a restructuring of it to be more accommodating for ci (right). Does this look okay?

Untitled Diagram

lresende:
I don't know if it makes a difference for your scenario, but on your diagram, `build server` is only the first command, while `install server` are the other two commands. 

Stepping back, what are you proposing here? Refactor the makefile so that users can still call `make lint` while the build would call `make prettier` and `make eslint`?

Also, this is an awesome diagram that MUST be added to some sort of documentation for us to use in the future.

bourdakos1:
> Stepping back, what are you proposing here? Refactor the makefile so that users can still call make lint while the build would call make prettier and make eslint?

@lresende Yes and this also makes it possible to call `make build-server` without linting. And building the ui without installing dependencies/linting etc...

lresende:
> > Stepping back, what are you proposing here? Refactor the makefile so that users can still call make lint while the build would call make prettier and make eslint?
> 
> @lresende Yes and this also makes it possible to call `make build-server` without linting. And building the ui without installing dependencies/linting etc...

As long we don't go against the contributor's expectations. I see this as, when I am developing my code and I build it, I want to see all the issues, including syntax errors, and lining problems... if we are decomposing the make file and the contributor will have to call multiple targets to accomplish this, I am -1... if we are decomposing the make file and the smaller tasks will be used only for the build, I am +1.

bourdakos1:
@lresende I think this is ready for review. Here are the graphs of `make install` and `make test`, let me know if I should change the names/grouping of any of the targets:

image

image

Note that linting will fail because we are stricter (we fail for warnings in ui), this will be gone once #1757 is merged

Makefile
```diff
@@ -124,10 +124,14 @@ install-extensions:
 build-jupyterlab:
 	jupyter lab build
 
-install-server: lint-server build-server install-server-package # Install backend
+prepare-server:
+	pip install --upgrade pip wheel
+
+only-install-server: prepare-server build-server install-server-package
+
+install-server: lint-server only-install-server # Install backend
```
lresende, line 132: do we really need to start breaking things to this granularity? 
linting on the server is 13 seconds, and the build runs in the background...

bourdakos1: it's helpful to have linting separate because if it fails it won't fail all part of ci only the linting section

bourdakos1: we don't need to get extremely granular, but I think that linting, unit tests, and integration tests should all have their own "step" and they shouldn't be duplicated across "jobs"

bourdakos1:
I accidentally pushed the "remove log" commit to the wrong branch

lresende (approved):

Issue #1696, opened 2021-05-25
ELYRA_RUNTIME_ENV variable is no longer set during pipeline execution or export


ptitzler:
**Describe the issue**

I've noticed that the `ELYRA_RUNTIME_ENV` variable that was introduced in https://github.com/elyra-ai/elyra/pull/1668 is no longer set to any value. The problem surfaces in all three runtime environment types (local/kfp/aa).

**To Reproduce**
1. Install 2.3/master
2. Create a pipeline
3. Create a notebook | script that dumps the environment variables
    ```
    import os
    print(os.environ.get('ELYRA_RUNTIME_ENV'))
    ```
4. Add notebook | script to pipeline
5. Run pipeline and inspect notebook output

**Screenshots or log output**

**Expected behavior**

`ELYRA_RUNTIME_ENV` is set properly

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 2.3/master (pulled 5/25)

ptitzler:
(I updated the original comment with exact recreate steps)
It took me a while to understand the root cause: 

- The environment variable detection code discovers that the notebook | script references the `ELYRA_RUNTIME_ENV` variable, and thus adds it to the list of environment variables for each node. 

![image](https://user-images.githubusercontent.com/13068832/119562214-f03a7000-bd5a-11eb-8c34-701b8a1b38fe.png)

- A user should never set this _"system-owned"_ variable (it'd defeat its purpose) and it remains therefore unset `ELYRA_RUNTIME_ENV=` when the pipeline runs or is exported.

From a code perspective:

- https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_local.py#L121-L122  (and the same for the other processors)
  ```
  envs['ELYRA_RUNTIME_ENV'] = "local"  # sets
  envs.update(operation.env_vars_as_dict()) # overrides with empty value  (see definition below)
  ```
- https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline.py#L152 # sets the variable to the empty string because it was detected in the notebook | script

ptitzler:
We need to make sure that the user cannot explicitly or implicitly override this environment variable and any other environment variable that is owned by Elyra.

kevin-bates:
Ah - good point/catch.  I'll update the tests to handle this case as well.  We should be able to just reverse the initialization sequences such that operation-provided envs are set first and system-owned env values override.

ptitzler:
[...]

PR #1701, opened 2021-05-26
Set system-owned envs after user-provided envs

kevin-bates:
The changes introduced in #1668 to indicate the runtime name within an operation failed when an operation referenced `ELYRA_RUNTIME_ENV` on KFP and Local runtimes (Airflow was fine).  This corrects those issues and addresses test cases to better handle that scenario - since that _is_ the use case that is being addressed.

Extended the tests to include cases where the system-owned variables are set in the operation to ensure the system values are properly reflected.  Ran a test against KFP since the unit tests test this functionality at a lower level for the Kfp and Airflow processors.  (The local test can address this since it includes an end-to-end operation.)

Fixes: #1696

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -141,7 +141,7 @@ class Operation(object):
         """
         Operation stores environment variables in a list of name=value pairs, while
         subprocess.run() requires a dictionary - so we must convert.  If no envs are
-        configured on the Operation, an empty dictionary is return, otherwise envs
+        configured on the Operation, an empty dictionary is returned, otherwise envs
         configured on the Operation are converted to dictionary entries and returned.
         """
         envs = {}
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -141,8 +141,8 @@ class Operation(object):
         """
         Operation stores environment variables in a list of name=value pairs, while
         subprocess.run() requires a dictionary - so we must convert.  If no envs are
-        configured on the Operation, the existing env is returned, otherwise envs
-        configured on the Operation are overlayed on the existing env.
+        configured on the Operation, an empty dictionary is return, otherwise envs
+        configured on the Operation are converted to dictionary entries and returned.
         """
         envs = {}
         for nv in self.env_vars:
@@ -150,11 +150,10 @@ class Operation(object):
                 nv_pair = nv.split("=")
                 if len(nv_pair) == 2 and nv_pair[0].strip():
                     envs[nv_pair[0]] = nv_pair[1]
+                elif logger:
+                    logger.warning(f"Could not process environment variable entry `{nv}`, skipping...")
                 else:
-                    if logger:
-                        logger.warning(f"Could not process environment variable entry `{nv}`, skipping...")
-                    else:
-                        print(f"Could not process environment variable entry `{nv}`, skipping...")
+                    print(f"Could not process environment variable entry `{nv}`, skipping...")
         return envs
 
     @property
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -146,15 +146,33 @@ class Operation(object):
         """
         envs = {}
         for nv in self.env_vars:
-            if nv and len(nv) > 0:
-                nv_pair = nv.split("=")
+            if nv:
+                nv_pair = nv.split("=", 1)
                 if len(nv_pair) == 2 and nv_pair[0].strip():
-                    envs[nv_pair[0]] = nv_pair[1]
-                elif logger:
-                    logger.warning(f"Could not process environment variable entry `{nv}`, skipping...")
+                    if len(nv_pair[1]) > 0:
+                        envs[nv_pair[0]] = nv_pair[1]
+                    else:
+                        Operation._log_info(f"Skipping inclusion of environment variable: "
+                                            f"`{nv_pair[0]}` has no value...",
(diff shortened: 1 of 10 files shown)


lresende:
Should we allow the user to discover + set these at all?

kevin-bates:
> Should we allow the user to discover + set these at all?

System-owned properties, in general, were brought up on the issue (https://github.com/elyra-ai/elyra/issues/1696#issuecomment-848259469) and we should probably discuss it as something outside of this particular PR.

Generally speaking, I don't think users should be allowed to _set_ system-owned envs.   `ELYRA_RUNTIME_ENV` is a system-owned env that the system controls, yet for use by pipeline authors.  It should not be user-settable.   However, when we considered accessing a Gateway server from a (non-local) runtime, we were going to expose all of the gateway configuration "knobs" via `ELYRA_GATEWAY_` variables.  So those _could_ be considered something that should be user-settable.  But are they considered _system-owned_ in that case?  I think so.  So then it really becomes an issue as to how the application would know what is and what is not user-settable. Since we don't have any `ELYRA_`-prefixed envs that we think should be user-settable at this moment, we could make a statement that `ELYRA_`-prefixed envs are system-owned and non-user-settable.  But, envs make great "knobs" and we might want a user-settable, system-owned `ELYRA`-prefixed env someday.  :smile:

So perhaps the answer is:  _`ELYRA_`-prefixed envs are system-owned, and, yes, you _can_ set them.  However, non-user-settable system-owned envs will NOT honor your value and WILL BE overridden by the server._ Because they are truly system-owned.  In this case, the best "discovery" mechanism for user's setting system-owned envs will be the documentation.

(We also have some `AWS_` variables that should be considered system-owned, despite the fact that they derive their _values_ from runtime-configuration properties (i.e., user-controlled metadata).  These should not be settable outside of the values the _server_ pulls from the runtime configuration to set them - nor should they appear anywhere (so non-discoverable).)

Just my two cents.

ptitzler:
> we should probably discuss it as something outside of this particular PR.

I agree. Created https://github.com/elyra-ai/elyra/issues/1704 for follow-up

> the best "discovery" mechanism for user's setting system-owned envs will be the documentation.

Created https://github.com/elyra-ai/elyra/issues/1703 for follow-up.

elyra/pipeline/pipeline.py
```diff
@@ -141,20 +141,19 @@ def env_vars_as_dict(self, logger: Optional[Logger] = None) -> Dict:
         """
         Operation stores environment variables in a list of name=value pairs, while
         subprocess.run() requires a dictionary - so we must convert.  If no envs are
-        configured on the Operation, the existing env is returned, otherwise envs
-        configured on the Operation are overlayed on the existing env.
+        configured on the Operation, an empty dictionary is return, otherwise envs
```
ptitzler, line 144: typo

```suggestion
        configured on the Operation, an empty dictionary is returned, otherwise envs
```

kevin-bates: Done.

elyra/pipeline/pipeline.py
```diff
@@ -141,20 +141,19 @@ def env_vars_as_dict(self, logger: Optional[Logger] = None) -> Dict:
...
+        configured on the Operation, an empty dictionary is returned, otherwise envs
+        configured on the Operation are converted to dictionary entries and returned.
         """
         envs = {}
         for nv in self.env_vars:
             if nv and len(nv) > 0:
                 nv_pair = nv.split("=")
                 if len(nv_pair) == 2 and nv_pair[0].strip():
                     envs[nv_pair[0]] = nv_pair[1]
+                elif logger:
```
ptitzler, line 153: Follow up to this scenario:
 - User sets variable to `MYVAR=3=2`.
 - VPE does not raise a validation error
 - A warning message is logged and the variable ignored but the user never explicitly notified.

```
Could not process environment variable entry `MYVAR=3=2`, skipping...
```

Would it be better to 
- treat this as fatal error on the back-end
- add code to the validation engine to alert the user?

ptitzler: "add code to the validation engine to alert the user?" -> new issue

ptitzler:
Tested local/kfp/airflow and `ELYRA_RUNTIME_ENV`. Seems to work now as expected.

elyra/pipeline/pipeline.py
kevin-bates: I think we should preserve the intention of the user and not treat any inability to parse environment variables as fatal at this particular location.  Should we treat these as errors detected during validation?  Perhaps, but that's not in the scope of this issue.

Regarding `MYVAR=` scenarios.  This was discussed in today's scrum and the consensus was that this particular _non-value_ scenario should not result in `MYVAR`'s inclusion in the environment of the runtime.  However, an explicit, yet empty, scenario like `MYVAR=""` _should_ result in `MYVAR`'s inclusion in the environment.

If you're in agreement here, I can include this functionality in this PR - please advise.

ptitzler: > If you're in agreement here, 

Yes, that makes sense

How about handling multiple `=` signs in the backend code? https://github.com/elyra-ai/elyra/pull/1701#discussion_r639863979. Having thought about this some more I believe `A=2=3` could be a valid scenario and should not be ignored. I am thinking about the component library here, which uses environment variables extensively and might have a need to have an `=` sign somewhere in the value.

ptitzler (approved):
LGTM

lresende (approved):

PR #1705, opened 2021-05-26
Add make graph generation script

bourdakos1:
Since the handmade graph of the makefile targets seemed useful, I quickly made a script that generates something similar:

image

Usage:
```
yarn graph <make-target>
```

Example:
```
yarn graph install
```

(I'm not sure how to pass arguments with `make` to do something like `make graph install` )

### DIFF
--- a/etc/scripts/generate-make-graph.ts
+++ b/etc/scripts/generate-make-graph.ts
@@ -20,7 +20,7 @@ import path from 'path';
 
 import chalk from 'chalk';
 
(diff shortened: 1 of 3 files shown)


lresende:
@bourdakos1 Awesome, do you mind moving the script to https://github.com/elyra-ai/elyra/tree/master/etc/scripts, other then that LGTM

bourdakos1:
@lresende done 👍

lresende (approved):

PR #1707, opened 2021-05-27
Move note on pip version to top of pip section

mudspringhiker:
While installing Elyra using the installation instructions, I encountered the issue on pip version (which is found towards the end of the section on installation using pip) so it might be better to put the note on the pip version at the top. I merely changed the location of the "NOTE". Also, this is my introductory contribution. Thanks!


ptitzler (approved):
LGTM

PR #1708, opened 2021-05-27
Add support for pipeline properties

### LINKED ISSUE #602: Allow for display of pipeline | canvas widget properties in the Visual Pipeline Editor
As the creator of a notebook pipeline I would like to embed human-readable information in  my [notebook] flow that  enables other users to quickly determine what the purpose of (1) the pipeline and (2) the included notebooks is.  For illustrative purposes let's look at this pipeline, courtesy of https://github.com/CODAIT/covid-notebooks

image

Assuming user A has created this pipeline and shared it with user B (through whatever means), there's currently a potential information loss risk unless additional information is shared through other means (email, slack message, verbal communication, etc).
This information could entail:
 - what's the origin of the pipeline
 - what's the purpose of the pipeline
 - what's the purpose of a pipeline component (this information is likely embedded in the notebook but requires the user to open it to find out) 

Even if a creator currently chooses descriptive notebook names(like it was done here, e.g. `etl_us_data`), user B wouldn't know what `us_data` is referring to unless he/she opens the notebook or, in case of generic notebooks, inspects input parameter values or is aware of the pipeline origins. As of today, a pipeline creator would have to embed information about the origin and purpose of the pipeline in one of the notebooks (to make the pipeline "self-explaining"), which doesn't seem to be the appropriate place to convey the information, since a notebook might be part of many pipelines.

### LINKED ISSUE #1421: Support for global properties / env variables in pipeline editor
Some properties are confidential (eg. hostnames, usernames, passwords) - when a pipeline is shared, those have to be removed manually by opening each pipeline step's properties

A section for global properties attachable to the pipeline which are valid / accessible for all pipeline steps which can easily be removed/copy pasted

Writing a parser and removing env variables of the json file of the pipeline

### LINKED ISSUE #1699: Add pipeline properties
The pipeline editor should have the ability to add properties for an *entire pipeline* rather than just for *single nodes*.

Examples include: 
- Pipeline description
- Environment variables that apply to all nodes
- default runtime image
- ~~[not yet supported] mounted volumes~~
- ~~UI variables such as font size~~

Proposition for UI to edit these properties: add an additional tab into the right side pane, and the appearance of the editor would otherwise be similar to the node properties editor. 

Proposition for the definition of pipeline properties: same format as the node properties JSON (with adjustments as needed)

Thoughts on either design or additional properties that could be added? 

This supersedes the following issues: #602 #1421

marthacryan:
Adds the ability to edit pipeline properties (as opposed to *node* properties). Adds a tab to the right panel in the pipeline editor for the pipeline properties editor.

This relies on https://github.com/elyra-ai/pipeline-editor/pull/122 being merged / released, so I'll keep the blocked label / draft state until that's released. For now, it uses the code sandbox link so that testing is easier. 

Appearance:
image

Todo:
- [x] Add default runtime image field
- [x] Switch to using a text area (larger) rather than a text input for pipeline description
- [x] Add readonly fields like pipeline name / pipeline type
 
Fixes #1699
Fixes #1421
Closes #602 

           in the file; or

           it.


lresende:
@marthacryan Pipeline Editor 0.4.0 has been published now.

bourdakos1 (approved):
LGTM

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -164,20 +189,27 @@ const PipelineWrapper: React.FC<IProps> = ({
...
+        setUpdatedPipelineProperties(pipelinePropertiesCopy);
         changeHandler();
       })
       .catch(error => RequestErrors.serverError(error));
 
     return (): void => {
       currentContext.model.contentChanged.disconnect(changeHandler);
     };
-  }, []);
+  }, [updatedPipelineProperties]);
```
ajbozarth: This causes an infinitely looping call of this hook (and thus the api) since this var is set within this hook.

marthacryan:
> Big bug where the api in infinitely called once the properties is opened, otherwise the code looks good.

Fixed!

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -164,19 +189,27 @@ const PipelineWrapper: React.FC<IProps> = ({
...
         setUpdatedNodes(nodesCopy);
+        setUpdatedPipelineProperties(pipelinePropertiesCopy);
         changeHandler();
       })
       .catch(error => RequestErrors.serverError(error));
 
     return (): void => {
       currentContext.model.contentChanged.disconnect(changeHandler);
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bourdakos1, line 212: We shouldn't ignore these warnings

bourdakos1 (changes requested):

packages/pipeline-editor/src/PipelineEditorWidget.tsx
marthacryan: It caused an infinite loop when I added that dependency - what would you suggest as the solution for that?

ptitzler:
> The actual properties aren't functional though

What's the point of adding properties that have no effect? If there's no backend support they should not be exposed in the UI.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
marthacryan: This should be addressed in a follow-up PR that adds support for default runtime image - for now, this issue doesn't need to be addressed yet.

marthacryan:
Removing the env var and default runtime fields - planning for those to be added in a follow-up PR.

ajbozarth:
summary of the follow up plan for runtime images:

The backend will handle filling in the default runtime for nodes that are missing it. Also validation will only flag empty runtimes if a default is also not set.
As for the front end, we want to show a kinda of greyed out version of the default in node properties when it's not set so a user can see what runtime will be used

ajbozarth (approved):
lgtm

bourdakos1 (approved):

PR #1713, opened 2021-05-28
Support pipeline export for DEX/password-secured KF

### LINKED ISSUE #1712: Add uid/password-based DEX authentication support to generated KFP DSL
If Kubeflow is secured by DEX the currently exported Python DSL doesn't work out of the box because it does not include code that performs authentication. As a result the user has to add this code manually.

Pipeline export generates Python code that works if KF is secured by uid/password based DEX authentication. The relevant code is already included in https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_kfp.py#L490 and https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_kfp.py#L85-L110

Even though the code generator has access to the credentials, they should not be included in the generated code. Instead command line parameters should be used.

n/a

ptitzler:
This PR
  - supports pipeline export for KF target runtimes that are secured by DEX
  - closes #1712 

Notes:
 - The generated code does not work as one would expect if it is run from the JupyterLab Python editor. The reason is that `sys.argv` contains parameters that are "unexpected" from a user's point of view.
 - The changes have no impact on how pipelines are run from the VPE.
 - The changes have no impact on pipelines that are exported to KF YAML because the generated artifacts are the same for secured and unsecured environments.
  
 

- Created pipeline
- Created RTC for DEX-secured KF deployment
- Exported pipeline as Python DSL using above RTC
- Reviewed/linted generated Python code (no fatal errors)
- Ran generated code in a terminal window, testing the following scenarios
   - Invoke script with no parameters (OK - raises error/shows help)
   - Invoke script with uid parameter only (OK - raises error/shows help)
   - Invoke script with valid uid parameter and invalid password parameter  (OK - raises error)
   - Invoke script with invalid uid parameter and valid password parameter  (OK - raises error)
   - Invoke script with valid uid parameter and valid password parameter  (OK - uploads/runs pipeline)
- Created RTC for unsecured KF deployment
- Exported pipeline as Python DSL using above RTC
- Reviewed/linted generated python code (no fatal errors) 

 
           in the file; or

           it.


ptitzler:
Test artifacts (extract into workspace and export using VPE):  [i1712test.tar.gz](https://github.com/elyra-ai/elyra/files/6563529/i1712test.tar.gz)

ptitzler:
> The generated code does not work as one would expect if it is run from the JupyterLab Python editor. The reason is that `sys.argv` contains parameters that are "unexpected" from a user's point of view.

To illustrate:

image

kevin-bates:
Hi @ptitzler - thanks for the PR - it looks good on the surface. 

I really am more curious about your last comment.  I'm not following how anything run from the Lab python editor is germane to this.  The screenshot you display is the current `sys.argv` used to launch the kernel that is now the python REPL.

I understand the need to move the authentication generation code into the jinja template.  Can you share how a pipeline is invoked from the command line such that the jinja template comes into play?  Perhaps that's the link of understanding I'm missing.  Since it's a python command, were you expecting it can be run from the Python editor?  I'm not sure if general python applications can be run from the REPL.  Seems like you'd need to escape out of the REPL with `!python `.  Have you tried using a terminal that is also available via Lab (assuming you want the "one-stop-shop" approach of remaining in the browser)?

I guess I'm also asking for an example of `argv[0]` here: https://github.com/elyra-ai/elyra/pull/1713/files#diff-19f068ce06768ad5a740e4c3cfac478fc8f48eb3c091d717946f1fbf8b2854deR113

Thanks.

((There's a typo in the password placeholder argument (f-password).))

ptitzler:
> Can you share how a pipeline is invoked from the command line such that the jinja template comes into play?

The workflow I was trying to support is as follows:
 - user exports pipeline as Python DSL
 - user runs generated Python code to upload/run the pipeline 
    - in a terminal window in JL
    - in a terminal window outside of JL 
    - from the Python editor
 
> Perhaps that's the link of understanding I'm missing. Since it's a python command, were you expecting it can be run from the Python editor? 

Yes, exactly. As is the run feature seems to be too limited to be useful for anything else than a basic hello world app. (In the past I always ran the generated code that did not require command line parameters from the editor because it was a rather convenient way to do it.)

> Have you tried using a terminal that is also available via Lab (assuming you want the "one-stop-shop" approach of remaining in the browser)?

Yes, that worked as expected.

> I guess I'm also asking for an example of argv[0] here

None should be needed because `argv[0]` is the filename containing the code: 

```
(base) ~/elyra_dev/i1712/workspace $ python 1712_test.py 
Invocation: python 1712_test.py <kf-user-id> <f-password>
```

> ((There's a typo in the password placeholder argument (f-password).))

Fixed!

Issue #1703, opened 2021-05-26
Document development best practices for notebook | script based nodes


ptitzler:
**Is your feature request related to a problem? Please describe.**
In part prompted by https://github.com/elyra-ai/elyra/issues/1696 and related discussions.

**Describe the solution you'd like**

Add a new topic to the documentation (or augment existing documentation) that describes special considerations, such as proprietary environment variables (e.g. `ELYRA_RUNTIME_ENV`) that notebooks | scripts have access to during execution.

ptitzler:
Topic: directory layout to support running in containers (see #1706)

ptitzler:
- Topic: environment variables
- Topic: runtime images

ptitzler:
Topic: Elyra environment variables
  - `ELYRA_RUN_ID`

PR #1732, opened 2021-06-08
Add support for ELYRA_RUN_NAME environment variable

### LINKED ISSUE #1702: As a pipeline node, I want to have access to a unique pipeline run id
This feature request was initially raised by @simon-streamotion in https://github.com/elyra-ai/elyra/discussions/1686. 

- During [local, kfp, airflow] pipeline processing, each pipeline node shall have access to an identifier that is unique for each pipeline run irrespective of how it was started. For example, a pipeline run can be started via the Elyra GUI, via the Elyra CLI, via the KFP UI, via the KFP CLI, etc. This implies that the identifier value is _unknown_ until the pipeline is executed, and can therefore not be generated in advance, e.g. during pipeline pre-processing.
- The identifier shall be a string that is made available to every pipeline node via an environment variable, e.g. `ELYRA_RUN_ID`.
- If possible, that identifier shall include the pipeline name to enable the user to associate the identifier with the entity that is associated with it.

No workarounds come to mind.

ptitzler:
This PR introduces a new protected environment variable `ELYRA_RUN_NAME`, which is made available to notebooks | scripts during pipeline execution. The value is a unique string that changes for every pipeline run. For example
- run 1: `ELYRA_RUN_NAME`=`some_value`
- run 2: `ELYRA_RUN_NAME`=`some_other_value`

### Local runtime environment

`ELYRA_RUN_NAME` is set to `<pipeline-name>-<timestamp>`

Example value for pipeline file `give_run_name.pipeline`:

```
give_run_name-0617153527
```

### KFP runtime environment

`ELYRA_RUN_NAME` is set to the KFP `<run_name>`

Example value for pipeline file `give_run_name.pipeline`:

```
give_run_name-0617153855
```

See screen captures in https://github.com/elyra-ai/kfp-notebook/pull/102

### Airflow runtime environment

`ELYRA_RUN_NAME` is set to `<pipeline-name-timestamp>-<AA-DAG-run-id>`

Example value for pipeline file `give_run_name.pipeline`:

```
give_run_name-0617150717-scheduled__2021-06-16T00:00:00+00:00
```
Example DAG:
image

Task Info:

image

Job output (scheduled run):

image

Job output (subsequent manual run)

image

Fixes #1702

Notes:
- The actual value is the same for each node during pipeline execution.
- The actual value starts with the pipeline name, e.g. `my-pipeline-` followed by a sequence of characters that assure uniqueness.
- The trailing character sequence has either no semantic meaning (local execution) or is an run identifier (generated by KFP and AA). 
- The actual value _does not_ provide information about the order in which the runs where executed. 
- The actual value is not persisted by Elyra.
- If the user manually overrides the variable value in the node properties (e.g. sets `ELYRA_RUN_ID` to `bogusid`), the change is ignored, as it should be for all protected/system Elyra environment variables.

TODO:

- [x] Add documentation (https://github.com/elyra-ai/elyra/issues/1703)
- [x] Add support in local pipeline execution
- [x] Add support in KFP with Argo pipeline execution 
- [x] Add support in KFP with Tekton pipeline execution 
- [x] Add support in AA pipeline execution 

Dependencies:

- [x] [PR in `kfp-notebook` repo](https://github.com/elyra-ai/kfp-notebook/pull/102) (requires new published version of `kfp-notebook`)

- [Download/extract the test case archive](https://github.com/elyra-ai/elyra/files/6658749/1702_testcase.tar.gz)
  into a JL workspace
- Open `1702_test.pipeline` using the VPE (note that the comments indicate which nodes attempt to override the environment variable value - n/a for local)
- Run the pipeline locally, on KFP, and AA and inspect the outputs.
- [KFP/AA] Export the pipelines, upload, and run them

           in the file; or

           it.


ptitzler:
@kevin-bates For _local execution_ the proposed code generates id that looks like this: `1702_test2-65aff520-aedd-4c3d-9c31-9f3bb6e646ac` (`<pipeline-name>-<uuid>`). It might be a bit of overkill to use such a long value in light of the fact that the consumer would likely use it to generate unique file names, e.g. `model-1702_test2-65aff520-aedd-4c3d-9c31-9f3bb6e646ac`. If we use a formatted timestamp instead (`<pipeline-name>-<MMDDYYHHMMSS>`) we probably produce a reasonably safe unique id that is a lot shorter: `model-1702_test2-MMDDYYHHMMSS`. wdyt?

kevin-bates:
Yes, I agree.  I suppose this will make the value more like the "subsequent-run-pipeline-version" value used on non-local runtimes.

ptitzler:
> I suppose this will make the value more like the "subsequent-run-pipeline-version"

Exactly!

ptitzler:
_slightly off-topic_
Following up on the discussion during the community meeting I did a little prototyping and confirmed that the general approach (e.g. evaluating the KFP run id) used in this PR can also be used to store run output in a "sub-directory" in the object storage bucket. In this example the user ran the pipeline from the VPE and subsequently again manually from the KFP UI:
image

Content of one of the "sub-directories"

image

With just a few lines of code we can avoid that pipeline runs overwrite the output from prior runs.

kevin-bates:
Given this, do we still want to prefix the platform-provided run-id with our pipeline-name/version when composing `ELYRA_RUN_ID`, or just make `ELYRA_RUN_ID` essentially a UUID on all platforms (given `{RUN_ID_PLACEHOLDER}` and, presumably, `{{run_id}}` are also UUIDs)?
I believe it would be good to have a normalized variable for users to assume exist regardless of platforms, but prefixing that with what will likely be the name of the artifact's parent directory seems a little verbose.

ptitzler:
> Given this, do we still want to prefix the platform-provided run-id with our pipeline-name/version when composing ELYRA_RUN_ID, or just make ELYRA_RUN_ID essentially a UUID on all platforms.

My motivation for introducing the prefix was to give the user the ability to associate the "UUID" with a pipeline. The UUID can be something internal (like it is in KFP) that the user is not necessarily familiar with.

> (given {RUN_ID_PLACEHOLDER} and, presumably, {{run_id}} are also UUIDs)?

On Airflow (https://github.com/elyra-ai/airflow-notebook/pull/10#issue-666085843) it's unfortunately not a UUID. I'll take another look if there's something similar.

> I believe it would be good to have a normalized variable for users to assume exist regardless of platforms, but prefixing that with what will likely be the name of the artifact's parent directory seems a little verbose.

Could you please elaborate? Which artifact you are referring to?

To complicate things further:

For Kubeflow I managed to obtain the pipeline run name but it's done in a way that is not necessarily future proof because it relies on us utilizing static strings that are specific to Argo (`{{workflow.annotations.pipelines.kubeflow.org/run_name}}`) and Tekton (`$(context.pipelineRun.name)`).

image

From a user's point of view using the run name would likely be the most convenient solution because it has a meaning, whereas a UUID does not. However, it's possible that characters in the name might be illegal (depending on how the name is consumed, e.g. used as a file name), potentially requiring sanitization.

ptitzler:
Let me create a dedicated issue for the directory layout discussion to avoid confusion.

kevin-bates:
>> I believe it would be good to have a normalized variable for users to assume exist regardless of platforms, but prefixing that with what will likely be the name of the artifact's parent directory seems a little verbose.

> Could you please elaborate? Which artifact you are referring to?

The directory that contains the artifacts is based on the pipeline-name or pipeline-version, depending on whether this is the first or subsequent run respectively (e.g., `output_dir_test-0610134342`).  The ELYRA_RUN_ID is composed of the pipeline-version followed by the platform's notion of a _run id_ (e.g., `output_dir_test-0610134342-8b79eb94-93a4-4ed5-aa15-56dd86d1ccd1`).

In the use-case, the ELYRA_RUN_ID is used to store output artifacts, resulting in the following path in the respective bucket:

`output_dir_test-0610134342/output_dir_test-0610134342-8b79eb94-93a4-4ed5-aa15-56dd86d1ccd1`

with the first portion of the artifact directory being a duplicate of the parent directory, and I'm just wondering if a path like:

`output_dir_test-0610134342/8b79eb94-93a4-4ed5-aa15-56dd86d1ccd1`

would be more tolerable.  Given neither of these is user-friendly, I suppose this is more an issue of screen real estate.

By using the platform's run-id "in the raw", we could simply document what `ELYRA_RUN_ID` equates to on each platform w/o having to discuss the prefix.

ptitzler:
I see, thank you. We should be able to separate the value of `ELYRA_RUN_ID` (which the user could use for anything) and the value that's used to determine the location in COS to avoid duplication. That said ...

In fact, if we can obtain the run name we should consider exposing a variable named `ELYRA_RUN_NAME` instead of `ELYRA_RUN_ID`. For KFP the value is something the user entered, for AA it appears to be a generic string (see link in earlier comment).

kevin-bates:
That sounds good.  

> For KFP the value is something the user entered

how would this value change when submitted via KFP?

ptitzler:
> how would this value change when submitted via KFP?

See above screen capture. I uploaded an exported pipeline named `output_dir_test_2` using the KF central dashboard and ran it (using a one-off run). The run name `Run of ...` was generated by KFP and can be customized before submission. For repeating runs it generates names on the fly.

ptitzler:
Scope of this PR has changed to use the run _name_ value, which has more meaning to the user.

kevin-bates:
This looks good @ptitzler.   

Do you mind if I extend the local pipeline test framework to include validation of `ELYRA_RUN_NAME` before we merge?  I'll post a commit or if you'd like to do that, feel free.  I'm thinking we'd extend the framework to post a node's ERN env value to an output file (currently, I think the output files are empty)  that is specified as input to a subsequent node.  Nodes with inputs would validate the file's content with the env, while nodes w/o inputs would only ensure the prefixed (pipeline name) value matches the name.

ptitzler:
> Do you mind if I extend the local pipeline test framework

@kevin-bates of course not. whatever improves the code quality is very welcome!

ptitzler:
One question about a potential usability improvement for the Airflow value:

The current approach uses the DAG run id string, which contains formatting characters `-`, `:`, and `+`: 
```
give_run_name-0617150717-scheduled__2021-06-16T00:00:00+00:00
```

While the formatting makes the value easier to read, it might cause issues if one would try to use it as is. It'd be possible to have those removed, which would yield something like 
```
give_run_name-0617162405-20210618T130606
```

The only caveat I can think of is that by using this formatting the value would loose the information how the DAG was triggered (`scheduled` vs `manual`). Unfortunately I have not been able to obtain this piece of information from the metadata.

elyra/pipeline/processor_airflow.py
```diff
@@ -169,6 +169,14 @@ def _cc_pipeline(self, pipeline, pipeline_name):
                                                    cos_username=cos_username,
                                                    cos_password=cos_password)
 
+                # Generate unique ELYRA_RUN_NAME value and expose it as an
+                # environment variable in the container
+                # https://airflow.apache.org/docs/apache-airflow/1.10.12/_api/airflow/contrib/operators/kubernetes_pod_operator/index.html
+                # https://airflow.apache.org/docs/apache-airflow/1.10.12/macros-ref#default-variables
```
kevin-bates, line 175: I see `run_id` referenced in the second link, but I'm not able to determine what the first link is about.  Should we add some kind of description for how these links are helpful?

kevin-bates:
> While the formatting makes the value easier to read, it might cause issues if one would try to use it as is. It'd be possible to have those removed

I tend to agree that removing potentially problematic characters for things like filenames, etc, makes sense.  I'm curious how/where the reformatting would be applied though.  Since your approach also removes the 'scheduled' or 'manual' string it sounds like it's less flexible than, say, removing only `':'` and `'+'` characters (which I think is the minimally required set of characters we'd need to remove).

For my curiosity, where would this additional filtering occur?  And why couldn't the embedded "submission method" be retained?

(Ah, are you suggesting the submission method string be removed because it would perhaps interfere with the usability of whatever thing were to get created using this value?  I.e., it really has no meaning to anything other than the runtime.)

ptitzler:
> For my curiosity, where would this additional filtering occur? And why couldn't the embedded "submission method" be retained?

In https://github.com/elyra-ai/elyra/blob/c9b351c3a92af6eef20e1a70105a6e6db7783c5e/elyra/pipeline/processor_airflow.py#L178 we would specify a different macro variable [`ts_nodash`](https://airflow.apache.org/docs/apache-airflow/1.10.12/macros-ref#default-variables) that is replaced by Airflow at runtime. That variable does not contain the submission information and none of the other variables I checked seems to either.

> (Ah, are you suggesting the submission method string be removed because it would perhaps interfere with the usability of whatever thing were to get created using this value? I.e., it really has no meaning to anything other than the runtime.)

No, it's just a side effect of using a different variable. If we could retain it I would prefer to.

kevin-bates:
I see - thanks.  I agree that it would be useful to have.  If retained, it would be up to users that want to use it for other things to  scrub out the invalid chars appropriately - which seems like a bit of a hassle if we suspect this is a primary use-case for this.  So I'm almost inclined to go with the cleaner macro if that will satisfy this particular use-case.  (I bet there will be plenty of other clues present as to how the pipeline was submitted.)

elyra/pipeline/processor_airflow.py
ptitzler: Done

ptitzler:
I changed the variable for Airflow to `ts_nodash` and updated the information in the initial comment that describes the PR behavior.

ptitzler:
Looking at one more thing related to an issue with Tekton support. The `$(context.pipelineRun.name)` is equivalent to Argo's `{{ workflow.name }}`, which is different from `{{workflow.annotations.pipelines.kubeflow.org/run_name}}`.

kevin-bates (approved):
The changes look good.  I have not had a chance to run with these changes against KFP or Airflow, but have tested locally.  If someone needs me to run against the runtimes, it will be a day or two.

ptitzler:
I've added a fix to https://github.com/elyra-ai/kfp-notebook/pull/102 that resolves the Tekton issue I've discovered. The two workflow engines behave differently and we therefore have to take different approaches to collect the information we wanted to expose. I've tested in three KF environments (kfp/argo, kfp/argo, and kfp/tekton) and am now getting the expected results.

akchinSTC:
@marthacryan - mind checking out the CI test failures? thx

ptitzler:
The regression test failure is caused by the fact that the required `kfp-notebook` version is not yet published. (See initial comment indicating that this PR is requiring another PR) Therefore the following error is raised

```
__init__() got an unexpected keyword argument 'workflow_engine'
```

akchinSTC (approved):

PR #1744, opened 2021-06-10
Made code snippet tag text truncate if it's too long

### LINKED ISSUE #1395: Long tags in code snippet search overflows the button element
**Describe the issue**
Long tags in code snippet search overflows the button element when the side pane is narrower than the text

image

**To Reproduce**
1. Open the code snippet widget (click code snippets button on the side bar)
2. Either edit an existing snippet or create a new one
3. Add a tag with a long name and press enter to enable it
4. Save and close
5. Go back to the code snippet widget (side panel)
6. Click on the tag button to filter for tags
7. Check the long tag added 
8. Play with the side pane width (the problem happens on narrow states)

**Expected behavior**
The tag name should be contained in the button element, independent of the widget width.
If the full text can't fit, we could trim the end such as
`thisisanexampleofaverylongtag` --> `thisisanexample...`

**Environment information**
- Elyra version: 2.1.0rc2
- Operating system: MacOS
- Browser: Chrome

jdcai:

If the code snippets tag text is too long then the text will be truncated so that it will fit within the code snippet window.
I also added a title attribute to the tag so that if it's truncated you can hover over the tag and see the full use text.
There was also an issue where if a tag starts with a space " test" a checkmark would not show up next to it when it's selected which was fixed by using textContent instead of innerHTML.

Before:
image
image

After:
image
image
image

This was tested manually by adding tags to the code snippets window and selecting/unselecting them.
Useful tags for testing:
" wwwwwww wwwww wwwwwww wwwwww wwwwwww wwwww"
"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"
" test"
 
Fixes #1395

           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/FilterTools.tsx
+++ b/packages/metadata-common/src/FilterTools.tsx
@@ -33,6 +33,7 @@ interface IFilterMetadataState {
 const FILTER_OPTION = 'elyra-filter-option';
 const FILTER_TAGS = 'elyra-filter-tags';
 const FILTER_TAG = 'elyra-filter-tag';
+const FILTER_TAG_LABEL = 'elyra-filter-tag-label';
 const FILTER_CHECK = 'elyra-filter-check';
 const FILTER_TOOLS = 'elyra-filterTools';
 const FILTER_SEARCHBAR = 'elyra-searchbar';
@@ -112,9 +113,10 @@ export class FilterTools extends React.Component<
         className={`${FILTER_TAG} tag applied-tag`}
         id={'filter' + '-' + tag + '-' + index}
         key={'filter' + '-' + tag + '-' + index}
+        title={tag}
         onClick={this.handleClick}
       >
-        {tag}
+        <span className={FILTER_TAG_LABEL}>{tag}</span>
         <checkIcon.react
           className={FILTER_CHECK}
           tag="span"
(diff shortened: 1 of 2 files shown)


packages/metadata-common/src/FilterTools.tsx
```diff
@@ -134,16 +136,17 @@ export class FilterTools extends React.Component<
...
-        {tag}
+        <span className={FILTER_TAG_LABEL}>{tag}</span>
       </button>
     );
   }
 
   handleClick(event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void {
     const target = event.target as HTMLElement;
-    const clickedTag = target.innerText.trim();
+    const clickedTag = target.textContent ?? target.innerHTML;
```
karlaspuldaro, line 149: Maybe using just `target.textContent` would solve the issue here, I don't see a case it would fall into `target.innerHTML`

Now that you pointed out to the whitespace bug I think we should clean whitespaces from both ends of a text when user registers a new tag. This can be a separate issue.

jdcai: For textContent it can be null so I had to return a non null value in the case it is null. I think I could also do "" instead of target.innerHTML.

karlaspuldaro: Ok, yeah maybe empty string as a fallback might be better then, otherwise `innerHTML` will always resolve to the new span element
`<span class=\"elyra-filter-tag-label\"> whitespace</span>`

jdcai: okay I changed it to empty string

karlaspuldaro (approved):
Looks great! Thank you @jdcai

lresende (approved):

PR #1654, opened 2021-05-10
Add support for updating env vars

ajbozarth:
PR paired with https://github.com/elyra-ai/pipeline-editor/pull/116

### What changes were proposed in this pull request?
Add support for refreshing the env vars detected in a notebook or script in the properties for a node

The callback function is designed to support all future properties updates, not just env vars
 
Also includes Makefile  and doc updates for supporting parallel development of pipeline-editor

Fixes #710

Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #1631, opened 2021-04-29
Update UI of pipeline editor

marthacryan:
Fixes #1628.

New appearance of the runtimes dropdown:
image
image

Changes to using the JL variable for content font size, instead of code font size.
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #1779, opened 2021-06-16
Remove dead code

bourdakos1:
Fixes #1741

Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #1790, opened 2021-06-17
Update server-side properties with recent changes

ajbozarth:
While #1620 was in developement there was a couple changes made to
the propeties files on the front end that were missed when moving
the content to the backend (#1631 and #1654).

This adds back those missing changes.

This brings back the refresh env vars button that was lost.

Tested locally

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/resources/properties.json
+++ b/elyra/pipeline/resources/properties.json
@@ -67,7 +67,7 @@
           "default": "Filename"
         },
         "description": {
-          "default": "The path to the notebook file.",
+          "default": "The path to the file.",
           "placement": "on_panel"
         },
         "data": {
@@ -180,7 +180,8 @@
           "placement": "on_panel"
         },
         "data": {
-          "placeholder": "env_var=VALUE"
+          "placeholder": "env_var=VALUE",
+          "canRefresh": true
         }
       },
       {
--- a/elyra/pipeline/resources/properties.json
+++ b/elyra/pipeline/resources/properties.json
@@ -67,7 +67,7 @@
           "default": "Filename"
         },
         "description": {
-          "default": "The path to the file.",
+          "default": "The path to the notebook file.",
           "placement": "on_panel"
         },
         "data": {


ajbozarth:
Found this while testing #1779

elyra/pipeline/resources/properties.json
```diff
@@ -67,7 +67,7 @@
           "default": "Filename"
         },
         "description": {
-          "default": "The path to the notebook file.",
+          "default": "The path to the file.",
```
bourdakos1, line 70: I think this is dynamically handled in the server code

ajbozarth: I saw that, so I figured this a default/fallback value, so it should be generic. I changed this before finding that dynamic handling and just didn't change it back for that reason.

bourdakos1 (approved):

kiersten-stokes (commented):
Quick question -- when I remove an env var from a file, save it, and then go back to the properties to hit refresh, the removed value persists. Is this the functionality that we ultimately decided on? I vaguely remember a conversation about this. Happens with all file types.

Not sure if this is worth nitpicking right now though. Refresh works exactly as expected in the rest of the scenarios that I tested.

ajbozarth:
> Is this the functionality that we ultimately decided on?

This is the intended functionality, there was discussion about pruning unset vars that are no longer in the file, but that kind of validation was decided to be dealt with later alongside the new properties service we were planning

kiersten-stokes (approved):
LGTM!

lresende (approved):

lresende (changes requested):
@ajbozarth Please check test failures

ajbozarth:
I've pushed a fix for the test failure, though it makes me think the test is written poorly rather than properly failing. It checks that the api returns the same content as the json file on the backend. But the backend code always edits that content before returning. This test only currently passes because the content specific to the notebook endpoint is the default values in the file. I broke it by changing a default value to be generic. @lresende you wrote the test, was this intentional or should the test be updated? If so we should do that in a separate issue anyway.

lresende:
I did not look at what was failing? What test was it? Feel free to enhance tests at ay time :)

lresende (approved):

Issue #1784, opened 2021-06-17
Add pipeline json compatibility test


lresende:
Add pipeline json compatibility test to validate backward compatibility with previous versions:

Create a model pipeline that:
- Have at least three nodes
- Uses environment variables
- Include local file that is used by the node impl
- Produce a file that is used on a subsequent node impl
- Uses Notebook, Python Script, R Script nodes
- Use node content from another sibling folder

Add test case that creates a similar pipeline using the current UI and make sure the generated json is the same.

Issue #1693, opened 2021-05-24
Elyra 2.2.x cannot process Pipelines that were created using 2.3.x


ptitzler:
**Describe the issue**
See below

**To Reproduce**
1. Install Elyra 2.2.4
2. Follow the instructions in https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html#verifying-runtime-configurations
3. Open the `validate_python` pipeline. (The pipeline was created using a later release!)
4. See validation errors

**Screenshots or log output**

![image](https://user-images.githubusercontent.com/13068832/119413082-d25d0480-bca1-11eb-8126-f412fbedbedc.png)

![image](https://user-images.githubusercontent.com/13068832/119413170-f7517780-bca1-11eb-8ee9-b7477597f34a.png)

Note that the same pipeline "passes" 2.3.x validation.

**Expected behavior**
No validation errors should be displayed

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 2.2.4

bourdakos1:
looks like the `cpu`/`gpu`/`memory` are strings so would fail the `value > 0` check

bourdakos1:
If we used to persist cpu/gpu/memory as empty strings at some point we should probably bump the pipeline version to 4 and add a migration that purges empty strings and converts any strings that are supposed to be numbers to numbers

bourdakos1:
Oh misunderstood this is going from the new version to the old version. It looks like the component definition incorrectly sets the default value to an empty string

ptitzler:
Yep, several options:
 - bump pipeline version (least desirable option because we'd have to maintain 2.2.x and 2.3.x example pipelines)
 - 2.3.x sets those properties to something that 2.2.x can handle (change in 2.3 required)
 - 2.2.5+ can handle 2.3 formatted inputs (change in 2.2.4 required)

lresende:
2.3.x should be backward compatible in this scope.

bourdakos1:
yea, this should be an easy fix in 2.3 to use a default that is backwards compatible (it is a bug how it is currently handled in 2.3)

bourdakos1:
This will also need to be fixed in https://github.com/elyra-ai/elyra/pull/1620 

FYI @marthacryan @kiersten-stokes

ptitzler:
[...]

PR #1792, opened 2021-06-17
Add snapshot testing for pipeline files

bourdakos1:
This PR adds two tests:
1. Check the JSON of a pipeline that had all the nodes deleted
2. Check the JSON of a pipeline that had one node added

The bulk of this PR adds a snapshotting mechanism to make these tests easier to write and update.

## Basic usage
```js
cy.readFile('example.pipeline').matchesSnapshot();
```

The first run of this in a test will generate a snapshot of the pipeline file if it does not exist. If it exists, it will verify that the file matches the stored snapshot. If the test cannot find a snapshot in CI, the test will fail instead of generating a new snapshot.

In order to update a snapshot, delete the snapshot and re-run the integration tests for a new one to be generated.  

## Highlights of changes to pipeline json

changes going from 2.2.4 -> master (complex example):

These are some of the diffs for the snapshot of the complex pipeline example

the elyra-example-anaconda runtime didn't exist so I swapped it for the only only anaconda option:
```diff
- "runtime_image": "elyra/elyra-example-anaconda-py3:latest",
+ "runtime_image": "continuumio/anaconda3:2020.07",
```

added the new pipeline app_data:
```diff
+  "properties": {
+    "name": "simple",
+    "runtime": "Generic",
+  },
```

updated notebook description (python script description hasn't changed):
```diff
-  "description": "Notebook file",
+  "description": "Notebook",
```

When the node doesn't have any errors, it no longer has a null `invalidNodeError` property:
```diff
- "invalidNodeError": null,
```

There is a new label property:
```diff
+ "label": "",
```
> **Note:** Labels are handled differently in the new `pipeline-editor` previously we were manually setting the UI label, but now generate that dynamically. This will cause old labels from 2.2.4 pipelines to be destroyed.

I also removed links, because they are too hard to simulate in cypress (and this is hopefully something canvas already tests for)

           in the file; or

           it.


lresende (changes requested):
- Please add the `Developer's Certificate of Origin 1.1` to your pr
- We should have the scenario described in the issue #1784 :
   - Have at least three nodes
   - Uses environment variables
   - Include local file that is used by the node impl
   - Produce a file that is used on a subsequent node impl
   - Uses Notebook, Python Script, R Script nodes
   - Use node content from another sibling folder

lresende:
Could we actually have the tests in `tests/integration` and maybe something like `pipeline-format.ts` or `pipeline-backward-compatibility`.

lresende:
Also, I am not sure what the snapshot mechanism does, from `The first run of this in a test will generate a snapshot of the pipeline file if it does not exist. If it exists, it will verify that the file matches the stored snapshot.` it seems to imply that the snapshot can be created dynamically, but what I really think we want is to have a set of pipeline examples `certified` to be valid (e.g. test/resources/valid_empty_pipeline.json, test/resources/valid_3_node_pipeline) and make sure that creating a new pipeline via the UI from the build env generates an equivalent pipeline, and by equivalent, I mean it's exactly the same except for unique fields like node coordinates, ids, etc.

lresende:
From the changes highlighted:
- The first two I believe are only additions
- The third one might be a bug? 

```
Node properties no longer include unset properties in pipeline json:

-  "include_subdirectories": false,
-  "runtime_image": "",
```

Are these fields that have a default value and weren't changed by the user? 
runtime_image should be there otherwise pipeline can't be saved, no?

bourdakos1:
@lresende 

> Could we actually have the tests in tests/integration and maybe something like pipeline-format.ts or pipeline-backward-compatibility.

The test cases are currently in `tests/integration/pipeline.ts`, but I can move them into a new file with one of the suggested file names, do you have a preference?

> Also, I am not sure what the snapshot mechanism does

I normally see snapshot testing in UI unit tests, this is the description of snapshots from jest: "Snapshot tests are a very useful tool whenever you want to make sure your UI does not change unexpectedly."

I think we can reuse this concept for the pipeline json, because I think we basically want to make sure it doesn't change unexpectedly, right?

The main idea is that we would write an integration test that creates a pipeline, and then call `cy.readFile('example.pipeline').matchesSnapshot()` as the assertion. The first time you run the test, it would generate the actual snapshot, this is the "certified" example and thats what you see in the `snapshots/example.snap`. This snapshot is never supposed to change and if it does, it is a test failure.

> The third one might be a bug?

Yea, I wasn't sure if this should be considered a bug or not. The field `include_subdirectories` has a default value of false, so it is omitted unless manually edited. `runtime_image` is unset with no default so it really should be null instead of an empty string I think. Omitting these values is still backwards compatible because 2.2.4 will automatically fill  `include_subdirectories` and  `runtime_image` with false and an empty string respectively.

bourdakos1:
notes to self

```
elyra/elyra-example-anaconda-py3:latest
```

bourdakos1:
@lresende this should be good to go once https://github.com/elyra-ai/pipeline-editor/pull/136 is released and we replace the codesandbox version

bourdakos1:
@marthacryan @ajbozarth can you take a closer look at my file request logic to make sure I'm not missing anything

Makefile
```diff
@@ -100,6 +100,13 @@ dev-link:
 	cd node_modules/@elyra/pipeline-editor && jupyter labextension link --no-build .
 	cd node_modules/@elyra/pipeline-services && jupyter labextension link --no-build .
 
+dev-unlink:
+	yarn unlink @elyra/pipeline-services
+	yarn unlink @elyra/pipeline-editor
+	jupyter labextension uninstall @elyra/pipeline-services
+	jupyter labextension uninstall @elyra/pipeline-editor
```
ajbozarth, line 107: this is handled in `uninstall`, is there a reason we need another command?

packages/pipeline-editor/package.json
```diff
@@ -37,8 +37,8 @@
   },
   "dependencies": {
     "@elyra/metadata-common": "^2.3.0-dev",
-    "@elyra/pipeline-editor": "^0.7.0",
-    "@elyra/pipeline-services": "^0.7.0",
+    "@elyra/pipeline-editor": "https://pkg.csb.dev/elyra-ai/pipeline-editor/commit/42fac9a8/@elyra/pipeline-editor",
+    "@elyra/pipeline-services": "https://pkg.csb.dev/elyra-ai/pipeline-editor/commit/42fac9a8/@elyra/pipeline-services",
```
ajbozarth, line 41: is there a reason you're using the snapshots instead of the dev build? (just curious)

ajbozarth (commented):
I'm still trying to wrap my head around what the changes to `onFileRequested` do and why, but figured I should comment the note's I have while I do

packages/pipeline-editor/package.json
bourdakos1: I was using the dev build when testing locally, but I just wanted to make sure everything still worked in ci with the bundled version

Makefile
bourdakos1: I think it's convenient to have a command to just unlink, but if others don't find it useful I can remove

marthacryan: I think that'd be useful

ajbozarth: fair enough I guess make sure to add `dev-unlink` to `.PHONY` since it's a new command

bourdakos1: oh yea I think we also need to add dev-link to .PHONY too

ajbozarth: thx, I had added it, but I must've been lost in some merge or rebase along the way

lresende:
@bourdakos1 Where is this comparing the built pipeline with the `consumer-producer.pipeline` that was built on the previous Elyra release?

lresende:
The complex.snapshot in this pr is not compatible with Elyra 2.2.4. If you compare the producer-consumer.pipeline and the snapshot you will see that the issue reported in #1693 is still there and the comparison is not picking it up.

lresende:
You can launch Elyra 2.2.4 via docker with:
```
docker run -it -p 8888:8888 -v ${HOME}/opensource/jupyter-notebooks:/home/jovyan/work -v ${HOME}/Library/Jupyter:/home/jovyan/.local/share/jupyter -w /home/jovyan/work elyra/elyra:2.2.4 jupyter lab --debug
```

bourdakos1:
@lresende it did pick up that change as noted above, but I didn’t notice the bug. I think this issue was fixed at one point, but must have been re-introduced when moving the component definitions back to the server. I think the fix for this is changing the cpu/gpu type to number instead of string. 

I can roll back the snapshot so the tests fail until fixed

lresende (approved):
Thanks @bourdakos1 , LGTM now.

Issue #1773, opened 2021-06-15
Pipeline JSON Format Change - Version Next


lresende:
**Describe the issue**
These are some of the changes we might consider when bumping the pipeline version:

primary_pipeline[app_data][source] -> primary_pipeline[app_data][filename] -> Consistency with node format
"Notebooks" component id -> "notebook" component id

bourdakos1:
For the next version bump I think we should move all the node properties from directly in app_data to a properties object inside of app_data.

from this:
```json
"app_data": {
  "filename": "load_data.ipynb",
  "runtime_image": "amancevice/pandas:1.0.3",
  "ui_data": {
    "label": "Load Data",
    "image": "data:image/xxx",
    "x_pos": 50,
    "y_pos": 100,
    "description": "Notebook file"
  }
}
```

to this:
```json
"app_data": {
  "properties": {
    "filename": "load_data.ipynb",
    "runtime_image": "amancevice/pandas:1.0.3"
  },
  "ui_data": {
    "label": "Load Data",
    "image": "data:image/xxx",
    "x_pos": 50,
    "y_pos": 100,
    "description": "Notebook file"
  }
}
```

ptitzler:
`description` should  be a child of `properties` because it's not a UI property.

bourdakos1:
@ptitzler label and description are required to be in ui_data by canvas. (This description is different than the user editable description)

ptitzler:
Ah yes, okay. I thought this was was the user-provided value. Never mind.

lresende:
@bourdakos1 suggestion is already being done as part of #1882

PR #1801, opened 2021-06-21
Refactor parsing to be owned by processors

lresende:

This refactoring applies some separation of concerns into the implementation of the support for runtime-specific components. It also makes the Processor sort of the factory for the related objects, giving it the ability to have different registries and component types per processor. 

image

 
           in the file; or

           it.


elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,156 @@
...
+
+        return components
+
+    def get_properties(self, processor_type, component_id):
+        """
+        Return the properties JSON for a given component.
+        """
+
+        # Process component catalog
+        component_catalog = self._read_component_catalog()
```
kevin-bates, line 112: Perhaps we should make `component_catalog` and singleton object: `ComponentCatalog` who's init is `_read_component_catalog()`.  This would save duplicate reads here and in the above method - `get_all_components()`.

`ComponentCatalog` would provide single component get() method as well as a coroutine to get all the components.

`ComponentCatalog.get_component(id: str)` would return a `Component` instance and raise a `ValueError` if no component was found - just as below.  The `Component` instance would then know its reader and could provide 'location', 'body' and 'properties' properties.

Looks like `ComponentRegistry` could be easily converted to provide all of this functionality.

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,157 @@
...
+        if reader._type == "filename":
+            component_location = os.path.join(os.path.dirname(__file__), component_location)
+
+        component_body = reader.read_component_definition(component_location)
+        properties = self._parser.parse_component_properties(component_body, component_location)
+        properties['current_parameters']['component_source_type'] = reader._type
+
+        return properties
+
+    def _read_component_catalog(self):
```
kevin-bates, line 135:
```suggestion
    def _read_component_catalog(self) -> Dict:
```

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,157 @@
...
+        Read a component catalog and return its component definitions.
+        """
+
+        with open(self._component_catalog_location, 'r') as catalog_file:
+            catalog_json = json.load(catalog_file)
+
+        if 'components' in catalog_json.keys():
+            return catalog_json['components']
+        else:
+            return list()
```
kevin-bates, line 146: This should be an instance of a dictionary.

kevin-bates:
I think more functionality should be in a component and that's the center of this stuff.  If components were more stateful, then subsequent requests would avoid additional (and expensive) fetch/parse operations.

elyra/pipeline/component.py
```diff
@@ -0,0 +1,233 @@
...
+                    {"id": "component_source_type", "type": "controls", "parameter_refs": ["component_source_type"]}
+                ]
+            }
+        ]
+    },
+    "resources": {}
+}
+
+
+class Property(object):
```
kevin-bates, line 109: Should this be named `ComponentProperty` or something other than just `Property`?  That seems a bit too generic.

elyra/pipeline/component.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+
+    id: str
+    name: str
+    description: str
+    runtime: str
+    properties: List[Property]
+
+    def __init__(self, id: str, name: str, description: str, runtime: Optional[str] = None, properties:
+                 List[Property] = None):
```
kevin-bates, line 132:
```suggestion
                 Optional[List[Property]] = None) -> None:
```

Not sure what our strategy is wrt typing.  There are a number of signatures (including newly added ones) that don't have typing so its not clear how strict we should be about it.  Please advise.

elyra/pipeline/component_registry.py
```diff
@@ -133,61 +128,41 @@ def get_component(self, processor_type, component_id):
...
-
-            palette_json.append(component_json)
-        return palette_json
+        # Load jinja2 template
+        loader = PackageLoader('elyra', 'templates/components')
+        template_env = Environment(loader=loader)
+        template = template_env.get_template('canvas_palette_template.jinja2')
+
+        return template.render(components=components,
+                               num_components=len(components))
```
kiersten-stokes, line 165: Remove `num_components` here and use `components|length` in the template instead. Similar for `num_properties` in the properties palette function

kiersten-stokes:
In order to test the Airflow component split, an extra operator has to be added to `airflow_component_catalog.json`.

I've been testing with this one:
```
"slack-operator": {
  "name": "Slack Operator",
  "location": {
    "url": "https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/slack_operator.py"
  }
}
```

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,271 @@
...
+        adjusted_id = self._parser.get_adjusted_component_id(component_id)
+        component_entry = self.get_component_catalog_entry(adjusted_id)
+
+        # Get appropriate reader to read component_id definition
+        reader = self._get_reader(component_entry)
+
+        component_location = component_entry['location'][reader._type]
+        # Adjust location to dislay full path
+        if reader._type == "filename":
+            component_location = os.path.join(os.path.dirname(__file__), component_location)
```
kevin-bates, line 157: Are we going to have all file-based component definitions within our installation area?

Also, since `component_location` is already a function of `reader._type` (3 lines above), can we perform this additional prefixing when `component_entry['location']['filename']` is getting initialized and avoid this adjustment here?

kiersten-stokes: I think long term we probably don’t want components stored here, so you’re right to question it. 

I like the idea of doing the prefixing in init since it doesn’t fit very well! Then we can remove the duplicate code and still have the full file path displayed when passing parameters to the front end.

kevin-bates: I believe by housing component definitions within the installation area, any additional definitions (added post-installation) will be removed when uninstalled.  Given these probably have no use elsewhere, that may not be a big deal, but it's a bit of a nasty side-effect that users should be aware of.  If this approach is retained, it might be better to ask users to contribute their definitions - making them part of our release (which also could be nasty). :smile:

I agree with just moving forward with this approach for now and perhaps see what the "involvement" is like.

kevin-bates:
Just curious, do KFP and Airflow publish component definitions schemas by any chance?  Seems like there has to be something like that.

kiersten-stokes:
Surprisingly, I haven't come across any schemas in researching for the runtime-component-support PRs. KFP and Airflow each have a page or two in their documentation that addresses components/operators and what the general structure of each file should be, but no mention of a schema specifically.

elyra/pipeline/processor_airflow.py
```diff
@@ -45,6 +49,29 @@ class AirflowPipelineProcessor(RuntimePipelineProcess):
     def type(self):
         return self._type
 
+    @property
+    def component_registry(self) -> str:
```
kevin-bates, line 53: How does this property differ from `Component.registry_location` (which is the correct name IMO).

I think it might be better if this property was object-valued and held the instance of the `ComponentRegistry` that is instantiated during construction below.

elyra/pipeline/processor_airflow.py
```diff
@@ -45,6 +49,29 @@ class AirflowPipelineProcessor(RuntimePipelineProcess):
...
+        super().__init__(root_dir, **kwargs)
+        # then sys.prefix, where installed files will reside (factory data)
+        self._component_registry_location =  \
+            os.path.join(jupyter_core.paths.ENV_JUPYTER_PATH[0],
+                         'components',
+                         f"{self._type}_component_catalog.json")
+
+        if not os.path.exists(self._component_registry_location):
+            raise FileNotFoundError(f'Invalid component_id catalog path {self._component_registry_location}'
+                                    f' for {self._type} processor')
```
kevin-bates, line 70: Rather than use a local variable to house the location, the location composition and its existence check could be done from the superclass constructor since `self._type` will have the appropriate value.  The same could be done for the parser provided the instance is created prior to `super.__init__()`, then you could have the creation of the `ComponentRegistry` instance done in the superclass constructor.

Regardless, the two "component" properties (if they will remain) should be moved to the superclass anyway since kfp does the exact same thing, relatively speaking.

elyra/pipeline/processor_kfp.py
```diff
@@ -52,6 +57,29 @@ class KfpPipelineProcessor(RuntimePipelineProcess):
...
+            os.path.join(jupyter_core.paths.ENV_JUPYTER_PATH[0],
+                         'components',
+                         f"{self._type}_component_catalog.json")
+
+        if not os.path.exists(self._component_registry_location):
+            raise FileNotFoundError(f'Invalid component_id catalog path {self._component_registry_location}'
+                                    f' for {self._type} processor')
+
+        self._component_parser = KfpComponentParser()
+        self._component_registry = CachedComponentRegistry(self.component_registry, self.component_parser)
```
kevin-bates, line 81: What is the plan for allowing users to not use registry caching?  Seems like this needs to be configurable somehow.

elyra/pipeline/tests/test_component_parser_airflow.py
```diff
@@ -0,0 +1,96 @@
...
+        'name': 'Test Operator',
+        'type': FilesystemComponentReader._type,
+        'location': _get_resource_path('airflow_test_operator.py'),
+        'adjusted_id': ''
+    }
+    component_entry = SimpleNamespace(**entry)
+
+    parser = AirflowComponentParser()
+    component = parser.parse(component_entry)[0]
+    properties = ComponentRegistry.to_canvas_properties(component)
```
kevin-bates, line 57: I would suggest having a fixture of sorts that specifies the component registry, one with caching enabled, the other not.  Then each test that references the fixture will perform two passes and both forms of registries are exercised.

elyra/pipeline/component.py
kevin-bates: Per the in-person review, we'll tackle these in general following the release when we add tooling to validate, etc.

elyra/pipeline/tests/test_component_parser_airflow.py
kevin-bates: Since there really isn't a plan to allow switching between `ComponentRegistry` and `CacheComponentRegistry` and since all tests will complete within the default (and currently not configurable) 60-second TTL, we should update all the test references to use `CachedComponentRegistry`.

elyra/pipeline/component.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+class FilesystemComponentReader(ComponentReader):
+    """
+    Read a component_id definition from the local filesystem
+    """
+    type = 'filesystem'
+
+    def read_component_definition(self, component_id: str, location: str) -> str:
+        component_location = os.path.join(os.path.dirname(__file__), location)
```
kevin-bates, line 246: This will need to be coordinated with the change pending in #1854 that adds appropriate prefixing logic to the `location` value.

elyra/pipeline/processor_airflow.py
```diff
@@ -216,6 +244,27 @@ def _cc_pipeline(self, pipeline, pipeline_name):
                                                           operation)
 
             else:
+                # Change value of variables according to their type. String variables must include
+                # quotation marks in order to render properly in the jinja template and dictionary
+                # values must be converted from strings.
+                component = self._component_registry.get_component(operation.classifier)
+                for property in component.properties:
```
kevin-bates, line 251: We should use a different name here other than `property` as it conflicts with the built-in string.

elyra/pipeline/processor.py
```diff
@@ -148,30 +157,60 @@ def to_json(self):
...
 
+    def get_component_properties(self, component_id):
+        """
+        Retrieve runtime-specific component_id details if component_id is not one of the generic set
+        """
+
+        if component_id not in ('notebooks', 'python-script', 'r-script'):
+            return self._component_registry.get_component(component_id=component_id)
+
+        return component_id
```
kevin-bates, line 212: This doesn't look correct.  Please add type hints.

elyra/pipeline/processor.py
```diff
@@ -148,30 +157,60 @@ def to_json(self):
...
+
         return components
 
+    def get_component_properties(self, component_id):
+        """
+        Retrieve runtime-specific component_id details if component_id is not one of the generic set
+        """
+
+        if component_id not in ('notebooks', 'python-script', 'r-script'):
+            return self._component_registry.get_component(component_id=component_id)
```
kevin-bates, line 210: Should this call/method be changed to `get_component_propertes()`?  I'm confused how `get_component()` returns a component's propertes.

elyra/pipeline/handlers.py
```diff
@@ -97,37 +100,38 @@ async def post(self, *args, **kwargs):
...
             raise web.HTTPError(400, f"Invalid processor name '{processor}'")
 
-        properties = ComponentRegistry().get_properties(processor, component_id)
-        json_msg = json.dumps(properties)
+        if not component_id:
+            raise web.HTTPError(400, f"Invalid component_id id '{component_id}'")
 
-        self.set_status(200)
+        component: List[Component] = \
+            await PipelineProcessorManager.instance().get_component_properties(processor, component_id)
```
kevin-bates, line 133: This implies a list of components is returned.  Yet, if I'm tracking things properly, a `Component` is returned.  Which I also find confusing (noted previously) since this is all about getting a component's properties.

elyra/pipeline/handlers.py
```diff
@@ -97,37 +100,38 @@ async def post(self, *args, **kwargs):
...
 
+        self.set_status(200)
         self.set_header("Content-Type", 'application/json')
-        self.finish(json_msg)
+        self.finish(ComponentRegistry.to_canvas_properties(component))
```
kevin-bates, line 137: `to_canvas_properties` takes a `Component`, while `component` is typed as a List.

kevin-bates (commented):
I think things are getting closer, but I still have comments, some of which require changes.

elyra/pipeline/processor.py
lresende: it's some magic, it's related to "generic components" versus "runtime components"... it indeed feels strange, but I there is some logic to handle these as special cases... let me think if I could think about something better here.

elyra/pipeline/processor_airflow.py
lresende: Renamed the different incarnation of the property to all use the same name. We will try to add something to abstract the registry catalog and the additional changes will come with that pr.

elyra/pipeline/processor.py
kevin-bates: Thanks for adding type hints.  This needs to change for the case of the generic component.  

I suggest adding a `get_generic_component(component_id: str) -> Component:` method to `ComponentRegistry` to mirror the list-valued method `get_genertic_components()` then modify the method to:
```python
        component: Component = None
        if component_id not in ('notebooks', 'python-script', 'r-script'):
            component = self._component_registry.get_component(component_id=component_id)
        else:
            component = ComponentRegistry.get_generic_component(component_id=component_id)

        return component
```

Also, what happens if the `component_id` is invalid/not found?  Should the `get` methods return `None` or raise an exception?

If they return `None`, then the return value type hint needs to change to `Optional[Component]`.

elyra/pipeline/processor_local.py
```diff
@@ -57,6 +60,17 @@ def __init__(self, root_dir, **kwargs):
     def type(self):
         return self._type
 
+    @property
+    def registry_location(self) -> str:
```
kevin-bates, line 64: This property and `component_parser` should be removed and these should be defined on `RuntimePipelineProcessor` since they don't apply to local processors.  Then the constructors for `AirflowPipelineProcessor` and `KfpPipelineProcessor` should be refactored with most all of their content pushed to the (new) `RuntimePipelineProcessor` constructor.

I'm happy to add that commit if you like.

kevin-bates (approved):

elyra/pipeline/component.py
```diff
@@ -0,0 +1,297 @@
...
+    """
+    Read a component_id definition from a url
+    """
+    type = 'url'
+
+    def read_component_definition(self, component_id: str, location: str) -> str:
+        res = requests.get(location)
+        if res.status_code != HTTPStatus.OK:
+            self.log.error(f'Invalid location for component: {component_id} -> {location}')
+            raise FileNotFoundError(f'Invalid location for component: {component_id} -> {location}')
```
ptitzler, line 272: We should expose the HTTP error + message to simplify troubleshooting.

elyra/pipeline/component.py
```diff
@@ -0,0 +1,297 @@
...
+
+class FilesystemComponentReader(ComponentReader):
+    """
+    Read a component_id definition from the local filesystem
+    """
+    type = 'filesystem'
+
+    def read_component_definition(self, component_id: str, location: str) -> str:
+        component_location = os.path.join(os.path.dirname(__file__), location)
+        if not os.path.exists(component_location):
```
ptitzler, line 254: This returns true for files _and_ directories. It'd be safer to use `os.path.isfile`

elyra/pipeline/component.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        Find the proper reader based on the given registry component_id.
+        """
+        if not component_entry:
+            raise ValueError("Invalid null component_id")
+
+        try:
+            return self._readers.get(component_entry.type)
+        except Exception:
+            raise ValueError(f'Unsupported registry type {component_entry.type}.')
```
ptitzler, line 297: Should this be `reader type`?

elyra/pipeline/component_parser_airflow.py
```diff
@@ -0,0 +1,192 @@
...
+                                        type="string",
+                                        value=location,
+                                        description="The path to the component_id specification file.",
+                                        control="readonly",
+                                        required=True),
+                      ComponentProperty(ref="component_source_type",
+                                        name="Component Source Type",
+                                        type="string",
+                                        value=source_type,
+                                        description="",
```
ptitzler, line 179: Should this not be empty?

elyra/pipeline/component.py
lresende: No, it defines the type of the catalog entry, it has no knowledge of there is a reader to read it.

elyra/pipeline/component_parser_airflow.py
```diff
@@ -0,0 +1,192 @@
...
+                                        control="readonly",
+                                        required=True)]
+        return properties
+
+    def _read_component_definition(self, registry_entry):
+        """
+
```
ptitzler, line 186: Should not be empty

elyra/pipeline/component.py
```diff
@@ -0,0 +1,297 @@
...
+                 extension: str = None):
+        """
+        :param id: Unique identifier for a component_id
+        :param name: The name of the component_id for display
+        :param description: The description of the component_id
+        :param runtime: The runtime of the component_id (e.g. KFP or Airflow)
+        :param properties: The set of properties for the component_id
+        :type properties: List[ComponentProperty]
+        :param op: The operation name of the component_id; used by generic components in rendering the palette
+        """
```
ptitzler, line 184: Doc for `extension` parm is missing

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,214 @@
...
+
+class ComponentRegistry(LoggingConfigurable):
+    """
+    Component Registry, responsible to provide a list of available components
+    for each runtime. The registry uses component parser to read and parse each
+    component entry from the catalog and transform them into a component value object.
+    """
+
+    _generic_components: Dict[str, Component] = {
+        "notebooks": Component(id="notebooks",
```
ptitzler, line 34: Nit: key and id should be singular for consistency

elyra/pipeline/component.py
lresende: I believe we need this to support local directories as storage of filesystem components

elyra/pipeline/component_registry.py
lresende: Discussed this, but this would imply migration of old pipelines

lresende: Add here: https://github.com/elyra-ai/elyra/issues/1773

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,214 @@
...
+    def to_canvas_properties(component: Component) -> dict:
+        """
+        Converts registry components into appropriate canvas properties format
+        """
+        loader = PackageLoader('elyra', 'templates/components')
+        template_env = Environment(loader=loader)
+
+        # If component_id is one of the generic set, render with generic template,
+        # else render with the runtime-specific property template
+        if component.id in ('notebooks', 'python-script', 'r-script'):
```
ptitzler, line 122: Nit: not future proof. Should use key lookup  in `_generic_components` instead of hard-coding them here.

elyra/pipeline/handlers.py
```diff
@@ -97,37 +100,38 @@ async def post(self, *args, **kwargs):
...
-        self.log.info(f'Retrieving pipeline component properties for component {component_id}')
+        self.log.info(f'Retrieving pipeline component_id properties for component_id: {component_id}')
+
         if PipelineProcessorManager.instance().is_supported_runtime(processor) is False:
             raise web.HTTPError(400, f"Invalid processor name '{processor}'")
 
-        properties = ComponentRegistry().get_properties(processor, component_id)
-        json_msg = json.dumps(properties)
+        if not component_id:
+            raise web.HTTPError(400, f"Invalid component_id id '{component_id}'")
```
ptitzler, line 130: Should this be 404 - Not found?

elyra/pipeline/component_registry.py
lresende: we will enhance this in future prs

elyra/pipeline/processor.py
```diff
@@ -148,30 +158,48 @@ def to_json(self):
...
+            components.extend(self._component_registry.get_all_components())
+
         return components
 
+    def get_component(self, component_id: str) -> Component:
+        """
+        Retrieve runtime-specific component_id details if component_id is not one of the generic set
+        """
+
+        if component_id not in ('notebooks', 'python-script', 'r-script'):
```
ptitzler, line 198: Nit: not future proof. Should use key lookup  in `_generic_components` instead of hard-coding them here.

lresende: Will rework this in a future pr

elyra/pipeline/processor_airflow.py
```diff
@@ -216,6 +221,27 @@ def _cc_pipeline(self, pipeline, pipeline_name):
                                                           operation)
 
             else:
+                # Change value of variables according to their type. String variables must include
+                # quotation marks in order to render properly in the jinja template and dictionary
+                # values must be converted from strings.
+                component = self._component_registry.get_component(operation.classifier)
+                for component_property in component.properties:
+                    if component_property.ref in ['runtime_image', 'component_source', 'component_source_type']:
```
ptitzler, line 229: Nit: not future proof.

elyra/pipeline/processor_kfp.py
```diff
@@ -480,26 +486,48 @@ def _cc_pipeline(self,
...
             else:
                 component_source = {}
                 component_source[operation.component_source_type] = operation.component_source
 
-                # Build component task factory
+                # Change value of variables according to their type. Path variables should include
+                # the contents of the specified file and dictionary values must be converted from strings.
+                component = self._component_registry.get_component(operation.classifier)
+                for component_property in component.properties:
+                    if component_property.ref in ['runtime_image', 'component_source', 'component_source_type']:
```
ptitzler, line 498: See earlier comment.

elyra/pipeline/resources/palette.json
```diff
@@ -5,7 +5,7 @@
       "label": "Notebook",
       "image": "",
       "id": "notebooks",
-      "description": "Notebook",
+      "description": "Notebook file",
```
ptitzler, line 8: If this is the component description in the palette, it should probably read "Run Jupyter notebook" because a node implements a task, not a file of that type.

elyra/pipeline/resources/palette.json
```diff
@@ -53,7 +53,7 @@
       ]
     },
     {
-      "label": "Python Script",
+      "label": "Python",
       "image": "",
       "id": "python-script",
       "description": "Python Script",
```
ptitzler, line 59: If this is the component description in the palette, it should probably read "Run Python script" because a node implements a task, not a file of that type.

elyra/pipeline/resources/palette.json
```diff
@@ -104,7 +104,7 @@
       ]
     },
     {
-      "label": "R Script",
+      "label": "R",
       "image": "",
       "id": "r-script",
       "description": "R Script",
```
ptitzler, line 110: If this is the component description in the palette, it should probably read "Run R script" because a node implements a task, not a file of that type.

elyra/pipeline/resources/properties.json
```diff
@@ -67,7 +67,7 @@
           "default": "Filename"
         },
         "description": {
-          "default": "The path to the notebook file.",
+          "default": "The path to the Notebook file.",
```
ptitzler, line 70: Nit: it should be lower case because it's not a name, place, ...

elyra/pipeline/processor_airflow.py
lresende: we will refactor the generic components in a future pr

elyra/pipeline/processor_kfp.py
lresende: we will refactor the generic components in a future pr

elyra/pipeline/resources/palette.json
lresende: @kiersten-stokes why we change these? Could we leave it the way it is on master?

elyra/pipeline/resources/palette.json
lresende: @kiersten-stokes why we change these? Could we leave it the way it is on master?

elyra/pipeline/component_parser_airflow.py
```diff
@@ -0,0 +1,192 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def get_adjusted_component_id(self, component_id):
+        # Component ids are structure differently in Airflow to handle the case
+        # where there are multiple classes in one operator file. The id queried
+        # must be adjusted to match the id expected in the component_entry catalog.
+        return component_id.split('_')[0]
+
+    def parse(self, registry_entry: dict) -> List[Component]:
```
akchinSTC, line 36: is `dict` the correct type for registry_entry?

elyra/pipeline/component_parser_airflow.py
```diff
@@ -0,0 +1,192 @@
...
+        components: List[Component] = list()
+
+        component_definition = self._read_component_definition(registry_entry)
+
+        # Adjust filename for display on frontend
+        if registry_entry.type == "filename":
```
akchinSTC, line 42: or should  use registry_entry['type'] instead?

elyra/pipeline/component_parser_kfp.py
```diff
@@ -0,0 +1,164 @@
...
+class KfpComponentParser(ComponentParser):
+    _type = "kfp"
+
+    def __init__(self):
+        super().__init__()
+
+    def get_adjusted_component_id(self, component_id):
+        return component_id
+
+    def parse(self, registry_entry: dict) -> List[Component]:
```
akchinSTC, line 33: same in here ^^

elyra/pipeline/resources/palette.json
kiersten-stokes: I think the change was to force the names to match how the properties for the generic components were originally rendered. Specifically, the filename parameter, which has the below description:

`The path to the [filetype] file.`

I guess the best solution would be to just remove the word `file` from that description and we can change the labels back to how they were originally.

elyra/pipeline/resources/palette.json
kiersten-stokes: I opened #1859 to address this since I saw your comment on the pipeline editor PR. It can definitely just as easily be changed here, I just wasn't sure if it would have an unintended effects on the front end.

elyra/pipeline/component_registry.py
```diff
@@ -0,0 +1,214 @@
...
+            template = template_env.get_template('generic_properties_template.jinja2')
+            properties_json = template.render(component=component)
+        else:
+            template = template_env.get_template('canvas_properties_template.jinja2')
+            properties_json = template.render(properties=component.properties)
+
+        return properties_json
+
+    def _read_component_registry(self) -> Dict:
```
akchinSTC, line 131:
```suggestion
    def _read_component_registry(self) -> List:
```

elyra/pipeline/resources/palette.json
lresende: So, it is an issue in Master today, right? so maybe leave them as is in Master and use #1859 to make changes on places as needed?

elyra/pipeline/resources/palette.json
lresende: Same answer as above

elyra/pipeline/resources/properties.json
kiersten-stokes: Building on my comment above (re:changing the label `Python` back to `Python Script` and same for R), we can then easily change the case here by changing the generic_properties jinja template to render the lowercase version of the label instead. 

`{{ component.name|lower }}`

Issue #1797, opened 2021-06-18
Custom component properties can break Elyra


ptitzler:
**Describe the issue**
Reverse engineering the current pipeline file structure for typed pipelines, I've noticed that component properties appear to be not scoped in a safe way, which can result in corrupted pipeline files if a component property name is identical to a Elyra/reserved property name.

Excerpt from a file. Note that `app_data` stores reserved properties and component properties as siblings.

```
       {
          "id": "5d121620-6873-4aa1-a85f-2a864485e194",
          "type": "execution_node",
          "op": "filter-text-using-shell-and-grep-ptitzler",
          "app_data": {
            "component_source": "/opt/anaconda3/envs/components/lib/python3.7/site-packages/elyra/pipeline/resources/filter-text-local.yaml",
            "runtime_image": "quay.io/ptitzler/alpine",
            "component_source_type": "filename", # <------ reserved property
            "elyra_path_text": "",
            "pattern": "Apache",       # <------ component property
            "ui_data": {
```

Unless I am missing something we need to store the properties in a separate scope.

ptitzler:
Further testing surfaced that some sort of prefix is added to component properties if there are collisions. That approach can be exploited though and a component definition can therefore break Elyra:

![image](https://user-images.githubusercontent.com/13068832/122606953-a62d6d00-d02e-11eb-85a1-0198299673aa.png)

Here's a working example: [breaking-component.yaml.gz](https://github.com/elyra-ai/elyra/files/6678908/breaking-component.yaml.gz)

As is, I don't think the current component data structure in the pipeline file is robust enough and future proof.

kevin-bates:
I agree with @ptitzler's assessment.  I think introducing a scope/namespace within `app_data` would be sufficient.  I think the only _sibling-level_ attributes at the level would be `component_source` and `component_source_type`.  All other attributes would be embedded within a `component_properties` object_valued (dict) attribute.

Using the following two-node pipeline snippet as an example, what today looks like this...
```JSON
        {
          "id": "ca95fd8a-776b-43ae-a73e-e99906b80935",
          "type": "execution_node",
          "op": "serve-pytorch-model-seldon-core",
          "app_data": {
            "component_source": "https://raw.githubusercontent.com/kubeflow/pipelines/master/components/ibm-components/ffdl/serve/component.yaml",
            "runtime_image": "docker.io/aipipeline/ffdl-serve:latest",
            "component_source_type": "url",
            "model_id": "dddd",
[...]

PR #1828, opened 2021-06-23
Add ‘extensions’ key/value pair to palette.json for generic components

kiersten-stokes:
For the generic file-based components, an `extensions` key is added to the palette JSON. @marthacryan feel free to comment to add details and motivation

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/resources/palette.json
+++ b/elyra/pipeline/resources/palette.json
@@ -6,6 +6,7 @@
       "image": "",
       "id": "notebooks",
       "description": "Run notebook file",
+      "extensions": [".ipynb"],
       "node_types": [
         {
           "id": "",
--- a/elyra/pipeline/resources/palette.json
+++ b/elyra/pipeline/resources/palette.json
@@ -57,6 +57,7 @@
       "image": "",
       "id": "python-script",
       "description": "Run Python script",
+      "extensions": [".py"],
       "node_types": [
         {
           "id": "",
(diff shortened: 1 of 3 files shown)


kiersten-stokes:
After discussion with Martha, it looks like there may be more that needs to be done on the frontend here.

elyra/pipeline/resources/palette.json
```diff
@@ -6,6 +6,7 @@
       "image": "",
       "id": "notebooks",
```
kevin-bates, line 7: Unrelated to this PR, but does anyone know why the `id` for `notebooks` plural when all others are singular?  Even the description field is singular. Unless it has far-reaching implications to other things (presumably on the front-end), it seems like this should be updated for the sake of consistency.  I have no idea where this value is used.

kiersten-stokes: We definitely check against the `id` in the backend in a few places, but could easily change that. I don't think there are any additional implications for the backend

But yes, not sure if there is a frontend necessity or not. If not, could be good to rope into the changes to #1797

marthacryan:
Just pushed a fix for this - to test, try right-clicking on the canvas, and select "New Node from File". If the python, R, and notebook files are available in the dialog, the changes are working. Thanks!

kiersten-stokes:
This is working for me right now as-is! My only additional comment is that the way that this is done right now will change when #1801 is merged. It will be a fairly easy backend change again when the time comes

PR #1854, opened 2021-06-30
Update kfp component files to include optional parameter

akchinSTC:
- Update the kfp component definitions to align with best practices
and include descriptions and source license information
- Set directory as source when reading local custom components
- Remove runtime type from reader abstract method

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component.py
+++ b/elyra/pipeline/component.py
@@ -242,7 +242,7 @@ class ComponentReader(LoggingConfigurable):
         return self.type
 
     @abstractmethod
-    def read_component_definition(self, component_id: str, location: str) -> str:
+    def read_component_definition(self, component_id: str, location: str, runtime_type: str) -> str:
         raise NotImplementedError()
 
 
@@ -252,8 +252,9 @@ class FilesystemComponentReader(ComponentReader):
     """
     type = 'filename'
 
-    def read_component_definition(self, component_id: str, location: str) -> str:
-        component_location = os.path.join(os.path.dirname(__file__), location)
+    def read_component_definition(self, component_id: str, location: str, runtime_type: str) -> str:
+        component_resource_dir = "resources/" + runtime_type
+        component_location = os.path.join(os.path.dirname(__file__), component_resource_dir, location)
         if not os.path.exists(component_location):
             self.log.error(f'Invalid location for component: {component_id} -> {component_location}')
             raise FileNotFoundError(f'Invalid location for component: {component_id} -> {component_location}')
@@ -268,7 +269,7 @@ class UrlComponentReader(ComponentReader):
     """
     type = 'url'
 
-    def read_component_definition(self, component_id: str, location: str) -> str:
+    def read_component_definition(self, component_id: str, location: str, runtime_type: str) -> str:
         res = requests.get(location)
         if res.status_code != HTTPStatus.OK:
             self.log.error (f'Invalid location for component: {component_id} -> {location} (HTTP code {res.status_code})')  # noqa: E211 E501
--- a/elyra/pipeline/component.py
+++ b/elyra/pipeline/component.py
@@ -242,7 +242,7 @@ class ComponentReader(LoggingConfigurable):
         return self.type
 
     @abstractmethod
-    def read_component_definition(self, component_id: str, location: str, runtime_type: str) -> str:
+    def read_component_definition(self, registry_entry: dict) -> str:
         raise NotImplementedError()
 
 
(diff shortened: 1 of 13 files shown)


elyra/pipeline/resources/kfp/kubeflow_serve_model_using_kfserving/component.yaml
```diff
@@ -0,0 +1,43 @@
+name: Kubeflow - Serve Model using KFServing
```
ptitzler, line 1:
```suggestion
name: Kubeflow - Serve model using KFServing
```

elyra/pipeline/resources/kfp/filter_text_using_shell_and_grep/component.yaml
```diff
@@ -0,0 +1,25 @@
+name: Filter text using shell and grep
```
ptitzler, line 1:
```suggestion
name: Filter text 
description: Filter text using shell and grep
```

elyra/pipeline/registry.py
```diff
@@ -793,8 +793,8 @@ def _get_reader(self, component):
         Find the proper reader based on the given registry component.
         """
         try:
-            component_type = list(component['path'].keys())[0]
-            return self.readers.get(component_type)
+            reader_type = list(component['path'].keys())[0]
```
ptitzler, line 796: Since the change is not related to the PR title it should be listed in the PR description

ptitzler:
Could we add a comment at the beginning of each example component file:
- Since Elyra re-uses existing components that were published by somebody else, we should document the raw URL where the YAML file originally came from , e.g. 
  ```
  # Component source location: https://...
  ```
- Often there is not enough information in the specification for the user to make an informed decision about valid values. If one exists, we should  reference a web page (if applicable) that contains more details. For example, https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/kfserving/README.md
   ```
   # Component details: https://
   ```

etc/config/components/kfp_component_catalog.json
```diff
@@ -3,19 +3,19 @@
     "run-notebook-using-papermill": {
       "name": "Run Notebook Using Papermill",
       "path": {
-        "url": "https://raw.githubusercontent.com/kubeflow/pipelines/1.4.1/components/notebooks/Run_notebook_using_papermill/component.yaml"
+        "filename": "resources/kfp/run_notebook_using_papermill/component.yaml"
```
kevin-bates, line 6: I would suggest we move the `resources/kfp` portion of these into the server code that, today, uses `__file__` in #1801.  That way, _when_ we move these files out of our installed location, we only need to update the server code that concatenates to `__file__` and don't have to change both the server and the catalog file entries themselves.

lresende:
@akchinSTC Could you please update the license and list the origin + license of these component files

elyra/pipeline/resources/kfp/filter_text_using_shell_and_grep/component.yaml
```diff
@@ -0,0 +1,26 @@
+# Component source location: https://raw.githubusercontent.com/kubeflow/pipelines/master/components/sample/Shell_script/component.yaml
+name: Filter text
```
ptitzler, line 2: Please add a `description` (see earlier comment)

etc/config/components/kfp_component_catalog.json
ptitzler: I agree

elyra/pipeline/registry.py
```diff
@@ -768,7 +772,10 @@ def get_properties(self, processor_type, component_id):
 
             component_path = component['path'][reader._type]
             if reader._type == "filename":
-                component_path = os.path.join(os.path.dirname(__file__), component_path)
+                component_resource_dir = ""
+                if processor_type in ['kfp', 'airflow']:
+                    component_resource_dir = "resources/" + processor_type
```
kevin-bates, line 777: I'm confused between `processor_type` and `parser._type` and would prefer that we consolidate these if possible.  Since 'local' is not a formal processor type, I suspect it's an invariant condition that this method is ever called when `processor_type == 'local'` - is that a correct statement?

We should also try to avoid the `in` clause, otherwise we'll require code changes on each new processor.

This question seems like one fundamental to the overall design, so I'm not sure it can be addressed here.

Also, I believe this area of the code is being changed in #1801 (and eliminated), so it may be moot altogether.

kiersten-stokes: Yeah, you're correct that the way we implement this particular portion of the code will be changed once the refactoring is done. It will remove the need to use `in` and we'll put the onus on the component parser to determine it's own path to the local file, then just pass that path to the reader

akchinSTC: will wait for the refactoring to complete then to make the changes, thx

lresende:
@akchinSTC @kiersten-stokes Could you please resolve conflicts.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -13,15 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import os
-from typing import List
-
 import yaml
+from typing import List
```
kiersten-stokes, line 17: Per the linter's suggestion

```suggestion

from typing import List
```

kiersten-stokes (changes requested):
I think the linting issue below is the only outstanding thing for this PR (I say as if I didn't introduce the problem...)

elyra/pipeline/component.py
```diff
@@ -242,7 +242,7 @@ def type(self) -> str:
         return self.type
 
     @abstractmethod
-    def read_component_definition(self, component_id: str, location: str) -> str:
+    def read_component_definition(self, component_id: str, location: str, runtime_type: str) -> str:
```
lresende, line 245: The reader is agnostic of the runtime, so this parameter should not be added.

lresende (changes requested):
We need to make the registry aware of the additional location where components are located either by convention and/or by explicitly adding things on the registry JSON with the proper type and not let a reader calculate the location based on the type of the runtime, because then that reader looses its flexibility

elyra/pipeline/component.py
kiersten-stokes: Woops, looks like I kept this in accidentally. Pushing fix now

kiersten-stokes (approved):
LGTM! Tested an example for each runtime and each succeeded

lresende (approved):
I am not sure if for file-based directory of components we need to list one by one in the registry, but for this PR I guess it's ok to avoid delays, but we will need to revisit this in the future.

Issue #1783, opened 2021-06-17
Validate 2.2/2.3 pipeline handling


ptitzler:
Create a pipeline that:
- Have at least three nodes
- Uses environment variables
- Include local file that is used by the node impl
- Produce a file that is used on a subsequent node impl
- Uses Notebook, Python Script, R Script nodes
- Use node content from another sibling folder

Verify that the following scenarios are handled as expected:

### create pipeline using v2.2.4

- [x] VPE: pipeline opens in generic VPE in 2.3
- [ ] VPE: pipeline requires upgrade to v2.3 (pipeline is explicitly saved as generic pipeline)
- [x] VPE: pipeline runs as expected locally
- [x] VPE: pipeline runs as expected on KF
- [x] VPE: pipeline runs as expected on AA
- [x] VPE: pipeline exports and runs as expected on KF
- [x] VPE: pipeline exports and runs as expected on AA
- [ ] CLI: 2.2.4 pipeline cannot be run without prior upgrade (behavior up for discussion)
- [ ] CLI: 2.2.4 pipeline cannot be submitted without prior upgrade (behavior up for discussion)

### create generic pipeline using v2.3.x

- [ ] 2.2.4 VPE: pipeline does not open in VPE (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be run (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be submitted (unsupported version)

### create KF pipeline using v2.3.x

- [ ] 2.2.4 VPE: pipeline does not open in VPE (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be run (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be submitted (unsupported version)

### create AA pipeline using v2.3.x

- [ ] 2.2.4 VPE: pipeline does not open in VPE (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be run (unsupported version)
- [ ] 2.2.4 CLI: pipeline cannot be submitted (unsupported version)

karlaspuldaro:
In summary, the items without check mark didn't meet the expectations:
- A pipeline created in 2.2.4 can be accessed as generic in 2.3.0 editor, run/submitted from CLI, no upgrade request anywhere
- Any pipeline (generic, aa, kfp) created in 2.3.0 can be accessed in the 2.2.4 editor, in my tests, they all show as invalid for missing cpu/memory fields marked as required in previous version
[...]

PR #1860, opened 2021-07-02
Enable migration from previous pipeline versions

ajbozarth:
Migration was moved to pipeline-services but not hooked back into the editor, this reimplements that functionality

Required upstream PR in pipeline-editor: https://github.com/elyra-ai/pipeline-editor/pull/140

Fixes https://github.com/elyra-ai/elyra/issues/1783

 
           in the file; or

           it.


ajbozarth:
Some functionality is still broken:
- migrating from v0 to v1 throws an error in the console and fails to migrate
- migrate dialog will often open more than once (migration still works as intended, just a confusing UX)
- migration doesn't save and closing a migrated pipeline doesn't prompt for save on close

lresende:
You will need to also update https://github.com/elyra-ai/elyra/blob/30290ad11794e4a4bf205515596b03de4d741176/elyra/cli/pipeline_app.py#L34

ajbozarth:
Pushed fixes for:
- migrating from v0 to v1 throws an error in the console and fails to migrate
- migration doesn't save and closing a migrated pipeline doesn't prompt for save on close

Still need to address
- migrate dialog will often open more than once (migration still works as intended, just a confusing UX)

And the sister PR needs its tests updated.

Makefile
```diff
@@ -85,10 +85,10 @@ lint-server: test-dependencies
 	flake8 elyra
 
 prettier-ui:
-	yarn prettier:check
+	yarn prettier
 
 eslint-ui:
-	yarn eslint:check --max-warnings=0
+	yarn eslint
```
bourdakos1, line 91: is there a reason we are changing this?

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -332,15 +336,55 @@ const PipelineWrapper: React.FC<IProps> = ({
...
+          console.log('migrating pipeline');
+          setPipeline(
+            migrate(pipeline, pipeline =>
+              PipelineService.setNodePathsRelativeToPipeline(
+                pipeline,
+                contextRef.current.path
+              )
+            )
+          );
+          contextRef.current.model.dirty = true;
```
bourdakos1, line 370: You should avoid calling `setPipeline` directly. We have the react state of the pipeline connected to file changes so using `contextRef.current.model.fromString("<pipeline-string>")` will handle the dirty state and the react pipeline state

bourdakos1 (changes requested):

kevin-bates:
Heads up. The work taking place in #1882 will require some changes as well.  To summarize those changes, parameters specific to the component (node) will be pushed down a level (regardless of component type) into a stanza named (TBD) `component_parameters`.  Only system-level/non-component properties will reside as siblings to `component_parameters`.  

For those pipelines prior to 3.0, we will need to identify at least one of the following: properties that should be located in the new `component_parameters` stanza, or system-level properties that should **not** be located in the new stanza.  I personally suspect the latter set might be easier and more stable when taking into account possible hotfixes that could hypothetically be applied, thereby preventing updates to migration code.  But that's purely hypothetical.

kevin-bates:
Since all of the pipelines in https://github.com/elyra-ai/elyra/tree/master/elyra/pipeline/tests/resources/sample_pipelines will require updates to complete #1882, and those updates are exactly what should result from migration, it would be awesome if we could apply the migration to those prior to completion of #1882 and commit those updated files as part of the #1882 PR.  This would imply a slight window between this PR and #1882 in which a migrated pipeline would be unusable until #1882 is merged, but I believe that could be coordinated rather easily and we can benefit in a couple of ways - practical migration testing and time/stability savings of manual updates.

Makefile
ajbozarth: yeah, I'm unsure why this was changed, I would've not approved it myself, it removes functionality that was intentional. Changing this to check just makes lint fail with no help to the user. Prettier is meant to fix and the fix flag on eslint only make minor intentional changes, the rest are still just displayed, though on a re-read I should leave `--max-warnings=0` as that is useful to add. The use of fix here matches how jupyterlab lint functions

packages/pipeline-editor/src/PipelineEditorWidget.tsx
ajbozarth: gotcha, I'll switch to using that

Makefile
bourdakos1: Maybe we need to distinguish between what's run on ci vs locally? I don't think that eslint/prettier should stop a local build. On ci we want it to fail, if changes need to be made to the code

ajbozarth:
Based on Webex convo with @kevin-bates I will be removing the update to v4 from this PR and we will make that switch alongside writing the migration in  #1882 where the actual changes that need migration are added.

kevin-bates:
Related to this https://github.com/elyra-ai/elyra/pull/1860#issuecomment-877290570, @kiersten-stokes and I decided it would be best to manually update the server-based test pipeline files now (in this [commit](https://github.com/elyra-ai/elyra/pull/1882/commits/574311cebbabb1233682818c78c26d2b24968c8a)) since not having them updated was interfering with our ability to validate the actual changes for #1882.

Makefile
```diff
@@ -85,10 +85,10 @@ lint-server: test-dependencies
 	flake8 elyra
 
 prettier-ui:
-	yarn prettier:check
+	yarn prettier
 
 eslint-ui:
-	yarn eslint:check --max-warnings=0
+	yarn eslint --max-warnings=0
```
bourdakos1, line 91: can you rebase? this should be fixed in master now

bourdakos1 (approved):

Makefile
ajbozarth: done

marthacryan:
I'm using an old pipeline from the https://github.com/CODAIT/covid-notebooks repo and I'm seeing the dialog show up twice: 
Jul-13-2021 14-35-41

marthacryan:
Oh oops didn't see that that was a known issue ^^ I'm fine with that being a follow-up PR.

marthacryan (approved):
Tried locally and other than the issue with the dialog showing up more than once, this looks good

lresende (approved):

PR #1866, opened 2021-07-06
[HOTFIX] Fix integration tests after pipeline-editor palette changes

bourdakos1:

           in the file; or

           it.


marthacryan:
There's an actual bug with the run / export button due to the changes we made on the pipeline editor - the validation is looking for a field that has been re-arranged in the node definitions here: https://github.com/elyra-ai/pipeline-editor/blob/54b5a1e0c41d540b001b9145c328bae1186c1ed8/packages/pipeline-services/src/validation/index.ts#L62

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -82,6 +82,67 @@ export const commandIDs = {
   addFileToPipeline: 'pipeline-editor:add-node'
 };
 
+const createPalette = (categories: any[]): any => {
+  const palette = {
+    version: '3.0' as '3.0',
+    categories: categories ?? []
+  };
+
+  for (const category of categories) {
```
lresende: We shouldn't do this on the front-end, all info required should come from the services. @kiersten-stokes

kiersten-stokes: Yeah, if we need any changes to the palette that is being returned from the call to `/elyra/pipeline/components/[runtime]/[component-id]/properties`, I'm happy to make them!

bourdakos1: @lresende we have to piece the palette together at the moment because the `pipeline-editor` expects the palette to have all the properties embedded.

We originally did this because the pipeline needs all properties to validate. However, I want to change this to be a bit more async so that we can lazily fetch properties for validation

tests/integration/pipeline.ts
```diff
@@ -429,6 +429,7 @@ describe('Pipeline Editor tests', () => {
   it('kfp pipeline should display custom components', () => {
     cy.createPipeline({ type: 'kfp' });
     cy.openPalette();
+    cy.get('.palette-flyout-category[value="Kubeflow Pipelines Nodes"]');
```
karlaspuldaro, line 432: We need to add a similar selector in https://github.com/elyra-ai/elyra/blob/de163f3d1db186d4b4a45b8d60d6018aeb585140/tests/integration/pipeline.ts#L456

`cy.get('[value="Apache Airflow Nodes"]').click();`

lresende (approved):

PR #1882, opened 2021-07-08
Restructure pipeline JSON to prevent custom components from breaking

### LINKED ISSUE #1748: Pipeline version bump must be performed to prevent failures in legacy releases
**Describe the issue**
 Elyra 2.3rc1 produces pipeline files with version identifier 3.0, which is the same 2.x releases use. With the introduction of runtime specific pipeline files a version bump must be performed to prevent failures if a user opens a 2.3.x pipeline file in Elyra < 2.3.

```
{
  "doc_type": "pipeline",
  "version": "3.0",
```

### LINKED ISSUE #1773: Pipeline JSON Format Change - Version Next

### LINKED ISSUE #1797: Custom component properties can break Elyra

### LINKED ISSUE #1892: Support node labels for custom components
As a user I want to specify a custom label for nodes that are implemented using custom components. This functionality is already supported for nodes that are implemented using generic components.

image

image

- The node's properties view displays an editable `label` text fields.
- On the canvas the node's label is displayed instead of the component name if a label is defined.

image

### LINKED ISSUE #1907: Pipeline format potential issues
I think we have too much info duplicated across the pipeline file.

For example runtime: 
- `app_data.ui_data.runtime.display_name`
- `app_data.runtime`
- `app_data.properties.runtime`

I think that anything we keep in `ui_data` should safely be deleted and re-generated by the client. This isn't the case for runtime, if deleted it turns the pipeline into a generic pipeline.

both of the pipeline properties `name` and `runtime` seem redundant. 
- `name` is updated when the pipeline is renamed, but only when the pipeline is closed and then reopened. 
- `runtime` never changes and seems completely unused, the "runtime" in the properties editor is the runtime from `ui_data`

sample empty KFP pipeline for reference:
```json
{
  "doc_type": "pipeline",
  "version": "3.0",
  "json_schema": "http://api.dataplatform.ibm.com/schemas/common-pipeline/pipeline-flow/pipeline-flow-v3-schema.json",
  "id": "elyra-auto-generated-pipeline",
  "primary_pipeline": "primary",
  "pipelines": [
    {
      "id": "primary",
      "nodes": [],
      "app_data": {
        "ui_data": {
          "runtime": {
            "display_name": "Kubeflow Pipelines"
          },
          "comments": []
        },
        "version": 3,
        "runtime": "kfp",
        "properties": {
          "name": "untitled66",
          "runtime": "Kubeflow Pipelines"
        }
      },
      "runtime_ref": ""
    }
  ],
  "schemas": []
}
```

### LINKED ISSUE #1912: Airflow runtime components are showing runtime image as text field
**Describe the issue**
Airflow runtime components are showing runtime image as text field instead of allowing user to select from configured runtime images.

image

If we are showing the `runtime image` field, we should show it as a drop-down list as in the generic components.

### LINKED ISSUE #1932: Palette JSON format and dynamic categories support
`components` and `categories` both have the same base metadata:
- label
- image
- id
- description

`components` have some additional info:
- op
- parameter_refs
- extensions

Example palette:
```ts
{
  version: "3.0",
  categories: [
    {

      // category metadata
      label: "<CATEGORY_LABEL>",
      image: "<CATEGORY_IMAGE>",
      id: "<CATEGORY_ID>",
      description: "<CATEGORY_DESCRIPTION>",
   
      // List of all components for the group/category
      node_types: [
        {
          op: "<COMPONENT_OP>",
          description: "<COMPONENT_DESCRIPTION>",
          id: "<COMPONENT_ID>", // I don't think this is actually used for anything
          image: "<COMPONENT_IMAGE>",
          label: "<COMPONENT_LABEL>",
          type: "execution_node",
          inputs: [
            {
              id: "inPort",
              app_data: {
                ui_data: {
                  cardinality: {
                    min: 0,
                    max: -1,
                  },
                  label: "Input Port",
                },
              },
            },
          ],
          outputs: [
            {
              id: "outPort",
              app_data: {
                ui_data: {
                  cardinality: {
                    min: 0,
                    max: -1,
                  },
                  label: "Output Port",
                },
              },
            },
          ],
          parameters: {},
          app_data: {
            parameter_refs: <COMPONENT_PARAMETER_REFS>,
            extensions: <COMPONENT_EXTENSIONS>,
            image: "<COMPONENT_IMAGE>",
            ui_data: {
              description: "<COMPONENT_DESCRIPTION>",
              label: "<COMPONENT_LABEL>",
              image: "<COMPONENT_IMAGE>",
              x_pos: 0,
              y_pos: 0,
            },
          },
        },
      ],
    },
  ],
};
```

kiersten-stokes:
Fixes #1797
Fixes #1748 
Fixes #1773 
Fixes #1907
Fixes #1912
Fixes #1892 
Fixes #1932

This PR will restructure the format of the pipeline JSON and how it is parsed on the backend. Frontend changes and changes to the palette/properties JSON that is sent to the frontend may also require changes. Will definitely require changes to tests.

- [x] Change palette/properties JSON sent **from** backend
- [x] Change pipeline JSON sent **to** backend (frontend work)
    - [x] Adjust for new properties JSON format sent from backend component parsers (see [comment](https://github.com/elyra-ai/elyra/pull/1882#issuecomment-878438488) below)
    - [x] Adjust for new pipeline JSON format expected by backend processors (see [comment](https://github.com/elyra-ai/elyra/pull/1882#issuecomment-878433258) below)
- [x] Incorporate changes to fix #1773 
- [x] Change parsing of pipeline JSON in backend processors
- [x] Fix backend tests
- [x] Migration work
- [x] Manual testing

Removes the runtime-specific parameters `component_source_type` and `runtime image` from the UI, and changes the `component_source` parameter to a relative path if the component specification is filesystem-based. 

Changes the component `id` of notebooks from `notebooks` to `notebook` for consistency. 

Adds a `GenericOperation` subclass to the `Operation` class for generic components, which all have the same attributes/properties (`filename`, `runtime image`, `env_vars`, etc.).

Tests will have to be rewritten.

### Acceptance criteria

Please validate by migrating and properly running:
- https://github.com/CODAIT/covid-notebooks
- https://github.com/CODAIT/flight-delay-notebooks
- https://github.com/lresende/elyra-sample/tree/master/consumer-producer/pipelines
- https://github.com/elyra-ai/examples/tree/master/pipelines/dax_noaa_weather_data

           in the file; or

           it.


kiersten-stokes:
Based on some offline discussions, we're updating the pipeline JSON sent from the frontend to the backend in the following way. The first example is a generic component and the second example is a runtime-specific component.

```diff
{
"id": "4fcb7b56-7450-454f-9f4a-be454847c6cf",
"type": "execution_node",
"op": "execute-notebook-node",
"app_data": {
     "label": "",
-    "filename": "elyra/my_pipeline/node1.ipynb",
-    "runtime_image": "continuumio/anaconda3:2020.07",
-    "cpu": "",
-    "gpu": "",
-    "memory": "",
-    "outputs": [],
-    "env_vars": [], 
-    "dependencies": [],
-    "include_subdirectories": false
+    "component_parameters": {
+       "filename": "elyra/my_pipeline/node1.ipynb",
+       "runtime_image": "continuumio/anaconda3:2020.07",
+       "cpu": "",
+       "gpu": "",
+       "memory": "",
+       "outputs": [],
+       "env_vars": [], 
+       "dependencies": [],
+       "include_subdirectories": false
+    },
    "ui_data": {
              "label": "node1.ipynb",
              "description": "Notebook"
     }
},
{
"id": "ca95fd8a-776b-43ae-a73e-e99906b80935",
"type": "execution_node",
"op": "serve-pytorch-model-seldon-core",
"app_data": {
     "label": "",
-    "model_id": "dddd",
-    "deployment_name": "dddd",
-    "model_class_name": "ModelClass",
-    "model_class_file": "model_class.py",
-    "serving_image": "aipipeline/seldon-pytorch:0.1",
+    "component_parameters": {
+        "model_id": "dddd",
+        "deployment_name": "dddd",
+        "model_class_name": "ModelClass",
+        "model_class_file": "model_class.py",
+        "serving_image": "aipipeline/seldon-pytorch:0.1"
+    }
      "ui_data": {
              "label": "Serve PyTorch Model - Seldon Core",
              "description": "Serve PyTorch Models remotely as web service using Seldon Core"
       }
}
```

kiersten-stokes:
Additionally, the properties JSON sent from the backend to the frontend for display on the canvas is changing in the following way. The `parameter_ref` and the `group_info` `id` for every property besides the 'system-wide properties' (only `label` currently) will be prefixed with `elyra_`, e.g. `elyra_filename`, `elyra_serving_image`. 

This is to prevent collisions between our 'system-wide properties' and the properties given in the runtime-specific component. When displaying these properties and when sending them to the backend on pipeline submit/export, the `elyra_` prefix should be removed.

bourdakos1:
@ajbozarth when migrating the label we will need to do something like this:

```
if app_data.ui_data.label === app_data.filename
  app_data.label = ""

else
  app_data.label = app_data.ui_data.label
```

kevin-bates:
Hmm - Why not always default to the ui_data.label value?   Doesn't it default to app_data.filename today?

kiersten-stokes:
> Does this entry need an `"id":` key?

@kevin-bates I was trying to get that change committed before you noticed 😆

bourdakos1:
> Hmm - Why not always default to the ui_data.label value? Doesn't it default to app_data.filename today?

@kevin-bates `app_data.label` with only ever be what the user types into the textbox, so it will never be the filename (unless the user manually types in the filename)

`app_data.ui_data.label` should be considered as transient because it is part of `ui_data`. It will most likely always be there, but if all the `ui_data` objects are deleted out of the pipeline, the server should still be able to process it and the UI should be able to render it

The `app_data.ui_data.label` is set to `app_data.label ?? app_data.filename ?? component_name` on every render cycle so that it shows the proper label on the node. The server should use similar logic when a resolved label is needed for the node

bourdakos1:
I think that `component_parameters` should be treated as a black box, so we shouldn't have any hardcoded references to things inside of it. If we adhere to that, I think that we would need to move `filename` out of `component_parameters` onto the same level as `label`. Thoughts?

ptitzler:
> move filename out of component_parameters onto the same level as label

I don't think we can because conceptually the file name is a component parameter that identifies the file that this component is processing.

kevin-bates:
I believe `filename` needs to remain in `component_parameters` since it only applies to _generic components_ and things outside of `component_parameters` span all component "types".

elyra/pipeline/tests/test_component_parser_airflow.py
```diff
@@ -58,7 +58,7 @@ def test_parse_airflow_component_file():
 
     # Ensure component parameters are prefixed (and system parameters are not), and hold correct values
     assert properties_json['current_parameters']['label'] == ''
-    assert properties_json['current_parameters']['component_source'] == entry.location
+    assert properties_json['current_parameters']['component_source'] == entry['location']
```
kevin-bates, line 61: I think this meant to be `component_entry.location` or scrap the simple namespace thing altogether.

elyra/pipeline/pipeline.py
```diff
@@ -16,47 +16,46 @@
...
 
 
 class Operation(object):
     """
-    Represents a single operation in a pipeline
+    Represents a single operation in a pipeline representing a third-party component
     """
 
-    standard_node_types = ["execute-notebook-node", "execute-python-node", "exeucute-r-node"]
+    generic_node_types = ["execute-notebook-node", "execute-python-node", "exeucute-r-node"]
```
akchinSTC, line 30:
```suggestion
    generic_node_types = ["execute-notebook-node", "execute-python-node", "execute-r-node"]
```
typo

bourdakos1:
> I believe filename needs to remain in component_parameters since it only applies to generic components and things outside of component_parameters span all component "types". That said, I completely agree that component_parameters should be treated as a black box.

Ah, yea I see what you're saying. I was considering things outside `component_parameters` as system properties that _could_ span all "types", but don't have to. I do understand they need to span some sort of grouping otherwise there wouldn't be a way to determine which components get which property. Even `generic components` wouldn't be a good grouping because we could have generic components that don't need a `filename` in the future.

It's just a bit messy because the UI has to dig into `component_parameters` to find `filename` and any other component could have a `filename` that could be anything (an array for example, even though thats unlikely)

I think we have the "file based nodes" identified through the `extensions` field? I pretty sure if the component has `extensions` defined it needs the `filename` property. Can we generate the filename field outside of `component_parameters` if the component has `extensions` defined?

elyra/pipeline/pipeline.py
kiersten-stokes: Thank you!

elyra/pipeline/tests/test_component_parser_airflow.py
kiersten-stokes: Ah completely missed this! Yeah I should have accessed `component_entry.location`, not sure why I didn't

bourdakos1:
On yesterday's call we discussed a proposal for a new "pointer" system for enabling richer features in Elyra without giving special treatment to our in-house components.

One feature we need is the ability to hydrate nodes from dragging and dropping a file onto the canvas. In order for this to work, we need to be able to set a node property on the pipeline json to hold the file path information once the node has been created. However, we want to be able to treat `component_parameters` as a black box so the UI only sees this when looking to set a filepath field:
```ts
{
  label: "",
  component_parameters: {
    ??????????
    ??????????
    ??????????
  }
}
```

Since, it could be slightly unsafe to blindly set one of these `component_parameters` one of the solutions we discussed was hoisting out a system owned field that we could reliably set:
```ts
{
  label: "",
  filename: "books/fun.ipynb",
  component_parameters: {
    ??????????
    ??????????
    ??????????
  }
}
```

However, we decided against this because we didn't want to start leaking parameters out of the black box. Additionally, this wouldn't be flexible enough to enable the discussed feature on components we don't own.

Before diving into the proposed solution, I will briefly outline `palette`, `properties` and `pipeline`.

## Palette
The `palette` describes the basic metadata for all components (name, op, icon, description, etc...). Along with an optional `extensions` field that holds a list of extensions that this component could be activated for.

Example:
```ts
{
  version: "3.0",
  categories: [
    {
      ...<category-info>...
      node_types: [
        {
          op: "execute-notebook-node",
          label: "Notebook",
          extensions: [".ipynb"],
          ...<other-node-info>...
        }
      ]
   }
}
```

## Properties
Each component has a `properties`. This is the CommonProperties schema. The `current_parameters` field describes all the available properties and what value they are initialized to when a node is created. All properties specific to the component are prefixed with `elyra_`. This `elyra_` prefix only lives in the CommonProperties schema and shouldn't show up anywhere else.

Example:
```ts
{
  current_parameters: {
    label: "",
    elyra_filename: "",
    elyra_runtime_image: "",
    elyra_cpu: null,
    elyra_gpu: null,
    ...<all-the-other-properties>...
  },
  ...<rest-of-common-properties-schema>...
}
```

## Pipeline
The `pipeline` stores all the initial properties from `current_parameters` when a node is created and updates them if a user makes edits through the properties editor. When these properties are stored in the pipeline, the `elyra_` prefix is stripped and all properties with said prefix and placed in a `component_parameters` object.

Example:
```ts
// app_data
{
  "label": "",
  "component_parameters": {
    "filename": "books/fun.ipynb",
    "runtime_image": "continuumio/anaconda3:2020.07",
    // cpu is null, so doesn't show up
    // gpu is null, so doesn't show up
    ...<all-the-other-properties>...
  }
}
```

## Solution
The solution we came up with only requires a minor modification to the `palette`. We will add a `filehandler_parameter_ref` field to the `palette` which will tell Elyra which field in `component_parameters` to use to store the file path information. Since `filehandler_parameter_ref` is not part of CommonProperties the parameter_ref will **NOT** include the `elyra_` prefix.

Here is an example showcasing how the `palette`, `properties` and `pipeline` would look if we wanted to add this file handling feature to the `run-notebook-using-papermill` component.

palette:
```diff
{
  version: "3.0",
  categories: [
    {
      ...<category-info>...
      node_types: [
        {
          op: "run-notebook-using-papermill",
          label: "Run notebook using papermill",
+          extensions: [".ipynb"],
+          filehandler_parameter_ref: "notebook",
          ...<other-node-info>...
        }
      ]
   }
}
```

properties:
```diff
// nothing changes
{
  current_parameters: {
    label: "",
    elyra_notebook: "",
    ...<all-the-other-properties>...
  },
  ...<rest-of-common-properties-schema>...
}
```

pipeline:
```ts
// nothing changes
{
  "label": "",
  "component_parameters": {
    "notebook": "books/fun.ipynb",
    ...<all-the-other-properties>...
  }
}
```

kevin-bates:
@bourdakos1 - thank you for the great explanation - it is extremely helpful.

> We will add a filehandler_parameter_ref field to the palette which will tell Elyra which field in component_parameters to use to store the file path information.

Is `filehandler_parameter_ref` conditionally present in the palette.json only when a) a file-based parameter exists on the component and b) that parameter can be associated with one or more extensions that are specified in the (already conditional) `extensions` property?

bourdakos1:
> Is filehandler_parameter_ref conditionally present in the palette.json only when a) a file-based parameter exists on the component and b) that parameter can be associated with one or more extensions that are specified in the (already conditional) extensions property?

I think it would normally only be there for case "b". It could technically be there for "a", but that wouldn't enable the drag and drop functionality

kevin-bates:
Cool.  I think making its presence conditional on both (file-based prop WITH extension association) is what we should do until there's some functionality that can happen w/o extension association.

ajbozarth:
I've just pushed the initial working front end support. It is dependent on https://github.com/elyra-ai/pipeline-editor/pull/147 and will require you have that branch checked out locally and run `make clean dev-link install` in each repo.

I have not yet looked at the front end test and would not be surprised if I broken them. I will be looking into updating the tests tomorrow.

elyra/pipeline/component.py
```diff
@@ -171,11 +171,13 @@ class Component(object):
     runtime: str
     op: str
     properties: List[ComponentParameter]
-    extension: str
+    extensions: str
```
kevin-bates, line 174: This should be typed as `Optional[List[str]]`

elyra/pipeline/component.py
```diff
@@ -202,7 +204,20 @@ def __init__(self, id: str, name: str, description: Optional[str], source_type:
...
+        # Error checking for extensions & filehandler_parameter_ref depends on the component source type
+        if self._source_type == "elyra":
+            if not extensions:
+                raise ValueError("Invalid generic component: Missing field 'extensions'.")
+            filehandler_parameter_ref = "filename"
+        else:
+            if extensions and not filehandler_parameter_ref:
+                raise ValueError("Invalid component: A component for which file extensions are \
+                                 supplied must include the parameter_ref of the first parameter \
+                                 that is associated with that extensinon.")
```
kevin-bates, line 217: This feels a little strict to me.  I think we'll always be passing the extensions for our 'elyra' components and, someday, new ones may not use extensions.  I also think a missing `filehandler_parameter_ref` is more of a warning when there are extensions specified than a fatal error.  The component should still function, just that drag and drop can't be supported.  However, even when specified, the custom component's drag/drop behavior will likely be superseded by the drag/drop behavior of generic components unless the specified extensions are different.   Therefore, I think reducing this if/else to a warning message if extensions are set and fh_p_r is not set is more appropriate.  Perhaps the warning could read something like:
 
 "Component '{op}' specifies extensions '{extensions}' but no 'filehandler_parameter_ref' value and cannot participate in drag and drop functionality as a result."

elyra/pipeline/component.py
```diff
@@ -240,8 +255,12 @@ def properties(self):
         return self._properties
 
     @property
-    def extension(self):
-        return self._extension
+    def extensions(self):
```
kevin-bates, line 258: Can we make a pass on this file and add type hints?  `-> Optional[List[str]]`

elyra/pipeline/component_registry.py
```diff
@@ -41,21 +41,21 @@ class ComponentRegistry(LoggingConfigurable):
                               op="execute-notebook-node",
                               source_type="elyra",
                               source="elyra",
-                              extension=".ipynb"),
+                              extensions=".ipynb"),
```
kevin-bates, line 44: Convert to list: `[".ipynb"]`

elyra/templates/components/canvas_palette_template.jinja2
```diff
@@ -10,8 +10,9 @@
     {% if component.runtime %}
       "runtime": "{{ component.runtime }}",
     {% endif %}
-    {% if component.extension %}
-      "extensions": ["{{ component.extension }}"],
+    {% if component.extensions %}
+      "extensions": ["{{ component.extensions }}"],
```
kevin-bates, line 14: Should probably double-check that because `component.extensions` is a list, this doesn't produce a nested list (e.g., `[['.ipynb']]`).  If so, it would be preferable if we could reduce this to `"{{ component.extensions }}"` rather than rebuild the list iteratively.  (Sorry, I'm not that familiar with jinja functionality.)

This seems like a good thing for a test assertion.

kiersten-stokes: I found a nice jinja keyword to make this work! Agreed in converting `extensions` to a list

bourdakos1:
Do we want to create a `parameter_refs` object to hold the parameter refs?

```ts
parameter_refs: {
  filehandler: "filename",
  something_else: "example"
}

// vs 

filehandler_parameter_ref: "filename",
something_else_parameter_ref: "example"
```

elyra/pipeline/component.py
kiersten-stokes: Great point!

kevin-bates:
> Do we want to create a parameter_refs object to hold the parameter refs?

This strikes me as a good idea that would save a migration down the road.  I'm good with making this look like:
```
parameter_refs: {
  filehandler: "filename"
}
```
where `something_else` could be additive (and not trigger a migration).

Thanks for raising this @bourdakos1!

elyra/pipeline/component.py
```diff
@@ -119,39 +110,39 @@ def __init__(self, ref: str, name: str, type: str, value: str, description: str,
...
 
     @property
-    def items(self):
+    def items(self) -> List[str]:
         return self._items
 
     @property
-    def required(self):
+    def required(self) -> bool:
         if self._required:
```
kevin-bates, line 146: we could simplify this to `return bool(self._required)` since that will also treat `None` values as False.

elyra/pipeline/component.py
```diff
@@ -205,63 +186,67 @@ def __init__(self, id: str, name: str, description: Optional[str], source_type:
...
         return self._id
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self._name
 
     @property
-    def description(self):
+    def description(self) -> str:
```
kevin-bates, line 209: Looks like description is optional, so this should be `-> Optional[str]`

elyra/templates/components/canvas_palette_template.jinja2
```diff
@@ -11,7 +11,7 @@
       "runtime": "{{ component.runtime }}",
     {% endif %}
     {% if component.extensions %}
-      "extensions": ["{{ component.extensions }}"],
+      "extensions": {{ component.extensions|tojson }},
```
kevin-bates, line 14: Nice!

kevin-bates (commented):
Awesome - thanks for the type hints!  Just had a couple of minor nits.

elyra/pipeline/component.py
kiersten-stokes: Not sure how I missed that!

ajbozarth:
I've pushed my initial work to support the front end tests. As filename is still broken at the moment I am unsure if I will need to do more work once that's in. I also included the fix for multiple migration dialogs showing up.

bourdakos1:
I pushed some changes to the frontend that fixes the issues we were having. The fix is toggle-able with a `HACK` variable. Once the backend changes are ready we can test by setting `const HACK = false` and if it all works we can delete the fix code

elyra/pipeline/component.py
```diff
@@ -153,7 +153,7 @@ class Component(object):
 
     def __init__(self, id: str, name: str, description: Optional[str], source_type: str,
                  source: str, runtime: Optional[str] = None, op: Optional[str] = None,
-                 properties: Optional[List[ComponentParameter]] = None,
+                 category: str = None, properties: Optional[List[ComponentParameter]] = None,
```
kevin-bates, line 156:
```suggestion
                 category: Optional[str] = None, properties: Optional[List[ComponentParameter]] = None,
```
Should this parameter be named `category_id` or `category_name` or leave alone?  Not sure what the best answer is.    Since these also have a label, I think `category_id` would be the least confusing.

elyra/pipeline/component.py
```diff
@@ -230,6 +231,10 @@ def op(self) -> Optional[str]:
         else:
             return self._id
 
+    @property
+    def category(self) -> str:
```
kevin-bates, line 235:
```suggestion
    def category(self) -> Optional[str]:
```

elyra/pipeline/component_registry.py
```diff
@@ -41,21 +41,31 @@ class ComponentRegistry(LoggingConfigurable):
...
                               source="elyra",
-                              extensions=[".r"])}
+                              extensions=[".r"],
+                              category="generic")}
+
+    _generic_category: dict = SimpleNamespace(**({
+        "id": "generic",
+        "label": "Generic Components",
+        "image": "",
+        "description": "<GENERIC_CATEGORY_DESCRIPTION>"
```
kevin-bates, line 67: Is this just a placeholder or tied to some magic?

elyra/pipeline/component_registry.py
```diff
@@ -101,6 +111,23 @@ def get_component(self, component_id) -> Component:
             component = next(iter(component), None)
         return component
 
+    def get_categories(self) -> List[dict]:
+        category_entries: list = list()
+        with open(self._component_registry_location, 'r') as catalog_file:
```
kevin-bates, line 116: I feel like this should be baked into `_read_component_registry()` registry and let that return a tuple of components and categories.  I also wonder if the `_update_cache()` method would call `_read_component_registry()` directly, then perform what `get_all_xxx` was doing.  That way, we're only reading the once on each cache refresh (every 60 seconds!).  Probably not a big deal now, but when processing dozens (or hundreds) of components, it might be a good optimization.

elyra/pipeline/component_registry.py
```diff
@@ -202,13 +236,15 @@ class CachedComponentRegistry(ComponentRegistry):
 
     _cache: List[Component] = list()
```
kevin-bates, line 237: With the introduction of `_categories` I feel like `_cache` should be named something like `_cached_components` along side `_cached_categories`.

elyra/pipeline/component_registry.py
```diff
@@ -202,13 +236,15 @@ class CachedComponentRegistry(ComponentRegistry):
...
     _last_updated = None
+    _categories: list = list()
 
     def __init__(self, component_registry_location: str, parser: ComponentParser, cache_ttl_in_seconds: int = 60):
         super().__init__(component_registry_location, parser)
         self.cache_ttl_in_seconds = cache_ttl_in_seconds
 
         # Initialize the cache
         self.get_all_components()
+        self.get_categories()
```
kevin-bates, line 247: Should these two lines get replaced with `self._update_cache()`?

elyra/pipeline/component_registry.py
```diff
@@ -223,8 +259,15 @@ def get_component(self, component_id: str) -> Component:
         cached_component = next((component for component in self._cache if component.id == component_id), None)
```
kevin-bates, line 259: Looks like this returns `None` if there isn't a component_id match.  If so, please update the return type to `Optional[Component]`

elyra/pipeline/component_registry.py
```diff
@@ -223,8 +259,15 @@ def get_component(self, component_id: str) -> Component:
...
+            self._update_cache()
+
+        return self._categories
+
     def _update_cache(self):
         self._cache = super().get_all_components()
+        self._categories = super().get_categories()
         self._last_updated = time.time()
 
     def _is_cache_expired(self):
```
kevin-bates, line 273:
```suggestion
    def _is_cache_expired(self) -> bool:
```

elyra/pipeline/handlers.py
```diff
@@ -109,7 +109,9 @@ async def get(self, processor):
             raise web.HTTPError(400, f"Invalid processor name '{processor}'")
 
         components: List[Component] = await PipelineProcessorManager.instance().get_components(processor)
-        palette_json = ComponentRegistry.to_canvas_palette(components)
+        categories: list = await PipelineProcessorManager.instance().get_categories(processor)
```
kevin-bates, line 112: `List[Dict]`

elyra/pipeline/processor.py
```diff
@@ -209,6 +215,14 @@ def get_component(self, component_id: str) -> Component:
 
         return ComponentRegistry.get_generic_component(component_id)
 
+    def get_categories(self) -> list:
+        categories: list = [ComponentRegistry.get_generic_category()]
```
kevin-bates, line 219: these types should probably be `List[Dict]`

elyra/pipeline/component.py
kiersten-stokes: I agree on `category_id`!

elyra/pipeline/component_registry.py
kiersten-stokes: Per SCRUM, we'll update the descriptions in the catalogs to "Components that are supported by [runtime platform]"

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -468,12 +442,17 @@ const PipelineWrapper: React.FC<IProps> = ({
       const node = pipeline.pipelines[0].nodes.find(
         (node: any) => node.id === data.selectedObjectIds[i]
       );
-      if (!node || !node.app_data || !node.app_data.filename) {
+      if (
+        !node ||
+        !node.app_data ||
+        !node.app_data.component_parameters ||
+        !node.app_data.component_parameters.filename
```
bourdakos1, line 449:
```ts
!node?.app_data?.component_parameters?.filename
```

packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -482,47 +461,18 @@ const PipelineWrapper: React.FC<IProps> = ({
...
       }
-      if (node.app_data.memory === null) {
-        delete node.app_data.memory;
+      if (node.app_data.component_parameters.memory === null) {
+        delete node.app_data.component_parameters.memory;
       }
-      if (node.app_data.gpu === null) {
-        delete node.app_data.gpu;
+      if (node.app_data.component_parameters.gpu === null) {
+        delete node.app_data.component_parameters.gpu;
```
bourdakos1, line 471: should we enumerate all keys and delete any null entries?

bourdakos1 (changes requested):

bourdakos1:
One of the integration tests are failing, because of a missing "operation name" field:

Pipeline Editor tests -- should run pipeline after adding runtime image (failed) (attempt 2)

ajbozarth:
@bourdakos1 I'm still working on the tests, those are not merged yet

bourdakos1:
@ajbozarth  This is just broken in general, the integration tests are running fine and failing where expected. This is a real bug

kiersten-stokes:
> @ajbozarth This is just broken in general, the integration tests are running fine and failing where expected. This is a real bug

@bourdakos1 you're right just pushed a fix!

bourdakos1:
@kiersten-stokes oh is "operation name" supposed to be the label? 

If so, you should check `app_data.label` then `app_data.component_parameters[<filehandler-param-ref>]` if both of those are empty, it should fall back to the component name (e.g. Notebook)

kevin-bates:
> If so, you should check app_data.label then app_data.component_parameters[<filehandler-param-ref>] if both of those are empty, it should fall back to the component name (e.g. Notebook)

@kiersten-stokes and I just had a sidebar and have discovered that the pipeline being submitted in this particular test is invalid.  It's version is listed as '4', yet it still holds an `app_data.ui_data.label` value that has not been transferred to `app_data.label`.

```jSON
            "nodes": [
              {
                "id": "a6d8506d-2faf-46f1-8b9e-bee82895fc29",
                "type": "execution_node",
                "op": "execute-notebook-node",
                "app_data": {
                  "component_parameters": {
                    "filename": "examples/data 2.ipynb",
                    "runtime_image": "amancevice/pandas:1.1.1",
                    "outputs": [],
                    "env_vars": [],
                    "dependencies": [],
                    "include_subdirectories": false
                  },
                  "label": "",
                  "ui_data": {
                    "label": "data 2.ipynb",
                    "image": "data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2216%22%20viewBox%3D%220%200%2022%2022%22%3E%0A%20%20%3Cg%20class%3D%22jp-icon-warn0%20jp-icon-selectable%22%20fill%3D%22%23EF6C00%22%3E%0A%20%20%20%20%3Cpath%20d%3D%22M18.7%203.3v15.4H3.3V3.3h15.4m1.5-1.5H1.8v18.3h18.3l.1-18.3z%22%2F%3E%0A%20%20%20%20%3Cpath%20d%3D%22M16.5%2016.5l-5.4-4.3-5.6%204.3v-11h11z%22%2F%3E%0A%20%20%3C%2Fg%3E%0A%3C%2Fsvg%3E%0A",
                    "x_pos": 102,
                    "y_pos": 105.5,
                    "description": "Run notebook file"
                  }
                },
```
We also discussed having the server go and dereference the `parameter_ref.filehandler` pointer that you suggested, and that is not viable today with the current structure of the software.

Since the server must assume that all pipeline payloads are at the current version and all current versions have a value for `app_data.label`,  I think we need to take closer look at how this particular test pipeline was "migrated".  If that was accomplished manually, the pipeline should be fixed.  If migrated via the migration utility, then I believe there's more work involved.  

There's also the distinct possibility that we're simply caught in between of evolving software and may need to ignore this until all the dust has settled.

cc: @ajbozarth

ajbozarth:
`app_data.ui_data.label` is not being removed during migration, `app_data.label` is just being set to it. Since `ui_data` is transient I didn't both to clean up it's content during migration.

packages/pipeline-editor/src/PipelineEditorWidget.tsx
kevin-bates: Was there a reason why only the "resource-based" parameters were deleted before this change?  Making this general will potentially delete all component parameters and I'm not sure what side effect such a change may have when considering any component (generic or custom).

bourdakos1: I think only resource Params are potentially null at the moment, but also wanted to double check

elyra/pipeline/component_parser_airflow.py
```diff
@@ -144,40 +148,26 @@ def _parse_properties(self, registry_entry, component_definition, component_clas
             if match:
                 type = match.group(1).strip()
 
-            properties.append(ComponentProperty(ref=arg,
-                                                name=arg,
-                                                type=type,
-                                                value=default_value,
-                                                description=description,
-                                                control_id=control_id))
+            properties.append(ComponentParameter(ref=arg,
```
lresende, line 151: why are we calling this ref? and not id? I know this might be overriding things, etc but we should be consistent...

lresende: I have a local change to rename this back to id, but will not push before discussing it.

elyra/pipeline/component.py
```diff
@@ -171,12 +172,12 @@ class Component(object):
     runtime: str
     op: str
     properties: List[ComponentParameter]
-    extensions: str
+    extensions: Optional[List[str]]
```
lresende: Why is this becoming a list? Are we assuming that, if multiple extensions, the "running" aspects would still be same?

elyra/pipeline/component.py
```diff
@@ -194,41 +176,83 @@ def __init__(self, id: str, name: str, description: Optional[str], runtime: Opti
...
+    def op(self) -> Optional[str]:
         if self._op:
             return self._op
         else:
             return self._id
 
     @property
-    def extension(self):
-        return self._extension
+    def category_id(self) -> Optional[str]:
```
lresende, line 235: Is category really optional? Can we have a component without a category? Would the UI break in this case?

elyra/pipeline/component_registry.py
```diff
@@ -62,22 +79,22 @@ def __init__(self, component_registry_location: str, parser: ComponentParser):
     def registry_location(self) -> str:
         return self._component_registry_location
 
-    def get_all_components(self) -> List[Component]:
+    def get_all_components(self) -> Tuple[List[Component], List[Dict]]:
```
lresende, line 82: What is being returned here?  Is this really the best design to implement?

lresende (changes requested):
Left some comments, but want to understand the choice of design with the support for categories

lresende:
Overall functionality looks good, have tested some scenarios with CLI and fixed an issue there and have tested a few pipelines created in 2.2.4 and migrated and run locally and on kfp. I had a few questions above around some implementation details.

elyra/pipeline/component.py
kevin-bates: No idea.  This was listed as `Optional` because the constructor uses `None` as a default value.  The effect on the UI was probably not attempted.

elyra/pipeline/component.py
kevin-bates: IIRC, this was made a list since it (ultimately) corresponds to a list-valued attribute in what becomes the palette.json under `app_data.extensions:`.  I think the only assumption is that it enables support for a list of extensions, rather than one, and matches the frontend expectations (of a list).

> if multiple extensions, the "running" aspects would still be same?

 Since `Component` is really a client-side (pre-execution) thing used to support drag and drop, I actually don't think this comes into play.  `Component`'s runtime counter-part is the `Operation` (`GenericOperation`) - which doesn't reflect anything about the extension since it would reference the actual file itself.

elyra/pipeline/component_registry.py
kevin-bates: > Is this really the best design to implement?

This seems a little harsh - no?  Here's why this was changed to a tuple is due to the late-breaking discovery of a malformed palette that existed since custom components were introduced into master a few weeks ago: https://github.com/elyra-ai/elyra/pull/1882#discussion_r673228696  

However, I didn't fully convey how best to implement this workaround.  Now, after looking closer, the solution will be more involved due to the additional parsing that is required for components and the "adjusted_id" - that appears to only apply to airflow components. That said, since only the `CachedComponentRegistry` is used in this "design", the callers get the correct result despite the tuple.  I think what _should_ happen is the parser code  necessary only in `get_all_components()` but not `get_all_categories()` should be moved into the read method so that it returns "final results".  Then the base `ComponentRegistry` will just continue reading the file, once for each call to `get_all_components()` and `get_all_categories()`, while the `CachedComponentRegistry` will call the read method directly so that the file only gets read once (every 60 seconds).  What is causing trouble with that "workaround" is with the way `get_component()` works with its need to handle `adjusted_id`.  

I do have a question: do we also need to handle `adjusted_id` in `get_all_components()`?

This is worth further discussion, preferably via a virtual call.

elyra/pipeline/component.py
```diff
@@ -24,21 +25,11 @@
...
-    type: str
-    value: str
-    description: str
-    control: str
-    control_id: str
-    items: list
-    required: bool
-
     def __init__(self, ref: str, name: str, type: str, value: str, description: str, required: bool = False,
                  control: str = "custom", control_id: str = "StringControl", items: List[str] = []):
```
kevin-bates, line 34: The default value for `items` needs to be updated or removed.  Having a mutable object as a default value will result in side effects.

I'll post a commit to address this.

elyra/pipeline/component_registry.py
lresende: Sorry, I didn't mean to be harsh, but the API seemed a little strange, and the unpacking requirement... wanted basically to start a discussion and agree we could handle it on a call (maybe as an item on the scrum one).

elyra/pipeline/component.py
```diff
@@ -44,12 +44,12 @@ def __init__(self, ref: str, name: str, type: str, value: str, description: str,
...
 
-        if not ref:
-            raise ValueError("Invalid component: Missing field 'ref'.")
+        if not id:
+            raise ValueError("Invalid component: Missing field 'id'.")
         if not name:
             raise ValueError("Invalid component: Missing field 'name'.")
 
-        self._ref = ref
+        self._ref = id
```
kevin-bates, line 52: Shouldn't the attribute and property names be updated to reflect `id` rather than `ref`?

lresende (approved):

bourdakos1 (approved):

PR #1916, opened 2021-07-14
Set 'required' attribute default to True for KFP properties

### LINKED ISSUE #1896: KFP component specification parser interprets `optional` property incorrectly for inputs
**Describe the issue**

The code in https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/component_parser_kfp.py#L77-L80 

```
 # Determine whether parameter is optional
 required = False
 if "optional" in param and not param.get('optional'):
     required = True
```

does not properly evaluate the `optional` input property:

> `optional`: Specifies if input is optional or not. This is of type **Bool**, and defaults to **False**. Only valid for inputs.

Reference:

https://www.kubeflow.org/docs/components/pipelines/reference/component-spec/#interface

kiersten-stokes:
Fixes #1896. 

Changes the default value for KFP `ComponentProperty`'s `required` attribute to `True` during component parsing according to the following source: https://www.kubeflow.org/docs/components/pipelines/reference/component-spec/#interface

Adds new logic to an existing test to check against this case. Three new properties are added to the test component YAML, one of which is explicitly required, one of which is explicitly optional, and one of which does not specify and should therefore be set as required.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_parser_kfp.py
+++ b/elyra/pipeline/component_parser_kfp.py
@@ -70,10 +70,11 @@ class KfpComponentParser(ComponentParser):
             if "description" in param:
                 description = param.get('description')
 
-            # Determine whether parameter is optional
-            required = False
-            if "optional" in param and not param.get('optional'):
-                required = True
+            # KFP components default to being required unless otherwise stated.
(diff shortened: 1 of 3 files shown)


kevin-bates (approved):
Looks good - thanks.

lresende (approved):

PR #1918, opened 2021-07-14
Merge kfp-notebook to Elyra repo

### LINKED ISSUE #1132: Refactor kfp-notebook into elyra
Introducing the airflow notebook operator, also introduces a lot of duplicate code, since both the kfp-notebook and airflow-notebook operators use, for the most part, the same image configuration files and scripts. 
This issue is to cover the work regarding moving the kfp-notebook operator in https://github.com/elyra-ai/kfp-notebook into elyra.

- Extend the bootstrapper.py script base class to cover both kfp and airflow operations
- Move and merge existing unit and integration tests
- Separate airflow and kfp runtimes under elyra/pipeline
- Move configuration scripts and files and general housekeeping under etc/

lresende:
Fixes #1132 
 
           in the file; or

           it.


lresende:
Have submitted a few pipelines to KFP and they seem to be working end-to-end as before the merge.

elyra/kfp/operator.py
```diff
@@ -0,0 +1,358 @@
...
+                                                                     'elyra/{branch}/elyra/kfp/bootstrapper.py'.
+                                                                     format(org=ELYRA_GITHUB_ORG,
+                                                                            branch=ELYRA_GITHUB_BRANCH))
+ELYRA_REQUIREMENTS_URL = os.getenv('ELYRA_REQUIREMENTS_URL', 'https://raw.githubusercontent.com/{org}/'
+                                                             'elyra/{branch}/etc/kfp/requirements-elyra.txt'.
+                                                             format(org=ELYRA_GITHUB_ORG,
+                                                                    branch=ELYRA_GITHUB_BRANCH))
+
+
+class ExtendedContainerOp(ContainerOp):
```
kevin-bates, line 58: `ExtendedContainerOp` seems too obvious and not very informative.  How about `ElyraGenericOp` or `ElyraFileOp`.  Since, I suspect, nearly every operator extends `ContainerOp`, I don't think we need to preserve that aspect of things. `ElyraGenericOp` implies it's one of our generic operators, although, per the previous discussion, those may not always be file-based, as they are today - thus, the `ElyraFileOp`.

kevin-bates (approved):
If this gets rid of kfp-notebook and the associated release-coordination hassles, I'm all for it!  Was the reason we didn't do this initially because we didn't want to include elyra on the container image?

I just had the one comment about the name - `ExtendedContainerOp`, but am approving assuming that can be updated.

elyra/pipeline/tests/test_processor_airflow.py
```diff
@@ -185,7 +185,7 @@ def test_create_file(monkeypatch, processor, parsed_pipeline, parsed_ordered_dic
         for node in pipeline_json['pipelines'][0]['nodes']:
             for i in range(len(file_as_lines)):
                 # Matches an op with a node ID
-                if "notebook_op_" + node['id'].replace("-", "_") + " = NotebookOp(" in file_as_lines[i]:
+                if "notebook_op_" + node['id'].replace("-", "_") + " = ExtendedContainerOp(" in file_as_lines[i]:
```
akchinSTC, line 188:
```suggestion
                if "notebook_op_" + node['id'].replace("-", "_") + " = NotebookOp(" in file_as_lines[i]:
```

elyra/templates/airflow/airflow_template.jinja2
```diff
@@ -2,7 +2,7 @@ from airflow import DAG
 {% if cos_secret %}
 from airflow.kubernetes.secret import Secret
 {% endif %}
-from airflow_notebook.pipeline import NotebookOp
+from airflow_notebook.pipeline import ExtendedContainerOp
```
akchinSTC, line 5:
```suggestion
from airflow_notebook.pipeline import NotebookOp
```

elyra/templates/airflow/airflow_template.jinja2
```diff
@@ -45,7 +45,7 @@ op_{{ operation.id|regex_replace }} = {{ operation.name }}(namespace='{{ namespa
                                                                     {{ param }}={{ value }},
         {% endfor %}
     {% else %}
-op_{{ operation.id|regex_replace }} = NotebookOp(name='{{ operation.notebook|regex_replace }}',
+op_{{ operation.id|regex_replace }} = ExtendedContainerOp(name='{{ operation.notebook|regex_replace }}',
```
akchinSTC, line 48:
```suggestion
op_{{ operation.id|regex_replace }} = NotebookOp(name='{{ operation.notebook|regex_replace }}',
```

akchinSTC (changes requested):
lets revert these back to the original class name, and we can update the airflow notebook to a better name the next time we need to cut a release.

elyra/kfp/operator.py
akchinSTC: +1, a little too on the nose

lresende: I am not sure I wanted to use Elyra on the name of the op, would ExtendedFileOp or just FileOp be ok?

kevin-bates:
This looks similar to #1132.  @akchinSTC - does this PR sufficiently address the issue?

elyra/kfp/operator.py
kevin-bates: I understand not wanting `Elyra` in the op name.  `FileOp` seems fine.  @ptitzler had suggested `RunFileOp` in a sidebar and I like that too and that's what happens.  I guess in that sense `ExecuteFileOp` also works.  I don't think `Extended` fits this context.

akchinSTC:
looks like it should satisfy the issue, the airflow uses its own copy of the boostrapper script but as far as I can tell there isnt any different between the two, dont think there should be considering the script is running inside an isolation container/pod and doesnt contain any runtime specific properties.
we should update the airflow-notebook as well next go around to pull scripts from elyra as well. I think we might be able to drop the airflow package req as well

akchinSTC (approved):

kevin-bates:
The final changes look good - this will be nice!

PR #1936, opened 2021-07-20
Improve how type information is inferred during component parsing

### LINKED ISSUE #1926: Incorrect property type being inferred from airflow operator
**Describe the issue**
See : https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/http_operator.py

Traced the issue to headers param is defaulted to None with no type hints available. Our component parser looks like it resolves in the following order: type hint in init method followed by type in docstring. In this special case, the type is a sentence describing the parameter, resulting in falling back to a default of `string`. 

**To Reproduce**
1. create a new airflow pipeline file in VSE
2. open palette and drag the http_operator
3. look at the properties, see `header` as type string and not dict

**Screenshots or log output**

**Expected behavior**
The parser should parse the operator constructor types correctly but given this special case and lack of typing, it will be difficult to fix without upstream changes or hosting of the operator ourselves

**Deployment information**
 - Elyra version: [e.g. 1.5.3]
 - Operating system: [e.g. macos, linux]
 - Installation source: [e.g. PyPI, conda, from source, official container image, custom container image]
 - Deployment type: [e.g. local installation, Docker, Kubernetes, Kubeflow [notebook server] , Open Data Hub]

 - Local execution
 - Kubeflow Pipelines (provide version number, whether multi-user auth enabled)
 - Apache Airflow (provide version number)
 
**Runtime configuration settings**
If the issue is related to pipeline execution, document the runtime configuration settings from the Elyra UI, omitting confidential information.

kiersten-stokes:
Fixes #1926. 

This PR moves the assignment of component parameter `type`, `default value`, and `control_id` out of `ComponentParameter` init into a `ComponentParser` method. It adds additional logic to cover cases where the type information parsed from the component specification is not 'standard' or simple, e.g. type (as parsed) is something like `an array of strings`.

This also amends the type hint that is added to the parameter description to be the type as-parsed rather than the type determined after standardization. The idea here is that the type given in the specification will provide more context to the user as to what is expected.

Screenshots to be added.

Tests are passing, but we will want to add cases to existing tests where the parameter information parsed from the spec is of a non-standard type.
 
           in the file; or

           it.


elyra/pipeline/component.py
```diff
@@ -61,49 +63,15 @@ def __init__(self, ref: str, name: str, type: str, value: str, description: str,
...
-            type = 'string'
-            if not value:
-                value = ''
-        elif type_lowered in ['int', 'integer', 'number']:
-            control_id = "NumberControl"
-            if not value:
-                value = 0
-        elif type_lowered in ['bool', 'boolean']:
-            control_id = "BooleanControl"
+        if control_id == "BooleanControl":
```
kiersten-stokes, line 66: On looking at this further, I think this block can be removed as it will be covered during the properties JSON template render.

elyra/pipeline/component.py
```diff
@@ -360,13 +341,9 @@ def determine_type_information(self, type: str):
...
         elif type_lowered in ['dict', 'dictionary']:
             type_lowered = "dictionary"
-            default_value = ''
         elif type_lowered in ['list', 'set', 'array', 'arr']:
             type_lowered = "list"
-            default_value = ''
-        elif type_lowered in ['file']:
-            default_value = ''
-        else:
+        elif type_lowered not in ['file']:
```
kevin-bates, line 346: This is correct, but it requires the reader to think about the `not` condition and can be done simply because there's only one entry in the 'file' list.  Should another type mapping be introduced or we find multiple mappings for 'file' someday, then we'd need to abandon this approach.  I think this might be easier to read and maintain at the expense of some redundancy:
```python
    elif type_lowered in ['file']:
        type_lowered = 'file'
    else:
        type_lowered = 'string'
```

kiersten-stokes: I had this thought as well! Will change back to a more readable format

kevin-bates (approved):
LGTM - thank you.

PR #1948, opened 2021-07-22
Minor changes to component & category parsing as a followup to palette restructuring

kiersten-stokes:
A followup to the change in palette format returned to the frontend as of #1882 merge.

In discussing with @kevin-bates, we found that there were some areas of opportunity to improve upon variable names and type hints in the component registry/parsers, and to streamline the return of categories along with components. 

- [x] Change the name `adjusted_id` to `catalog_entry_id` in order to better reflect the relationship between the two id's for (Airflow) components that define multiple operators per file
- [x] Add type hints to all component parser files
- [x] Add type hints to all registry files
- [x] Push parsing into `_read_component_registry` in order to restructure how the registry returns both components and categories

Existing `pytest` suite should cover these scenarios.

 
           in the file; or

           it.


elyra/pipeline/component_registry.py
```diff
@@ -79,40 +80,32 @@ def __init__(self, component_registry_location: str, parser: ComponentParser):
...
-            component = self._parser.parse(component_entry)
-            if component:
-                components.extend(component)
 
-        return components, category_entries
+        components: List[Component] = self._read_component_registry().get('components')
+        return components
 
-    def get_component(self, component_id) -> Component:
+    def get_component(self, component_id: str) -> Component:
```
kevin-bates, line 91: I still don't understand why this method can't just look up its `id` in the result of `get_all_components()`.  Are the components in the list returned from `get_all_components()` different than the component returned from this method?

elyra/pipeline/component_registry.py
```diff
@@ -255,14 +267,17 @@ def get_component(self, component_id: str) -> Optional[Component]:
...
         if self._is_cache_expired():
             self._update_cache()
 
         return self._cached_categories
 
     def _update_cache(self):
-        self._cached_components, self._cached_categories = super().get_all_components()
+        registry = super()._read_component_registry()
+        self._cached_components = registry.get('components')
+        self._cached_categories = registry.get('categories')
```
kevin-bates, line 279: I'm not able to place a comment on `CachedComponentRegistry.get_component()` directly, so adding one here.

Per my previous comment, I'm fairly certain that `CachedComponentRegistry.get_component()` is returning a different result than `ComponentRegistry.get_component()` because this one is behaving as I'd suspect and just using the results of `get_all_components()` to locate the specific component.

elyra/pipeline/component_registry.py
```diff
@@ -176,52 +173,67 @@ def _read_component_registry(self) -> Tuple[List, List]:
...
-            self.log.error(f"Component with ID '{component_id}' could not be found in the " +
-                           f"{self._component_registry_location} component_id catalog.")
-            raise ValueError(f"Component with ID '{component_id}' could not be found in the " +
-                             f"{self._component_registry_location} component_id catalog.")
-
-        return component_entry
+    def _read_registry_for_component(self, queried_component_id: str, catalog_entry_id: str) -> Component:
+        with open(self._component_registry_location, 'r') as catalog_file:
+            catalog_json = json.load(catalog_file)
+            component_entry = dict(catalog_json['components'].items()).get(catalog_entry_id)
```
kevin-bates, line 214: Isn't `catalog_json['components']` already a dictionary?  This statement seems equivalent to:
```python
component_entry = catalog_json['components'].get(catalog_entry_id)
```

elyra/pipeline/component_registry.py
kiersten-stokes: That would be much easier. We currently don't store the results of `get_all_components`, but obviously if we did, we could access it on a call to `get_component` (like how it's done in the cache). Should we go ahead and do that? I just wasn't sure if there were other potential implications of this option vs. opening/reading the registry again

elyra/pipeline/component_registry.py
```diff
@@ -176,52 +173,67 @@ def _read_component_registry(self) -> Tuple[List, List]:
...
+            if not component_entry:
+                self.log.error(f"Component with ID '{queried_component_id}' could not be found in the " +
+                               f"{self._component_registry_location} component_id catalog.")
+                raise ValueError(f"Component with ID '{queried_component_id}' could not be found in the " +
+                                 f"{self._component_registry_location} component_id catalog.")
+
+            component_type = next(iter(component_entry.get('location')))
```
kevin-bates, line 221: Is this statement extracting the first (and only) key name within the location stanza? I.e.,
```python
component_type = component_entry.get('location').keys()[0]
```

 I find the `next(iter(` stuff difficult, but that's due to my unfamiliarity with Python.  I think a comment would be helpful here.

elyra/pipeline/component_registry.py
kevin-bates: Ah, right.  So you're saving the parsing of the other components and `parse()` loads the component as part of its operation.  That makes sense.  Please leave as-is.

elyra/pipeline/component_registry.py
kevin-bates: Per Kiersten's previous response, my assumption was incorrect.  This is moot.

elyra/pipeline/component_registry.py
kiersten-stokes: Wow 😄 not sure what I was thinking there

elyra/pipeline/component_registry.py
kevin-bates: I just checked that the two statements are only logically equivalent (`keys()` returns a `dict_keys` object which is not scriptable), so please add a comment that you're going after the key name for the entry in `location`.

kevin-bates: I wonder if this variable should be named `location_type` since it's value will be either `"url"` or `"filename"`?

kevin-bates (approved):
Thanks Kiersten - these changes help improve maintainability quite a bit.

PR #1950, opened 2021-07-22
Update Airflow operation naming for uniqueness

akchinSTC:
When two operations share the same name Airflow is
is unable to render the DAG flow correctly. This fix
appends a numberical prefix to the end of the op name
so it will resolve correctly

The following separate operations will result in a incorrectly rendered DAG in airflow
```
op_2f3f6243_82ef_43ee_af09_9888a5dfbc30 = NotebookOp(
    name="load_data",
    namespace="default",
    task_id="load_data",
    notebook="flight-delay-notebooks/notebooks/load_data.py",
    image="amancevice/pandas:1.1.1",
 ....
    
op_b614bf55_c127_413a_935a_175d7afa7f0e = NotebookOp(
    name="load_data",
    namespace="default",
    task_id="load_data",
    notebook="flight-delay-notebooks/notebooks/load_data.py",
    image="amancevice/pandas:1.1.1"
,........
```
appending a numerical suffix to the op name resolves the issue

```
op_b614bf55_c127_413a_935a_175d7afa7f0e = NotebookOp(
    name="load_data_2",
    namespace="default",
    task_id="load_data_2",
    notebook="flight-delay-notebooks/notebooks/load_data.py",
    image="amancevice/pandas:1.1.1"
,........
```

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_airflow.py
+++ b/elyra/pipeline/processor_airflow.py
@@ -202,7 +202,7 @@ class AirflowPipelineProcessor(RuntimePipelineProcessor):
                 names = [op['notebook'] for op in target_ops]
                 while unique_operation_name in names:
                     unique_name_counter += 1
-                    unique_operation_name = ''.join([unique_operation_name, '_', str(unique_name_counter)])
+                    unique_operation_name = ''.join([operation.name, '_', str(unique_name_counter)])
 
                 target_op = {'notebook': unique_operation_name,
                              'id': operation.id,
--- a/elyra/pipeline/processor_airflow.py
+++ b/elyra/pipeline/processor_airflow.py
@@ -197,7 +197,14 @@ class AirflowPipelineProcessor(RuntimePipelineProcessor):
                             image_instance.metadata.get('pull_policy'):
                         image_pull_policy = image_instance.metadata['pull_policy']
 
-                target_op = {'notebook': operation.name,
+                unique_name_counter = 1
+                unique_operation_name = operation.name
+                names = [op['notebook'] for op in target_ops]
+                while unique_operation_name in names:
+                    unique_name_counter += 1
+                    unique_operation_name = ''.join([unique_operation_name, '_', str(unique_name_counter)])
+
+                target_op = {'notebook': unique_operation_name,
                              'id': operation.id,
                              'filename': operation.filename,
                              'runtime_image': operation.runtime_image,


kevin-bates:
This strikes me as something we should generally enforce during pipeline creation (irrespective of runtime).  I think auto-renaming could introduce confusion to users trying to correlate their renamed nodes to those in the pipeline.  I understand this being a stop-gap until a better solution can be implemented, but I believe the user should be solely responsible for a node's name.

elyra/pipeline/processor_airflow.py
```diff
@@ -197,7 +197,14 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
 
-                target_op = {'notebook': operation.name,
+                unique_name_counter = 1
+                unique_operation_name = operation.name
+                names = [op['notebook'] for op in target_ops]
+                while unique_operation_name in names:
+                    unique_name_counter += 1
+                    unique_operation_name = ''.join([unique_operation_name, '_', str(unique_name_counter)])
+
+                target_op = {'notebook': unique_operation_name,
```
ptitzler, line 207: In your example `name` and `task_id` were updated to generate a unique name:
```
    name="load_data_2",
    namespace="default",
    task_id="load_data_2",
```

Is `operation. id` (line below) not the same as `task_id` and should therefore also be updated?

ptitzler:
DAGs should complete successfully for these conflicting scenarios:
- [ ] two generic nodes with the same filename and no labels (file1/nolabel + file1/nolabel)
- [ ] two generic nodes with the same filename and different labels (file1/label1 + file1/label2)
- [ ] two generic nodes with the same filename and the same labels (file1/label + file1/label)
- [ ] two generic nodes with different filenames and no labels (file1/nolabel + file2/nolabel)
- [ ] two generic nodes with different filenames and the same labels (file1/label + file2/label)
- [ ] a generic node and an rts node with the same labels (file1/label + op/label)

elyra/pipeline/processor_airflow.py
akchinSTC: operation.id is a UUID generated by the frontend (i think) that we use to give the notebook op instance variable a distinct name in the template. I dont think we use it anywhere else.  e.g. op_2f3f6243_82ef_43ee_af09_9888a5dfbc30

akchinSTC:
> This strikes me as something we should generally enforce during pipeline creation (irrespective of runtime). I think auto-renaming could introduce confusion to users trying to correlate their renamed nodes to those in the pipeline. I understand this being a stop-gap until a better solution can be implemented, but I believe the user should be solely responsible for a node's name.

For now we can do this as a stop gap until we think of something more "elegant". Will be including a `warning` as well in the validation service PR to alert users we are renaming some conflicting node names.

elyra/pipeline/processor_airflow.py
akchinSTC: airflow runtime specific components also appear to use a different naming scheme that uses a timestamp to generate a unique task name for the DAG.

kiersten-stokes (commented):
It doesn't really affect the functionality, but I noticed when testing with three (or more) operations of the same name we get the following:

Screen Shot 2021-07-23 at 2 54 31 PM

Functionality looks good in all my test cases, though!

ptitzler:
> we get the following:

ouch

akchinSTC:
@kiersten-stokes - pushing a fix for that shortly

kiersten-stokes (approved):
LGTM!

PR #1957, opened 2021-07-23
Fix computation of absolute path for url-based KFP components

kiersten-stokes:
This fixes a bug where the absolute path was being appended to KFP runtime-specific components for both filepath and url-based components. 

- Add a new url-based KFP component to the KFP component catalog file. 
- Add the component to a pipeline and export or submit.
- Export/submission succeeds.

Two new tests added (one with a filename-based component source and one with a url-based source) to confirm that KFP custom components are being successfully converted into their corresponding `TaskSpec` objects.

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_kfp.py
+++ b/elyra/pipeline/processor_kfp.py
@@ -517,7 +517,10 @@ class KfpPipelineProcessor(RuntimePipelineProcessor):
                         operation.component_params[component_property.ref] = ast.literal_eval(op_property)
 
                 # Get absolute path of component source
-                component_path = os.path.join(os.path.dirname(__file__), "resources", component.source)
+                component_path = component.source
+                if component.source_type == "filepath":
+                    component_path = os.path.join(os.path.dirname(__file__), "resources", component_path)
+
                 component_source = {}
                 component_source[component.source_type] = component_path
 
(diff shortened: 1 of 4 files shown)


lresende:
@kiersten-stokes will be adding a test

kiersten-stokes:
I'm not sure if the tests added in the latest commit are kosher or not. Some details below.

The goal of these tests is to determine whether KFP custom components are created properly from their respective sources (one filename-based, one url-based). I played around with calling the processor's `cc_pipeline()` directly, but this required a that a runtime config metadata instance be defined. So, instead, I've created a function `_mock_cc_pipeline()` that performs the pertinent set of operations for a KFP custom component (some extraneous steps have been removed, but the rest of the function remains the same as that in `processor_kfp.py`).

If there is some other workaround where a metadata instance should be created first, and then `_cc_pipeline` run directly, I'm happy to make the change!

kiersten-stokes:
Thanks @akchinSTC for pointing out a much easier way to implement the tests! Now, the tests use a pytest fixture to mock a runtime config, and `_cc_pipeline()` is called directly.

akchinSTC (approved):

ptitzler (approved):
LGTM

PR #1958, opened 2021-07-23
Make 'name' an optional property when parsing a component registry entry

kiersten-stokes:
No error is thrown when accessing an entry from the component registry that does not include the `name` key. This allows a component registry entry like the below to be parsed with no difference in functionality.

```
"run-notebook-using-papermill": {
  "location": {
    "filename": "run_notebook_using_papermill/component.yaml"
  },
  "category": "kfp"
}
```

This also removes the `name` key from the KFP and Airflow component catalogs to reduce confusion.

We may want to discuss whether we want to remove `name` and all uses of it in the catalogs and during parsing since we do not currently use this field for anything. The idea behind including the name field sort of rests on the assumption that eventually we may have a GUI where users can add new components to the catalog and provide their own UI-friendly name.

- Delete the `name` key/value pair from an entry in either component catalog
- Rebuild and launch JupyterLab
- Drag the changed component onto a pipeline
- No error is thrown

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_registry.py
+++ b/elyra/pipeline/component_registry.py
@@ -178,7 +178,7 @@ class ComponentRegistry(LoggingConfigurable):
                     # TODO Add error checking for category here or elsewhere
                     entry = {
                         "id": component_id,
-                        "name": component_entry["name"],
+                        "name": component_entry.get("name"),
                         "type": component_type,
                         "location": component_location,
                         "adjusted_id": None,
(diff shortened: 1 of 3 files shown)


ptitzler:
Can you please remove the `name` property from https://github.com/elyra-ai/elyra/blob/master/etc/config/components/kfp_component_catalog.json and https://github.com/elyra-ai/elyra/blob/master/etc/config/components/airflow_component_catalog.json

ptitzler:
> Can you please remove the name property 

My main concern is that users will see the property and assume that the specified identifier is exposed/used somewhere, but that is currently not the case.

ptitzler (approved):
LGTM

PR #1970, opened 2021-07-27
Reverse how description and type information appear in node properties

kiersten-stokes:
Fixes the issue seen in the below screenshot where component description and type are in the opposite position as expected.

127157010-3ebc0b75-75d9-4122-8184-dfb9b8c11326

Creates a description of the format `[parsed description] (type: [parsed type])`, as shown in below screenshot. 

Screen Shot 2021-07-27 at 11 53 23 AM

Also adds a clause to the description parsing regex for Airflow that accounts for the case where a parameter is given a description (`:param [param_name]: ...`) but not a type (`:type [param_name]: ...`) in the operator class docstring.
Tests are added that check that the descriptions and types given in the component specification appear as expected.

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_parser_airflow.py
+++ b/elyra/pipeline/component_parser_airflow.py
@@ -148,7 +148,7 @@ class AirflowComponentParser(ComponentParser):
                 description = match.group(1).strip().replace("\"", "'")
 
             # Amend description to include type information
-            description = self._get_description_with_type_hint(type, description)
+            description = self._format_description(description=description, type=type)
 
             type, control_id, default_value = self.determine_type_information(type)
 
--- a/elyra/pipeline/component_parser_airflow.py
+++ b/elyra/pipeline/component_parser_airflow.py
@@ -142,13 +142,13 @@ class AirflowComponentParser(ComponentParser):
 
             # Search for :param [param] in class doctring to get description
             description = ""
-            param_regex = re.compile(f":param {arg}:" + r"([\s\S]*?(?=:type|:param))")
+            param_regex = re.compile(f":param {arg}:" + r"([\s\S]*?(?=:type|:param|\"\"\"|'''|\.\.))")
             match = param_regex.search(class_content)
             if match:
                 description = match.group(1).strip().replace("\"", "'")
 
             # Amend description to include type information
-            description = self._get_description_with_type_hint(description, type)
+            description = self._get_description_with_type_hint(type, description)
 
             type, control_id, default_value = self.determine_type_information(type)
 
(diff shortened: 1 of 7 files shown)


elyra/pipeline/component_parser_airflow.py
```diff
@@ -142,13 +142,13 @@ def _parse_properties(self, component_definition: str, component_class: str) ->
...
             description = ""
-            param_regex = re.compile(f":param {arg}:" + r"([\s\S]*?(?=:type|:param))")
+            param_regex = re.compile(f":param {arg}:" + r"([\s\S]*?(?=:type|:param|\"\"\"|'''|\.\.))")
             match = param_regex.search(class_content)
             if match:
                 description = match.group(1).strip().replace("\"", "'")
 
             # Amend description to include type information
-            description = self._get_description_with_type_hint(description, type)
+            description = self._get_description_with_type_hint(type, description)
```
kevin-bates, line 151: I wonder if this should be named `_format_description()` since its not really _getting_ anything, and, as was a natural instinct, have `description` first, then `type`.  I think using keyword arguments might help:
```python
def _format_description(description: str, type: str) -> str:
```
Since the kfp parser could call this with a `None` description (airflow defaults to `""`), I would update its call such that the dictionary `get()` method uses a default value of `""` to ensure its at least the empty string.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -72,7 +72,7 @@ def _parse_properties(self, component_yaml: Dict[str, Any]) -> List[ComponentPar
             type = param.get('type', 'string')
 
             # Set description and include parsed type information
-            description = self._get_description_with_type_hint(param.get('description'), type)
+            description = self._get_description_with_type_hint(type, param.get('description'))
```
kevin-bates, line 75: Per the previous comment about renaming the method and reordering parameters, it might be worthwhile adding a default value to the getter here.

lresende:
Would we benefit from using named type/value like in (type=type, description=param.get('description')) which would have avoided the issue in the first place even if in the wrong order?

kevin-bates:
> Would we benefit from using named type/value like in (type=type, description=param.get('description')) which would have avoided the issue in the first place even if in the wrong order?

Yes, but since one argument (`description`) _was_ optional, the order (to the best of my understanding) was dictated at the method's definition since optional arguments must follow required arguments.  The last commit eliminates that being the case and introduces the keywords at the call - which I agree is better.  If we wanted to make `description` optional again, we'd need to change the order on the method definition but continue using the keywords.

Seems like we should be getting `description` exactly how we are getting `type` (in both cases), then scrap the keywords to eliminate the redundancy - but that's well into the nit-realm.

kevin-bates (approved):

PR #1973, opened 2021-07-28
Fix ci dialog selector

bourdakos1:

 
           in the file; or

           it.


tests/integration/pipeline.ts
```diff
@@ -280,7 +280,7 @@ describe('Pipeline Editor tests', () => {
 
     // close tabs
     cy.closeTab(-1); // notebook tab
-    cy.findByRole('button', { name: /discard/i }).click();
+    cy.findByRole('button', { name: /cancel/i }).click();
```
karlaspuldaro, line 283: We need the discard button to actually close these tabs and clean the env for the test below.
I think the issue is either this `Save your work` dialog is not being displayed anymore (since there hasn't been changes to the open files when we're closing the tabs, even though the dialog still showed up when this test was written). If that's the case, we can remove these lines.

Another idea is separating these tests in two each: one to test the pipeline in root dir, another for the pipeline in a subfolder scenario. That way we don't need to manually close the tabs between them.

bourdakos1: It's seems that in the latest version of JupyterLab the `Discard` button doesn't exist anymore. I haven't tested this locally yet

image

bourdakos1: oh wait, this looks like a submit dialog, not close, now I'm confused haha

bourdakos1: looks like the `save your work` dialog doesn't show anymore, so I removed the discard click

karlaspuldaro (approved):
LGTM

PR #1980, opened 2021-07-28
Document backend build

karlaspuldaro:
Add instructions to build Elyra's server for backend development

Ran `make` to list new command
 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ prepare-server:
 
 only-install-server: prepare-server build-server install-server-package
 
-install-server: lint-server only-install-server ## Install backend
+install-server: lint-server only-install-server ## Build and install backend only
 
 install-server-package:
 	pip install --upgrade --upgrade-strategy $(UPGRADE_STRATEGY) --use-deprecated=legacy-resolver dist/elyra-*-py3-none-any.whl
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ prepare-server:
 
 only-install-server: prepare-server build-server install-server-package
 
-install-server: lint-server only-install-server # Install backend
+install-server: lint-server only-install-server ## Install backend
 
 install-server-package:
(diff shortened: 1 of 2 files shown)


docs/source/developer_guide/development-workflow.md
```diff
@@ -128,7 +128,13 @@ You can install Elyra using a local build of @elyra/pipeline-editor with:
 make clean dev-link install
 ```
 
-### Incremental Development
+### Back-end Development
```
lresende, line 131: I would describe this in more details similar to the one below (--dev-lint) like when to use, what it does, etc

Makefile
```diff
@@ -136,7 +136,7 @@ prepare-server:
 
 only-install-server: prepare-server build-server install-server-package
 
-install-server: lint-server only-install-server # Install backend
+install-server: lint-server only-install-server ## Install backend
```
kevin-bates, line 139: Since the `install` target is described as "Build and install", it seems like this description should be "Build and install backend only".

This also leads to one asking whether there should be a similar _frontend-only_ target.  I.e., do we need to expose `install-ui`?

Also, in looking at the targets, `check-install` seems generally useful and could be helpful in confirming a dev-based installation.  Not sure if we should document that more formally (i.e., mention it in the docs and add it to the default `make` output).

lresende: lol, I told @karlaspuldaro these questions were going to come :) I will let you guys decide.

karlaspuldaro: As discussed in today's meeting to gather an overall team opinion, it was decided
1. not to target `install-ui` since `watch` seems more useful for frontend dev and it's already documented
2. `check-install` is a good candidate to be exposed and documented, maybe it needs a better name, but for now we'll create a separate issue for it

kevin-bates (approved):
Thanks Karla.

ajbozarth (approved):

PR #1984, opened 2021-07-28
Fix error case where error handler throws error

ajbozarth:
Currently when this error handling is called the following error is sometimes thrown:

```
ERROR elyra/kfp/tests/test_bootstrapper.py::test_main_method - AttributeError: 'NoneType' object has no attribute 'decode'
```

 
           in the file; or

           it.

### DIFF
--- a/elyra/kfp/tests/test_bootstrapper.py
+++ b/elyra/kfp/tests/test_bootstrapper.py
@@ -54,7 +54,11 @@ def start_minio():
             check=True)
         time.sleep(3)
     except subprocess.CalledProcessError as ex:
-        raise RuntimeError(f'Error executing process: {ex.stderr.decode("unicode_escape")}') from ex
+        if ex.stderr:
+            error = ex.stderr.decode("unicode_escape")
+        else:
+            error = ''
+        raise RuntimeError(f'Error executing process: {error}') from ex
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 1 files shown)


elyra/kfp/tests/test_bootstrapper.py
```diff
@@ -54,7 +54,11 @@ def start_minio():
             check=True)
         time.sleep(3)
     except subprocess.CalledProcessError as ex:
-        raise RuntimeError(f'Error executing process: {ex.stderr.decode("unicode_escape")}') from ex
+        if ex.stderr:
+            error = ex.stderr.decode("unicode_escape")
```
kevin-bates, line 58: I'm fairly certain we can abandon the `.decode("unicode_escape")` suffix altogether since all strings are unicode in Python 3 anyway.  `decode()` is essentially unnecessary.

As a result, we could revert to the original statement with something like:
```python
raise RuntimeError(f"Error executing process: {ex.stderr or '')}") from ex
```

A quick test of this yields the following in the output...
```
        except subprocess.CalledProcessError as ex:
>           raise RuntimeError(f"Error executing process: {ex.stderr or ''}") from ex
E           RuntimeError: Error executing process:
```
which indicates that `ex.stderr` was `None`.

ajbozarth: If that's ok then I'll update it to that then

ajbozarth: pushed

kevin-bates (approved):

Issue #1761, opened 2021-06-14
2.3/rc1: VPE component nodes need to expose output parameters


ptitzler:
**Describe the issue**

Currently the VPE does not expose KFP component outputs, making it impossible for users to choose them as inputs for subsequent nodes. In this example, `get lines` produces a file that `count lines` needs to access.

![image](https://user-images.githubusercontent.com/13068832/121972336-a7f5e880-cd2f-11eb-9c28-4d597c8eb804.png)

A working DSL implementation of this pipeline can be found here: https://github.com/ptitzler/kfp-component-tests/blob/main/pipelines/create_and_run_pipeline-1.py

**To Reproduce**
1. Register https://raw.githubusercontent.com/ptitzler/kfp-component-tests/main/example-1/component.yaml as a KFP component (`Get Lines`)
2. Register https://raw.githubusercontent.com/ptitzler/kfp-component-tests/main/example-2/component.yaml as a KFP component (`Count Lines`)
3. Create a pipeline from the two components as shown above.

**Expected behavior**
A user should be able to select _any_ compatible output from a parent node as an input.

ptitzler:
Example scenarios (`Cn` = `<component N>`, `On` = `<output N>`, `In` `<input N>`)
- `C1()` -> `C2(I1)`  => user can only enter the value of I1 because C1 does not declare any outputs
-  `C1(O1,O2)` -> `C2 (I1)`  => For C2 user can
      - do nothing 
      - enter the value of `I1`
      - select `O1` or `O2` as the value of `I1`
-  `C1(O1,O2)` -> `C2 (I1,O3)` -> `C3(I2)`  => for C3 user can
      -  do nothing 
      -  enter the value of `I2`
      -  select `O1` or `O2` as the value of `I2`
      - select `O3` as the value of `I2`

kevin-bates:
Thank you - this is helpful. I see that you added 'do nothing' to the first case - which was my question after reading the email-notification (original) version.

So it seems our general idea of "all outputs are available as inputs" will need much greater scrutiny since all inputs must be explicitly declared.

Also wondering how wild-carding would work.  I suppose `inputPath` could be a file containing a set of other references.  This would only apply to our built-in components of course.  (I actually thought this comment was on #1765, so I apologize for the cross-posting kind of thing.)

ptitzler:
> I see that you added 'do nothing' to the first case - which was my question after reading the email-notification (original) version.

I added this to cover optional input parameters.

[...]

PR #1986, opened 2021-07-29
Convert to relative path for all file-based node properties

### LINKED ISSUE #1991: Filepath handling for generic and custom components
How to handle conversion of properties with paths: pipeline-relative -> workspace-relative

-more info coming soon-

marthacryan:
Updates the path conversion between relative to pipeline and relative to workspace to run on all node file path node properties, not just the generic cases. This is meant to allow custom components that "represent" a file(s) to be portable across instances the same as the generic components.

Fixes #1991 

**Note**: This only converts string fields, not string array fields since the conversion should only occur on file fields that "represent" a node, which an array such as dependencies do not.
 
           in the file; or

           it.


packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -375,19 +375,22 @@ export class PipelineService {
...
       }
       for (const param of params) {
-        if (param.data?.format === 'file') {
+        if (
+          param.data?.format === 'file' &&
+          param.custom_control_id === 'StringControl'
+        ) {
           node.app_data.component_parameters[
-            param.parameter_ref
+            param.parameter_ref.substring(6)
```
kevin-bates, line 390: I'm curious what the 6th index is about and my hunch is that this must be skipping over the `elyra_` prefix.  If so, would it make sense to define a constant (e.g., `ELYRA_PREFIX_LEN`) for this value where a comment could then be added.

kiersten-stokes (approved):
This is working for me! Example screenshots below of a pipeline consisting of a generic `Notebook` component and a `Run notebook using papermill` component that both use a relative path to the notebook in question.

Generic node properties:
Screen Shot 2021-07-29 at 12 58 05 PM

Papermill component node properties:
Screen Shot 2021-07-29 at 12 58 23 PM

Success on KFP Dashboard:
Screen Shot 2021-07-29 at 1 01 27 PM

packages/pipeline-editor/src/PipelineService.tsx
marthacryan: Good idea - will do!

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -27,6 +27,8 @@ export const RUNTIMES_NAMESPACE = 'runtimes';
 export const RUNTIME_IMAGES_NAMESPACE = 'runtime-images';
 export const COMPONENTS_NAMESPACE = 'components';
 
+const ELYRA_PREFIX_LEN = 6;
```
ajbozarth, line 30: I'm not sure I like this way of handling the prefix, @bourdakos1 this is different from how pipeline-editor handles it, should we do it like that side or just use this implementation?

packages/pipeline-editor/src/PipelineService.tsx
```diff
@@ -370,18 +372,31 @@ export class PipelineService {
...
+        return n.op === node.op;
+      });
+      const params = nodeDef?.app_data?.properties?.uihints?.parameter_info;
+      if (!params) {
+        continue;
+      }
+      for (const param of params) {
+        if (
+          param.data?.format === 'file' &&
+          param.custom_control_id === 'StringControl'
```
ajbozarth, line 389: Why do we need to check this as well? is there a case where the format is `file` but it's not a string?

ajbozarth (changes requested):
A few comments, also is there a reason we are only updating `setNodePathsRelativeToWorkspace` and not `setNodePathsRelativeToPipeline` those two functions should be the inverse of each other. I plan to check this out and test it locally, but I'm unsure how it works without updating `setNodePathsRelativeToPipeline` as well

packages/pipeline-editor/src/PipelineService.tsx
bourdakos1: We can’t just strip off the first 6 characters, only a subset of the property IDs will be prefixed so we need to check that the property actually has the prefix

bourdakos1 (changes requested):
There’s a couple issues I see here, I’ll follow up with a full review soon

packages/pipeline-editor/src/PipelineService.tsx
bourdakos1: I don't think that we should check either of these. I would say this is a case where we only would update the path of the property that the `filehandler` parameter ref is pointing to

bourdakos1: I realize the papermill op doesn't have `filehandler` though

bourdakos1 (commented):
After discussion, I think we can move forward with this. I think the only thing we need to change is the prefix.

also if possible, is there a better way to check StringControl instead of explicitly checking the control?

packages/pipeline-editor/src/PipelineService.tsx
bourdakos1: @ajbozarth we have to check `StringControl` because `StringArrayControl` can also have `format: "file"` and we don't want to convert any of those

bourdakos1: @marthacryan can we add a comment explaining this?

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth: My next update will change this to use a util function similar to how pipeline-editor handles it

packages/pipeline-editor/src/PipelineService.tsx
ajbozarth: Closing this thread since I now understand the purpose

packages/pipeline-editor/src/PipelineService.tsx
kevin-bates, line 30: Could we add a comment here indicating what this is about?  Perhaps something like: 
```
/*
 * Generic component properties are prefixed with 'elyra_' in order to distinguish themselves from custom component properties.
 */
```

This kind of comment could go at the point of use instead.

@kiersten-stokes - is that wording correct?

ajbozarth: I haven't pushed my updates yet, but this line will be gone and replaced with a util function with a comment similar to the one you gave

ajbozarth:
I pushed my updates, I'm leaving this in draft for now because I have yet to test them (only build)

packages/pipeline-editor/src/PipelineService.tsx
kiersten-stokes: That sounds good! If we wanted to be more specific, maybe something like "Generic component properties and certain meta-properties (e.g., 'label') that apply to both generic and custom components are prefixed..." and so on as above

packages/pipeline-editor/src/utils.ts
```diff
@@ -105,4 +105,15 @@ export default class Utils {
 
     return widget;
   };
+
+  /**
+   * Strips the `elyra_` prefix from a property key if it exists
+   */
```
kevin-bates, line 111: I think it would be better to state _why_ this is necessary.  The code and method name already do a good job of _what_ is happening.

I found @kiersten-stokes' https://github.com/elyra-ai/elyra/pull/1986#discussion_r698831913 helpful.

ajbozarth:
Based on discussions with @akchinSTC in a Webex this PR becomes unnecessary based on a new understanding of kfp components as found during work on #1761 

In particular we have found that kfp components will never have "representative" files, they will always get a file from a previous node, and nodes can only "represent" files found in locations like s3, not locally.

PR #1974, opened 2021-07-28
Refactor component location to jupyter/shared/components/runtime

lresende:
Fixes #1965

 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #1990, opened 2021-07-29
Re-enable tests for constructing KFP pipeline with custom components

### LINKED ISSUE #1981: Server tests are failing when calling _cc_pipeline directly
**Describe the issue**
The 2 new tests introduced in #1957 are failing as a result of changes made in #1968. The `get_display_name()` function introduced in the latter PR expects a `ContainerOp` object whereas the tests create a `TaskSpec` object. The disconnect appears to happen due to some of the internals of the KFP compiler that we do not have access to.

Related to #1756 as well.

**Screenshots or log output**
Test failure output is below.
<pre>
Error constructing component Filter text: 'TaskSpec' object has no attribute 'set_display_name'</pre>
</details>

**Expected behavior**
Eventually, all invocations of _cc_pipeline() should produce the same output. For now, we'll skip these tests.

**Deployment information**
 - Elyra version: latest master

kiersten-stokes:
Fixes #1981 as a follow up to #1982. 

Instead of calling `_cc_pipeline()` directly, the pipeline YAML is compiled and saved to `tmpdir` for the comparison of expected vs. actual results.

~~Also changes the url for the tested component to reflect the relocation of component specs in #1974.~~

After discussion, a new test resource was added for the `Filter text` component specification. This way, the url-based component test will now be based on this static version of the component, and any changes made to the component spec stored in `etc/config/components` will not affect this particular test.

The test `test_processor_kfp.py::test_processing_url_runtime_specific_component` will fail until this is merged, as it requires the component spec to be present in the pipeline test resources folder on master.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/tests/test_processor_kfp.py
+++ b/elyra/pipeline/tests/test_processor_kfp.py
@@ -13,13 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import ast
 import os
 import tarfile
 from unittest import mock
 
 from kfp import compiler as kfp_argo_compiler
-from kfp.components.structures import TaskSpec
 import pytest
+import yaml
 
 from elyra.metadata.metadata import Metadata
 from elyra.pipeline.component import Component
@@ -192,11 +193,10 @@ def test_collect_envs(processor):
     assert 'USER_NO_VALUE' not in envs
 
 
-@pytest.mark.skip(reason="failing - need to evaluate function at kfp compiler level")
-def test_processing_url_runtime_specific_component(monkeypatch, processor, sample_metadata):
+def test_processing_url_runtime_specific_component(monkeypatch, processor, sample_metadata, tmpdir):
     # Assign test resource location
-    url = 'https://raw.githubusercontent.com/elyra-ai/elyra/master/elyra/' \
-          'pipeline/resources/kfp/filter_text_using_shell_and_grep/component.yaml'
+    url = 'https://raw.githubusercontent.com/elyra-ai/elyra/master/' \
+          'etc/config/components/kfp/filter_text_using_shell_and_grep.yaml'
 
     # Instantiate a url-based component
     component = Component(id="filter-text",
@@ -212,12 +212,17 @@ def test_processing_url_runtime_specific_component(monkeypatch, processor, sampl
     processor._component_registry._cached_components = [component]
 
     # Construct hypothetical operation for component
+    operation_name = "Filter text test"
+    operation_params = {
+        "text": "path/to/text.txt",
+        "pattern": "hello"
+    }
     operation = Operation(id='filter-text-id',
                           type='execution_node',
                           classifier='filter-text',
-                          name='Filter text',
+                          name=operation_name,
                           parent_operation_ids=[],
-                          component_params={"text": "resources/components/filter.txt", "pattern": "hello"})
+                          component_params=operation_params)
 
     # Build a mock runtime config for use in _cc_pipeline
     mocked_runtime = Metadata(name="test-metadata",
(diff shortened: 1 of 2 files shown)


elyra/pipeline/tests/test_processor_kfp.py
```diff
@@ -192,11 +193,10 @@ def test_collect_envs(processor):
...
 
 
-@pytest.mark.skip(reason="failing - need to evaluate function at kfp compiler level")
-def test_processing_url_runtime_specific_component(monkeypatch, processor, sample_metadata):
+def test_processing_url_runtime_specific_component(monkeypatch, processor, sample_metadata, tmpdir):
     # Assign test resource location
-    url = 'https://raw.githubusercontent.com/elyra-ai/elyra/master/elyra/' \
-          'pipeline/resources/kfp/filter_text_using_shell_and_grep/component.yaml'
+    url = 'https://raw.githubusercontent.com/elyra-ai/elyra/master/' \
+          'elyra/pipeline/tests/resources/components/filter_text.yaml'
```
kevin-bates, line 199: I guess this test won't succeed until it's merged.  Have you run a test that (temporarily) points to a different branch?  (I really don't want to have to introduce env overrides for branches, so am okay waiting and fixing if broken.)

elyra/pipeline/tests/test_processor_kfp.py
```diff
@@ -212,12 +212,17 @@ def test_processing_url_runtime_specific_component(monkeypatch, processor, sampl
     processor._component_registry._cached_components = [component]
 
     # Construct hypothetical operation for component
+    operation_name = "Filter text test"
```
kevin-bates, line 215: I see `operation_name` includes a " test" suffix yet the corresponding yaml doesn't include such a string.  Is this `operation_name` equated to anything in the component.yaml or does this value just get rendered as `pipelines.kubeflow.org/task_display_name` below?

elyra/pipeline/tests/test_processor_kfp.py
```diff
@@ -230,42 +235,63 @@ def test_processing_url_runtime_specific_component(monkeypatch, processor, sampl
...
-                          source="kfp/filter_text_using_shell_and_grep/component.yaml",
+                          source=relative_path,
                           properties=[],
                           catalog_entry_id="")
 
     # Replace cached component registry with single filename-based component for testing
     processor._component_registry._cached_components = [component]
 
     # Construct hypothetical operation for component
+    operation_name = "Filter text test"
```
kevin-bates, line 284: Given this component is remaining in its present location and (absolutely) doesn't contain a " test" suffix, I suspect this answers my previous questions and `operation_name` is just a rendered entity.

kevin-bates (approved):
This looks good. I had one question (which I left in place) that I believe was subsequently answered later in the review.  Should the answer be different, I may change my review. :smile:

elyra/pipeline/tests/test_processor_kfp.py
kiersten-stokes: Yes, I tested against the resource in my branch (`url = https://raw.githubusercontent.com/kiersten-stokes/elyra/server-test/elyra/pipeline/tests/resources/components/filter_text.yaml`) and it's passing!

kevin-bates:
The current server test failure is expected due to the way the test is structured (needs to reference a resource in the github master branch) but has been tested locally.  As a result, I think this will need to be one of those PRs merged with a known test failure - provided `elyra/pipeline/tests/test_processor_kfp.py::test_processing_url_runtime_specific_component` is the only test failing.

elyra/pipeline/tests/test_processor_kfp.py
kiersten-stokes: Good question. `operation_name` corresponds to what would be the `label` field in an equivalent pipeline JSON. So you're right that it's not in the YAML. In that way, I guess these two tests sort of confirm a handful of things about the output of the `_cc_pipeline()` function (and after pipeline compilation):

- `label` is assigned properly by `set_display_name()
- the YAML is properly loaded in `load_component()` and the spec location is therefore correct
- the corresponding `ContainerOp` object is created properly by the factory function and the component parameters passed to it are as expected

PR #1995, opened 2021-07-30
Add Describe Command to CLI

### LINKED ISSUE #1831: CLI: Add  "describe pipeline" command
Elyra recently introduced additional pipeline metadata that a CLI user likely has an interest in. Currently a user has to either use the VPE or browse the `.pipeline` file to access this information, which is not user friendly.

```
$ elyra-pipeline describe <file.pipeline>
```

would display in a human readable format:
- the pipeline version
- the pipeline description (if defined)
- the pipeline type
- dependencies (e.g. list of custom components that are utilized)

```
$ elyra-pipeline describe --json <file.pipeline>

{
  "version": "4",
  "description": "pipeline description text",
  "type": "generic",
  ...
}
```

would return the same information in machine-readable JSON format. An easy way to test that the generated JSON can be consumed, run 

```
$ elyra-pipeline describe --json mypipeline.pipeline | jq .
```

Constraints:
 - Returns none zero return code if an error is encountered (e.g. no file specified, invalid file specified)

19690ao:

These changes close #1831, adding a `describe` command to the CLI.
```
$ elyra-pipeline describe <path_to_pipeline>
```
Our expected output if [this file](https://github.com/elyra-ai/examples/blob/master/pipelines/run-generic-pipelines-on-kubeflow-pipelines/hello-generic-world.pipeline) was placed in the path:
```
Name: hello-generic-world
Description: A generic pipeline tutorial
Type: Generic
Version: 4
Nodes: 4
File Dependencies: 
    None Listed
```
Any missing and/or empty fields will be listed as "None".
The command has the optional flag `--json`.
```
$ elyra-pipeline describe --json <path_to_pipeline>
```
This will return the same information in JSON format. Any fields that are missing/empty will also be undefined in the output.
This is the `--json` output for the same pipeline file:
```
{
    "name": "hello-generic-world",
    "description": "A generic pipeline tutorial",
    "type": "Generic",
    "version": 4,
    "nodes": 4
}
```

I've added a few tests based on the tests that were already in the `test_pipeline_app` file for the 'run' command.
 
           in the file; or

           it.

### DIFF
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -329,9 +329,9 @@ def describe(pipeline_path):
     pipeline_definition = \
         _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
 
-    print_banner(pipeline_definition)
-
     click.echo(pipeline_definition)
+
+    click.echo(type(pipeline_definition))
 
     print_banner("Elyra Pipeline Description Complete")
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -317,7 +317,7 @@ def run(pipeline_path):
               'json_option',
               is_flag=True,
               required=False,
-              help='Returns pipeline file in a machine-readable JSON format')
+              help='Display pipeline summary in JSON format')
 @click.argument('pipeline_path')
 def describe(json_option, pipeline_path):
     """
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -330,9 +330,9 @@ def describe(json_option, pipeline_path):
 
     indent_length = 4
     for current_pipeline in pipeline_definition["pipelines"]:
-        pipeline_keys = ["name", "description", "type", "version", "nodes", "dependencies"]
+        pipeline_keys = ["name", "description", "type", "version", "nodes", "file dependencies"]
 
-        iter_keys = {"dependencies"}
+        iter_keys = {"file_dependencies"}
 
         describe_dict = OrderedDict()
 
@@ -352,23 +352,24 @@ def describe(json_option, pipeline_path):
 
         describe_dict["nodes"] = len(current_pipeline.get("nodes", []))
 
-        describe_dict["dependencies"] = set()
+        describe_dict["file_dependencies"] = set()
         for node in current_pipeline["nodes"]:
             if "dependencies" in node["app_data"]["component_parameters"]:
                 for dependency in node["app_data"]["component_parameters"].get("dependencies", []):
-                    describe_dict["dependencies"].add(f"{dependency}")
+                    describe_dict["file_dependencies"].add(f"{dependency}")
 
         if not json_option:
             for key in pipeline_keys:
+                readable_key = ' '.join(key.title().split('_'))
                 if key in iter_keys:
-                    click.echo(f"{key.title()}:")
+                    click.echo(f"{readable_key}:")
                     if describe_dict.get(key, set()) == set():
                         click.echo(f"{' ' * indent_length}None")
                     else:
                         for item in describe_dict.get(key, ["None"]):
                             click.echo(f"{' ' * indent_length}{item}")
                 else:
-                    click.echo(f"{key.title()}: {describe_dict.get(key, 'None')}")
+                    click.echo(f"{readable_key}: {describe_dict.get(key, 'None')}")
         else:
             for key in iter_keys:
                 describe_dict[key] = list(describe_dict[key])
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -331,43 +331,32 @@ def describe(json_option, pipeline_path):
         _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
 
     if not json_option:
-        click.echo("Name:")
-        for current_pipeline in pipeline_definition["pipelines"]:
-            click.echo("  " + str(current_pipeline["app_data"]["properties"]["name"]))
-
-        click.echo("Version:")
-        for current_pipeline in pipeline_definition["pipelines"]:
-            click.echo("  " + str(current_pipeline["app_data"]["version"]))
-
-        has_description = False
+        indent_length = 2
 
         for current_pipeline in pipeline_definition["pipelines"]:
-            if current_pipeline["app_data"]["properties"]["description"] != "":
-                if not has_description:
-                    click.echo("Description:")
-                    has_description = True
-                click.echo("  " + str(current_pipeline["app_data"]["properties"]["description"]))
+            pipeline_data = current_pipeline["app_data"]
 
-        if not has_description:
-            click.echo("No Description")
+            click.echo("Name: " + str(pipeline_data["properties"]["name"]))
 
-        click.echo("Type:")
-        for current_pipeline in pipeline_definition["pipelines"]:
-            click.echo("  " + str(current_pipeline["app_data"]["properties"]["runtime"]))
+            click.echo("Version: " + str(pipeline_data["version"]))
 
-        has_dependencies = False
+            click.echo("Description: " + str(pipeline_data["properties"]["description"]))
 
-        for current_pipeline in pipeline_definition["pipelines"]:
+            click.echo("Type: " + str(pipeline_data["properties"]["runtime"]))
+
+            has_dependencies = False
+
+            click.echo("Dependencies:")
+
             for node in current_pipeline["nodes"]:
                 if len(node["app_data"]["component_parameters"]["dependencies"]) >= 1:
                     if not has_dependencies:
-                        click.echo("Dependencies:")
                         has_dependencies = True
                     for dependency in node["app_data"]["component_parameters"]["dependencies"]:
-                        click.echo("  " + str(dependency))
+                        click.echo((indent_length * " ") + str(dependency))
 
-        if not has_dependencies:
-            click.echo("No Dependencies")
+            if not has_dependencies:
+                click.echo((indent_length * " ") + "None")
     else:
         click.echo(json.dumps(pipeline_definition, indent=2))
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -332,7 +332,7 @@ def describe(pipeline_path):
     click.echo(type(pipeline_definition))
 
     for key in pipeline_definition:
-        click.echo(pipeline_definition[key])
+        click.echo(f"{key}: " + str(pipeline_definition[key]))
 
     print_banner("Elyra Pipeline Description Complete")
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -317,9 +317,14 @@ def run(pipeline_path):
 @click.argument('pipeline_path')
 def describe(pipeline_path):
     """
-    Display pipeline description in a human readable format
+    Describe a pipeline in a human readable format
     """
+
+    click.echo()
+
     print_banner("Elyra Pipeline Describe")
+
+    _validate_pipeline_file(pipeline_path)
 
 
 pipeline.add_command(submit)
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -324,15 +324,12 @@ def describe(pipeline_path):
 
     print_banner("Elyra Pipeline Describe")
 
-    _validate_pipeline_file(pipeline_path)
+    _validate_pipeline_file_extension(pipeline_path)
 
     pipeline_definition = \
         _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
 
-    for key in {"version", "doc_type"}:
-        click.echo(f"{key}:\n" + str(pipeline_definition[key]))
-
-    click.echo("Dependencies:")
+    click.echo("dependencies:")
     for current_pipeline in pipeline_definition["pipelines"]:
         for node in current_pipeline["nodes"]:
             for dependency in node["app_data"]["dependencies"]:
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -326,6 +326,11 @@ def describe(pipeline_path):
 
     _validate_pipeline_file(pipeline_path)
 
+    pipeline_definition = \
+        _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
+
+    print_banner(pipeline_definition)
+
     print_banner("Elyra Pipeline Description Complete")
 
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -344,8 +344,9 @@ def describe(json_option, pipeline_path):
 
         properties = pipeline_data.get("properties", dict())
 
-        # If the name is actually "None", it will seem as if there is no name
-        # The same can be said for all fields
+        # If the name is actually the same as the blank_field, it will seem as if there is no name
+        # The same can be said for all single fields
+        # The same issue happens with iterable fields and the blank_list
         describe_dict["name"] = properties.get("name")
 
         describe_dict["description"] = properties.get("description")
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -329,7 +329,7 @@ def describe(json_option, pipeline_path):
     _validate_pipeline_file_extension(pipeline_path)
 
     pipeline_definition = \
-        _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
+        _preprocess_pipeline(pipeline_path, runtime="None", runtime_config="None")
 
     indent_length = 4
     for current_pipeline in pipeline_definition["pipelines"]:
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -329,23 +329,34 @@ def describe(pipeline_path):
     pipeline_definition = \
         _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
 
-    click.echo("version:")
+    click.echo("Version:")
     for current_pipeline in pipeline_definition["pipelines"]:
         click.echo(current_pipeline["app_data"]["version"])
 
-    click.echo("description:")
+    click.echo()
+
+    click.echo("Description:")
     for current_pipeline in pipeline_definition["pipelines"]:
         click.echo(current_pipeline["app_data"]["properties"]["description"])
 
-    click.echo("type:")
+    click.echo()
+
+    click.echo("Type:")
     for current_pipeline in pipeline_definition["pipelines"]:
         click.echo(current_pipeline["app_data"]["properties"]["runtime"])
 
-    click.echo("dependencies:")
+    click.echo()
+
     for current_pipeline in pipeline_definition["pipelines"]:
         for node in current_pipeline["nodes"]:
-            for dependency in node["app_data"]["component_parameters"]["dependencies"]:
-                click.echo(dependency)
+            if len(node["app_data"]["component_parameters"]["dependencies"]) != 0:
+                click.echo("Dependencies:")
+                for dependency in node["app_data"]["component_parameters"]["dependencies"]:
+                    click.echo(dependency)
+            else:
+                pass
+
+    click.echo()
 
     print_banner("Elyra Pipeline Description Complete")
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -332,7 +332,7 @@ def describe(json_option, pipeline_path):
 
     blank_field = "Not Specified"
 
-    blank_list = "None Listed"
+    blank_list = ["None Listed"]
     for current_pipeline in pipeline_definition["pipelines"]:
         pipeline_keys = ["name", "description", "type", "version", "nodes", "file dependencies"]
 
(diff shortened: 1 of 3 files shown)


kevin-bates:
Hi @19690ao - thank you for contributing this pull request.  Since this is still in draft state and in progress, I thought I'd just give a couple of brief comments that might help save time later and/or start a discussion.

1. We should include the name of the pipeline - which, in your example, would come from the `name:` field with a value of `"hello-generic-world"`.
2. I believe the order of information should be, `name`, `description`, `type`, `version`, and whatever else. 
3. I think it would be useful to have, at a minimum, a node count, but we could also dump the node details (perhaps as an option)
4. Personally, I'd like to see the values on the same lines as their field name - when scalar (single) values are present.  List-based fields (like dependencies, nodes, etc.) would need some indentation.

This would then yield something like the following:
```
Name: hello-generic-world
Description: A generic pipeline tutorial
Type: Generic
Version: 4
Nodes: 4
Dependencies: 
    None
```
Was the idea with dependencies that they would essentially represent the _accumulated_ set of dependencies across all nodes?  I think if we offered a _details view_ we could just include dependencies (along with the other per-node attributes) as part of that view.

One approach that might be interesting to try is to use an [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict) where you add items in the order you want them printed, then use `json.dumps(dict)` to display the results rather than print as you go.  That way would could "build and dump" - which would likely wind up being more maintainable since the prints are scattered across the gathering logic.

lresende:
When using this [pipeline](https://github.com/lresende/elyra-sample/blob/master/producer-consumer/pipelines/producer-consumer.pipeline) for test I am getting the following:
```
 elyra-pipeline describe ~/opensource/jupyter-notebooks/elyra-pipelines-examples/producer-consumer/pipelines/producer-consumer.pipeline 
Version:
  4
  File "/Users/lresende/opt/anaconda/envs/dev/bin/elyra-pipeline", line 8, in <module>
    sys.exit(pipeline())
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/lresende/opt/anaconda/envs/dev/lib/python3.7/site-packages/elyra/cli/pipeline_app.py", line 328, in describe
    if current_pipeline["app_data"]["properties"]["description"] != "":
KeyError: 'description'
```

You probably should use `get('description') ` which will return an empty string if the property is not there. Also, just checking for the given string (e.g. if current_pipeline["app_data"]["properties"]["description"]) will check for empty string and none giving you a little more safe guard.

lresende:
For the output:
`elyra-pipeline describe ~/opensource/jupyter-notebooks/elyra-pipelines-examples/producer-consumer/pipelines/producer-consumer.pipeline`

```
Version:
  4
Description:
  Foo
Type:
  Generic
Dependencies:
  setup.txt
```

I would try to use less space, and one suggestion would be:
```
Type: Generic
Version:  4
Description:
  Foo
Dependencies:
  setup.txt
```

lresende:
@ptitzler for the `--json` option, do we want to use that to display the full pipeline or have something similar to the `human readable` format but in JSON format? If the latter, maybe we could add a `--print` or something to actually display the full pipeline? not sure as one can easily see the `file contents in `json` by catting the pipeline file.

elyra/cli/pipeline_app.py
```diff
@@ -258,5 +258,61 @@ def run(pipeline_path):
...
+
+        click.echo("Type:")
+        for current_pipeline in pipeline_definition["pipelines"]:
+            click.echo("  " + str(current_pipeline["app_data"]["properties"]["runtime"]))
+
+        has_dependencies = False
+
+        for current_pipeline in pipeline_definition["pipelines"]:
+            for node in current_pipeline["nodes"]:
+                if len(node["app_data"]["component_parameters"]["dependencies"]) >= 1:
```
lresende, line 303: dependencies might not be there or be none, use get('dependencies')

elyra/cli/pipeline_app.py
```diff
@@ -258,5 +258,61 @@ def run(pipeline_path):
...
+                    if not has_dependencies:
+                        click.echo("Dependencies:")
+                        has_dependencies = True
+                    for dependency in node["app_data"]["component_parameters"]["dependencies"]:
+                        click.echo("  " + str(dependency))
+
+        if not has_dependencies:
```
lresende, line 310: If we are collecting `dependencies` from each node, should we also collect `outputs`?  And should it be grouped by node?

lresende:
@19690ao overall it looks very good, thanks for the PR. I found a couple of small issues mostly about safeguarding the read of JSON properties and had a couple of questions more towards @ptitzler. As for fixing the issues mentioned above, maybe we could add test cases to validate these won't regress.

elyra/cli/pipeline_app.py
kevin-bates: +1 to using `node["app_data"]["component_parameters"].get("dependencies")`.  I would also add that you really don't need to check that length is >= 1 since a zero-length dependency list won't get into the `for` loop anyway.  As a result, this `if` block could be replaced with:
```python
for dependency in node["app_data"]["component_parameters"].get("dependencies", []):
    click.echo("  " + str(dependency))
```

ptitzler:
> @ptitzler for the `--json` option, do we want to use that to display the full pipeline or have something similar to the `human readable` format but in JSON format? 

The latter. I've updated the original issue with a clarifying example output. The idea was that this would enable users to filter pipeline files. For example, by querying the `version` property, one could identify those pipelines that are backlevel and need to be migrated to the latest version. 

(A dump of the entire pipeline file content doesn't really enable the user to do anything new that cannot be done already.)

ptitzler:
Clarification: When I created the original enhancement request I was primarily thinking about pipeline properties, not node-level properties (such as component parameters or outputs). For example, displaying a list of nodes with their associated component types would the user understand which components the pipeline requires to be catalogued in the registry.

My main concern about dumping node-level properties is that the output becomes very quickly unreadable for humans.

elyra/cli/pipeline_app.py
ptitzler: see clarifying comment

elyra/cli/pipeline_app.py
```diff
@@ -300,5 +301,66 @@ def run(pipeline_path):
...
+        for node in current_pipeline["nodes"]:
+            if "dependencies" in node["app_data"]["component_parameters"]:
+                for dependency in node["app_data"]["component_parameters"].get("dependencies", []):
+                    if describe_dict.get("dependencies") is None:
+                        describe_dict["dependencies"] = []
+                    describe_dict["dependencies"].append(f"{dependency}")
+        if not json_option:
+            for key in pipeline_keys:
+                if key in list_keys:
+                    click.echo(f"{key.title()}")
```
kevin-bates, line 354: I think these should have a trailing `:` as well.
```suggestion
                    click.echo(f"{key.title()}:")
```

elyra/cli/pipeline_app.py
```diff
@@ -300,5 +301,66 @@ def run(pipeline_path):
...
+@click.argument('pipeline_path')
+def describe(json_option, pipeline_path):
+    """
+    Display pipeline description in a human readable format
+    """
+
+    _validate_pipeline_file_extension(pipeline_path)
+
+    pipeline_definition = \
+        _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
```
kevin-bates, line 319: Forcing a `local` runtime (and runtime_config) causes the preprocessing to fail on non-generic pipelines (i.e., pipelines that are specific to either `airflow` or `kfp`) because `local` doesn't match the runtime validation within the preprocessing and leads to this:
```
$ elyra-pipeline describe ~/notebooks/elyra/my_pipeline/my_kfp_only.pipeline 
  File "/opt/anaconda3/envs/elyra-dev/bin/elyra-pipeline", line 8, in <module>
    sys.exit(pipeline())
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/elyra/cli/pipeline_app.py", line 319, in describe
    _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
  File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/elyra/cli/pipeline_app.py", line 140, in _preprocess_pipeline
    runtime_description = primary_pipeline['app_data']['ui_data']['runtime']['display_name']
KeyError: 'runtime'
```
I would change those `runtime` and `runtime_config` parameters to `None` for now and this issue will need to be resolved in another PR.

There should be a way to discover the runtime name or separate that validation out of the preprocessing so that options like`describe` can just display what's in the pipeline.

I'm opening a separate issue for the validation.

cc: @lresende

kevin-bates (commented):
This is looking good - thank you.  Just had a few more comments and there's an issue with non-local (non-generic) pipelines that needs to get resolved.

elyra/cli/pipeline_app.py
kevin-bates: @19690ao - I've addressed the above by allowing the `runtime` and `runtime-config` parameters to be optional.  Please rebase your changes against the master branch and update your call to `_preprocess_pipeline()` to only take the pipeline filename.  You should find the runtime-config value via `pipeline_data['app_data'].get('runtime')`.  Note that this value may be None if the runtime is generic.  The actual runtime name should be located via `pipeline_data['app_data']['properties'].get('runtime')`.  This value should exist for all pipelines and will be one of 'Generic', 'Kubeflow Pipelines', and 'Apache Airflow'.

19690ao: @kevin-bates - Thanks. I've rebased my local branch. I'm still having an issue describing non-generic pipelines.

> ... update your call to `_preprocess_pipeline()` to only take the pipeline filename.

Currently, if I change that line to only take the `pipeline_path` as an argument, running a kfp file through the describe command is currently giving me this output:
```
Error: This pipeline requires an instance of Kubeflow Pipelines runtime configuration.
The specified configuration 'None' is for None runtime.  
```
I was under the impression that the changes made were meant to allow the _preprocess_pipeline function to work without the need of specifying the runtime/runtime_config. I'm not quite sure how to run the function at the moment.

kevin-bates: Ok - I'll need to poke around at this relative to your PR.  It looks like `_validate_pipeline_runtime()` needs to flip its preference and default to returning True and be strict with its definition of _invalid_.  I will come up with a definition of the function that you can use and include in your PR.  It's critical that `describe` not have to know about the runtime.

kevin-bates:
@19690ao - I've just created a pull request against your `description` branch which makes the necessary adjustments to runtime validation to tolerate cases when the target runtime value is not specified - such as `describe` operations.

karlaspuldaro:
We should probably also update the [Elyra CLI documentation](https://elyra.readthedocs.io/en/latest/user_guide/command-line-interface.html) to expose the new command.

lresende:
Other than docs? Is the functionality all here? Should we remove this from "WIP"?

ptitzler:
> We should probably also update the [Elyra CLI documentation](https://elyra.readthedocs.io/en/latest/user_guide/command-line-interface.html#working-with-pipelines) to expose the new command.


and https://elyra.readthedocs.io/en/latest/user_guide/pipelines.html (source: https://github.com/elyra-ai/elyra/blob/master/docs/source/user_guide/pipelines.md)

elyra/cli/pipeline_app.py
```diff
@@ -311,5 +310,69 @@ def run(pipeline_path):
...
+            for key in pipeline_keys:
+                if key in iter_keys:
+                    click.echo(f"{key.title()}:")
+                    if describe_dict.get(key, set()) == set():
+                        click.echo(f"{' ' * indent_length}None")
+                    else:
+                        for item in describe_dict.get(key, ["None"]):
+                            click.echo(f"{' ' * indent_length}{item}")
+                else:
+                    click.echo(f"{key.title()}: {describe_dict.get(key, 'None')}")
```
ptitzler, line 369: See above --> `None`

elyra/cli/pipeline_app.py
```diff
@@ -311,5 +310,69 @@ def run(pipeline_path):
...
+        else:
+            for key in iter_keys:
+                describe_dict[key] = list(describe_dict[key])
+            click.echo(json.dumps(describe_dict, indent=indent_length))
```
ptitzler, line 373: The JSON dump currently encodes all values as string, even though some of them are always numeric (e.g. `nodes` or `version`). 

```
$ elyra-pipeline describe no_comment.pipeline --json
{
    "name": "drop_even_rows",
    "description": "None",
    "type": "Kubeflow Pipelines",
    "version": "4",
    "nodes": "1",
    "dependencies": []
}
```

Also note that `None/not set` (the value of `description` in above example) cannot be programmatically distinguished from a  user provided value "`None`". Therefore it would be better to encode `None/not set` as an empty string `""` to make that distinction possible.

elyra/cli/pipeline_app.py
```diff
@@ -311,5 +310,69 @@ def run(pipeline_path):
...
+@click.command()
+@click.option('--json',
+              'json_option',
+              is_flag=True,
+              required=False,
+              help='Returns pipeline file in a machine-readable JSON format')
+@click.argument('pipeline_path')
+def describe(json_option, pipeline_path):
+    """
+    Display pipeline description in a human readable format
```
ptitzler, line 322: We should reword the help text to remove ambiguity (description != pipeline description) and remove the reference to the output format because multiple formats (text/JSON) are supported.

```suggestion
    Display pipeline summary
```

elyra/cli/pipeline_app.py
```diff
@@ -311,5 +310,69 @@ def run(pipeline_path):
...
+
+    indent_length = 4
+    for current_pipeline in pipeline_definition["pipelines"]:
+        pipeline_keys = ["name", "description", "type", "version", "nodes", "dependencies"]
+
+        iter_keys = {"dependencies"}
+
+        describe_dict = OrderedDict()
+
+        properties = current_pipeline["app_data"]["properties"]
```
ptitzler, line 337: This code fails for empty pipelines (pipelines without a node):

```
$ elyra-pipeline describe empty.pipeline 
  File "/opt/anaconda3/envs/cli-describe/bin/elyra-pipeline", line 8, in <module>
    sys.exit(pipeline())
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/anaconda3/envs/cli-describe/lib/python3.7/site-packages/elyra/cli/pipeline_app.py", line 337, in describe
    properties = current_pipeline["app_data"]["properties"]
KeyError: 'properties'

```

ptitzler (changes requested):

elyra/cli/pipeline_app.py
```diff
@@ -311,5 +310,69 @@ def run(pipeline_path):
...
+
+        describe_dict["description"] = str(properties.get("description", "None"))
+
+        describe_dict["type"] = str(properties.get("runtime", "None"))
+
+        describe_dict["version"] = str(pipeline_data.get("version", "None"))
+
+        describe_dict["nodes"] = str(len(current_pipeline.get("nodes", [])))
+
+        describe_dict["dependencies"] = set()
```
ptitzler, line 353: We should clarify that the listed dependencies are only _file dependencies_.

```
$ elyra-pipeline describe dependency.pipeline 
Name: untitled
Description: None
Type: Kubeflow Pipelines
Version: 4
Nodes: 1
File dependencies:
    dependency.cfg
```

```
$ elyra-pipeline describe dependency.pipeline --json
{
    "name": "untitled",
    "description": "None",
    "type": "Kubeflow Pipelines",
    "version": "4",
    "nodes": "1",
    "file_dependencies": [
        "dependency.cfg"
    ]
}

```

elyra/cli/pipeline_app.py
19690ao: I'm currently testing empty pipelines using this pipeline with no nodes that recently got added to the `test_pipeline_app.py` file:
```
{
  "doc_type": "pipeline",
  "version": "3.0",
  "id": "0",
  "primary_pipeline": "1",
  "pipelines": [
    {
      "id": "1",
      "nodes": [],
      "app_data": {
        "runtime":"",
        "version": 4,
        "properties": {
          "name": "generic",
          "runtime": "Generic"
        }
      }, 
      "schemas": []
    }
  ]
}
```
This passes, so I'm not sure what an empty pipeline file actually looks like. If you could show me what the `empty.pipeline` you're testing with looks like, I shouldn't have a problem testing and resolving this on my own.

As a side note, if the CLI commands are expected to deal with a pipeline that's "more empty" than what the tests contain, the tests may be insufficient. All the current test pipelines in the `test_pipeline_app.py` do what I expect.

ptitzler: The best way to produce an empty pipeline is to create a `.pipeline` file using the Visual Pipeline Editor in JupyterLab. Fwiw I used the Airflow pipeline editor. If you save that pipeline without adding any nodes you should be able to reproduce the scenario that wasn't working for me. (The issue is recreatable using any empty pipeline of any type generic/kfp/airflow.)

> As a side note, if the CLI commands are expected to deal with a pipeline that's "more empty" than what the tests contain, the tests may be insufficient.

In general the expectation would be that the code gracefully handles common [problem] scenarios:
 - input is not a file
 - input file is not valid JSON
 - input file does not contain the expected property names that the code relies on to produce the output

19690ao: > * input file does not contain the expected property names that the code relies on to produce the output

That explains a lot. Thanks for the help. I just pushed out some changes that make it impossible to run into key errors as long as `_preprocess_pipeline(pipeline_path)` always returns a `dict`.

I'm still going to make that empty file and test for its formatting, but it shouldn't cause an error anymore.

elyra/cli/pipeline_app.py
19690ao: My most recent changes make it so that the new output for your example will be:
```
$ elyra-pipeline describe no_comment.pipeline --json
{
    "name": "drop_even_rows",,
    "type": "Kubeflow Pipelines",
    "version": "4",
    "nodes": "1",
    "dependencies": []
}
```
The JSON output won't even contain the keys that would be None/null as long as the original `.pipeline` didn't have the key either. This way there will be a difference in output between a pipeline with `""` as the description and the pipelines that have no `"description"` key. 
Does this sound good?

ptitzler: That would be another option. Two comments though:

```
"name": "drop_even_rows",,
```

Should not end with two `,,`

```
"dependencies": []
```

For the sake of consistency the same approach would have to be taken for all _optional_ properties, such as the file dependencies.

19690ao: > ```
> "name": "drop_even_rows",,
> ```
> 
> Should not end with two `,,`

Yeah, that was a typo. My bad. It should be one comma in the CLI.

I've just pushed the change that makes sure any empty iterable values (like the list of dependencies) won't be in the JSON output.

elyra/cli/pipeline_app.py
kevin-bates: Please note that I've opened #2066 to remove this "empty pipeline ambiguity" we're seeing.  This particular test artifact should be updated (not necessary for this PR - although that might prevent/eliminate possible "defensive" coding techniques) to reflect an "explicitly-saved" pipeline.

lresende (approved):

kevin-bates:
I just noticed that `describe` is not validating the pipeline.  I figured `_preprocess_pipeline()` would be doing that.  I can understand wanting to still display results, even when the pipeline is invalid, but I think we should let folks know their pipeline is invalid when they ask for its description.

PR #2005, opened 2021-08-02
Add missing node label for single file pipeline submission

### LINKED ISSUE #2003: Validation of standalone generic component throws error on "Run as pipeline"
**Describe the issue**
Opening a Notebook, python script or R script and selecting `Run as pipeline` throws an error related to a missing label field in the pipeline JSON.

**To Reproduce**
1. Open a notebook or script
2. Click on `Run as pipeline`
3. Submit to any runtime
4. See error below

**Screenshots or log output**
<pre>
[E 2021-08-02 14:42:19.509 ServerApp] Uncaught exception POST /elyra/pipeline/schedule?1627933339498 (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/elyra/pipeline/schedule?1627933339498', version='HTTP/1.1', remote_ip='::1')
      File "/Users/kierstenstokes/miniconda3/envs/3.0-test/lib/python3.9/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/Users/kierstenstokes/miniconda3/envs/3.0-test/lib/python3.9/site-packages/elyra/pipeline/handlers.py", line 111, in post
      File "/Users/kierstenstokes/miniconda3/envs/3.0-test/lib/python3.9/site-packages/elyra/pipeline/validate.py", line 122, in validate
        await self._validate_node_properties(pipeline=pipeline, response=response, pipeline_runtime=pipeline_runtime,
      File "/Users/kierstenstokes/miniconda3/envs/3.0-test/lib/python3.9/site-packages/elyra/pipeline/validate.py", line 254, in _validate_node_properties
        node_label = node_data['ui_data'].get('label')
    KeyError: 'ui_data'
</pre>
</details>

Screen Shot 2021-08-02 at 3 41 10 PM

**Expected behavior**
No error

**Deployment information**
- Elyra version: 3.0.0rc3 (from pypi) and latest master

ajbozarth:
- Add node label to single pipeline submission template
- Update validation checks for single pipeline submission

Fixes #2003

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/validate.py
+++ b/elyra/pipeline/validate.py
@@ -276,7 +276,7 @@ class PipelineValidationManager(SingletonConfigurable):
                                                                   response=response)
 
                         # Check label against kfp naming standards
-                        if pipeline_runtime == 'kfp' and filename != node_label:
+                        if pipeline_runtime == 'kfp' and node_label and filename != node_label:
                             self._validate_label(node_id=node['id'], node_label=node_label, response=response)
                         if dependencies:
                             notebook_root_relative_path = os.path.dirname(filename)
--- a/elyra/pipeline/validate.py
+++ b/elyra/pipeline/validate.py
@@ -121,7 +121,8 @@ class PipelineValidationManager(SingletonConfigurable):
                                            pipeline_execution=pipeline_execution)
         await self._validate_node_properties(pipeline=pipeline, response=response, pipeline_runtime=pipeline_runtime,
                                              pipeline_execution=pipeline_execution)
-        self._validate_pipeline_graph(pipeline=pipeline, response=response)
+        if not self._is_standalone_submission(pipeline_json['pipelines'][0]['app_data']['source']):
+            self._validate_pipeline_graph(pipeline=pipeline, response=response)
 
         return response
 
@@ -240,6 +241,7 @@ class PipelineValidationManager(SingletonConfigurable):
         :param response: ValidationResponse containing the issue list to be updated
         """
         pipeline_json = json.loads(json.dumps(pipeline))
+        pipeline_source = pipeline_json['pipelines'][0]['app_data'].get('source')
 
         for single_pipeline in pipeline_json['pipelines']:
             node_list = single_pipeline['nodes']
@@ -249,7 +251,7 @@ class PipelineValidationManager(SingletonConfigurable):
             components = ComponentRegistry.to_canvas_palette(component_list, categories)
             for node in node_list:
                 if node['type'] == 'execution_node':
-                    if self._is_legacy_pipeline(pipeline):
+                    if self._is_legacy_pipeline(pipeline) and not self._is_standalone_submission(pipeline_source):
                         node_data = node['app_data']
                         node_label = node_data['ui_data'].get('label')
                     else:
@@ -647,3 +649,11 @@ class PipelineValidationManager(SingletonConfigurable):
             if parameter['parameter_ref'] == node_property:
                 return parameter['data']['required']
         return False
+
+    def _is_standalone_submission(self, pipeline_source: str):
+        """
+        Determine if the submission is standalone e.g. The Button
+        :param pipeline_source: the pipeline source file e.g. example.pipeline, example.py etc.
+        :return:
+        """
+        return os.path.splitext(pipeline_source)[1] in ['.ipynb', '.r', ".py"]
--- a/elyra/pipeline/validate.py
+++ b/elyra/pipeline/validate.py
@@ -115,13 +115,14 @@ class PipelineValidationManager(SingletonConfigurable):
 
         pipeline_execution = pipeline_json['pipelines'][0]['app_data'].get('runtime')  # local, kfp, airflow
         pipeline_runtime = self._get_runtime_schema(pipeline, response)
+        pipeline_source = pipeline_json['pipelines'][0]['app_data'].get('source')
 
         self._validate_pipeline_structure(pipeline=pipeline, response=response)
         await self._validate_compatibility(pipeline=pipeline, response=response, pipeline_runtime=pipeline_runtime,
                                            pipeline_execution=pipeline_execution)
         await self._validate_node_properties(pipeline=pipeline, response=response, pipeline_runtime=pipeline_runtime,
                                              pipeline_execution=pipeline_execution)
-        if not self._is_standalone_submission(pipeline_json['pipelines'][0]['app_data']['source']):
+        if not self._is_standalone_submission(pipeline_source):
             self._validate_pipeline_graph(pipeline=pipeline, response=response)
 
         return response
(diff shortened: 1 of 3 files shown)


akchinSTC:
Tested against notebook, R and Python script "standalone button" and passing
image

elyra/pipeline/validate.py
```diff
@@ -249,7 +251,7 @@ async def _validate_node_properties(self, pipeline: dict, pipeline_runtime: str,
             components = ComponentRegistry.to_canvas_palette(component_list, categories)
             for node in node_list:
                 if node['type'] == 'execution_node':
-                    if self._is_legacy_pipeline(pipeline):
+                    if self._is_legacy_pipeline(pipeline) and not self._is_standalone_submission(pipeline_source):
```
kevin-bates, line 254: Before this change, only a "legacy pipeline" could get into this block.  Now a legacy pipeline can fall into the else block if it happens to be a "standalone submission" - which seems a bit odd, but I guess that's the issue being resolved, although this _also_ changes where the `node_data` is derived relative to standalone submissions.  Was that also an issue?

Might it be less confusing to determine `node_label`, for all invocation "styles", by first looking at `['ui_data']['labe'l']` (and tolerating `ui_data`'s absence, then (and only then) fallback to `['app_data']['label']` since `ui_data` is the first source of truth and will not be present in standalone submissions.

```python
if node_data['ui_data']:  # If present, always the source of truth
    node_label = node_data['ui_data'].get('label')
else:
    node_label = node['app_data'].get('label')
```
Also, shouldn't the label always have a value and should that be checked here (or soon) as part of the validation?

elyra/pipeline/validate.py
```diff
@@ -276,7 +278,7 @@ async def _validate_node_properties(self, pipeline: dict, pipeline_runtime: str,
                                                                   response=response)
 
                         # Check label against kfp naming standards
-                        if pipeline_runtime == 'kfp' and filename != node_label:
+                        if pipeline_runtime == 'kfp' and node_label and filename != node_label:
```
kevin-bates, line 281: Per above, it seems like `node_label` should be a required value, especially since it's used in all the validation messaging that happens prior to this point.

elyra/pipeline/validate.py
akchinSTC: thanks for the review, I will make the updates shortly.
Need to better organize the scenarios..=|

elyra/pipeline/validate.py
akchinSTC: yep, this is going to be removed. thx

elyra/pipeline/validate.py
akchinSTC: moving the standalone submission to its own block. 
adding the label logic to the legacy conditional block...weird thing...in a migrated pipeline from 3->4, are we allowed to have two label fields in the node? 
image

elyra/pipeline/validate.py
```diff
@@ -115,13 +115,15 @@ async def validate(self, pipeline: Dict) -> ValidationResponse:
 
         pipeline_execution = pipeline_json['pipelines'][0]['app_data'].get('runtime')  # local, kfp, airflow
         pipeline_runtime = self._get_runtime_schema(pipeline, response)
+        pipeline_source = pipeline_json['pipelines'][0]['app_data'].get('source')
```
lresende, line 118: Do we still have/use source? Wasn't this changed as part of https://github.com/elyra-ai/elyra/issues/1773?

lresende:
Maybe it is still too early and I need more coffee, but why these changes are on the validation service? Shouldn't a pipeline always be validated going through the same code block and not be dependent on where or how it has been submitted? Isn't this a matter of adjusting the template used to submit the "single-node" pipeline?

elyra/pipeline/validate.py
kevin-bates: Yes, this is possible.  This is why `['app_data']['ui_data']['label']` is the _first_ source of truth as that will always contain the correct value - when it exists.  However, there are some circumstances (like "standalone submission") when it won't exist, in which case `['app_data']['label']` should be used.  My comment was to not look at the "submission style" and always reach for `['app_data']['ui_data']['label']` while tolerating its absence.

elyra/pipeline/validate.py
```diff
@@ -656,4 +663,6 @@ def _is_standalone_submission(self, pipeline_source: str):
         :param pipeline_source: the pipeline source file e.g. example.pipeline, example.py etc.
         :return:
         """
-        return os.path.splitext(pipeline_source)[1] in ['.ipynb', '.r', ".py"]
+        if pipeline_source:
+            return os.path.splitext(pipeline_source)[1] in ['.ipynb', '.r', ".py"]
```
kevin-bates, line 667: Please use consistent quotes.

elyra/pipeline/validate.py
```diff
@@ -251,9 +252,15 @@ async def _validate_node_properties(self, pipeline: dict, pipeline_runtime: str,
...
             for node in node_list:
                 if node['type'] == 'execution_node':
-                    if self._is_legacy_pipeline(pipeline) and not self._is_standalone_submission(pipeline_source):
+                    if self._is_standalone_submission(pipeline_source):
+                        node_data = node['app_data'].get('component_parameters')
+                        node_label = node['app_data'].get('label')
+                    elif self._is_legacy_pipeline(pipeline):
                         node_data = node['app_data']
-                        node_label = node_data['ui_data'].get('label')
+                        if node_data.get('ui_data'):  # If present, always the source of truth
```
kevin-bates, line 260: I guess I figured this block should be used unconditionally since it's always correct irrespective of submission style.  Then only `node_data` would vary (which is unfortunate - see next).  

To @lresende's point, would updating the "template" for standalone submission to build `['app_data']['component_parameters']` then allow all of this to be agnostic as to how a pipeline's content got here?

elyra/pipeline/validate.py
akchinSTC: as of today, the payload that I receive still contains source
image

elyra/pipeline/validate.py
akchinSTC: I would be all for consistency in updating the template to align better with how full pipelines are being submitted

elyra/pipeline/validate.py
kevin-bates: `source` seems useful.  Is the pipeline filename present under any other field?

kevin-bates:
These changes look good.  I still see the `source` field accessed in the top-level `validate()` method in order to conditionally call `_validate_pipeline_graph()`.  However, in looking at that method, it appears the "orphaned node" warning is only produced when there are multiple nodes in the pipeline - which shouldn't ever be the case with standalone submissions.  

If `_validate_pipeline_graph()` can be called on standalone pipelines, then both the access of the `source` field and the `_is_standalone_submission()` method could be removed leaving all pipeline submission "styles" as equivalent wrt to validation.

akchinSTC:
Updated the circular checks to account for standalone submissions and removed the `is_standalone` function

akchinSTC (approved):

kevin-bates:
Thanks for the updates Alan - the last set of changes look good.

PR #2007, opened 2021-08-03
Remove slack operator base class definition from palette

akchinSTC:
The base class definition is not `usable` and is not needed when
constructing an Airflow DAG in the visual editor and should be
removed

           in the file; or

           it.

### DIFF
--- a/etc/config/components/airflow/slack_operator.py
+++ b/etc/config/components/airflow/slack_operator.py
@@ -20,10 +20,8 @@
 
 import json
 
-from airflow.models import BaseOperator
+from airflow.operators.slack_operator import SlackAPIOperator
 from airflow.utils.decorators import apply_defaults
-from airflow.hooks.slack_hook import SlackHook
-from airflow.exceptions import AirflowException
 
 
 class SlackAPIPostOperator(SlackAPIOperator):
--- a/etc/config/components/airflow/slack_operator.py
+++ b/etc/config/components/airflow/slack_operator.py
@@ -26,66 +26,6 @@ from airflow.hooks.slack_hook import SlackHook
 from airflow.exceptions import AirflowException
 
 
-class SlackAPIOperator(BaseOperator):
-    """
-    Base Slack Operator
-    The SlackAPIPostOperator is derived from this operator.
-    In the future additional Slack API Operators will be derived from this class as well
-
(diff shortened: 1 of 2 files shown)


etc/config/components/airflow/slack_operator.py
```diff
@@ -26,66 +26,6 @@
 from airflow.exceptions import AirflowException
 
 
-class SlackAPIOperator(BaseOperator):
```
kevin-bates, line 29: I'm not sure how/when these classes get instantiated, but it seems that if the `BaseOperator` needed to be imported previously, then `SlackAPIOperator`'s import would be necessary for this change.  Looks like `SlackHook` and `AirflowException` imports can also be removed.

akchinSTC: These classes arent instantiated locally, if I understand correctly, more being used as a source of information/data to populate the component palette properties and to construct the DAG using the airflow template. As long as the operators are defined correctly on the deployment side, where they are instantiated, a local change here should be ok

lresende: Are you saying that, any customizations we make locally are lost because the server side installed component is used instead? Can we then actually just query which components are available in the server?

akchinSTC: Sort of.
We can make for lack of a better term `cosmetic` modifications like this removal because all we are interested in right now are the `arguments` needed to make the operator function. This in turn is the next item in question, that the `arguments` we are interested in, HAVE to be aligned and the same as the operator being used in the deployment.  So any "customizations" will need to fit within the confines of the source operator class definition in the airflow deployment, or included in the DAG itself.

lresende: yeah, that is sort of thing we need to further discuss/investigate in the upcoming releases.

lresende (approved):

etc/config/components/airflow/slack_operator.py
kevin-bates: I still feel, for the sake of correctness and good example, these _cosmetic_ modifications should compile if their original incantations compile.

akchinSTC: opened two issues for this #2008 and #2009

kevin-bates (approved):
LGTM

PR #2034, opened 2021-08-12
Proposal to relax kfp sdk dependency

### LINKED ISSUE #2033: Kfp depedency is too strict
**Describe the issue**
Hi folks, kfp dependency is currently pinned to 1 single dependency which makes hard to perform upgrades when working on shared environments between kfp and elyra pipelines. Would it be possible to relax the depdency instead to take a range to accept semver valid versions of kfp sdk as they should be forward compatible?

This is blocking us from being able to easily upgrade kfp in our monorepo.

casassg:

Relax kfp requirement in elyra releases as described in #2033

Change is trivial and should be noop until there's a new kfp sdk release.

May be worth considering how often to bump kfp top constraint as well.

Closes #2033

- [x]  Manual Testing with kfp sdk 1.7.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup_args = dict(
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp==1.6.3',
+        'kfp>=1.6.3<1.7',
         'kfp-tekton==0.8.1',
         # Airflow runtime dependencies
         'pygithub',
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup_args = dict(
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp>=1.6.3<1.7',
+        'kfp~=1.6.3',
         'kfp-tekton==0.8.1',
         # Airflow runtime dependencies
         'pygithub',
--- a/setup.py
+++ b/setup.py
@@ -87,8 +87,8 @@ setup_args = dict(
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp~=1.6.3',
-        'kfp-tekton==0.8.1',
+        'kfp>=1.6.3<2.0',
+        'kfp-tekton~=0.8.1',
         # Airflow runtime dependencies
         'pygithub',
         'black'


lresende:
@tomlyn I remember there was some correlation between the KFP and Tekton versions, Is the range being proposed here ok?

casassg:
Seems so - https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/setup.py#L57

That said, pip should resolve this nicely. I would recommend making the dep less strict here.

Also, is there a way we can move kfp-tekton to an extra instead of install_requires?

lresende:
Ideally, if we move `kfp-tekton` out of the required dependency, we would need to better handle the possible runtime-related issues such as a user trying to submit a pipeline to a tekton runtime, but the package is not installed.

casassg:
I would think it should not be an issue, extras is exactly for those cases, where some stuff may break if you try to use X feature but dont install X corresponding extra. Adding correct documentation with this explaining that you need extra tekton for those times you want to use kfp on tekton should probs work. It's a backwards incompatible change so I understand the reticency.

However, this would make it easier to consume elyra specially for those who are not using kfp-tekton. As we can upgrade kfp to newer version as it's a more stable package with garanteed forward compatible API.

Having all those deps in a single package makes it hard to use in monorepos or large requirements environments. Airflow, Beam, and others splitted it up similarly for similar reasons.

casassg:
#2036

Does this work?

setup.py
```diff
@@ -87,7 +87,7 @@
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp==1.6.3',
+        'kfp>=1.6.3<1.7',
```
kevin-bates, line 90: I'd prefer we use the `~` operator for this, which would be equivalent...
```suggestion
        'kfp~=1.6.3',
```

Tomcli:
> @Tomcli I remember there was some correlation between the KFP and Tekton versions, Is the range being proposed here ok?

This range should be mostly fine, we have this strict kfp version for kfp-tekton because there were new bugs introduced back in the kfp 1.3.0 release when using func_to_container_op. So we decided to force users to install the tested stable version to minimize the possible bugs that might be introduced in the newer releases.

kevin-bates (commented):
I agree with this change.  In fact, I'd like to see us convert **all** _three-part_ `==` requirements to `~=` so that we pick up security and bug fixes by default.  Yes, there could be instances where a patch fix breaks downstream dependencies, but that's a bug on the upstream component and should get resolved quickly.  Having the ability to capture security fixes, however, feels more important to me.

casassg:
We should try to relax a bit to allow users to upgrade KFP between minor releases when possible here as well, but can be done as a separate commit after KFP releases 1.7

Tomcli:
Just FYI, KFP SDK releases not longer follow their github releases. So they already have the kfp 1.7.0 sdk release despite their github is still having pre-release.

setup.py
```diff
@@ -87,7 +87,7 @@
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp>=1.6.3<1.7',
+        'kfp~=1.6.3',
         'kfp-tekton==0.8.1',
```
kevin-bates, line 91: Should we do the same with `kfp-tekton`? (`~=0.8.1`). That's the only remaining `==` requirement we have.

casassg:
Damn, well, if the changes are not backwards incompatible (aka Elyra doesnt use any new method introduced by KFP 1.7) we should be good to add it in the accepted range

ptitzler:
> (aka Elyra doesnt use any new method introduced by KFP 1.7) 
Not yet.

> we should be good to add it in the accepted range
I would agree _if_ 1.7.0 passes a smoke test.

casassg:
Would running the tests using GitHub actions be enough? If so, I can solve run it quickly :D (just need to upgrade it on the setup.py I imagine)

lresende:
What are the next steps here? Are we relaxing the kfp-tekton dependency and merging this while we discuss other changes that would come in other prs? or are we waiting for a 1.7 discussion/smoke test before merging this?

kevin-bates:
> Would running the tests using GitHub actions be enough?

The tests currently do not exercise the kfp API, so `1.7` support should be manually verified at this time.

> What are the next steps here?

We should make the appropriate updates that would include the tolerance of minor `kfp` versions (assuming manual testing via 1.7 is successful) and go ahead and relax the `kfp-tekton` version to tolerate patches and security fixes as suggested here:  https://github.com/elyra-ai/elyra/pull/2034#discussion_r687907376

Since it sounds like we're going to tolerate minor kfp versions, I guess we'd want `'kfp>=1.6.3,<2.0'` so as to still prevent major release inclusion.  Else we could remove the cap altogether.

Does that sound right?

ptitzler:
> The tests currently do not exercise the kfp API, so 1.7 support should be manually verified at this time.

I confirmed that the example pipelines work as expected using `kfp==1.7.0` and `kfp-tekton==0.8.1`. I've also inspected the change log for that kfp release and didn't notice anything related to features we are using.

It should therefore be okay to go with

```
'kfp-tekton~=0.8.1'
'kfp>=1.6.3,<2.0'
```

setup.py
```diff
@@ -87,7 +87,7 @@
         'websocket-client',
         'yaspin',
         # KFP runtime dependencies
-        'kfp==1.6.3',
+        'kfp~=1.6.3',
         'kfp-tekton==0.8.1',
```
casassg, line 91:
```suggestion
        'kfp>=1.6.3<2.0',
        'kfp-tekton~=0.8.1',
```

casassg:
Updated! Thanks for testing!

kevin-bates (approved):
Thanks @casassg!

lresende (approved):

PR #2038, opened 2021-08-12
Standardize processing for custom component parameters of type dictionary and list

### LINKED ISSUE #2020: airflow/email_operator triggers "malformed node or string" error during compilation
**Describe the issue**

image

**To Reproduce**
1. Install Elyra 3.0
2. Launch JupyterLab
3. Create AA pipeline
4. Add and configure email operator
4. Try to run the pipeline. See error

**Log output**
<pre>
[E 2021-08-06 12:16:23.700 ServerApp] Uncaught exception POST /elyra/pipeline/schedule?1628277383647 (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='POST', uri='/elyra/pipeline/schedule?1628277383647', version='HTTP/1.1', remote_ip='::1')
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/elyra/pipeline/handlers.py", line 119, in post
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/elyra/pipeline/processor.py", line 119, in process
        res = await asyncio.get_event_loop().run_in_executor(None, processor.process, pipeline)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/concurrent/futures/thread.py", line 57, in run
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/elyra/pipeline/processor_airflow.py", line 81, in process
        pipeline_filepath = self.create_pipeline_file(pipeline=pipeline,
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/elyra/pipeline/processor_airflow.py", line 318, in create_pipeline_file
        target_ops = self._cc_pipeline(pipeline, pipeline_name)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/site-packages/elyra/pipeline/processor_airflow.py", line 261, in _cc_pipeline
        operation.component_params[component_property.ref] = ast.literal_eval(op_property)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/ast.py", line 99, in literal_eval
        return _convert(node_or_string)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/ast.py", line 98, in _convert
        return _convert_signed_num(node)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/ast.py", line 75, in _convert_signed_num
        return _convert_num(node)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/ast.py", line 66, in _convert_num
        _raise_malformed_node(node)
      File "/opt/anaconda3/envs/ga_30/lib/python3.8/ast.py", line 63, in _raise_malformed_node
        raise ValueError(f'malformed node or string: {node!r}')
    ValueError: malformed node or string: <_ast.BinOp object at 0x7f865b947fd0>
</pre>
</details>

**Deployment information**
 - Elyra version: 3.0
 - Installation source: PyPI

kiersten-stokes:
Fixes #2020 

The following changes have been made:

- `_process_list_value()` added to `RuntimePipelineProcessor` to take component parameters of standardized type `list` and convert the user-entered string value to a python `list` data structure
  - ~user can now enter a comma- or semicolon-delimited list of strings (or a singular value)~
  - ~e.g., a user can enter the following value into the `to` parameter field of the `EmailOperator`: `email1@gmail.com, email2@gmail.com`, and it will be converted appropriately to `["email1@gmail.com", "email2@gmail.com"]` before rendering in the Airflow DAG template~
  - After discussion, only basic conversion is done: user-entered string value is stripped and checked for square brackets, then `ast.literal_eval()` is attempted
    - If `literal_eval` fails, the stripped user-entered string is sent as the value for component execution
- `_process_dictionary_value()` added to take component parameters of standardized type `dictionary` and convert the user-entered string value to a python `dict` data structure
  - ~this one is a bit trickier than the `list` conversion with more room for error, but it does cover some basic scenarios~
  - ~e.g., if a user doesn't put quotation marks around the keys of a dictionary input (`{key1: 2, key2: "value"}`) or if they don't include curly braces (`"key1": 2, "key2": "value"`), it will still be converted propertly to `{"key1": 2, "key2": "value"}`~
  -   After discussion, only basic conversion is done: user-entered string value is stripped and checked for curly braces, then `ast.literal_eval()` is attempted
       - If `literal_eval` fails, the stripped user-entered string is sent as the value for component execution

Both Airflow and KFP processors call these functions in the `_cc_pipeline` function during component parameter standardization. Because Airflow (`AirflowPipelineProcessor`) requires some additional processing for these values, it extends the `RuntimePipelineProcessor` class methods listed above. Airflow processors now call the superclass first, then include surrounding quotation marks on strings so that they render appropriately in the DAG.

Component parameters of type `number`, `boolean`, and  `string` do not require a conversion function because they are sent in the proper format in the pipeline JSON.
 
Tests were added to `test_processor_airflow.py` and `test_processor_kfp.py` to ensure the functions `_process_list_value()` and `_process_dictionary_value()` produce the correct output.
 
           in the file; or

           it.


elyra/pipeline/pipeline.py
```diff
@@ -169,6 +172,86 @@ def _scrub_list(dirty: Optional[List[Optional[str]]]) -> List[str]:
...
+                    val = val.strip(" '\"")
+                    if val.startswith('{') and val.endswith('}'):
+                        val = Operation.process_dictionary_value(val, logger=logger)
+                    elif val.startswith('[') and val.endswith(']'):
+                        val = Operation.process_list_value(val)
+
+                converted_dict[key.strip(" '\"")] = val
+
+        if not isinstance(converted_dict, dict):
+            Operation._log_info(f"Could not convert parameter value '{value}' to dictionary",
```
kiersten-stokes, line 222: I need to determine if this case is ever reached with some of the latest changes I made. If not, will need to remove this block, change the type hints, and remove the logic in the processors that sends the user-entered string to the appropriate constructor if a dictionary could not be formed.

kiersten-stokes:
There is some discussion here worth having about the balance between how much conversion to do for dictionaries vs. how much responsibility to put onto the user in entering the value as the component expects it.

Some options that I can see:

1. Insist that the user enters a dictionary in correct python and/or JSON format and throw an error if not
    - benefits: less conversion and room for error on our end
    - risks: this may stricter than it needs to be in some scenarios (e.g. if the component supports various input format, then we specifically are not allowing some of them)

2. Do as much conversion as possible before sending the dictionary value off to the appropriate runtime (this is currently where this PR stands, pending review)
    - benefits: more flexible for user
    - risks: if the user makes a syntax mistake that we don't catch (somewhat likely because the line has to be drawn somewhere), the pipeline may fail later on down the line

3. Some middle ground: have a minimal amount of conversion logic (e.g. _only_ attempt a `literal_eval` and a `json.loads`) and send the exact user-entered string if both fail

elyra/pipeline/pipeline.py
kevin-bates: I agree that it's impossible for `converted_dict` to not be a `dict` at this point by virtue of L197 and recommend the cleanup that is suggested.

I suppose you could, however, wind up with an empty dictionary when the input value is non-None and non-dict.  Perhaps that's the case in which a fallback to the input value should occur.  Hmm, in that case, you probably want to check the recursed call's return value against `None` so that you can break out of the `for` loop.

I also wonder if, in these situations, we just return the input value rather than `None` and change the return type to `Union[dict, str]` along with logging a **warning** (rather than info).  Of course, this would mean that the recursed return value would check for a string instance rather than `None` to determine if `break` is required...
```python
val = Operation.process_dictionary_value(val, logger=logger)
if isinstance(val, str):
    break
```

I think a test that takes a vanilla string value (that doesn't contain a colon) would be a good test.  Based on the above discussion, that test should return the input, and log a warning (although the latter can be difficult to verify).

kevin-bates (commented):
Hi @kiersten-stokes - this is a great PR - excellent write and tests!  I just had a suggestion or two regarding your question about the dictionary conversion.

Regarding how far we need to go with this, I think what you have here is more than sufficient and a quick look at the tests show what support can be expected.  Thank you!

ptitzler:
> There is some discussion worth having here about the balance between how much conversion to do for dictionaries 

I believe the same can be said about lists. Even though https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/email_operator.py accepts `elem1;elem2` as valid input (according to the embedded doc), the same might not be true for other components. In addition, another components might consider `elem1 elem2` (space separated) a valid list-type input, which would require an update to Elyra's preprocessing logic and render input that would otherwise be considered invalid, valid. Since there are likely many more special cases, we should consider 
 - whether a pass-through should be good enough
 - whether only explicitly typed parameters are enforced

kevin-bates:
Hmm - that's a good point.  Between those two choices and given the possibility that there could be any number of separators permitted (or excluded - e.g., semi-colon separation may not be warranted), perhaps blind _`ast.eval`-else-pass-through_ is the best thing we can do (and should do).

kiersten-stokes:
Based on discussion with @ptitzler and @kevin-bates, we will be moving forward with option number 3 from my comment above:
> Some middle ground: have a minimal amount of conversion logic (e.g. only attempt a literal_eval and a json.loads) and send the exact user-entered string if both fail

There are just too many variables and too much room for error to do anything else. We will strip the user-entered string, check if it is surrounded by square brackets for list or curly braces for dictionary and run `literal_eval` if so, and the fallback case will be to use the user-entered string. Any 'unusual' scenarios are therefore up to the user to enter correctly

elyra/pipeline/processor_airflow.py
```diff
@@ -236,29 +235,27 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
-                        # Get corresponding component_property value from parsed pipeline and convert
-                        if not op_property:
-                            op_property = "[]"
-                        operation.component_params[component_property.ref] = ast.literal_eval(op_property)
+                        # Add surrounding quotation marks to string value for correct rendering
+                        # in jinja DAG template
+                        operation.component_params[component_property.ref] = json.dumps(property_value)
+                    elif component_property.type == 'dictionary':
+                        processed_value = operation.process_dictionary_value(property_value, logger=self.log)
+                        operation.component_params[component_property.ref] = processed_value
```
kiersten-stokes, line 255: Now that we may end up returning a string value, I'll have to add some additional logic here to put surrounding quotation marks around strings such that they are rendered properly in the Airflow DAG

kevin-bates: Interesting.  If that's the case (i.e., the need for runtime-specific differences), then we should probably move these kinds of methods from `Operation` to `RuntimePipelineProcessor` and let the airflow processor first call the superclass, then apply any additional quotations.  (Assuming I'm understanding the context)

kiersten-stokes: Ah that is a good point! Let me just double check that there isn't anything I'm overlooking, and if not, I'll make that change

kevin-bates (approved):
Looks good @kiersten-stokes - thank you.

PR #2043, opened 2021-08-13
Move kfp-tekton to extras

### LINKED ISSUE #2036: Improve requirements for elyra wheel

As a user, I want to use Elyra w other systems and dependencies. However, currently Elyra adds a lot of dependencies even when I don't necessarily need them to use the parts that I need. This makes it hard to use Elyra as I quite often may run into dependency conflicts.

This is made even worse in the case of monorepos where third party management must be done in a lock step release. As users who want to use Elyra will get stuck in old versions of dependencies, unless Elyra is upgraded.

Break up install_requires into multiple extra _require instead so that as a user, I can explicitly say what part of Elyra I need to be installed w the wheel.

casassg:
Move kfp-tekton to extras to avoid locking kfp upgrades when tekton does not needed to be used.

Fixes #2036


elyra/kfp/operator.py
```diff
@@ -254,7 +254,12 @@ def __init__(self,
             run_name_placeholder = '{{workflow.annotations.pipelines.kubeflow.org/run_name}}'
             self.container.add_env_variable(V1EnvVar(name='ELYRA_RUN_NAME',
                                                      value=run_name_placeholder))
-        else:
+        elif workflow_engine and workflow_engine.lower() == 'tekton':
```
kevin-bates, line 257: This all looks good.  Just wondering if we should precede this `if` block with `if not workflow_engine:` and raise a `ValueError` that it's required? This will clean up the `if/elif` statements.  Then add an `else` with another `ValueError` that an unsupported workflow engine was detected?

ptitzler:
To also support this in the UI - namely preventing users from creating (or selecting) a runtime configuration for an engine that is not available - perhaps we should also add a method that returns the available kfp engine types?  (Front end work that takes advantage of this could be delivered in a separate PR.)

kevin-bates:
> perhaps we should also add a method that returns the available kfp engine types

Since this is an enum that is currently expressed in the schema, I'd rather address this using an approach similar to [metadata instance classes](https://github.com/elyra-ai/elyra/pull/725) (i.e., _schema instance classes_), rather than add an entire endpoint that essentially says "the schema in your hand right now is wrong, use this one-off approach to fix it".  

The schema would also include a `schema_instance_class` entry as a potential sibling to its [`metadata_instance_class`](https://github.com/elyra-ai/elyra/blob/71cfb3fe77110a77325ab7211ccd71576e92ff7a/elyra/metadata/schemas/metadata-test.json#L7) such that when the `SchemaManager` loads the schema, it will give that class instance a chance to alter the schema via its `post_load()` method.  Since schemas are currently read-only, we don't need to implement `pre_save()` and `pre_delete()` ... for now. However, this should be viewed as a step forward towards _bring your own schema_.

Here's the current schema for `engine`:

```JSON
        "engine": {
          "title": "Kubeflow Pipelines engine",
          "description": "The Kubeflow Pipelines engine in use ('Argo' or 'Tekton')",
          "type": "string",
          "enum": ["Argo", "Tekton"],
          "default": "Argo",
          "uihints": {
            "field_type": "dropdown",
            "category": "Kubeflow Pipelines"
          }
        },
```

It should be noted that, in addition to an update of the `enum` attribute, the `default` attribute is also a candidate for update and we should probably relax the `description` to not include the _current_ options.

If we took this approach, zero changes are necessary on the frontend since it will always be handed accurate schema.

ptitzler:
> [...] rather address this using an approach similar to metadata instance classes [...]

Sounds good. I opened https://github.com/elyra-ai/elyra/issues/2046 for tracking

elyra/pipeline/processor_kfp.py
```diff
@@ -34,11 +34,17 @@
 from kfp import components as components
 from kfp.aws import use_aws_secret  # noqa H306
 from kfp_server_api.exceptions import ApiException
-from kfp_tekton import compiler as kfp_tekton_compiler
-from kfp_tekton import TektonClient
 import requests
 from urllib3.exceptions import LocationValueError
 from urllib3.exceptions import MaxRetryError
+try
```
kevin-bates, line 40:
```suggestion
try:
```

kevin-bates:
@casassg - could you please make the required updates and remove this from WIP?  #2062 addresses the UI concerns raised by @ptitzler in https://github.com/elyra-ai/elyra/pull/2043#issuecomment-899532878 so I think we should be good to go.

I needed to apply your "import logic" in `processor_kfp.py` in #2062, so one of us will likely need to resolve a merge conflict or two.

Thanks.

casassg:
Had some troubles finding a way for make to continue picking up the right wheel. I think we are good now! @kevin-bates lmk if there's anything else!

kevin-bates (approved):
LGTM - thanks @casassg.

If we add any other `extra_requires` entries that are runtime related (i.e., not test or docs), we'll likely want an `all` entry that comprises the others.  But that isn't necessary until the next one. :smile:

casassg:
Added it since it was easy enough to add 😄

setup.py
```diff
@@ -39,6 +39,11 @@
 metadata_path = './etc/config/metadata/runtime-images/*.json'
 settings_path = './etc/config/settings/*.json'
 
+extra_requires = {
+    'kfp-tekton': ['kfp-tekton~=0.8.1',]
+}
+extra_requires['all'] = list(set(sum(extra_requires.values(), [])))
```
kevin-bates, line 45: Sorry, I had a typo in my "approval" comment and find `extra_requires` confusing given the _proper_ name is `extras_require` and used as well.  As a result, I think it might be better to name this something like `runtime_extras` to convey these would only apply to runtime packages.

I do like the way it accumulates the entries and adds `'all'` though.

kevin-bates (approved):

kevin-bates (approved):
Thank you for the update!

ptitzler (approved):
LGTM; works as expected

Issue #1924, opened 2021-07-15
Multiple dialogs sometime open when migrating a pipeline


ajbozarth:
**Describe the issue**
Sometime the migrate pipeline dialog is opened multiple times. This is a UX issue only since the functionality is still only run once, the extra dialogs only confuse the user and don't affect the pipeline migration.

**To Reproduce**
1. Open a pipeline created with an older pipeline version
2. The migrate pipeline dialog will open, click a button 
3. A migrate pipeline dialog will still be open. If you selected migrate button the migration will have happened behind the dialog. If you selected cancel then the pipeline will already be closed. 
4. If you click migrate again nothing happens, but if you click cancel it will try to close the pipeline and you would be prompted to save.

**Expected behavior**
The migrate dialog should only show once and no second dialog should be seen once a choice is made

ajbozarth:
Afaik the `onError` function that handles the migration prompt is being called multiple times in `pipeline-editor`. This is most likely because it is called within this `useEffect` https://github.com/elyra-ai/pipeline-editor/blob/f1360ee09f424cf13ca1609ff67663468e01a5ee/packages/pipeline-editor/src/PipelineEditor/index.tsx#L211 which is most likely being triggered more than once.

karlaspuldaro:
Would this be related to #863 ?

bourdakos1:
You need to wrap `onError` in `useCallback` to memoize it, otherwise `onError` will be a new variable every render which can cause unnecessary `useEffect` re-runs

ajbozarth:
> You need to wrap onError in useCallback to memoize it, otherwise onError will be a new variable every render which can cause unnecessary useEffect re-runs

Ah that makes sense, I'll fix this once I'm done writing the tests for https://github.com/elyra-ai/pipeline-editor/pull/147

ajbozarth:
After some research and discussion with @bourdakos1 this is a symptom of a larger issue that will require a refactor of a lot of our hooks. So we will address this in 3.1 instead. Specifically the use of SWR is causing a lot of hooks to be run over and over when it isn't necessary, one symptom of which is that the update of runtimeImages from unknown to it's actual value via SWR updates pipeline which reruns open, which errors again and opens a migrate dialog even though one is already open.

kevin-bates:
> So we will address this in 3.1 instead. 

This seems significant.  I'm assuming you meant `3.0.1` based on the milestone update.

[...]

PR #2045, opened 2021-08-16
Block pipeline-editor-extension from showing more than one error dialogs at a time

bourdakos1:
I wasn't able to fix the rerending issues in time, but this should also fix the multiple migration dialogs problem 

fixes: #1924

 
           in the file; or

           it.


packages/pipeline-editor/src/PipelineEditorWidget.tsx
```diff
@@ -299,10 +299,15 @@ const PipelineWrapper: React.FC<IProps> = ({
...
   const onError = useCallback(
-    (error?: Error): void => {
+    async (error?: Error): Promise<void> => {
+      if (isDialogAlreadyShowing.current) {
+        return; // bail, we are already showing a dialog.
+      }
+      isDialogAlreadyShowing.current = true;
       if (error instanceof PipelineOutOfDateError) {
-        showDialog({
+        const result = await showDialog({
```
ajbozarth, line 310: Is there a reason you're removing the Promise then handling and replacing it with await result handling? It seems unnecessary for this boolean flag solution to this problem.

ajbozarth (commented):
I'll try this locally later today, but from a code perspective I'm mixed on this. I raised one specific issue in a comment below, but I already had considered this fix previously and though we had rejected it in a previous discussion with the team (or at least the front end side). If the refactor is going to take a long time I'm willing to let this solution go in to fix it in the sort term though

ajbozarth (approved):
LGTM, `isDialogAlreadyShowing` feels a little over-verbose of a var name though 😄

PR #2068, opened 2021-08-23
Fix mismatched runtime handling for run command

### LINKED ISSUE #2054: 'elyra-pipeline run' CLI command: improve incompatible pipeline error message
**Describe the issue**

If a user executes the `elyra-pipeline run` command and specifies an RTS pipeline file the following error output is produced:

```
$ elyra-pipeline run kfp.pipeline 

────────────────────────────────────────────────────────────────
 Elyra Pipeline Local Run
────────────────────────────────────────────────────────────────

Error: Invalid runtime configuration: local
 Invalid runtime configuration: local
 No such instance named 'local' was found in the runtimes namespace.

```

The error message isn't providing the user with the information needed to resolve the issue. The wording should be changed to indicate that this <runtime type> pipeline needs to be run using the `elyra-pipeline submit` command.

Example output for a similar user error scenario as produced by `elyra-pipeline submit`:

```
$ elyra-pipeline submit empty-airflow.pipeline --runtime-config my_kubeflow_test_instance

────────────────────────────────────────────────────────────────
 Elyra Pipeline Submission
────────────────────────────────────────────────────────────────

Error: This pipeline requires an instance of Apache Airflow runtime configuration.
The specified configuration 'my_kubeflow_test_instance' is for Kubeflow Pipelines runtime.
```

**Deployment information**
 - Elyra version: 3.0/master

ptitzler:

Fixes #2054

Verified that the `elyra-pipeline run` command returns the expected results:

```
$ elyra-pipeline run generic.pipeline  # correct
   ...
   Elyra Pipeline Local Run Complete

$ elyra-pipeline run kfp.pipeline  # incorrect invocation
   ...
   Error: This pipeline requires an instance of Kubeflow Pipelines runtime configuration. The provided configuration 'local' is for local runtime.

$ elyra-pipeline run airflow.pipeline # incorrect invocation
   ...
   Error: This pipeline requires an instance of Apache Airflow runtime configuration. The provided configuration 'local' is for local runtime.
```

Confirmed that `describe` and `submit` commands still behave as expected.

 
           in the file; or

           it.

### DIFF
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -160,8 +160,6 @@ def _preprocess_pipeline(pipeline_path: str,
         primary_pipeline["app_data"]["runtime"] = runtime
     if runtime_config:
         primary_pipeline["app_data"]["runtime-config"] = runtime_config
-    else:
-        primary_pipeline["app_data"]["runtime-config"] = 'local'
 
     return pipeline_definition
 
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -142,10 +142,11 @@ def _preprocess_pipeline(pipeline_path: str,
 
     if not _validate_pipeline_runtime(primary_pipeline, runtime):
         pipeline_runtime_display_name = _get_runtime_display_name(primary_pipeline['app_data']['runtime'])
-        provided_runtime_display_name = _get_runtime_display_name(_get_runtime_type(runtime_config))
-        raise click.ClickException(
-            f"This pipeline requires an instance of {pipeline_runtime_display_name} runtime configuration.\n"
-            f"The specified configuration '{runtime_config}' is for {provided_runtime_display_name} runtime.")
+        msg = f"This pipeline requires an instance of {pipeline_runtime_display_name} runtime configuration.\n"
+        if runtime_config:
+            provided_runtime_display_name = _get_runtime_display_name(_get_runtime_type(runtime_config))
+            msg = f"{msg}The specified configuration '{runtime_config}' is for {provided_runtime_display_name} runtime."
+        raise click.ClickException(msg)
 
     # update pipeline transient fields
     primary_pipeline["app_data"]["name"] = pipeline_name
@@ -155,6 +156,8 @@ def _preprocess_pipeline(pipeline_path: str,
         primary_pipeline["app_data"]["runtime"] = runtime
     if runtime_config:
         primary_pipeline["app_data"]["runtime-config"] = runtime_config
+    else:
+        primary_pipeline["app_data"]["runtime-config"] = 'local'
 
     return pipeline_definition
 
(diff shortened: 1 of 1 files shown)


elyra/cli/pipeline_app.py
```diff
@@ -301,7 +304,7 @@ def run(pipeline_path):
     _validate_pipeline_file_extension(pipeline_path)
 
     pipeline_definition = \
-        _preprocess_pipeline(pipeline_path, runtime='local', runtime_config='local')
+        _preprocess_pipeline(pipeline_path, runtime='local')
```
lresende, line 307: Why remove the explicit `runtime_config` here? it sort of explicitly say that run is only happening locally.

lresende:
Would modifying the functions `_get_runtime_type` and `_get_runtime_display_name` to handle local be another option to avoid the error in question?

ptitzler:
> Would modifying the functions `_get_runtime_type` and `_get_runtime_display_name` to handle local be another option to avoid the error in question?

Perhaps, but wouldn't that introduce an unwanted ambiguity and potentially lead to errors if the user defines a runtime configuration named `local`? We had already contemplated doing that for ODH deployments where KFP and JL/Elyra run in the same Kubernetes cluster.

lresende:
I took a quick look and currently the UI will submit pipelines to run in "local mode" by setting both the `runtime` and `runtime_config` to `local` and I would prefer to keep different clients consistent. 

How about we do something like:

```
diff --git a/elyra/cli/pipeline_app.py b/elyra/cli/pipeline_app.py
index c9949aee..3347926f 100644
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -48,6 +48,9 @@ def _get_runtime_type(runtime_config: Optional[str]) -> Optional[str]:
     if not runtime_config:
         return None
 
+    if runtime_config == 'local':
+        return runtime_config
+
     try:
         metadata_manager = MetadataManager(namespace='runtimes')
         metadata = metadata_manager.get(runtime_config)
@@ -60,6 +63,9 @@ def _get_runtime_display_name(schema_name: Optional[str]) -> Optional[str]:
     if not schema_name:
         return None
 
+    if schema_name == 'local':
+        return schema_name
+
```

lresende (changes requested):
See comment above.

ptitzler:
> See comment above.

changed

akchinSTC:
@lresende - needs review =)

elyra/cli/pipeline_app.py
```diff
@@ -155,6 +160,8 @@ def _preprocess_pipeline(pipeline_path: str,
         primary_pipeline["app_data"]["runtime"] = runtime
     if runtime_config:
         primary_pipeline["app_data"]["runtime-config"] = runtime_config
+    else:
```
lresende, line 163: I still don't feel comfortable defaulting here... have you seen a case where we are not passing the proper 'local' for local runs?

elyra/cli/pipeline_app.py
```diff
@@ -155,6 +160,8 @@ def _preprocess_pipeline(pipeline_path: str,
...
+        primary_pipeline["app_data"]["runtime-config"] = 'local'
```
lresende, line 164: I still don't feel comfortable defaulting here... have you seen a case where we are not passing the proper 'local' for local runs?

ptitzler: The goal was to prevent a fatal downstream error if no runtime config value is specified _and_ subsequently an attempt is made to run the pipeline. I reverted the change as requested.

lresende (approved):

Issue #2073, opened 2021-08-25
run local pipeline demo has error 


HGladiator:
os: win10
python:3.7.10
elyra:3.0.1

subprocess.CalledProcessError: Command '['python3', 'E:\\Code\\Elyra_examples/pipelines/setup_validation/python_script.py', '--PYTHONHOME', 'E:\\Code\\Elyra_examples/pipelines/setup_validation']' returned non-zero exit status 9009.

pls help me! thx!

ptitzler:
Transferred issue to the main repo, since this appears to be a setup problem.

ptitzler:
Can you please
- Confirm whether you tried to run a custom pipeline - not any of the existing examples?
- If you've used your own pipeline, could you please share the `.pipeline` file? The path to the Python script `E:\Code\Elyra_examples/test/print_version.py` appears to be invalid - it should not contain any forward slashes. Did you start JupyterLab in `E:\Code\Elyra_examples`  and created `print_version.py` in the `test` subdirectory? I don't have access to a Windows environment to confirm whether this generally works as expected. It's possible that somehow OS specific path handling is used somewhere in the Elyra code that we'd need to locate and fix.
- Confirm that python3 is in your PATH

ptitzler:
> Possible bug: https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/processor_local.py#L291 sets PYTHONHOME to the script's directory location, which doesn't seem right. Perhaps the intention was to set PYTHONPATH?

@lresende @kevin-bates

kevin-bates:
I looked at this when you last at-ed me and agree that this is doesn't seem right.  I will need to defer to @lresende who authored that portion of things.  I suspect this was done in order to make files in the local directory available, but prepending the current `PYTHONPATH` with the script directory is likely the better way to go.

I wasn't aware envs could be added to the `python` command as options like that (and it doesn't appear to work).  Here's a little script that dumps the value of `PYTHONHOME` using `print(f"PYTHONHOME: {os.getenv('PYTHONHOME')}")`. I don't currently have `PYTHONHOME` set in my environment...
```bash
$ echo $PYTHONHOME

$ python abc_test.py --PYTHONHOME=/tmp
PYTHONHOME: None
```

ptitzler:
[...]

PR #2077, opened 2021-08-26
Fix path issue and capture problem determination data for script execution

ptitzler:
Based on research performed for #2073

For local pipeline execution:
- Capture invocation information to aid with troubleshooting if generic components fail to execute
- Fix path handling for Windows platform
- Remove no-op parameter from Python script execution invocation

Example output (user runs pipeline that processes an r script but r is not installed):

```
[E 2021-08-30 08:58:13.875 ServerApp] Error executing rrrrrr.r: [Errno 2] No such file or directory: 'Rscript': 'Rscript'
[I 2021-08-30 08:58:13.876 ServerApp] Data capture for previous error:
     command: Rscript /Users/patti/elyra_dev/i2073/workspace/rrrrrr.r
     working directory: /Users/patti/elyra_dev/i2073/workspace
     environment variables: {'TMPDIR': '/var/folders/h8/l3qh8v192rv_vvggc6wyphzc0000gn/T/', '__CFBundleIdentifier': 'com.apple.Terminal', 'XPC_FLAGS': '0x0', 'TERM': 'xterm-256color', 'SSH_AUTH_SOCK': '/private/tmp/com.apple.launchd.UPRFGPMBxC/Listeners', 'XPC_SERVICE_NAME': '0', 'TERM_PROGRAM': 'Apple_Terminal', 'TERM_PROGRAM_VERSION': '440', 'TERM_SESSION_ID': 'D7E10F30-16B5-4397-A607-1E9F8A19A84F', 'SHELL': '/bin/zsh', 'HOME': '/Users/patti', ...}
```

Example output (user runs pipeline that executes a Python script but the `python3` executable in the PATH -- simulated by changing the command to `python4` from `python3`):

```
[E 2021-09-02 07:48:35.677 ServerApp] Error executing a-script.py: [Errno 2] No such file or directory: 'python4': 'python4'
[I 2021-09-02 07:48:35.679 ServerApp] Data capture for previous error:
     command: python4 /Users/patti/elyra_dev/i2073/workspace/a-script.py
     working directory: /Users/patti/elyra_dev/i2073/workspace
     environment variables: {'TMPDIR': 
```

(MacOS only - no access to Windows) Confirmed that pd information is logged in the following scenarios:
- Run R script in local environment where R is not installed
- Run Python script in local environment where `python3` is not in the PATH

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor_local.py
+++ b/elyra/pipeline/processor_local.py
@@ -23,6 +23,7 @@ from subprocess import run
 import time
 from typing import Dict
 from typing import List
+from typing import Optional
 
 from jupyter_server.gateway.managers import GatewayClient
 import papermill
@@ -171,7 +172,10 @@ class FileOperationProcessor(OperationProcessor):
             raise ValueError(f'Not a file: {filepath}')
         return filepath
 
-    def log_and_raise(self, file_name: str, ex: Exception) -> None:
+    def log_and_raise(self,
+                      file_name: str,
+                      ex: Exception,
+                      data_capture_msg: Optional[str]) -> None:
         """Log and raise the exception that occurs when processing file_name.
 
         If the exception's message is longer than MAX_ERROR_LEN, it will be
@@ -179,6 +183,8 @@ class FileOperationProcessor(OperationProcessor):
         will be logged.
         """
         self.log.error(f'Error executing {file_name}: {str(ex)}')
+        if data_capture_msg:
+            self.log.info(data_capture_msg)
         truncated_msg = FileOperationProcessor._truncate_msg(str(ex))
         raise RuntimeError(f'({file_name}): {truncated_msg}') from ex
 
@@ -263,6 +269,10 @@ class ScriptOperationProcessor(FileOperationProcessor):
 
         argv = self.get_argv(filepath)
         envs = OperationProcessor._collect_envs(operation, elyra_run_name)
+        data_capture_msg = f"Data capture for previous error:\n"\
+                           f" command: {' '.join(argv)}\n"\
+                           f" working directory: {file_dir}\n"\
+                           f" environment variables: {envs}"
         t0 = time.time()
         try:
             run(argv, cwd=file_dir, env=envs, check=True, stderr=PIPE)
@@ -270,7 +280,7 @@ class ScriptOperationProcessor(FileOperationProcessor):
             error_msg = str(cpe.stderr.decode())
             self.log.error(f'Error executing {file_name}: {error_msg}')
             # Log process information to aid with troubleshooting
-            self.log.info(f'FFDC: \n argv: {argv}\n cwd: {file_dir}\n env:{envs}')
+            self.log.info(data_capture_msg)
 
             error_trim_index = error_msg.rfind('\n', 0, error_msg.rfind('Error'))
             if error_trim_index != -1:
@@ -278,7 +288,7 @@ class ScriptOperationProcessor(FileOperationProcessor):
             else:
                 raise RuntimeError(f'({file_name})') from cpe
         except Exception as ex:
-            self.log_and_raise(file_name, ex)
+            self.log_and_raise(file_name, ex, data_capture_msg)
 
         t1 = time.time()
         duration = (t1 - t0)
--- a/elyra/pipeline/processor_local.py
+++ b/elyra/pipeline/processor_local.py
@@ -300,7 +300,7 @@ class PythonScriptOperationProcessor(ScriptOperationProcessor):
     _script_type = 'Python'
 
     def get_argv(self, file_path) -> List[str]:
-        return ['python3', file_path, '--PYTHONHOME', os.path.dirname(file_path)]
+        return ['python3', file_path]
 
 
 class RScriptOperationProcessor(ScriptOperationProcessor):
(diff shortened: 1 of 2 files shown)


elyra/pipeline/processor_local.py
```diff
@@ -269,6 +269,8 @@ def process(self, operation: GenericOperation, elyra_run_name: str):
         except CalledProcessError as cpe:
             error_msg = str(cpe.stderr.decode())
             self.log.error(f'Error executing {file_name}: {error_msg}')
+            # Log process information to aid with troubleshooting
+            self.log.info(f'FFDC: \n argv: {argv}\n cwd: {file_dir}\n env:{envs}')
```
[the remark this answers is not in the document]
ptitzler: See updated message text. I also added code that properly logs this information on other platforms, where the error is not raised via `CalledProcessError`.

kevin-bates (approved):
LGTM. Thanks Patrick.

kevin-bates:
> Run Python script in local environment where python3 is not in the PATH

I don't see any accommodations for a missing `python3` executable.

ptitzler:
> > Run Python script in local environment where python3 is not in the PATH
> 
> I don't see any accommodations for a missing `python3` executable.

I updated the description to properly reflect what has been tested. The PR does not attempt to resolve the problem scenario that `python3` is not accessible.

kevin-bates:
Thanks for the update. I think we can still run into an issue on Windows here. If we want to defer that portion of the issue (which _was_ the issue) I'm okay so long as another issue is created. 
I'm also okay updating this issue and can probably look into a fix later today if that's warranted.

ptitzler:
> If we want to defer that portion of the issue (which was the issue) I'm okay so long as another issue is created.

Even though I didn't state that explicitly, that's the reason why this PR does not close the issue using one of the magic keyword(s).

I didn't attempt to fix that specific problem because
- A basic platform dependent solution (use `python` on win, `python3` elsewhere)  would likely not be sufficient because (1) there are Python installations on windows that install `python3` and (2) `python` is not guaranteed to be at version 3, which is why we changed the executable name in the first place.
- I don't have an environment to verify that a potential generic proposal (try `python3` followed by `python`) would work as expected.

I also looked into updating the documentation to state what the prerequisites are but there's currently no topic that would lend itself.

kevin-bates:
I'm in the process of provisioning a Windows server to take a closer look at the python vs. python3 issue.

kevin-bates:
This is interesting.  Although python3.exe can be found using where, and that directory is in the PATH, it is not found from the active conda env, and the active conda env doesn't include a python3.exe:
```
(base) C:\Users\Administrator>echo %PATH%
C:\Users\Administrator\anaconda3;C:\Users\Administrator\anaconda3\Library\mingw-w64\bin;C:\Users\Administrator\anaconda3\Library\usr\bin;C:\Users\Administrator\anaconda3\Library\bin;C:\Users\Administrator\anaconda3\Scripts;C:\Users\Administrator\anaconda3\bin;C:\Users\Administrator\anaconda3\condabin;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0;C:\WINDOWS\System32\OpenSSH;C:\Users\Administrator\AppData\Local\Microsoft\WindowsApps;.

(base) C:\Users\Administrator>where python
C:\Users\Administrator\anaconda3\python.exe
C:\Users\Administrator\AppData\Local\Microsoft\WindowsApps\python.exe

(base) C:\Users\Administrator>where python3
C:\Users\Administrator\AppData\Local\Microsoft\WindowsApps\python3.exe

(base) C:\Users\Administrator>python --version
Python 3.8.8

(base) C:\Users\Administrator>python3 --version
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
```

I'd like to see how [`shutil.which`](https://docs.python.org/3.3/library/shutil.html#shutil.which) behaves here and that might be a simple solution.  Can't get to this today though..

kevin-bates:
In looking into `shutil.which`, I think the solution, for all platforms, is to simply use the python executable that is currently running the jupyter server: `sys.executable`.  We know that _that_ python is an appropriate version and available.

This may need to be revisited if we ever wanted to introduce _local runtime configurations_ (which I think we may want to look into someday), but, for local processing right now its arguably the most correct thing we can do.

elyra/pipeline/processor_local.py
```diff
@@ -288,7 +300,7 @@ class PythonScriptOperationProcessor(ScriptOperationProcessor):
     _script_type = 'Python'
 
     def get_argv(self, file_path) -> List[str]:
-        return ['python3', file_path, '--PYTHONHOME', os.path.dirname(file_path)]
+        return ['python3', file_path]
```
kevin-bates, line 303: Per this comment, I think we should go with `sys.executable` here.
```suggestion
        return [f"{sys.executable}", file_path]
```

elyra/pipeline/processor_local.py
```diff
@@ -23,6 +23,7 @@
 import time
 from typing import Dict
 from typing import List
+from typing import Optional
```
kevin-bates, line 26: `sys` is needed for the `sys.executable` reference below.
```suggestion
from typing import Optional
import sys
```

kevin-bates:
I've successfully tested the use of `sys.executable` on Windows (having removed python3.exe).  The local unit test will exercise for Linux, so I think we're good to go.

PR #2082, opened 2021-08-27
Pipeline definition abstraction

### LINKED ISSUE #2067: Provide abstraction layer for manipulating pipeline definition (json) files in Elyra source code
Currently client code (in the broader sense), e.g. `elyra-pipeline`, has to manually traverse the json structures in `.pipeline` files, which may change over time. 

There are a several drawbacks to not having an abstraction for the pipeline definition file:
- Logic is duplicated in multiple places
- Maintenance is hard as code is duplicated
- Susceptibility to bugs as corner cases logic needs to be considered by multiple consumers
- etc

To avoid duplication of this kind of code and reduce the maintenance effort, Elyra should provide class to properly retrieve fields from the pipeline definition json structure. 

Example:

Instead of:

```
for pipeline in pipeline_definition["pipelines"]:
    if pipeline['id'] == primary_pipeline_key:
        primary_pipeline = pipeline

primary_pipeline["app_data"]["version"]
```

Have something like:

```
pipeline_definition.get_field("version")
```
or
```
pipeline_definition.version
```

In the case above, the logic would have to handle things like finding the primary pipeline, validating the `app_data` structure is available, validate `field name` is available and retrieve its value.

Possible operations to abstract:
- Discover primary pipeline
- isEmpty
- get pipeline field
- get node (name/id) field
- etc

Possible consumers:
- Pipeline CLI
- Pipeline Validation service

Note that, this is to be used on preliminary areas that the existing [Pipeline Parser](https://github.com/elyra-ai/elyra/blob/31c3816acbfe41737ef5a2f0627c90419175f0a6/elyra/pipeline/parser.py#L30) and [Pipeline value object](https://github.com/elyra-ai/elyra/blob/31c3816acbfe41737ef5a2f0627c90419175f0a6/elyra/pipeline/pipeline.py#L305) can't be used. So, not a lot of logic and assumptions should be in this abstraction layer to avoid duplication with the Pipeline parser/value object.

### LINKED ISSUE #2076: Conflicting algorithm is used to calculate node label value
**Describe the issue**
A clear and concise description of what the issue is. There are several locations in the Elyra code where the label value of a node is calculated from information stored in the `.pipeline` file. Currently there are two conflicting implementations.

-  `app_data.label`  takes precedence over `app_data.ui_data.label`
https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/parser.py#L132-L135

-  `app_data.ui_data.label` takes precedence over `app_data.label`
https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/validation.py#L239

(I'm not sure where the calculation is done for the display in the VPE.)

To avoid any differences in behavior should any assumptions change (as to what property value is set) the same algorithm should be used everywhere.

lresende:
Fixes #2067
Fixes #2076
Related to #2088 (might need more investigation)

           in the file; or

           it.


elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class Pipeline(NodeBase):
+
+    def __init__(self, pipeline_definition: Dict, node: Dict):
+        super().__init__(pipeline_definition, node)
+
+    @property
+    def version(self) -> int:
+        return int(self._node['app_data'].get('version'))
```
kevin-bates, line 74: I understand that this resides on a `Pipeline` entry, but think it's more helpful on the `PipelineDefinition` and "adjacent" to `schema_version` and named `pipeline_version`.  This property will be confusing (and won't exist) when multi-pipeline definitions (supernodes) are better supported.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def runtime(self) -> str:
+        return self._node['app_data'].get('runtime') or 'local'
```
kevin-bates, line 78: Thinking the same here.  This won't fly for multi-pipeline definitions.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def type(self):
+        type_description_to_type = {'Kubeflow Pipelines': 'kfp',
```
kevin-bates, line 82: Ditto here.

`type` is really overused - could this be `runtime_type` or `pipeline_type`.   A help string would be useful.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+                                    'Generic': 'generic'}
+
+        pipeline_type_description = self._node['app_data']['properties'].get('runtime')
+        if pipeline_type_description not in type_description_to_type.keys():
+            raise ValueError(f'Unsupported pipeline runtime: {pipeline_type_description}')
+        return type_description_to_type[pipeline_type_description]
+
+    @property
+    def nodes(self) -> list:
+        if 'nodes' not in self._node:
```
kevin-bates, line 93: We should only create these Nodes once if `self._nodes is None` or something, rather than every time the `nodes` property is accessed.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+        self._node['app_data']['properties'][key] = value
+
+    def validate(self) -> list:
+        validation_issues = list()
+        if 'properties' not in self._node['app_data']:
+            validation_issues.append("Node is missing 'properties' field.")
+        elif len(self._node['app_data']['properties']) == 0:
+            validation_issues.append("Pipeline has zero length 'properties' field.")
+
+        return validation_issues
```
kevin-bates, line 124: Seems like a `Pipeline`'s validation should call `validate()` on each of its `Node` instances.

Again, we should consider only performing validation when "dirty".

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class Node(NodeBase):
+    def __init__(self, pipeline_definition: Dict, node: Dict):
+        super().__init__(pipeline_definition, node)
+
+    @property
+    def type(self) -> str:
+        return self._node.get('type')
```
kevin-bates, line 133: `node_type` for the same reasons as `Pipeline.type`.  Not sure what this is.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def op(self) -> str:
+        return self._node.get('op')
+
+    @property
+    def label(self) -> str:
+        return self._node['app_data']['ui_data'].get('label', self._node['app_data']['label'])
```
kevin-bates, line 141: Thank goodness!  Might need to flip this based on recent conversations, but at least its only here for the backend! :smile:

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+    def get_component_parameter(self, key: str, default_value=None) -> Any:
+        self._validate()
+        return self._node['app_data']['component_parameters'].get(key, default_value)
+
+    def set_component_parameter(self, key: str, value: Any):
+        if not key:
+            raise ValueError("Key is required")
+
+        if not value:
+            raise ValueError("Value is required")
```
kevin-bates, line 152: Third occurrence of this pattern.  Perhaps a `_vallidate_key_value()` helper in `NodeBase`?
```python
def _validate_key_value(self, key: str, value: Any) -> None:
        if not key:
            raise ValueError("Key is required")

        if not value:
            raise ValueError("Value is required")
```

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+        self._node['app_data']['component_parameters'][key] = value
+
+    def validate(self) -> list:
+        validation_issues = list()
+        if "component_parameters" not in self._node['app_data']:
+            validation_issues.append("Node is missing 'component_parameters' field")
+
+        return validation_issues
```
kevin-bates, line 161: I think the `validate()` methods on `NodeBase` instances should take the list of validation issues and `PipelineDefinition.validate()` constructs that list and calls `validate(issues_list)` on each `Pipeline` - which then calls `validate(issues_list)` on each of its `Node` instances.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        The constructor enables either passing a pipeline path or the content of the pipeline definition.
+
+        """
+        if not pipeline_path and not pipeline_definition:
+            # at least one parameter should be provided
+            raise ValueError("At least one parameter must be provided ('pipeline_path' or 'pipeline_definition')")
+        if pipeline_path and pipeline_definition:
+            # only one parameter should be provided
+            raise ValueError("Only one parameter should be provided ('pipeline_path' or 'pipeline_definition')")
```
kevin-bates, line 180: I think one way to handle this exclusive-or kind of behavior would be something like:
```python
    def __init__(self, pipeline_source: Union[str, Dict]):
        if isinstance(pipeline_source, str):
             # supporting loading pipeline from file
            if not os.path.exists(pipeline_source):
                raise ValueError(f"Pipeline file not found: '{pipeline_source}'\n")

            with open(pipeline_source) as f:
                try:
                    self._pipeline_definition = json.load(f)
                except ValueError as ve:
                    raise ValueError(f"Pipeline file is invalid: \n {ve}")
        elif isinstance(pipeline_source, dict):
            # supporting passing the pipeline definition directly
            self._pipeline_definition = pipeline_source
        else:
            raise TypeError(f"Parameter '{pipeline_source}' was not a string or dictionary instance.")
```

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+        if 'pipelines' not in self._pipeline_definition:
+            raise ValueError("Pipeline is missing 'pipelines' field.")
+        elif len(self._pipeline_definition['pipelines']) == 0:
+            raise ValueError("Pipeline has zero length 'pipelines' field.")
+
+        pipelines: list = list()
+        for pipeline in self._pipeline_definition['pipelines']:
+            pipelines.append(Pipeline(pipeline))
+
+        return pipelines
```
kevin-bates, line 215: Same a `nodes()` - we should consider building this only once.

We could also consider making these (this and `PIpeline.nodes`) coroutines so that they don't require accumulation.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+        if 'primary_pipeline' not in self._pipeline_definition:
+            validation_issues.append("Could not determine the primary pipeline.")
+        elif not isinstance(self._pipeline_definition["primary_pipeline"], str):
+            validation_issues.append("Field 'primary_pipeline' should be a string.")
+
+        primary_pipeline = self._get_pipeline_definition(self._pipeline_definition.get('primary_pipeline'))
+        if not primary_pipeline:
+            validation_issues.append("No primary pipeline was found")
+        else:
+            # Validate primary pipeline structure
```
kevin-bates, line 258: I think this is where we call `validate(issues_list)` on each pipeline entry.  A pipeline entry should determine if its _the_ primary pipeline in its constructor, then the code below can just be part of the Pipeline.validate() method and conditional if it's the primary pipeline.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+                validation_issues.append("At least one node must exist in the primary pipeline.")
+            if 'app_data' not in primary_pipeline:
+                validation_issues.append("Primary pipeline is missing the 'app_data' field.")
+            elif 'version' not in primary_pipeline['app_data']:
+                validation_issues.append("Primary pipeline is missing the 'version' field.")
+
+        return validation_issues
+
+    def is_valid(self) -> bool:
+        return len(self.validate()) == 0
```
kevin-bates, line 269: Again, not sure we want to validate the entire definition just to see if it's valid or not.

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def dict(self) -> Dict:
+        return self._pipeline_definition
```
kevin-bates, line 273: `to_dict()`

elyra/pipeline/pipeline_definition.py
```diff
@@ -0,0 +1,281 @@
...
+        return self._node['app_data'].get(key, default_value)
+
+    def set(self, key: str, value: Any):
+        if not key:
+            raise ValueError("Key is required")
+
+        if not value:
+            raise ValueError("Value is required")
+
+        self._node['app_data'][key] = value
```
kevin-bates, line 58: Hmm, should each setter method also update the corresponding location in `self._pipeline_definition`?  This will prevent the two renderings from getting out of sync and avoid the need to regenerate instances on each list accessor.

elyra/pipeline/pipeline_definition.py
kevin-bates: This isn't compatible with BYO runtimes.  It kinda seems like we should have a way to enumerate all  `elyra.runtime.processors` where each processor implementation returns the _name_ of the runtime. E.g., `KfpPipelineProcessor.runtime_name` returns 'Kubeflow Pipelines'.  Where are these strings coming from today?  Is that the schema's `display_name`?
https://github.com/elyra-ai/elyra/blob/9645b8e796306ecc700a1263e1b438e6d73f8169/elyra/metadata/schemas/kfp.json#L5

Seems like we need a method that performs this enumeration (and adds the 'Generic' entry I guess).

kevin-bates (commented):
Hey Luciano - this is going to be a huge help in eliminating redundancy and bugs that come with it!  Thanks for doing this.

Overall, I'm hoping we can add help strings and type hints on each of the methods.

Also, I'm not clear on the usage of this, particularly with respect to updates and validation - so many of my comments might come across as subjective since I'm trying to avoid performance concerns.  Since this is still single-user those concerns might be overblown, but that's the "hat" I had on at the time.  Thanks again.

elyra/pipeline/validation.py
```diff
@@ -187,122 +183,107 @@ def _validate_pipeline_structure(self, pipeline: dict, response: ValidationRespo
...
             # Checks pipeline node types are compatible with the runtime selected
-            for single_pipeline in pipeline_json['pipelines']:
-                pipeline_id = single_pipeline['id']
-                node_list = single_pipeline['nodes']
-                for node in node_list:
-                    if node['type'] == "execution_node" and node['op'] not in supported_ops:
-                        node_label = node['app_data']['ui_data'].get('label', node['app_data']['label'])
+            for pipeline in pipeline_definition.pipelines:
+                for node in pipeline.nodes:
+                    if node.type == "execution_node" and node.op not in supported_ops:
```
kevin-bates, line 220: Why isn't this just part of Pipelinedfintion.Pipeline.Node.validate()?  I figured the PD would also validate since it has `validate()` methods at every level.

elyra/pipeline/parser.py
```diff
@@ -30,37 +31,38 @@ def __init__(self, root_dir="", **kwargs):
...
 
-        pipeline_object = Pipeline(id=primary_pipeline_id,
-                                   name=PipelineParser._get_app_data_field(primary_pipeline, 'name', 'untitled'),
+        pipeline_object = Pipeline(id=primary_pipeline.id,
+                                   name=primary_pipeline.name,
                                    runtime=runtime,
                                    runtime_config=runtime_config,
                                    source=source)
-        self._nodes_to_operations(pipeline_definitions, pipeline_object, primary_pipeline['nodes'])
+        self._nodes_to_operations(pipeline_json, pipeline_object, primary_pipeline.to_dict().get('nodes'))
```
kevin-bates, line 61: Shouldn't this use the helper classes?
```suggestion
        self._nodes_to_operations(pipeline_definition, pipeline_object, primary_pipeline.nodes))
```

elyra/pipeline/parser.py
```diff
@@ -30,37 +31,38 @@ def __init__(self, root_dir="", **kwargs):
...
         return pipeline_object
 
     def _nodes_to_operations(self,
-                             pipeline_definitions: Dict,
+                             pipeline_definition: Dict,
```
kevin-bates, line 65: Shouldn't this use the helper classes?
```suggestion
                             pipeline_definition: PipelineDefintion,
```

elyra/pipeline/parser.py
```diff
@@ -30,37 +31,38 @@ def __init__(self, root_dir="", **kwargs):
...
+        runtime_config = primary_pipeline.runtime_config
         if not runtime_config:
             raise ValueError("Invalid pipeline: Missing runtime configuration.")
 
-        source = PipelineParser._get_app_data_field(primary_pipeline, 'source')
+        source = primary_pipeline.source
 
-        pipeline_object = Pipeline(id=primary_pipeline_id,
-                                   name=PipelineParser._get_app_data_field(primary_pipeline, 'name', 'untitled'),
+        pipeline_object = Pipeline(id=primary_pipeline.id,
```
kevin-bates, line 56: Here we go.  Which `Pipeline` is this?  (I know it's the one in `pipeline.py` and not in `pipeline_definition.py`, but we now have another ambiguity.)

kevin-bates (commented):
Hmm.  Before continuing much longer, I think we need to make a commitment to one data object model or another. Having pipeline definitions that are pure JSON or `PipelineDefinition` and multiple `Pipeline` class definitions, using `Operation` or `Node` instances, or pure JSON is a recipe for heartburn.

elyra/pipeline/validation.py
lresende: this is not structure validation, but actually, check each op against the list of components registered to the runtime, so it's more like a validation piece. Would you agree?

kevin-bates: validation of the structure and validation of the values needs to occur.  I'm just asking why this code isn't more like: 
```python
for pipeline in pipeline_definition.pipelines:
    pipeline.validate()
```
and `pipeline.validate()` would look more like:
```python
    # validate pipeline data
    for node in self.nodes:
        node.validate()
```
and `node.validate() would look more like:
```python
    if self.type == "execution_node" and self.op not in supported_ops:
```

elyra/pipeline/parser.py
```diff
@@ -30,43 +31,44 @@ def __init__(self, root_dir="", **kwargs):
...
+        self._nodes_to_operations(pipeline_definition, pipeline_object, primary_pipeline.nodes)
         return pipeline_object
 
     def _nodes_to_operations(self,
-                             pipeline_definitions: Dict,
+                             pipeline_definition: PipelineDefinition,
                              pipeline_object: Pipeline,
                              nodes: List[Dict],
-                             super_node: Optional[Dict] = None) -> None:
+                             super_node: Node = None) -> None:
```
kevin-bates, line 71:
```suggestion
                             super_node: Optional[Node] = None) -> None:
```

elyra/pipeline/parser.py
kevin-bates: What does _this_ `Pipeline` value object have that the `PropertyDefinition.primary_pipeline` value object doesn't have?

I can see there being a difference for `Operations` to `Nodes` but perhaps we could merge the Operation functionality into Nodes or let the single `Pipeline` instance have runtime-representations of Nodes as a parallel list of Operations?  (Or have a Node hold its `Operation` instance?)

kevin-bates (commented):
This is coming along and should improve maintainability and supportability.  Thank you.

My review isn't complete but I had to move to other things for now.

elyra/pipeline/parser.py
lresende: I would like to consider merging these, but as another pr/initiative

elyra/cli/pipeline_app.py
```diff
@@ -62,6 +67,9 @@ def _get_runtime_display_name(schema_name: Optional[str]) -> Optional[str]:
         # Cannot use metadata manager to determine the display name.
         return schema_name
 
+    if schema_name == 'local':
+        return schema_name
```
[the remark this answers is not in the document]
lresende: Sorry, merge side effect.

kevin-bates (approved):
I'm approving this knowing there are follow-on changes, but this initiates the general idea and its merge will unblock other activity.

Some changes we should consider:
1. Complete replacement of the _other_ Pipeline class.  Need to figure out the best way to handle Operator vs. Node.
1. Rename of `type` across the various objects.
1. Effective means for determining configured pipelines rather than hard-coding

Issue #1880, opened 2021-07-08
Add GUI for managing custom pipeline components


ptitzler:
**Is your feature request related to a problem? Please describe.**

As a user I want to 
 - add custom components to the registry
 - edit custom components
 - delete custom components from the registry

Currently these tasks can only be accomplished manually and are not supported by the GUI or CLI.

### Add a custom component to the registry

User workflow:
 - Choose a component runtime platform from list of supported platforms
    - Kubeflow Pipelines
    - Apache Airflow
 - Provide component metadata
    - Component label (for rendering; optional; default extracted from component definition)
    - Component description (for rendering; optional; default extracted from component definition)
 - Provide required component information
    - For KFP: 
       - component type: one of CT1 or CT2
          - CT1: [containerized component](https://www.kubeflow.org/docs/components/pipelines/sdk/component-development/)
          -  CT2: [Python function-based component](https://www.kubeflow.org/docs/components/pipelines/sdk/v2/python-function-components/)  -> requires https://github.com/elyra-ai/elyra/issues/1920 
        - component source:
          - CT1:
             - Interface definition (YAML file) location
             - Container image (required, type: `str`)  -> explicit default to "KFP default image"
          - CT2:
             - Component definition (Python file) location (required)
             - Container image (required, type: `str`)  -> explicit default to "KFP default image"
             - Packages to install (optional, type `List[str]`)
          - CT3:
             -  Apache Airflow operator definition  location (required)
             -  TBD
    - For Apache Airflow:
      - component source:
             - Component definition (Python file) location
             - Container image
- Save component
   - Required actions:
      - validate  existence of definition
      - validate that definition is valid (can be parsed and the required information extracted)
      - validate definition's optional parameters

TODO:
 - text/free form container image entry or use runtime-image (metadata managed)?

ptitzler:
Refer to https://github.com/elyra-ai/elyra/discussions/1881 for temporary workaround.

ptitzler:
https://github.com/ptitzler/kfp-component-tests/tree/main/python_components contains a few introductory examples + explanations

ptitzler:
Starting Elyra 3.3 custom components can be managed using the GUI. Refer to [this documentation topic for details](https://elyra.readthedocs.io/en/latest/user_guide/pipeline-components.html#managing-pipeline-components)

Issue #1873, opened 2021-07-07
Add `manage components` topic content to user guide


ptitzler:
Follow-up to https://github.com/elyra-ai/elyra/issues/1838:
- Add component to registry
- Edit component entry
- Remove component from registry

Requires https://github.com/elyra-ai/elyra/issues/1880

Workaround is documented in https://github.com/elyra-ai/elyra/discussions/1881

PR #2083, opened 2021-08-30
Migrate component registry to the metadata service

### LINKED ISSUE #1878: Palette node categories should be defined by backend
Currently, the categories / all of the palette information outside of the node definitions is defined by the frontend (and is somewhat hard-coded), but this should be moved to the backend.

### LINKED ISSUE #2031: Migrate component registry to metadata service
The component registries are currently stored in `etc/config/components`. Every time a change is made to the registry, a rebuild is required. Additionally, it is not very user-friendly to add or remove components from the registry JSON files, as seen in discussion #1881.

 As we look toward adding a GUI for managing custom components (#1880), it could make sense to move the implementation of the registries to the metadata service. This also has the added benefit of improved consistency among stored information.

# Design Considerations
I think this issue can be broken down into the following 3 parts:

1. Completely reorganizing the component registry to point to the location (directory or url) of a *list* of component specifications
2. Moving the component registry to the metadata service
3. Handling component categories

Description and design considerations and questions for each to follow.

---

### 1. Restructuring component registry
**Motivation**

Allows for a user to easily add more components, rather than having to define each individual component one-by-one. This will pair nicely with the ability for users to disable/not display certain components in the palette (issue #2009).

As a standalone feature, this may be of more 'moderate' importance, but if we intend to support users customizing their palette components via moving the component registries to metadata, then it would probably be best to implement this first so as to not cause confusion when we change things down the line.

**High-level direction**

Currently, the component registries (one for each runtime) contain a list of components and the location of each component spec in an 'unofficial' JSON format like the below.

```json=
 "components": {
    "bash-operator": {
      "location": {
        "url": "https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/bash_operator.py"
      },
      "category": "airflow"
    },
    "email-operator": {
      "location": {
        "url": "https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/email_operator.py"
      },
      "category": "airflow"
    }
}
```

Instead, we want to move to having user-defined component registries, where only one location is specified per registry "location specifier" and that specifier points to one or several component spec files to be parsed. To start, a location specifier will be one of three types filename, url, and directory, as shown below. See [comment below](https://github.com/elyra-ai/elyra/issues/2031#issuecomment-901278580) for more details.

```json=
{
  "display_name": "KFP Preloaded Components",
  "metadata": {
    "description": "Preloaded components that are supported by Kubeflow Pipelines",
    "runtime": "kfp",
    "locations": [
      {
        "location_path": "kfp",    *this is a relative directory for our preloaded components
        "location_type": "directory",
        "categories": ["Preloaded KFP"]
      },
      {
        "location_path": "https://some/url/here/component.yaml",
        "location_type": "url",
        "categories": ["Preloaded KFP"]
      },
      {
        "location_path": "some/path/to/component.yaml", *this is a relative path for our preloaded components
        "location_type": "filename",
        "categories": ["Preloaded KFP"]
      }
    ]
  },
  "schema_name": "component-registry"
}
```

Requirements:

- Remove current component catalogs (`airflow_component_catalog.json` and `kfp_component_catalog.json`) and instead create a component-registry metadata instance using a schema like the above JSON, with a directory location of where the preconfigured component specs reside (probably an `ENV_JUPYTER_PATH` in our case)
    - More schema info in next section
    - Maybe named something like `elyra-preconfigured-kfp` and `elyra-preconfigured-airflow`
- Removal of the component catalogs will push assigning the id until after spec content read, rather than getting it from the catalog directly as it is done now
  - We will need a new heuristic for assigning component ids: maybe `[filename w/o extension]_[component name]`, which should provide enough of an assurance against repeats
- Will probably have to remove the ability to parse just a single Airflow class in order to support the multi-valued type registries (directory-based or github repo-based, e.g.)
  - This shouldn't cause much of a latency issue since we can expect an operator file to only define a handful of classes
  - This will also remove the `catalog_entry_id` attribute on `Component` among other minor changes to existing functions, such as the function that retrieves only a singular component (again, due to caching and how `get_component` currently works, this shouldn't cause a huge latency issue)
- Assignment of the required `reader` (based on registry location type) will need to be pushed into `component_registry.py` and out of the parser classes (this helps to clear up the function of the parsers as well, which is simply "to parse" given content)

Questions:
- What is the best way to grab the details/location of the individual component specs from a given multi-valued registry location?
- How will we handle what a user could enter for a directory for file-based registries? Will the directory have to be somewhere accessible from the JL root dir?
  - Handling our preconfigured components will be different from handling the user-created registries (same as with preconfigured runtime-images, but functionality with parsing registries is much more involved)
- Will all components in given registry be required to be only-KFP or only-Airflow? If not, how to determine which parser will be used to parse the component specs? (allowing this sort of defeats the purpose of having per-processor parsing)

---

### 2. Moving component registry to metadata

**Motivation**

Allows a user to add/modify/delete their own component registries, as explained in the above section. This is fairly high importance/impact as it will significantly improve the user experience for customizing their component list.

(If we decide to not implement component registry definitions and instead stick to defining individual components and their locations, most things in this section still apply with minor changes to the schema and other details.)

**High-level direction**

The potential design raises some questions because components (and the preprocessing required to define them) are quite different from the other metadata namespaces that we currently store and access.

At high level, I think it might be best to keep much of the component parsing logic as-is. The metadata service has its own fetch endpoints, but we can't really use these due to the difference mentioned above. I think the best way to proceed would be to continue to use the existing palette and properties API endpoints (in `pipeline/handlers`) that invoke the processors to parse and return their component details. The majority of changes will occur in the `component_registry.py` `_read_component_registry` function, which will call `MetadataManager.get_all()` to retrieve registry details and loop through specs to construct `Component` objects as it does currently. And then the Metadata API endpoints to add/modify (PUT method), or delete (DELETE method) can be used for these additional operations.

Requirements:

- Create a `component-registries` namespace 
- Create a `component-registries` schema and add to `metadata/schemas`
    - Proposal:
    ```json=
    {
      "display_name": "Built-in KFP Components",
      "metadata": {
        "runtime":  "kfp",       *any supported runtime processor
        "description": "Some KFP components",
        "locations": {
             "location_path": "some/path/to/kfp/registry/filename.yaml",
             "location_type": "file",     *one of ['file', 'url', 'directory'],
             "categories": ["category_name1", "category_name2"],  *any arbitrary string names that will translate into category names
        }        
      },
      "schema_name": "component-registry"
    }
    ```
- Add JSON schema instances for each predefined component registries and move to `etc/config/metadata/component-registries`, as explained in the above section (these will be loaded into the share folder on build as they are currently)
- Add a GUI (not strictly necessary to start, but obviously is the big user-experience piece)
- We can also handle cache update considerations in a future PR, but we will eventually want to ensure that when a user adds/updates a component registry that the new components populate in the palette ASAP

Questions/Considerations:

- The given proposal confines users to assign the same category to every component for a the given location -> is this acceptable?
- The given proposal confines users to assign the same processor/runtime to every component in the given registry -> is this acceptable? If not, further questions are raised as to how runtime would have to be determined and how to do per-processor parsing

---

### 3. Handling component categories

**Motivation**

Allows a user to assign multiple categories to a single component or set of components.

Moderate importance. Would allow users to organize components in a way that makes sense to them.

**High-level direction**

We want to highly simplify the way that categories are fetched and rendered today, but also include the option for the user to enter some list of categories (that will be translated to a list of strings).

Currently, we only have two component categories: `KFP` and `Airflow`. We can have the preconfigured components assigned to these categories, with the option for the user to edit the registry definition to change the category.

Requirements:

- Remove the `ComponentCategory` class and all functions that fetch or create such an instance
- Change the `categories` attribute on a `Component` object to be a list of strings
- Change how the palette is rendered in `to_canvas_palette` and in the jinja template to organize components by category (components that have more than one category will therefore be rendered multiple times)

Questions/Considerations:

- Removing the `ComponentCategory` class will result in losing the ability to add a description to the category (viewable when hovering over a component in the palette)
- What category name do we want to assign to components that have no assigned category?

---

### Additional changes required
- Documentation will have to be updated (see discussion linked above)
- Tests will have to be updated; major changes will be required for the component parser tests, and minor changes to the handlers and other pipeline tests may be required based on the new approach
- The validation service will have to be tweaked so as not to refer to `ComponentCategory` objects (details to be added)
- `processor_kfp.py` may require minor changes to ensure that filename-based components are loaded correctly according to their stored path

kiersten-stokes:
Fixes #2031.
Fixes #1878.

This PR has been modified from the design laid out in #2031. 

High-level changes:

- A new metadata schemaspace is added: `component-registries`
  - The `component-registry` schema (`elyra/metadata/schemas/component-registry.json`) has been updated based on discussion to include a list of component definition locations of a single type (either `url`, `filename`, or `directory`) 
    - Note that this makes it impossible to have a single registry for all of our pre-loaded Airflow components, as 5 of them come from URL-based sources and 1 comes from a filesystem-based source (the `SlackOperator`)
      - We now have 2 Airflow registries: one for the URL-based components and one for the filename-based component
- The `ComponentRegistry` now fetches all instances of component-registries and filters according to the runtime (recall that each runtime processor contains its own `[Cached]ComponentRegistry`), then loops through each to construct individual component entries for parsing
  - This replaces the previous loop that constructed component entries from the runtime component catalogs
  - The `_read_component_registry()` function includes the bulk of the changes because adding support for multi-valued component registries required reorganizing how and when component entries are constructed and sent for parsing
    - One side effect is that the component reader assignment is pulled out of the `ComponentParser` class and into the `ComponentRegistry` class; this has the added benefit of giving the `ComponentParser` class a clearer purpose (it now _only_ parses a given component definition, it does not do any reading)
    - Another side effect of the new registry logic is that we lose the ability to parse for specific Airflow classes (in the case where one file contains multiple operators), the `catalog_entry_id` attribute of a `Component` can now be removed, as can the logic to parse for only a given class/operator only
    -  `_read_component_registry()`has been renamed to `_read_component_registries()`
- New `ComponentReader` classes were added to `component.py` that support reading multi-valued locations
  - The `DirectoryComponentReader` class is the only one implemented right now
  - The `GitHubComponentReader` class can be implemented in a followup PR
- Component catalogs were removed and replaced with equivalent schema instances of `component-registry`, which now reside in `etc/config/metadata/component-registries`
- The `ComponentCategory` class has been removed in favor of being a string-value on the `Component` object
  - The `Component` `categories` attribute is now a list of strings
  - The `to_canvas_palette()` function then organizes components according to their category in order for proper rendering of the palette template
  - Support is included for components that do not specify a category, and are automatically grouped into a category currently named "No Category"
- Provides a GUI for registries

Documentation will have to be updated, specifically discussion #1881. We should wait on these updates until functionality is finalized. The bulk of future documentation changes will likely occur once #1880 is supported in full. This PR provides a GUI to get the user started in the meantime.

- The existing component parser tests have been updated
- 4 new tests have been added (2 for Airflow and 2 for KFP) that test:
  - addition/modification/deletion of a component registry and the resulting components
  - parsing of the new directory-based registry type
 
           in the file; or

           it.


lresende:
We might want to validate the schema with the front-end guys to see if they would need to enhance support on the metadata editor. Below is what I see with the current schema:

image

kiersten-stokes:
> We might want to validate the schema with the front-end guys to see if they would need to enhance support on the metadata editor. Below is what I see with the current schema:

Agreed! The schema can be further refined if we like its current general format, and then from there see if we need additional support (or if there is some way we could work around additional support).

kevin-bates:
This is probably due to the `locations` property, which is both multi-valued and object-valued.   I suspect the latter is the issue.

kiersten-stokes:
So I prototyped pulling type into the main set of `metadata` properties and adjusted our preloaded registries accordingly. Some notes:

- I think we'll still require some frontend support to get this "array of paths" input type working (though probably a much simpler change than the previous structure would have)
- We may want to look at how to best do `Categories` -- I currently have the "categories" key set up to be `"field_type": "tags"` in the schema, but this causes the frontend to display as `Tags` instead of `Categories` 
- We sort of lose the ability to support our preloaded Airflow operators in a single registry (since we use mostly URL-based operators, but one file-based operator as well)
  - Solutions include having 2 separate registries for Airflow preloaded operators or to move all desired preloaded operators into a directory (like how we do for KFP)
- Logic has barely changed -- some minor simplification was made possible with this new schema

elyra/metadata/schemas/component-registry.json
```diff
@@ -0,0 +1,75 @@
...
+          "title": "Description",
+          "description": "Description of this Component Registry",
+          "type": "string"
+        },
+        "runtime": {
+          "title": "Runtime",
+          "description": "The runtime associated with this Component Registry",
+          "type": "string"
+        },
+        "type": {
```
kevin-bates, line 40: Could we call this `location_type` since `type` is already overused?

elyra/pipeline/component.py
```diff
@@ -114,21 +114,27 @@ class Component(object):
     Represents a generic or runtime-specific component
     """
 
-    def __init__(self, id: str, name: str, description: Optional[str], source_type: str,
-                 source: str, catalog_entry_id: str, runtime: Optional[str] = None, op: Optional[str] = None,
-                 category_id: Optional[str] = None, properties: Optional[List[ComponentParameter]] = None,
+    def __init__(self, id: str, name: str,
+                 description: Optional[str],
+                 source_type: str,
+                 source: str,
```
kevin-bates, line 120: I think we should try to match schema (for schema match code) so would recommend `location_type` and `location` for these.

elyra/pipeline/component.py
```diff
@@ -114,21 +114,27 @@ class Component(object):
...
+                 op: Optional[str] = None,
+                 categories: Optional[List[str]] = None,
+                 properties: Optional[List[ComponentParameter]] = None,
                  extensions: Optional[List[str]] = None,
                  parameter_refs: Optional[dict] = None):
         """
         :param id: Unique identifier for a component
         :param name: The name of the component for display
         :param description: The description of the component
+        :param source_type: Indicates the type of component definition resource location; one of ['url', filename']
```
kevin-bates, line 131: Should probably update the enum to include "'directory'".

elyra/pipeline/component.py
```diff
@@ -141,11 +147,10 @@ def __init__(self, id: str, name: str, description: Optional[str], source_type:
         self._description = description
         self._source_type = source_type
         self._source = source
-        self._catalog_entry_id = catalog_entry_id
 
         self._runtime = runtime
         self._op = op
-        self._category_id = category_id
+        self._category_ids = categories
```
kevin-bates, line 153: We can remove the conditional expression in the `categories` property handler via an `or` clause.
```suggestion
        self._category_ids = categories or []
```

elyra/pipeline/component.py
```diff
@@ -275,73 +236,158 @@ class ComponentReader(LoggingConfigurable):
     def type(self) -> str:
         return self.type
 
+    @property
+    def base_type(self):
+        return self.type
```
kevin-bates, line 241: what's the difference between `type` and `base_type`?

kevin-bates, line 241: What's the difference between `type` and `base_type`?

elyra/pipeline/component.py
```diff
@@ -275,73 +236,158 @@ class ComponentReader(LoggingConfigurable):
...
+
+            for filename in os.listdir(absolute_path):
+                if filename.endswith(tuple(parser_file_types)):
+                    absolute_paths.append(os.path.join(absolute_path, filename))
+
+        return absolute_paths
+
+    @property
+    def base_type(self):
+        return super().type
```
kevin-bates, line 322: Hmm - so this is returning `filename` - is that correct?  We might want a comment about that.

elyra/pipeline/component.py
```diff
@@ -275,73 +236,158 @@ class ComponentReader(LoggingConfigurable):
...
+    Read component definitions from a github repo
+    """
+    type = 'github'
+
+    def get_list_of_paths(self, paths: List[str], parser_file_types: List[str]) -> List[str]:
+        pass
+
+    @property
+    def base_type(self):
+        return super().type
```
kevin-bates, line 359: Same as Directory, and returns 'url'.  Please add a comment.

elyra/pipeline/component.py
```diff
@@ -275,73 +236,158 @@ class ComponentReader(LoggingConfigurable):
...
 class ComponentParser(LoggingConfigurable):  # ABC
-    _readers = {
-        FilesystemComponentReader.type: FilesystemComponentReader(),
-        UrlComponentReader.type: UrlComponentReader()
-    }
+    _type = None
+    _file_types = None
+
+    @property
+    def type(self) -> str:
```
kevin-bates, line 367: YAT - yet another 'type'.

kevin-bates: Looking below, it would be awesome if we could rename this to `runtime_platform` or `component_platform` or `platform` or ... anything but `type`.

elyra/pipeline/component_parser_airflow.py
```diff
@@ -164,21 +152,15 @@ def get_runtime_specific_properties(self) -> List[ComponentParameter]:
...
-        """
-        reader = self._get_reader(registry_entry)
-        component_definition = reader.read_component_definition(registry_entry)
-
-        return component_definition
+        return [
+            ComponentParameter(
+                id="runtime_image",
+                name="Runtime Image",
+                type="string",
```
kevin-bates, line 159: I'm on a mission - thou shall not use the name `type` for anything! :smile:
```suggestion
                data_type="string",
```

elyra/pipeline/component_registry.py
```diff
@@ -151,134 +170,104 @@ def to_canvas_properties(component: Component) -> Dict:
...
+            raise
+        except Exception:
+            raise
+
+        for registry in runtime_registries:
+            registry_name = registry['display_name']
+            self.log.debug(f"Component registry: processing components in registry '{registry_name}'")
+
+            # Assign reader based on the type of the registry entry (file, directory, url)
+            registry_type = registry['metadata']['type'].lower()
```
kevin-bates, line 197: Is this `location_type`?

kevin-bates (commented):
Thanks @kiersten-stokes.  I think we (as a project) should focus on removing redundant names of things.  In this case, I'm referring to the use of `type` in LOTS of places - many of which are inherited, so previously existed well before these changes.  Folks should be able to look at code and know what variables are and having too many things with the same names makes that extremely difficult.

elyra/pipeline/component.py
kiersten-stokes: Good question. `type` corresponds to either `url`, `directory`, or `filename` (and whichever other reader classes we introduce eventually). It is used to assign the appropriate reader for that class since they each have their own considerations when accessing locations. 

`base_type` is one of either `filename` or `url` (where directory's `base_type` is `filename` and an eventual github registry's `base_type`, e.g., would be `url`). This is what is assigned to the component as it's `source_type`. The reason for this is that the `source_type` field is used during processing of the KFP components. We look up the component and it's `source_type` in the registry cache, then the component is loaded via `load_component(url='someurl')` or `load_component(filename='somefile')`

elyra/pipeline/component.py
kiersten-stokes: Good point. Assuming my comment above explaining the difference between `type` and `base_type` doesn't change the way we want to do things, I'll definitely add that!

elyra/pipeline/component.py
kiersten-stokes: Agreed!

elyra/pipeline/component_registry.py
kiersten-stokes: Yes!

elyra/pipeline/component.py
kevin-bates: I see. So perhaps we could change `type` to `location_type` and `base_type` to `resource_type` with the explanation relative to the new names.  `location_type` indicates how to locate resources while `resource_type` indicates how to read the resource.

elyra/pipeline/component.py
kevin-bates: Since we're adding GitHub location support, we should add `GitHub` to the enum for `location_type` in the schema.

ptitzler:
- Could we use a different/unique icon? 

   image

- Could we also add an icon to the VPE tool bar to be consistent?
   image

ptitzler:
Trying to get a head start on https://github.com/elyra-ai/elyra/issues/1873 (documenting this), I looked at the UI proposal briefly and have a few usability concerns about the properties view. 

image.

- Currently the wording in the UI proposal seems to imply that the entity being created is a component registry. This is not the case though if one adds a single component.
- Runtime should be an enum, not a free form text to reduce the potential for errors
- Tags should be grouped with name/description, as it is done for all other metadata types, e.g. RTCs
   
   image

kevin-bates:
> Runtime should be an enum, not a free form text to reduce the potential for errors

Good point.  This will require a [`SchemaFilter`](https://github.com/elyra-ai/elyra/blob/39776b449abd0e7829e57e89f2d9a9fe78c093af/elyra/metadata/schema.py#L34) for this schema.  The question is, what do you we use as a placeholder for the `enum` in the _factory_ schema file since it must be valid schema from the factory and `enum` entries must have at least one value.

Regardless of the approach, we should introduce the `"$comment"` entry to let readers know this particular field is dynamic.  (The front-end JSON schema-driven editor should ignore `"$comment"` entries.

Here's the current _factory schema_ entry...
```JSON
        "runtime": {
          "title": "Runtime",
          "description": "The runtime associated with this Component Registry",
          "type": "string"
        },
```
Here are some choices:

- Use a placeholder of `"local"`.  This is not technically incorrect.
```JSON
        "runtime": {
          "title": "Runtime",
          "description": "The runtime associated with this Component Registry",
          "type": "string",
          "$comment": "This enum is dynamically generated to contain the available runtime values.",
          "enum": ["local"]
        },
```

- Use a well-known placeholder name like `"{schema-filter-value}"`. 
```JSON
        "runtime": {
          "title": "Runtime",
          "description": "The runtime associated with this Component Registry",
          "type": "string",
          "$comment": "This enum is dynamically generated to contain the available runtime values.",
          "enum": ["{schema-filter-value}"]
        },
```
Actually, the templated string doesn't need to be "well-known" at all as only the registered schema filter needs to know that it should be _replaced_. That is, nothing in the metadata service would act on `{schema-filter-value}`.

Of these two options, I prefer the latter.

elyra/metadata/schemas/component-registry.json
```diff
@@ -0,0 +1,75 @@
...
+          "title": "Location Type",
+          "description": "The type of location for this registry",
+          "type": "string",
+          "enum": ["URL", "Filename", "Directory"],
```
kevin-bates, line 44: Are we supporting GitHub?

kiersten-stokes: I know we want to eventually. I think in the issue, it says that we'll set the groundwork with directory, filename, and url types for now, and add GitHub support after. If it's trivial to support, I definitely think we should try to get it into this PR. But if not and we want to get this into 3.1.0, I can do an immediate follow up to add GitHub as well

elyra/pipeline/component.py
```diff
@@ -223,133 +232,179 @@ def _log_warning(msg: str, logger: Optional[Logger] = None):
...
+        raise NotImplementedError()
 
-    @property
-    def description(self) -> Optional[str]:
-        return self._description
+    @abstractmethod
+    def get_list_of_paths(self, paths: List[str], parser_file_types: List[str]) -> List[str]:
+        """
+        Returns a list of absolute paths to component specification file(s)
+        based on the array of potentially relative location given
```
kevin-bates, line 263:
```suggestion
        based on the array of potentially relative locations given
```

elyra/pipeline/component.py
```diff
@@ -223,133 +232,179 @@ def _log_warning(msg: str, logger: Optional[Logger] = None):
...
+        Read an absolute location to get the contents of a component
+        specification file
+        """
+        raise NotImplementedError()
 
-    @property
-    def description(self) -> Optional[str]:
-        return self._description
+    @abstractmethod
+    def get_list_of_paths(self, paths: List[str], parser_file_types: List[str]) -> List[str]:
```
kevin-bates, line 260: Do `parser_file_types` ever come into play?  ~I didn't catch their use in the existing implementations.~

`parser_file_types` appear to only apply to directory readers (and perhaps GitHub once implemented).  Should these values be fed into the appropriate constructors when the reader instance is instantiated?  Are these values static between a given component parser instance and its reader?  I think it would be good to remove these from the abstract method definition unless they apply to all instances.

Also, since "path" is very filesystem-ish, I wonder if a name like `get_absolute_locations` might be more generic since it could also apply to URLs?  It would take a list of "potentially relative" locations and return a list of absolute locations.

The directory-based readers would then apply the file-types filters that were provided at their construction.

marthacryan:
> * Currently the wording in the UI proposal seems to imply that the entity being created is a component registry. This is not the case though if one adds a single component.
Changed to "Pipeline Components"

> * Runtime should be an enum, not a free form text to reduce the potential for errors

Done

> * Tags should be grouped with name/description, as it is done for all other metadata types, e.g. RTCs

Done

> * For clarity the label should be "manage pipeline components" in the command list
>   image

Done

> * The tool tip text is inconsistent with the exiting tool tips being used for other Elyra tabs. Perhaps we should use "Pipeline components" to align with "Runtimes" and "Runtime images"
>   image

Done

> * The heading in the panel on the left should probably be "Pipeline components" to be consistent with the naming convention used for other metadata types.

Done

> * We need to add a link to the documentation topic, like it was done here: https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/runtime-image.json#L10 (link to https://elyra.readthedocs.io/en/stable/user_guide/pipeline-components.html)

@kiersten-stokes I think this needs to be added on the backend?

marthacryan:
Thoughts on icons? Here are a few options (feel free to check here for more options https://www.carbondesignsystem.com/guidelines/icons/library/): 
image
image
image

ptitzler:
> Thoughts on icons? Here are a few options (feel free to check here for more options https://www.carbondesignsystem.com/guidelines/icons/library/):

How about this, which does fit the bill _based on its description_ a bit more, because every component is executed in a container:
image

That said, from a depiction point of view there are several icons that would probably work just as well because they convey something similar.

packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -552,6 +559,22 @@ export class MetadataEditor extends ReactWidget {
...
         </div>
       );
+    } else if (uihints.field_type === 'array') {
+      return (
+        <ArrayInput
+          label={this.schema[fieldName].title}
+          description={this.schema[fieldName].description}
+          key={`${fieldName}TextInput`}
+          fieldName={fieldName}
+          defaultValues={this.metadata[fieldName] || defaultValue || []}
```
bourdakos1, line 569:
```suggestion
          defaultValues={this.metadata[fieldName] ?? defaultValue ?? []}
```

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+  List,
+  ListItem,
+  ListItemText,
+  TextField
+} from '@material-ui/core';
+import { FormHelperText } from '@material-ui/core';
+import produce from 'immer';
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+import React from 'react';
```
bourdakos1, line 31: I think we've been using `React.useXXX` in the main elyra repo
```suggestion
import * as React from 'react';
```

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+export const ArrayListItem: React.FC<IListItemProps> = ({
+  value,
+  isEditing,
+  placeholder,
+  onSubmit,
+  onCancel,
+  onDelete,
+  onEdit
+}: IListItemProps) => {
+  const inputRef = useRef<HTMLInputElement>(null);
```
bourdakos1, line 109:
```suggestion
  const inputRef = React.useRef<HTMLInputElement>(null);
```

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  useEffect(() => {
```
bourdakos1, line 111:
```suggestion
  React.useEffect(() => {
```

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+  }, [isEditing]);
+
+  if (isEditing) {
+    return (
+      <div>
+        <TextField
+          inputProps={{ ref: inputRef }}
+          defaultValue={
+            typeof value === 'string'
+              ? value ?? ''
```
bourdakos1, line 126: if `value` is a string  we shouldn't need `?? ''`

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+              : JSON.stringify(value, null, '\t')
```
bourdakos1, line 127: Is there a reason why we are pretty stringifying this? I'm assuming this string will end up in another json string in which case it will end up a 1-liner anyway

packages/ui-components/src/StringArrayInput.tsx
```diff
@@ -0,0 +1,293 @@
...
+  return (
+    <ListItem
+      onDoubleClick={(): void => {
+        onEdit?.();
+      }}
+    >
+      <ListItemText style={{ whiteSpace: 'pre' }}>
+        {typeof value === 'string'
+          ? value ?? ''
+          : JSON.stringify(value, null, '\t')}
```
bourdakos1, line 171: same as above

bourdakos1 (changes requested):
looks good, just a couple nitpicks

packages/ui-components/src/StringArrayInput.tsx
marthacryan: Oh yeah this is leftover from my settings UI work, I'll remove it cause we don't need it here

bourdakos1 (approved):

marthacryan:
Still need to push some style changes

kevin-bates (commented):
First, I want to thank you for the excellent description - that was very helpful.  I just had a few comments, mostly on the minor side.  Relative to your description, this comment stood out to me:
> Note that this makes it impossible to have a single registry for all of our pre-loaded Airflow components, as 5 of them come from URL-based sources and 1 comes from a filesystem-based source (the SlackOperator); currently, I have only included 1 Airflow registry that contains the URL-based components

I think we should go ahead and add the second registry to ensure there aren't going to be any short-comings with that approach.  That said, I suppose this could be deferred to the next release assuming the lack of the slack operator is not a big deal since we're trying to wrap this up.  

Does adding additional registries for a given runtime introduce issues elsewhere?  Just curious why we didn't do this now.

Another comment in the description...
> Another side effect of the new registry logic is that we lose the ability to parse for specific Airflow classes (in the case where one file contains multiple operators)

What do we do in these cases?  Is one component the "chosen one"?

Thank you.

elyra/pipeline/component.py
```diff
@@ -223,133 +232,176 @@ def _log_warning(msg: str, logger: Optional[Logger] = None):
...
-            return f.read()
+            for filename in os.listdir(absolute_path):
+                if filename.endswith(tuple(self.file_types)):
+                    absolute_paths.append(os.path.join(absolute_path, filename))
+
+        return absolute_paths
+
+    @property
+    def resource_type(self):
+        return super().location_type
```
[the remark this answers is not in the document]
kiersten-stokes: Oh you're right -- I completely forgot to add this from a prior review! will do

etc/config/metadata/component-registries/elyra-airflow-preconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "display_name": "Airflow Preloaded Components",
+  "metadata": {
+    "description": "Preloaded components that are supported by Apache Airflow",
+    "runtime": "airflow",
+    "location_type": "URL",
```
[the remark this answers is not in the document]
kiersten-stokes: I like that!

etc/config/metadata/component-registries/elyra-kfp-preconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "display_name": "KFP Preloaded Components",
+  "metadata": {
+    "description": "Preloaded components that are supported by Kubeflow Pipelines",
+    "runtime": "kfp",
+    "location_type": "Directory",
+    "categories": ["Preloaded KFP"],
+    "paths": ["kfp"]
```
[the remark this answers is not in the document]
kiersten-stokes: Yeah this gets into that sticky situation where we load the preconfigured filename-based components from `share/jupyter/metadata/components` (they are moved there from `etc/config/components` on build). So the filename-based readers (and all subclasses) check first to see if the entered path is an absolute path, and if not, assumes the location to be at `os.path.join(ENV_JUPYTER_PATH[0], 'components', path)`. If the resulting path is still inaccurate, the file or directory is skipped and logged as such.

kiersten-stokes:
> > Note that this makes it impossible to have a single registry for all of our pre-loaded Airflow components, as 5 of them come from URL-based sources and 1 comes from a filesystem-based source (the SlackOperator); currently, I have only included 1 Airflow registry that contains the URL-based components
> 
> I think we should go ahead and add the second registry to ensure there aren't going to be any short-comings with that approach. That said, I suppose this could be deferred to the next release assuming the lack of the slack operator is not a big deal since we're trying to wrap this up.
> 
> Does adding additional registries for a given runtime introduce issues elsewhere? Just curious why we didn't do this now.

No, you can really have as many as you want! The only "downside" to having the 2 preconfigured registries for Airflow is just that it's somewhat inconsistent with KFP at that point. But on the other hand, it does provide a nice example of a filename-based registry so that we have representation for all 3 (current) types: `filename`, `directory`, and `url`. It would only take a few minutes to add, so I can easily do that now. 

> > Another side effect of the new registry logic is that we lose the ability to parse for specific Airflow classes (in the case where one file contains multiple operators)
> 
> What do we do in these cases? Is one component the "chosen one"?

I probably should have been more clear in that this doesn't have any noticeable side effects. Here's an example that might clarify the change: we used to be able to call `parse()` for a single component and it would only parse the python lines corresponding to that Operator class. Now, `parse()` can only be called for one file at a time (which may have multiple Operator classes) and will return a list of `Component` objects (one for each operator class found during parsing). That could potentially be an issue if a file has a huge number of classes/operators defined within it because there could be a bit of a lag, but we haven't found a case like that yet.

elyra/pipeline/component_registry.py
```diff
@@ -151,134 +166,111 @@ def to_canvas_properties(component: Component) -> Dict:
...
+            for path in component_paths:
+                # TODO Figure out what would be the best path to display to the user
+                # TODO when accessing the node properties panel, since components can
+                # TODO now come from myriad locations
+
+                # Read in contents of the component
+                component_definition = reader.read_component_definition(path)
+
+                component_entry = {
+                    "location_type": reader.resource_type,
```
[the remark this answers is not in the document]
kiersten-stokes: Yeah I had the same thought. I'll put on my list to open a new issue to infer `location_type`. We could probably use another issue to add GitHub repo support as well

etc/config/metadata/component-registries/elyra-kfp-preconfig.json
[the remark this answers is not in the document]
kiersten-stokes: If we wanted to display this in such a way that it might be less confusing for the user, we could potentially add some logic to (what is currently the) SchemaFilter where it just replaces the relative path with the absolute path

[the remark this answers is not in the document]
kevin-bates: That won't work because this is actually in the _instance_ of the schema, not the schema itself.  However, schemas can reference `metadata_class_name` values that essentially derive from [`Metadata`](https://github.com/elyra-ai/elyra/blob/7b286ecbcf9b709391da3693798fe2e2e8162cc9/elyra/metadata/metadata.py#L31) that have `post_load` and `pre_save` method overrides, so such a class _could_ look for these kinds of things - validate their location, and update the "relative location".  The metadata-test schemas [show an example](https://github.com/elyra-ai/elyra/blob/7b286ecbcf9b709391da3693798fe2e2e8162cc9/elyra/metadata/schemas/metadata-test.json#L7) if we want to entertain this.

kevin-bates (approved):
Thanks for the updates!

etc/config/metadata/component-registries/elyra-kfp-preconfig.json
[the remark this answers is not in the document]
kiersten-stokes: Oh right! Not sure what I was thinking there, I think I was confusing it with the `post_load` method itself. Anyway, yeah maybe just another thing to keep in mind for a follow up

PR #2086, opened 2021-08-31
Expose pipeline description to runtime environments

ptitzler:
Expose user-provided pipeline description in the runtime environments if one was provided, otherwise the current behavior is retained and a generic text displayed:  

image

Example with user-provided description in KFP:

image

Example with user-provided description in AA:

image

This is a follow-up to https://github.com/elyra-ai/elyra/pull/1708, which introduced VPE support for a pipeline description property:
 - extend pipeline parser to recognize a user-provided description
 - extend pipeline class by adding a description property (populated by the parser)
 - pass-through user-provided pipeline description to runtime environments (kfp/aa) instead of static string

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -377,7 +377,7 @@ class Pipeline(object):
         return self._operations
 
     @property
-    def description(self) -> str:
+    def description(self) -> Optional[str]:
         """
         Pipeline description
         """
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -312,7 +312,13 @@ class Pipeline(object):
     Represents a single pipeline constructed in the pipeline editor
     """
 
-    def __init__(self, id: str, name: str, runtime: str, runtime_config: str, source: Optional[str] = None):
+    def __init__(self,
+                 id: str,
+                 name: str,
+                 runtime: str,
+                 runtime_config: str,
+                 source: Optional[str] = None,
+                 description: Optional[str] = None):
         """
         :param id: Generated UUID, 128 bit number used as a unique identifier
                    e.g. 123e4567-e89b-12d3-a456-426614174000
@@ -322,6 +328,7 @@ class Pipeline(object):
                         e.g. kfp OR airflow
         :param runtime_config: Runtime configuration that should be used to submit the pipeline to execution
         :param source: The pipeline source, e.g. a pipeline file or a notebook.
+        :description: Pipeline description
         """
 
         if not name:
@@ -333,6 +340,7 @@ class Pipeline(object):
 
         self._id = id
         self._name = name
+        self._description = description
         self._source = source
         self._runtime = runtime
         self._runtime_config = runtime_config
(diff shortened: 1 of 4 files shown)


elyra/pipeline/pipeline.py
```diff
@@ -368,11 +376,19 @@ def runtime_config(self) -> str:
     def operations(self) -> Dict[str, Operation]:
         return self._operations
 
+    @property
+    def description(self) -> str:
```
kiersten-stokes, line 380: NIT: I think this should be:
```suggestion
    def description(self) -> Optional[str]:
```

kiersten-stokes (approved):
Tested with KFP and Airflow for both export and submit, and all are working as expected!

elyra/pipeline/processor_kfp.py
```diff
@@ -209,7 +209,10 @@ def process(self, pipeline):
             # Upload the compiled pipeline, create an experiment and run
 
             try:
-                description = f"Created with Elyra {__version__} pipeline editor using '{pipeline.source}'."
+                if pipeline.description is not None:
+                    description = pipeline.description
+                else:
+                    description = f"Created with Elyra {__version__} pipeline editor using `{pipeline.source}`."
```
kevin-bates, line 215: I'm curious why you didn't take the same approach as the other two locations and pre-assign the variable to avoid the `else` block?

Hmm - since the same string is generated when `pipeline.description` is not set and since both values (`__version__` and `pipeline.source`) are available from the property accssor, should the property accessor always return a value and we can eliminate these three locations altogether?

elyra/pipeline/pipeline.py
ptitzler: Thank you! Fixed!

elyra/pipeline/processor_kfp.py
ptitzler: > I'm curious why you didn't take the same approach as the other two locations and pre-assign the variable to avoid the else block?

No reason, but I've changed it now to be consistent. I also renamed the variable to be consistent.

> Hmm - since the same string is generated when pipeline.description is not set and since both values (__version__ and pipeline.source) are available from the property accessor, should the property accessor always return a value and we can eliminate these three locations altogether?

I hadn't thought about that. Our original intention was to always provide a "description" that would enable the user to associate a DAG in the runtime environment with a pipeline. Since descriptions were not supported at that time we decided to use the static string instead. If we were to alter the getter to return the static string this might create issues in the future should other pieces of code start utilizing `Pipeline` objects. 

For example, one could argue that users should always provide a pipeline description. If pipeline validation were to use the `Pipeline` class, it would not be able to determine whether a description was provided by the user and thus not produce a [best practice] hint.

kevin-bates: This particular `Pipeline` class is downstream from validation and represents the _runtime representation_ of the pipeline and contains `Operation` instances, rather than _nodes_.  The `Pipeline` class you're referring to (:smile: ) is the `Pipeline` class contained within the `PipelineDefinition` hierarchy that @lresende is introducing with #2082.  I don't think there are plans to utilize `PipelineDefinition.Pipeline` where `Pipeline.Pipeline` instances are used today (and certainly not the reverse).  As a result, we're past the point of needing to detect whether or not a (future) required value is missing.

lresende (approved):

PR #2094, opened 2021-09-02
Support data exchange between custom components

akchinSTC:
Update the component registry response with additional
type information for kfp specific components

**Component Registry**
Actual Output:
```
{
    "current_parameters": {
        "label": "",
        "component_source": "kfp/run_notebook_using_papermill.yaml",
        "elyra_notebook": "",
        "elyra_parameters": "{}",
        "elyra_packages_to_install": "[]",
        "elyra_input_data": "",
        "elyra_notebook_2": "",
        "elyra_output_data": ""
    },
    "parameters": [
        {
            "id": "label"
        },
        {
            "id": "component_source"
        },
        {
            "id": "elyra_notebook"
        },
        {
            "id": "elyra_parameters"
        },
        {
            "id": "elyra_packages_to_install"
        },
        {
            "id": "elyra_input_data"
        },
        {
            "id": "elyra_notebook_2"
        },
        {
            "id": "elyra_output_data"
        }
    ],
    "uihints": {
        "id": "nodeProperties",
        "parameter_info": [
            {
                "parameter_ref": "label",
                "control": "custom",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Label"
                },
                "description": {
                    "default": "A custom label for the node.",
                    "placement": "on_panel"
                },
                "data": {}
            },
            {
                "parameter_ref": "component_source",
                "control": "readonly",
                "label": {
                    "default": "Component Source"
                },
                "description": {
                    "default": "The path to the component specification file.",
                    "placement": "on_panel"
                },
                "data": {}
            },
            {
                "parameter_ref": "elyra_notebook",
                "control": "custom",
                "custom_control_id": "FoobarControl",
                "label": {
                    "default": "Notebook"
                },
                "description": {
                    "default": "Required. Notebook to execute. (type: JupyterNotebook)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "inputpath",
                    "required": true
                }
            },
            {
                "parameter_ref": "elyra_parameters",
                "control": "custom",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Parameters"
                },
                "description": {
                    "default": "Map with notebook parameter values. (type: JsonObject)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "string",
                    "required": true
                }
            },
            {
                "parameter_ref": "elyra_packages_to_install",
                "control": "custom",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Packages to install"
                },
                "description": {
                    "default": "Python packages to install (type: JsonArray)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "list",
                    "required": true
                }
            },
            {
                "parameter_ref": "elyra_input_data",
                "control": "custom",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Input data"
                },
                "description": {
                    "default": "Optional data that can be passed to notebook. In notebook, the INPUT_DATA_PATH variable will point to the data (if passed). (type: string)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "string",
                    "required": false
                }
            },
            {
                "parameter_ref": "elyra_notebook_2",
                "control": "readonly",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Notebook"
                },
                "description": {
                    "default": "Executed notebook. (type: JupyterNotebook)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "outputpath",
                    "required": true
                }
            },
            {
                "parameter_ref": "elyra_output_data",
                "control": "readonly",
                "custom_control_id": "StringControl",
                "label": {
                    "default": "Output data"
                },
                "description": {
                    "default": "Directory with any output data. In notebook, the OUTPUT_DATA_PATH variable will point to this directory, so that the notebook can write output data there. (type: string)",
                    "placement": "on_panel"
                },
                "data": {
                    "format": "outputpath",
                    "required": true
                }
            }
        ],
        "group_info": [
            {
                "id": "nodeGroupInfo",
                "type": "panels",
                "group_info": [
                    {
                        "id": "label",
                        "type": "controls",
                        "parameter_refs": [
                            "label"
                        ]
                    },
                    {
                        "id": "component_source",
                        "type": "controls",
                        "parameter_refs": [
                            "component_source"
                        ]
                    },
                    {
                        "id": "elyra_notebook",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_notebook"
                        ]
                    },
                    {
                        "id": "elyra_parameters",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_parameters"
                        ]
                    },
                    {
                        "id": "elyra_packages_to_install",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_packages_to_install"
                        ]
                    },
                    {
                        "id": "elyra_input_data",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_input_data"
                        ]
                    },
                    {
                        "id": "elyra_notebook_2",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_notebook_2"
                        ]
                    },
                    {
                        "id": "elyra_output_data",
                        "type": "controls",
                        "parameter_refs": [
                            "elyra_output_data"
                        ]
                    }
                ]
            }
        ]
    },
    "resources": {}
}
```

- [ ] `id`s have been made unique, to avoid collisions between identical component I/O names
When parsing component parameters to determine the type, `ui_hints``parameter_info``data``format` can be reused instead of creating a new `type` field. The value of of this field will largely remain as-as except for the following:
- [ ] `inputPath`-  to indicate that it will be a reference to a parent `outputPath`
- [ ] `outputPath` - to indicate that this field will provide this output for child nodes. Any component parameters with fields set to this will also be made `read-only` and provide an informational reference to what this component produces. 
- [ ] `foobarControl` controller ID placeholder for `inputPath` fields, until the controller is complete and given a proper name

UI
- [ ] Uniqueness of node labels will be handled by the front end to avoid potential conflicts
- [ ] Component Outputs will be displayed in read-only since the user cannot make any changes to them outside of the component definition, we expose them to provide users with info about what outputs the component provides

Validation
- [ ] Verify that I/O values are of the correct type for the field
- [ ] Verify that inputs, if local have valid paths to files, or valid references to parent outputs

Processor
- [ ] L511-514 - We have a custom type `file` that we are using to read in the contents of a file into `inputPath` (papermill and filter text components) which is not in line with best practices. We should remove this functionality and treat it as a type.
- [ ] Remove `pop` of input and output parameters for support of data exchange

           in the file; or

           it.


elyra/pipeline/component_parser_kfp.py
```diff
@@ -63,42 +65,67 @@ def _parse_properties(self, component_yaml: Dict[str, Any]) -> List[ComponentPar
...
+
+        all_params = {"inputs": input_params, "outputs": output_params}
+
+        # Loop through inputs and outputs and create custom properties
+        for param_type, params in all_params.items():
+            for param in params:
+                # KFP components default to being required unless otherwise stated.
+                # Reference: https://www.kubeflow.org/docs/components/pipelines/reference/component-spec/#interface
+                required = True
+                if "optional" in param and param.get('optional') is True:
```
kevin-bates, line 81: This clause is unnecessary since `None` will be returned if the key doesn't exist and `None` != `True`.
```suggestion
                if param.get('optional') is True:
```

elyra/pipeline/component_parser_kfp.py
```diff
@@ -63,42 +65,67 @@ def _parse_properties(self, component_yaml: Dict[str, Any]) -> List[ComponentPar
...
+                                                                parameter_type=data_type)
+
+                type, control, control_id, required_param, default_value = self.determine_type_information(data_type)
+                if not required_param:
+                    required = required_param
+
+                # Get value if provided
+                value = param.get('default', '')
+
+                ref_name = param.get('name').lower().replace(' ', '_')
```
kevin-bates, line 104: If `name` is required, this is fine, otherwise `AttributeError: 'NoneType' object has no attribute 'lower'` could be raised.
I'm assuming it's safe to assume `name` will be present.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+                        adjusted_type = f"{io_object_type}Path"
 
         return adjusted_type
+
+    def determine_type_information(self, parsed_type: str) -> Tuple[str, str, str, bool, Any]:
+        """
+        Takes the type information of a component parameter as parsed from the component
+        specification and returns a new type that is one of several standard options.
+
+        """
```
kevin-bates, line 184: It looks like the difference in the returned tuple and its super class method is this one returns an extra `control` and `required_param` value.

Just curious what issues would arise if the airflow parser happened to use this same method?  Would the input/output paths be a problem?  It seems like its callers could automatically ignore the non-applicable tuples.

(We might also consider returning a dictionary of these values instead since this length of tuple is a little unwieldy - or see comment below.)

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+            type_lowered = "boolean"
+            control_id = "BooleanControl"
+            default_value = False
+        elif type_lowered in ['dict', 'dictionary']:
+            type_lowered = "dictionary"
+        elif type_lowered in ['list', 'set', 'array', 'arr']:
+            type_lowered = "list"
+        elif type_lowered in ['inputpath']:
+            type_lowered = "inputpath"
+            control_id = "FoobarControl"
```
kevin-bates, line 223: Is this an actual control_id?

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+        elif type_lowered in ['inputvalue']:
+            type_lowered = "inputvalue"
+        elif type_lowered in ['outputpath']:
+            control = "readonly"
+            type_lowered = "outputpath"
+            required_param = False
```
kevin-bates, line 229: Pre the previous comment, It looks like only `outputpath` influences BOTH `control` and `required_param`, all other types correspond to `control == custom` and `required_param == True`.

Just wondering if only the kfp code could use that information after calling the superclass method and something like this...

```python
def determine_type_information(self, parsed_type: str) -> Tuple[str, str, Any]:

    control = "custom"
    required_param = True
    data_type, control_id, default_value = super().determine_type_information(data_type)
    if data_type == "outputpath":
        control = "readonly"
        required_param = False
```

elyra/pipeline/component_parser_kfp.py
akchinSTC: negative, just a placeholder until the UI portions are slotted in.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+            type_lowered = "number"
+            control_id = "NumberControl"
+            default_value = 0
+        elif any(word in type_lowered for word in ['bool', 'boolean']):
+            type_lowered = "boolean"
+            control_id = "BooleanControl"
+            default_value = False
+        elif type_lowered in ['dict', 'dictionary']:
+            type_lowered = "dictionary"
+        elif type_lowered in ['list', 'set', 'array', 'arr']:
```
kevin-bates, line 219: This block is missing the `'file'` case that's in the super() version.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+            type_lowered = "list"
+        elif type_lowered in ['inputpath']:
```
kevin-bates, line 221: This can only be `type_lowered == 'inputpath'` and reads better.  Same goes for `inputvalue`, `outputvalue` and `file` in superclass.

elyra/pipeline/component_parser_kfp.py
```diff
@@ -140,11 +167,67 @@ def _get_adjusted_parameter_fields(self,
...
+        for option in type_options:
+            if any(word + " of " in type_lowered for word in type_options):
+                reg = re.compile(f"({option}) of ")
+                match = reg.search(type_lowered)
+                if match:
+                    type_lowered = option
+                    break
+            elif option in type_lowered:
+                type_lowered = option
+                break
```
kevin-bates, line 199: 1. What is the purpose of the nested for loop across `type_options`?  Isn't the outer loop sufficient?  E.g.,
```python
if f"{option} of" in type_lowered:
```

2. Same goes for the regular expression stuff.  If the 'of' clause is in the parsed type comment don't we know `option` by virtue of the for loop?

3. Regarding the if/elif here.  If `{option} of` is in `type_lowered` then `{option}` is in `type_lowered`, so isn't it sufficient to only ask the latter?

kiersten-stokes: Just for some context, I think Sourcery refactored this clause from my initial implementation, leading to it needing some refinement like you mentioned. 

The only thing I would point out regarding bullet 3 is that the `elif` catches cases where ` of` doesn't exist in `type_lowered`, so I believe we still need it. E.g., to determine data types from descriptions such as `this parameter is an array`

kevin-bates: Thanks for the response Kiersten - that's helpful.

Regarding the 3rd item, I guess my point is that ONLY the `elif` clause is necessary since it's actually a subset of the `if` clause that checks for `of`.  That is, if `option of` is in `type_lowered` then `option` is in `type_lowered`, so isn't only the `elif` portion of those checks sufficient?

I feel like I'm missing something but can't see it.

kiersten-stokes: Ah yes, I see what you're saying! Yes, you're right. I agree that I feel like I might be missing some obscure case, but I think that's just a side effect of this Sourcery refactor making me second guess myself.

akchinSTC:
@ajbozarth 
Node property types can double as both their original primitive/non primitive types as well as an xcom reference for parent outputs. Can we have a custom controller that will have the option to take in both these values but only in an either/or scenario? e.g.
image

ptitzler:
Can we mark output properties more clearly? Currently they are rendered the same way that system properties are rendered, making it impossible to distinguish them.

akchinSTC:
> Can we mark output properties more clearly? Currently they are rendered the same way that system properties are rendered, making it impossible to distinguish them.

image

elyra/pipeline/kfp/component_parser_kfp.py
```diff
@@ -63,40 +63,60 @@ def _parse_properties(self, component_yaml: Dict[str, Any]) -> List[ComponentPar
...
+
+                # Get value if provided
+                value = param.get('default', '')
+
+                ref_name = param.get('name').lower().replace(' ', '_')
+                name = param.get('name')
+
+                if data_type_info.data_type == 'outputpath':
+                    ref_name = f"output_{ref_name}"
+                    description = f"{description}. This is a(n) output of this component."
```
ptitzler, line 109: Not sure if there's a case it'd be `a` instead of `an`, hence

```suggestion
                    description = f"{description}. This is an output of this component."
```

kevin-bates (approved):
Looks good - thanks to all.

ajbozarth:
A detailed in https://github.com/elyra-ai/pipeline-editor/pull/161 I've removed the front end code form this PR. As such this should be good to go. Though we probably want to open a follow up issue to update/add integration tests for this new feature

ptitzler:
As discussed in https://github.com/elyra-ai/pipeline-editor/pull/161#issuecomment-928173933 we should change how the UI indicates that a property is an output property. (The current approach is to append `(Output)` to the property name, which has a trickle down effect because consumers of the name would have to remove this suffix to only render the property name.)

etc/config/metadata/component-registries/elyra-kfp-url-preconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "display_name": "KFP Preloaded Components",
```
ptitzler, line 2: Should this be 
```suggestion
  "display_name": "KFP Preloaded Components - URL",
```

As is there are now two entries with the same name:

image


In addition `component-registries/elyra-kfp-preconfig.json` should probably be named `elyra-kfp-directory-preconfig.json` and use `KFP Preloaded Components - Directory` as display name.

ptitzler: @kiersten-stokes

kiersten-stokes: I like that idea -- it slipped my mind sometime in the last few weeks that the Download Component will have to be permanent now

ptitzler: I only brought this up because this is how it was done for Airflow and we should be consistent (:

kiersten-stokes: I completely agree! :)

kiersten-stokes:
Some things we may want to discuss/change for the inclusion of the component description in the node properties panel:

- I only included the `Component Description` field if a description is provided; the reason for this is that we currently don't really support descriptions for Airflow components so that field would be pretty worthless for an entire runtime right now 
- Related, I added a description to the KFP `Filter text` component
- I did not include a property description for the `Component Description` field since I think its pretty self-explanatory, but again if we want to include one for consistency, e.g., it's very easy to do so

ptitzler:
Thank you Kiersten!

> * I only included the `Component Description` field if a description is provided; the reason for this is that we currently don't really support descriptions for Airflow components so that field would be pretty worthless for an entire runtime right now and I think including the lack of a description would just draw attention to that

Good point. I think they way you have it now is good. 

> * Related, I added a description to the KFP `Filter text` component

👍 

> * I did not include a property description for the `Component Description` field since I think its pretty self-explanatory, but again if we want to include one for consistency, e.g., it's very easy to do so

👍 we are already space constraint in that panel so anything that does not provide value should be minimized/removed.

For one of the tutorial components this is what the node properties view looks like:

image

More than half of the visible screen is already taken up by system/read only metadata, which is going to lead to usability issues because scrolling will be required if a component defines more than two inputs/outputs.

Perhaps we should consider moving the component source metadata to the bottom until we have added a viewer  link to the component source (which will remove the need to display long URLs or path information)?

kiersten-stokes:
> Perhaps we should consider moving the component source metadata to the bottom until we have added a viewer link to the component source (which will remove the need to display long URLs or path information)?

I like that because I'm not sure how useful it is right now. This also will have a nice unintended side effect of really simplifying the properties template (at least for now).

kevin-bates:
> More than half of the visible screen is already taken up by system/read only metadata, which is going to lead to usability issues because scrolling will be required if a component defines more than two inputs/outputs.
>
> Perhaps we should consider moving the component source metadata to the bottom until we have added a viewer link to the component source (which will remove the need to display long URLs or path information)?

Could the _Component Description_ value always be placed immediately under the component's title (e.g., just below "Count Rows" above) and sans the "Component Description" text?
Could we make the component source information part of a hover thing over the title?

kiersten-stokes:
> Could the _Component Description_ value always be placed immediately under the component's title (e.g., just below "Count Rows" above) and sans the "Component Description" text? Could we make the component source information part of a hover thing over the title?

I like that too because I guess the read-only fields should probably be grouped either at the very beginning or the very end, and this will have that effect.

ptitzler:
First time trying to run the tutorial pipelineLooks like validation might have an issue. I just tried to run the (new) tutorial pipeline for the first time:

```
  File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 110, in post
  File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 152, in validate
    response=response)
  File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 319, in _validate_node_properties
    elif self._get_component_type(property_dict, node_property) == 'inputpath':
  File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 730, in _get_component_type
    return prop['data']['format']
KeyError: 'format'
```

elyra/pipeline/validation.py
```diff
@@ -697,6 +714,17 @@ def _is_required_property(self, property_dict: dict, node_property: str) -> bool
...
         for parameter in node_op_parameter_list:
-            if parameter['parameter_ref'] == node_property:
+            if parameter['parameter_ref'] == f"elyra_{node_property}":
                 return parameter['data']['required']
         return False
+
+    def _get_component_type(self, property_dict: dict, node_property: str) -> str:
+        """
+        Helper function to determine the type of a node property
+        :param property_dict: a dictioanry containing the full list of property parameters and descriptions
```
ptitzler, line 724:
```suggestion
        :param property_dict: a dictionary containing the full list of property parameters and descriptions
```

elyra/pipeline/validation.py
```diff
@@ -303,28 +309,40 @@ async def _validate_node_properties(self, pipeline_definition: PipelineDefinitio
...
-                            elif not isinstance(node.get_component_parameter(node_property),
-                                                type(property_dict['current_parameters']['elyra_' + node_property])):
-                                response.add_message(severity=ValidationSeverity.Error,
-                                                     message_type="invalidNodeProperty",
-                                                     message="Node property is incorrect type.",
-                                                     data={"nodeID": node.id,
-                                                           "nodeName": node_label,
-                                                           "propertyName": node_property})
+                            elif self._get_component_type(property_dict, node_property) == 'inputpath':
+                                print(node_property)
```
ptitzler, line 320: remove or user proper debug logging

elyra/pipeline/validation.py
```diff
@@ -303,28 +309,40 @@ async def _validate_node_properties(self, pipeline_definition: PipelineDefinitio
...
+                                print(node.get_component_parameter(node_property))
```
ptitzler, line 321: remove or use proper debug logging

ptitzler:
For reference, this is what the updated node properties panel looks like after the latest round of changes:

image

kevin-bates:
Thanks @ptitzler  (and @kiersten-stokes for making these changes).

When I suggested moving the component description _value_ to the top, I didn't intend for the "Component Description" text to be present.  If the "Component Description" text can't be removed, then I would vote for it to be moved to the bottom as well.  It just seems natural to have the description part of the primary data.   If moved to the bottom, I think it should precede the "Component Source" name/value pair.

ptitzler:
We've discussed the layout during today's dev meeting and settled on the following UI improvements:
- Remove "component description" label
- Add sections/dividers to separate inputs, outputs etc visually
- Remove descriptive text for "component source" (`The path ...`)
- Render the component source value as a single line with scroll bar (to avoid wrapping, which saves space on the screen)

image

elyra/pipeline/kfp/component_parser_kfp.py
```diff
@@ -126,25 +150,54 @@ def _read_component_yaml(self, registry_entry: SimpleNamespace) -> Optional[Dict
...
-                            list(arg.keys())[0] == f"{io_object_type}Path":
-                        adjusted_type = "file"
-
-        return adjusted_type
+        data_type_info = super().determine_type_information(parsed_type)
+        if data_type_info.undetermined:
+            if 'inputpath' in data_type_info.parsed_data:
+                data_type_info.data_type = 'inputpath'
+                data_type_info.control_id = "NestedEnumControl"
+                data_type_info.undetermined = False
```
bourdakos1, line 193: Can we add a line here that sets the default for `inputpath` to be undefined? 
```python
data_type_info.default_value = None
```

kevin-bates: I would agree.  In fact, it "feels" more appropriate to default `default_value` to `None` from the start since its data type is hinted as `Any` and, currently, uses the empty string as the default value - which is only compatible with strings.

Would using `None` for `'inputvalue'` and `'outputpath'` defaults also be appropriate?

I'm thinking the correct default, across all the types, is always `None` and we explicitly set the default to the empty string only where that's appropriate.

akchinSTC:
> First time trying to run the tutorial pipeline it looks like validation might have an issue:
> 
> ```
> Traceback (most recent call last):
>   File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
>     result = await result
>   File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 110, in post
>     response = await PipelineValidationManager.instance().validate(pipeline=pipeline_definition)
>   File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 152, in validate
>     response=response)
>   File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 319, in _validate_node_properties
>     elif self._get_component_type(property_dict, node_property) == 'inputpath':
>   File "/opt/anaconda3/envs/kfp-data-exchange/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 730, in _get_component_type
>     return prop['data']['format']
> KeyError: 'format'
> ```

@ptitzler - was this resolved? which tutorial pipeline specifically is failing? thx!

elyra/pipeline/kfp/component_parser_kfp.py
akchinSTC: sure thing

ajbozarth:
> was this resolved? which tutorial pipeline specifically is failing? thx! 
this was not resolved, I see it as well when I try to submit a pipeline

elyra/pipeline/kfp/component_parser_kfp.py
bourdakos1: @kevin-bates i would say no, but I’m not 100% sure. It’s dependent on the type of form control it’s assigned. Some controls can’t handle an undefined default.

bourdakos1: In this case it’s always assigned the NestedEnumControl which requires undefined as the default

ajbozarth:
I pushed a few changes to comply with my latest updates to https://github.com/elyra-ai/pipeline-editor/pull/161

elyra/pipeline/kfp/component_parser_kfp.py
ajbozarth: I added @bourdakos1 suggested change (and it solved our issue in https://github.com/elyra-ai/pipeline-editor/pull/161 ). @akchinSTC if you want to change it to a more generic fix feel free

ptitzler:
In `component_parser_kfp.py` we are currently only logging a debug message if an error was encountered trying to parse a component specification:

```
    def _read_component_yaml(self, registry_entry: SimpleNamespace) -> Optional[Dict[str, Any]]:
        """
        Convert component_definition string to YAML object
        """
        try:
            return yaml.safe_load(registry_entry.component_definition)
        except Exception as e:
            self.log.debug(f"Could not read definition for component at "
                           f"location: '{registry_entry.location}' -> {str(e)}")
            return None
```

At a minimum the message severity needs to be warning so that there is no need to enable debug to determine why a component doesn't show up in the palette.

ptitzler:
We forgot to list the outputs in the node's tool tip:

image

akchinSTC:
when creating a new pipeline and using a component with a parameter with type `inputpath` the structure in the payload is still populated with the `label` field resulting in a validation error. 
@ajbozarth or @bourdakos1  - have any ideas why this is still populating even after Alex's changes?
```
{
          "id": "57f4b104-8921-48cb-a48c-3bafdc92c1ed",
          "type": "execution_node",
          "op": "filter_text_using_shell_and_grep_Filtertext",
          "app_data": {
            "component_parameters": {
              "component_description": "Filter input text according to the given regex pattern using shell and grep.",
              "pattern": ".*",
              "output_filtered_text": "",
              "text": {
                "value": "00309868-9bbb-4148-8bbf-ab7bff56c103",
                "option": "elyra_output_data",
                "label": "Download data: Data"
              }
            },

**Comment by ptitzler** [comment]
Open feature request https://github.com/elyra-ai/elyra/issues/2183 to improve the node summary visualization in the Visual Pipeline Editor.

File: elyra/pipeline/pipeline_definition.py [hunk 15]
```diff
@@ -226,6 +227,15 @@ def subflow_pipeline_id(self) -> Pipeline:
         else:
             return None
 
+    @property
+    def get_component_links(self) -> List:
```
ajbozarth, line 231: should this function name include `input` since it only returns the inputs?

elyra/pipeline/validation.py
```diff
@@ -333,6 +333,14 @@ async def _validate_node_properties(self, pipeline_definition: PipelineDefinitio
                                                                        "nodeName": node_label,
                                                                        "propertyName": node_property,
                                                                        "keyName": key})
+                                link_ids = list(x.get('node_id_ref', None) for x in node.get_component_links)
+                                if node.get_component_parameter(node_property)['value'] not in link_ids:
+                                    response.add_message(severity=ValidationSeverity.Error,
```
ajbozarth, line 338: IIUC this only checks for direct parent nodes, but any upstream node is valid (ie parents of parents)

ajbozarth (commented):
a couple comments on the most recent commit

elyra/pipeline/pipeline_definition.py
akchinSTC: Im fine with whatever, but since the links are always in the inputs structure, figured Id name it this

elyra/pipeline/validation.py
akchinSTC: ahh yes, thx for the reminder, leme change this to look recursively

elyra/pipeline/validation.py
```diff
@@ -697,6 +722,33 @@ def _is_required_property(self, property_dict: dict, node_property: str) -> bool
...
+        :param parent_list: the list to add additional found parent node ids
+        :return:
+        """
+        for node_id in node_id_list:
+            node = pipeline_definition.get_node(node_id)
+            if node:
+                parent_list.append(node_id)
+                node_ids = list(x.get('node_id_ref', None) for x in node.get_component_links)
+                self._get_parent_id_list(pipeline_definition, node_ids, parent_list)
+        return parent_list
```
bourdakos1, line 754: I haven't tested this, but it looks like this validation comes before the circular reference check. So we should make sure this doesn't end up in an infinite loop. Also, not sure if this handles subflows?

akchinSTC:
circular dep validation checks have been updated to 'flatten' the graph and add edges correctly and the error message has been simplified due to the change in structure. Will need to follow up with updates for more specific details when adding the linkIDs/super node bindings. 
parentID list recursive check is currently being updated.

akchinSTC:
@ajbozarth, so it doesn't look like the output/inputpath dropdowns support supernodes right now. 
image

ajbozarth:
@akchinSTC I just pushed the updated pipeline-editor version, it builds are a quick test shows it has the latest changes.

elyra/pipeline/kfp/processor_kfp.py
```diff
@@ -566,14 +563,25 @@ def _cc_pipeline(self,
...
-                    operation.component_params_as_dict.pop("outputs")
+                    comp_spec_inputs = [inputs.name.lower().replace(" ", "_") for
+                                        inputs in factory_function.component_spec.inputs]
+
+                    # Remove inputs and outputs from params dict
+                    # TODO: need to have way to retrieve only required params
+                    parameter_removal_list = ["inputs", "outputs"]
+                    for component_param in operation.component_params_as_dict.keys():
+                        if component_param not in comp_spec_inputs:
+                            parameter_removal_list.append(component_param)
```
kiersten-stokes, line 574: Smart!

kiersten-stokes (approved):
I focused mainly on functionality for my review, testing various KFP pipelines comprising custom and generic components. All my tests went as expected!

elyra/pipeline/kfp/processor_kfp.py
```diff
@@ -450,6 +449,7 @@ def _cc_pipeline(self,
         # and its parent's outputs.
 
         PipelineProcessor._propagate_operation_inputs_outputs(pipeline, sorted_operations)
+        component_task_factories = {}
```
kevin-bates, line 452: Is there a reason why we aren't just using `target_ops` for this?  They're both have the same keys and values. 

I do see that `target_ops` tends to be exclusive to generic components while `component_task_factories` appears to be exclusive to custom components, but since operation ids (i.e., the keys to these dictionaries) will be unique across those two sets, I'm not sure `component_task_factories` is necessary.

If we do retain c-t-f, could you please add some comments surrounding its use?  I don't really follow the 'task' or 'factory' portion of its name.

elyra/pipeline/validation.py
```diff
@@ -697,6 +730,43 @@ def _is_required_property(self, property_dict: dict, node_property: str) -> bool
...
+        Helper function to determine the type of a node property
+        :param property_dict: a dictionary containing the full list of property parameters and descriptions
+        :param node_property: the property to look for
+        :return: the data type associated with node_property, defaults to 'string'
+        """
+        for prop in property_dict['uihints']['parameter_info']:
+            if prop["parameter_ref"] == f"elyra_{node_property}":
+                return prop['data'].get('format', 'string')
+
+    def _get_parent_id_list(self, pipeline_definition, node_id_list: list, parent_list: list) -> List:
```
kevin-bates, line 748:
```suggestion
    def _get_parent_id_list(self, pipeline_definition: PipelineDefinition, node_id_list: list, parent_list: list) -> List:
```

elyra/pipeline/validation.py
```diff
@@ -517,28 +553,25 @@ def _validate_pipeline_graph(self, pipeline: dict, response: ValidationResponse)
...
+        #     if cycle_counter not in link_dict_table:
+        #         link_dict_table[cycle_counter] = []
+        #     for i in range(size_of_cycle):
+        #         if i == size_of_cycle - 1:
+        #             link_dict_table[cycle_counter].append(self._get_link_id(pipeline, cycle[i], cycle[0]))
+        #         else:
+        #             link_dict_table[cycle_counter].append(self._get_link_id(pipeline, cycle[i], cycle[i + 1]))
+        #     cycle_counter += 1
+
+        # for cycle_number, cycle_link_list in link_dict_table.items():
```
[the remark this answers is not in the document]
akchinSTC: Ill block comment this my change. Used to be able to identify the linkid between nodes to send back as information but with the pipeline traversal with supernode changes and supernodes being bugged atm this portion is currently disabled

PR #2109, opened 2021-09-07
Bring your own Schemas and Schemaspaces

### LINKED ISSUE #2075: Bring Your Own Schemas and Schemaspaces
This issue presents a high-level design for how we could go about introducing _Bring Your Own Schema_ to applications and is considered a prerequisite to separating the Elyra metadata service into its own Jupyter server extension.  Along with BYO Schemas, it introduces _Schemaspaces_ - which can also be _BYO'd_ for application use. 

This topic relates to the following:
[Provide metadata service as separate server extension](https://github.com/elyra-ai/elyra/issues/701)
[Bring your own runtime (schema)](https://github.com/elyra-ai/elyra/discussions/1672)
[Schema Filters](https://github.com/elyra-ai/elyra/pull/2062)
[Add support for metadata instance classes](https://github.com/elyra-ai/elyra/pull/725)

# Bring Your Own Schemas and Schemaspaces
One of the primary items that must be defined before separating the metadata service into its own server extension is how schemas get introduced.  In addition, Elyra currently has a higher order of organization known as a _Namespace_ - which contains schemas and allows applications to group similar schemas under a namespace.  Since _Namespace_ is a commonly used term, we are introducing the term _Schemaspace_ which is less ambiguous.  We are also promoting _Schemaspace_ to an object rather than just a name that it (namespace) is today.  This allows applications to better organize a schemaspace's schemas and provide a display name and description to each.

## Schemaspaces
As noted above, a schemaspace currently identifies a list of schemas that will be scoped under that schemaspace.  That is, the schemas, and _their_ instances have a logical connection.  Schemaspaces will also have properties like: `id` (a v4 UUID), `name` (a _symbolic_, case-insensitive, string used primarily for lookups), `display_name` (indicating a human-readable, and case-sensitive, name of the schemaspace), and `description` (a string describing the purpose of the schemaspace).

### `Schemaspace`
Schemaspaces will be discovered using [`entrypoints`](https://pypi.org/project/entrypoints/) via the group name `metadata.schemaspaces` and must be subclassed from a new base class: `Schemaspace`.  `Schemaspace` will resemble something like the following:
```python
class Schemaspace(object):
    
    _id: str
    _name: str
    _display_name: str
    _description str
    _schemas: List[Dict]

    def __init__(self, *args, **kwargs):
         self._id = None
         self._name = None
         self._display_name = None
         self._description = ""
         self._schemas = []

    @property
    def id(self) -> str:
        """The id (uuid) of the schemaspace"""
        assert self._id is not None and len(self._id) > 0, "Property 'id' requires a value!")
        # validate that it is uuidv4 compliant
        assert self._validate_id(), f"The value of property 'id' ({self._id}) does not conform to a UUID!"
        return self._id

    @property
    def name(self) -> str:
        """The name of the schemaspace"""
        assert self._name is not None and len(self._name) > 0, "Property 'name' requires a value!")
        return self._name

    @property
    def display_name(self) -> str:
        """The display_name of the schemaspace"""
        return self._display_name or self.name

    @property
    def description(self) -> str:
        """The description of the schemaspace"""
        return self._description

    @property
    def schemas(self) -> List[Dict]:
        """Returns the schemas currently associated with this namespace"""
        return self._schemas

    def add_schema(self, schema: Dict) -> None:
        """Associates the given schema to this namespace"""
        assert isinstance(schema, dict, "Parameter 'schema' is not a dictionary!")
        self._schemas.append(schema)
```
Some observations from class `Schemaspace`. 
1. Only `id` and `name` properties require values.
1. Schemas are non-existent and get added during entrypoint processing (more on this below).  This enables an application to bring their own schema to an _existing_ schemaspace.
1. Generally speaking, the only thing required to implement a `Schemaspace` is to set the `self._id` and `self._name` attributes.
1. Regarding the `name` property, we are treating this like a _symbolic name_, meaning it will be used for lookup and be treated as case-insensitive.  Its pattern must adhere to alphanumeric, beginning with alpha, and embedded underscores and dashes are permitted.  The `display_name` should be used for richer display/naming characteristics.
1. We chose to add `id` properties to schemaspaces because they are absolute and unique.  That said, schemaspaces have a mirroring `name` property for usability.  UUIDs are also the canonical approach for other storage mechanisms like database (SQL and NOSQL) - which we don't want to preclude.

### Registration
As noted previously, schemaspace implementations will be discovered via entrypoints.  To be discoverable, the schemaspace implementation must be listed in the hosting package's entrypoints definitions and associated with the group `metadata.schemaspaces`.  Entries relative to today's _namespaces_ used in Elyra should resemble the following:
```python
        'metadata.schemaspaces': [
            'runtimes = elyra.metadata.schemaspaces:Runtimes',
            'runtimes-images = elyra.metadata.schemaspaces:RuntimeImages',
            'code-snippets = elyra.metadata.schemaspaces:CodeSnippets'
        ],
```
with the `Runtimes` schemaspace implementation closely resembling the following:
```python
class Runtimes(Schemaspace):

    RUNTIMES_SCHEMASPACE_ID = "130b8e00-de7c-4b32-b553-b4a52824a3b5"

    def __init__(self, *args, **kwargs):
         self._id = RUNTIMES_SCHEMASPACE_ID
         self._name = "runtimes"
         self._display_name = "Runtimes"
         self._description = "Schemaspace for instances of Elyra runtime configurations"
```
### Processing
When the `SchemaManager` starts and initializes itself, it will retrieve all schemaspaces registered to group `metadata.schemaspaces` and build a dictionary keyed by BOTH the loaded schemaspace `id` and `name` properties.  That is, the dictionary will have two keys pointing to the same instance.  This facilitates faster lookups by either `id` or `name`.  We will also maintain an `id-to-name` map for scenarios where the schemaspace's `name` is desired but only an `id` is provided.  The `values()` of this dictionary will also be used to quickly get a list of schemaspace names. 

Once all schemaspaces are loaded, `SchemaManager` will then load all schemas via group `metadata.schemas_providers` (discussed below).  For each schema loaded, `SchemaManager` will identify the schema's `schemaspace_id` and add that schema definition to the corresponding `Schemaspace` instance using `schemaspace.add_schema(schema)`.  All schemaspaces will be considered "loaded" once all `metadata.schemas_providers` entrypoints have been processed.

### Additional Schemaspaces
There will be no provisions for detecting and adding new schemaspaces to a running system.  This is something we could add in the future where the `SchemaManager` periodically cycles across the `metadata.schemaspaces` and detects new schemaspaces (and new schemas).  For now, we will require an application restart for additional schemaspaces to be picked up.

## Schemas
Like _schemaspaces_, _schemas_ will also be registered and discovered via [`entrypoints`](https://pypi.org/project/entrypoints/).  However, because schemas are _just_ JSON objects (represented in Python as a dictionary) and don't require methods - unlike `Schemaspace` instances, the instance loaded via the entrypoint will be a subclass of `SchemasProvider`.

### `SchemasProvider`
`SchemasProvider` (note the plural _schemas_) is an abstract base class responsible for defining the expected methods in which the `SchemaManager` can get the schema instances from the entrypoint implementation.  Because schemas are self-contained and contain a reference to their hosting schemaspace (see _Meta-schema_ below), `SchemasProvider` instances will implement a single method: `get_schemas()` that returns **all** schemas relative to that provider.   Here's a possible class definition:

```python
class SchemasProvider(ABC):

    @abstractmethod
    def get_schemas(self) -> List[Dict]:
        """Returns a list of schemas"""
        pass
```
Prior to these changes, [Schema Filters](https://github.com/elyra-ai/elyra/pull/2062) could be implemented to allow a schema to be updated to reflect runtime configurations.  For example, the `engine` property of the runtime schema `kfp` is associated with an enum whose values are determined based on installed packages.  The schema filter implementation would then be responsible for any _dynamic changes_ to reflect the current installation.  Since the `SchemasProvider` will be invoked to retrieve its schemas, any dynamic changes that are necessary can be implemented within the `get_schemas()` method itself - thereby eliminating the need for Schema Filters.

### Registration
As noted, schemas will be discovered via the entrypoint machinery.  However, unlike schemaspaces, where each entrypoint entry corresponds to a schemaspace and associated `Schemaspace` instance, **all** schemas can be associated to a single `SchemasProvider` instance.  Of course, a given package can choose to separate their schema discovery however they choose by defining additional entrypoint entries (and corresponding `SchemasProvider` instances).  Here's what an entrypoint entry for the Elyra-based schemas might look like:
```python
        'metadata.schemas_providers': [
            'elyra-schemas = elyra.metadata.schemasprovider:ElyraSchemasProvider'
        ],
```
and the `ElyraSchemasProvider` implementation resembling the following:
```python
class ElyraSchemasProvider(SchemasProvider):

    def get_schemas(self) -> List[Dict]:
        elyra_schemas = []
        schema_dir = os.path.join(os.path.dirname(__file__), 'schemas')
        schema_files = [json_file for json_file in os.listdir(schema_dir) if json_file.endswith('.json')]
        for json_file in schema_files:
            schema_file = os.path.join(schema_dir, json_file)
            with io.open(schema_file, 'r', encoding='utf-8') as f:
                schema_json = json.load(f)
            elyra_schemas.append(schema_json)

        return elrya_schemas
```

### Processing
As noted above in the _Processing_ section of _Schemaspace_, `SchemaManager` will first load all schemaspaces.  Next, it will load each of the entrypoints grouped by `metadata.schemas_providers`, each yielding a `SchemasProvider` instance.  Upon calling the instance's `get_schemas()` method, it will iterate on each schema, validate the schema against the meta-schema (see next section), and access its (required) `schemaspace_id` property.  This value will be used to lookup the `Schemaspace` instance and, when found, call that instance's `add_schema()` method to add the current schema to its schemaspace.  If the schemaspace is not found, a warning message will be issued and processing will continue.  This will continue for each of the returned schemas in each of the registered schemas providers.

Here's a rough algorithm depicting the complete processing...
```
On SchemaManager initialization
    For each schemaspace found in `metadata.schemaspaces`:
        record the `Schemaspace` instance in a dictionary indexed by the instance's `id` property.
    For each schemas provider found in `metadata.schemas_providers`:
        verify schemasProvider is an instance of `SchemasProvider`
        call the `SchemasProvider.get_schemas()` method to get a list of schemas
        For each schema in the resulting list:
            validate the schema against the meta-schema, 
            access the schema's `schemaspace_id` property,
            lookup the schemaspace instance in the dictionary from the first loop
            if found, call the schemaspace instance's `add_schema()` method.
            else, issue a warning and continue
```
### Meta-schema
We will need to define a _meta-schema_ that describes the structure of schema files.  Because we'd like to add id-based properties with a format of UUID, this meta-schema will be based on Draft 2019-09 since it introduces `"format": "uuid"`.  Note: A slightly newer draft (Draft 2020-12) has been published and we may want to adopt that instead.  
**UPDATE:** See https://github.com/elyra-ai/elyra/issues/2075#issuecomment-906725026.  We will continue using draft-07 and define a UUID pattern with similar semantics to 2019-09's `"format": "uuid"`.

Here's an example of a meta-schema definition that could be used to validate the schemas themselves:
```JSON
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "Meta-schema for metadata service schemas",
    "properties": {
        "schemaspace_id": {
            "description": "The ID of the schemaspace in which this schema and its instances resides.",
            "type": "string",
            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
        },
        "name": {
            "description": "The name of the schema.",
            "type": "string",
            "pattern": "^[a-z][a-z0-9-_]*[a-z0-9]$"
        },
        "id": {
            "description": "The ID of this schema.",
            "type": "string",
            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
        },
        "description": {
          "title": "Description",
          "description": "Description of this schema.",
          "type": "string",
          "maxLength": 80
        },
        "metadata_class_name": {
          "title": "Metadata Instance Class",
          "description": "Class used by metadata service to alter instances on load, save, and deletion.",
          "type": "string"
        },
        "properties": {
            "type": "object",
            "required": ["schema_name", "display_name", "metadata"],
            "additionalProperties": true
        }
    },
    "required": ["schemaspace_id", "name", "properties"]
}
```
As you can see, this meta-schema requires that all schemas instances have `schemaspace_id`, `name`, and `properties` properties.  In addition, it enforces that all schemas _define_ "schema_name", "display_name", "metadata" properties - which equates to these properties being present in all instances of that schema (i.e., all metadata instances).   Fortunately, `schema_name` can be a constant value (i.e., `const`) and `metadata` is really just the container for the instance properties, so really, only `display_name` requires a value on the instance (besides the values it specifies as required within the `metadata` stanza).

### Additional Schemas
Like for additional `schemaspaces`, there will be no provisions for detecting and adding new schemas to a running system.  This is something we could add in the future where the `SchemaManager` periodically cycles across the `metadata.schemas_providers` and detects new schemas.  For now, we will require an application restart for additional schemas to be picked up.

karlaspuldaro:
Closes #2075

           in the file; or

           it.

### DIFF
--- a/elyra/metadata/handlers.py
+++ b/elyra/metadata/handlers.py
@@ -218,7 +218,7 @@ class SchemaResourceHandler(HttpErrorMixin, APIHandler):
 
 class NamespaceHandler(HttpErrorMixin, APIHandler):
     """Handler for retrieving namespaces """
-    # Redirect to SchemaSpaceHandler
+    # Redirect to SchemaspaceHandler
 
     @web.authenticated
     async def get(self):
@@ -238,7 +238,7 @@ class NamespaceHandler(HttpErrorMixin, APIHandler):
         self.finish(namespace_model)
 
 
-class SchemaSpaceHandler(HttpErrorMixin, APIHandler):
+class SchemaspaceHandler(HttpErrorMixin, APIHandler):
     """Handler for retrieving schemaspaces """
 
     @web.authenticated
--- a/elyra/metadata/handlers.py
+++ b/elyra/metadata/handlers.py
@@ -218,6 +218,7 @@ class SchemaResourceHandler(HttpErrorMixin, APIHandler):
 
 class NamespaceHandler(HttpErrorMixin, APIHandler):
     """Handler for retrieving namespaces """
+    # Redirect to SchemaSpaceHandler
 
     @web.authenticated
     async def get(self):
@@ -235,3 +236,23 @@ class NamespaceHandler(HttpErrorMixin, APIHandler):
 
         self.set_header("Content-Type", 'application/json')
         self.finish(namespace_model)
+
+class SchemaSpaceHandler(HttpErrorMixin, APIHandler):
+    """Handler for retrieving schemaspaces """
+
+    @web.authenticated
+    async def get(self):
+
+        try:
+            schema_manager = SchemaManager.instance()
+            schemaspaces = schema_manager.get_schemaspaces()
+        except (ValidationError, ValueError) as err:
+            raise web.HTTPError(404, str(err)) from err
+        except Exception as err:
+            raise web.HTTPError(500, repr(err)) from err
+
+        schemaspace_model = dict()
+        schemaspace_model['schemas'] = schemaspaces
+
+        self.set_header("Content-Type", 'application/json')
+        self.finish(schemaspace_model)
--- a/elyra/metadata/handlers.py
+++ b/elyra/metadata/handlers.py
@@ -237,6 +237,7 @@ class NamespaceHandler(HttpErrorMixin, APIHandler):
         self.set_header("Content-Type", 'application/json')
         self.finish(namespace_model)
 
+
 class SchemaSpaceHandler(HttpErrorMixin, APIHandler):
     """Handler for retrieving schemaspaces """
 
--- a/elyra/metadata/schema.py
+++ b/elyra/metadata/schema.py
@@ -25,7 +25,7 @@ from traitlets.config import SingletonConfigurable
 from elyra.metadata.error import SchemaNotFoundError
 
 
-METADATA_TEST_NAMESPACE = "metadata-tests"  # exposed via METADATA_TESTING env
+METADATA_TEST_SCHEMASPACE = "metadata-tests"  # exposed via METADATA_TESTING env
 
 
 class SchemaManager(SingletonConfigurable):
@@ -35,63 +35,63 @@ class SchemaManager(SingletonConfigurable):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        # namespace_schemas is a dict of namespace keys to dict of schema_name keys of JSON schema
-        self.namespace_schemas = SchemaManager.load_namespace_schemas()
+        # schemaspace_schemas is a dict of schemaspace keys to dict of schema_name keys of JSON schema
+        self.schemaspace_schemas = SchemaManager.load_schemaspace_schemas()
 
-    def validate_namespace(self, namespace: str) -> None:
-        """Ensures the namespace is valid and raises ValueError if it is not."""
-        if namespace not in self.namespace_schemas.keys():
-            raise ValueError("Namespace '{}' is not in the list of valid namespaces: '{}'".
-                             format(namespace, self.get_namespaces()))
+    def validate_schemaspace(self, schemaspace: str) -> None:
+        """Ensures the schemaspace is valid and raises ValueError if it is not."""
+        if schemaspace not in self.schemaspace_schemas.keys():
+            raise ValueError("schemaspace '{}' is not in the list of valid schemaspaces: '{}'".
+                             format(schemaspace, self.get_schemaspaces()))
 
-    def get_namespaces(self) -> list:
-        return list(self.namespace_schemas.keys())
+    def get_schemaspaces(self) -> list:
+        return list(self.schemaspace_schemas.keys())
 
-    def get_namespace_schemas(self, namespace: str) -> dict:
-        self.validate_namespace(namespace)
-        schemas = self.namespace_schemas.get(namespace)
+    def get_schemaspace_schemas(self, schemaspace: str) -> dict:
+        self.validate_schemaspace(schemaspace)
+        schemas = self.schemaspace_schemas.get(schemaspace)
         return schemas
 
-    def get_schema(self, namespace: str, schema_name: str) -> dict:
-        self.validate_namespace(namespace)
-        schemas = self.namespace_schemas.get(namespace)
+    def get_schema(self, schemaspace: str, schema_name: str) -> dict:
+        self.validate_schemaspace(schemaspace)
+        schemas = self.schemaspace_schemas.get(schemaspace)
         if schema_name not in schemas.keys():
-            raise SchemaNotFoundError(namespace, schema_name)
+            raise SchemaNotFoundError(schemaspace, schema_name)
         schema_json = schemas.get(schema_name)
 
         return schema_json
 
-    def add_schema(self, namespace: str, schema_name: str, schema: dict) -> None:
+    def add_schema(self, schemaspace: str, schema_name: str, schema: dict) -> None:
         """Adds (updates) schema to set of stored schemas. """
-        self.validate_namespace(namespace)
-        self.log.debug("SchemaManager: Adding schema '{}' to namespace '{}'".format(schema_name, namespace))
-        self.namespace_schemas[namespace][schema_name] = schema
+        self.validate_schemaspace(schemaspace)
+        self.log.debug("SchemaManager: Adding schema '{}' to schemaspace '{}'".format(schema_name, schemaspace))
+        self.schemaspace_schemas[schemaspace][schema_name] = schema
 
     def clear_all(self) -> None:
         """Primarily used for testing, this method reloads schemas from initial values. """
-        self.log.debug("SchemaManager: Reloading all schemas for all namespaces.")
-        self.namespace_schemas = SchemaManager.load_namespace_schemas()
+        self.log.debug("SchemaManager: Reloading all schemas for all schemaspaces.")
+        self.schemaspace_schemas = SchemaManager.load_schemaspace_schemas()
 
-    def remove_schema(self, namespace: str, schema_name: str) -> None:
-        """Removes the schema entry associated with namespace & schema_name. """
-        self.validate_namespace(namespace)
-        self.log.debug("SchemaManager: Removing schema '{}' from namespace '{}'".format(schema_name, namespace))
-        self.namespace_schemas[namespace].pop(schema_name)
+    def remove_schema(self, schemaspace: str, schema_name: str) -> None:
+        """Removes the schema entry associated with schemaspace & schema_name. """
+        self.validate_schemaspace(schemaspace)
+        self.log.debug("SchemaManager: Removing schema '{}' from schemaspace '{}'".format(schema_name, schemaspace))
+        self.schemaspace_schemas[schemaspace].pop(schema_name)
 
     @classmethod
-    def load_namespace_schemas(cls, schema_dir: Optional[str] = None) -> dict:
-        """Loads the static schema files into a dictionary indexed by namespace.
+    def load_schemaspace_schemas(cls, schema_dir: Optional[str] = None) -> dict:
+        """Loads the static schema files into a dictionary indexed by schemaspace.
            If schema_dir is not specified, the static location relative to this
            file will be used.
            Note: The schema file must have a top-level string-valued attribute
-           named 'namespace' to be included in the resulting dictionary.
+           named 'schemaspace' to be included in the resulting dictionary.
         """
-        # The following exposes the metadata-test namespace if true or 1.
+        # The following exposes the metadata-test schemaspace if true or 1.
         # Metadata testing will enable this env.  Note: this cannot be globally
         # defined, else the file could be loaded before the tests have enable the env.
         metadata_testing_enabled = bool(os.getenv("METADATA_TESTING", 0))
 
-        namespace_schemas = {}
+        schemaspace_schemas = {}
         if schema_dir is None:
             schema_dir = os.path.join(os.path.dirname(__file__), 'schemas')
         if not os.path.exists(schema_dir):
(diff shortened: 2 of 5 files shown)


kiersten-stokes (approved):
This looks really good! I tested the following scenarios and didn't run into any issues:

- Created a new schemaspace, implemented `Schemaspace` and `SchemasProvider` classes and the schema JSON
- Tested CRUD operations via the `CLI` for my new schema
- Tested all the endpoints for the new schemaspace (querying on both name and id where possible)
- Tested some CRUD operations via the UI for existing schemaspaces that have frontend support (code snippets, runtime images, and registries)
- Added a new runtime schema (didn't implement the processor or anything but made all the required metadata-based changes that allowed it to show up in `runtime` schemaspace queries)

docs/source/developer_guide/metadata.md
```diff
@@ -148,5 +158,69 @@ tags = runtime_configuration.metadata['tags']
...
+entry_points={
+  'metadata.schemas_providers': [
+    'runtimes = elyra.metadata.schemasproviders:RuntimesSchemas',
+    'runtimes-images = elyra.metadata.schemasproviders:RuntimeImagesSchemas',
+    'code-snippets = elyra.metadata.schemasproviders:CodeSnippetsSchemas',
+    'component-registries = elyra.metadata.schemasproviders:ComponentRegistriesSchemas'  ]
+}
+```
+## Implementing Custom Metadata Classes
+Another area within the metadata service applications can influence behavior is by introducing their own sublcass of `Metadata` to wrap instance data.  When this is done, those implementations will be called prior to and after any create, update and delete operations, in addition to following any load (get) operations.  As a result, applications can influence what is persisted and what is retrieved.
```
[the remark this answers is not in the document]
kevin-bates: Good catch - I agree and will update relative to your suggestion.

Thank you for so thoroughly putting this through its paces - that is greatly appreciated!

ajbozarth:
I took a look through the frontend code and it looks good, I'll try to find time to actually test this locally and look into the rest of the code when I can

ajbozarth (approved):
I had a moment to do a quick check locally of the UI/UX and it still works as intended/previously.

ptitzler (changes requested):
Can you please retake `docs/source/images/metadata-service-relationships.png` in full screen mode? Currently the screenshot includes the spellchecker annotations:

image

docs/source/developer_guide/metadata.md
```diff
@@ -16,50 +16,68 @@ limitations under the License.
...
+Schemaspaces and their schemas can be dynamically introduced via entrypoints.  When the system starts, it will identify applicable entrypoints and load their corresponding objects to seed the list of _configured_ schemaspaces and schemas.
+
+### Metadata Components
+The Elyra metadata service is composed of several components, many of which were previously referenced, and each responsible for a different task.  
+![MetadataServiceRelationships](../images/metadata-service-relationships.png)
+
+The following briefly describes each of those components.
 
-### Metadata Services structure using the default file system store
+- **[Schemaspace](https://github.com/karlaspuldaro/elyra/blob/d345dbd415cbdb16c8d46f705bcd1b3b6ed54450/elyra/metadata/schema.py#L199)**: This is the top-most component and represents a _container_ of schemas.  It has `id` (UUID), `name` (symbolic name), `display_name` (descriptive name), and `description` properties, along with a set of schemas.  Instances of a schemaspace's schemas are logically grouped together in whatever storage model is in use.  Schemaspaces are discovered and loaded via entrypoints.
```
ptitzler, line 42: If you perform a global replace of 
```
https://github.com/karlaspuldaro/elyra/blob/d345dbd415cbdb16c8d46f705bcd1b3b6ed54450/elyra/metadata/
```
with
```
https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/
```

this link and subsequent link that currently point to the fork should be valid

docs/source/developer_guide/metadata.md
```diff
@@ -148,5 +158,69 @@ tags = runtime_configuration.metadata['tags']
...
+    'code-snippets = elyra.metadata.schemaspaces:CodeSnippets',
+    'component-registries = elyra.metadata.schemaspaces:ComponentRegistries'
+  ]
+}
+```
+As you can see, each entry has a _name_ that equates to a module and object/class.
+
+### SchemasProvider
+As mentioned above, `Schemaspace` instances do not know their set of schemas at the time of their construction.  Instead, the `SchemaManager` loads a different entrypoints group following the loading of schemaspaces named `'metadata.schemas_providers'`.  These entrypoints identify a `SchemasProvider` object and this object is responsible for returning schemas when called upon by the `SchemaManager`.
+`SchemasProvider` instances are concrete instances of the abstract base class `SchemasProvider`.  The class is really quite simple:
```
ptitzler, line 187: Can you reword this (`really quite simple`) please so this does not come across as judgemental?

docs/source/developer_guide/metadata.md
```diff
@@ -148,5 +158,69 @@ tags = runtime_configuration.metadata['tags']
...
+        pass
+```
+When called, the `SchemasProvider` instance is responsible for producing a list of schemas.  These schemas must conform to the metadata service's _meta-schema_ and, as a result, will include the schemaspace name and schemaspace id to which they correspond, in addition to a `name` property representing that schema's name.  The `SchemaManager` uses this meta-information to confirm the schemaspace has been loaded and register that schema with the schemaspace following its successful validation against the meta-schema.
+
+Some SchemasProviders may have schema information which is dynamic, based on the active configuration of the system.  In such cases, they are responsible for returning information pertinent to the current configuration.  For example, the _component-registry_ schema (of the _component-registries_ schemaspace) includes an enumerated property named `runtime`.  However, the values/choices included in the `enum` are a function of what packages are configured.  As a result, the `get_schemas()` method for the _component-registries_ `SchemasProvider` determines, based on installed packages, which entries to add to the `runtime` enumeration.
```
ptitzler, line 199: Missing markdown

```suggestion
Some `SchemasProvider`s may have schema information which is dynamic, based on the active configuration of the system.  In such cases, they are responsible for returning information pertinent to the current configuration.  For example, the _component-registry_ schema (of the _component-registries_ schemaspace) includes an enumerated property named `runtime`.  However, the values/choices included in the `enum` are a function of what packages are configured.  As a result, the `get_schemas()` method for the _component-registries_ `SchemasProvider` determines, based on installed packages, which entries to add to the `runtime` enumeration.
```

docs/source/developer_guide/metadata.md
```diff
@@ -148,5 +158,69 @@ tags = runtime_configuration.metadata['tags']
...
 
+#### Registration
+Unlike `Schemaspaces` and `SchemasProviders`, metadata class overrides are registered within their respective schema file via the `metadata_class_name` property.  The value of this property denotes the dotted notation corresponding to the location of the implementation, the hosting package of which must be installed and locatable.  (Note that someday, we may choose to change this value to be an entrypoint name of a well-documented entrypoint group.  In such cases, the value would not be "dotted" and, therefore, backwards compatible with the current approach.)
```
ptitzler, line 220:
```suggestion
Unlike `Schemaspaces` and `SchemasProvider`s, metadata class overrides are registered within their respective schema file via the `metadata_class_name` property.  The value of this property denotes the dotted notation corresponding to the location of the implementation, the hosting package of which must be installed and locatable.  (Note that someday, we may choose to change this value to be an entrypoint name of a well-documented entrypoint group.  In such cases, the value would not be "dotted" and, therefore, backwards compatible with the current approach.)
```

docs/source/developer_guide/metadata.md
kevin-bates: Rather than have the 's' in another font, I think using "`SchemasProvider` instances" may render better.

kevin-bates:
Thank you for your initial review @ptitzler - I believe I've addressed your comments but would appreciate an additional pass.

ptitzler (approved):
doc - LGTM

PR #2127, opened 2021-09-10
Add additional safeguards around list and dictionary types during processing

kiersten-stokes:
When testing the Airflow `BashOperator`, I came across a new error while the pipeline was running. In the node properties, the `env` parameter was left empty, resulting in the below log output that indicates a dictionary value is expected. I'm almost positive that we've tested this operator before with an empty value and have not had this error result.

<details><summary>Airflow Log Output</summary>
<pre>
[2021-09-10 16:34:07,681] {taskinstance.py:1150} ERROR - 'str' object has no attribute 'update'
  File "/home/airflow/.local/lib/python3.6/site-packages/airflow/models/taskinstance.py", line 984, in _run_raw_task
    result = task_copy.execute(context=context)
  File "/home/airflow/.local/lib/python3.6/site-packages/airflow/operators/bash_operator.py", line 123, in execute
    env.update(airflow_context_vars)
AttributeError: 'str' object has no attribute 'update'
</pre>
</details>

This adds a check to the `_process_list_value()` and `process_dictionary_value()` processor functions for empty strings or a  string of `"None"`. 

Because list and dictionary-typed values are considered strings for the purposes of the node properties panel, the pipeline payload contains string values for these parameters when submitting/exporting.  If the user has not entered a value for a list or dictionary-typed parameter or has explicitly entered the value of `None` in the node properties panel, the payload will include `""` or `"None"` for that parameters value, respectively. This PR converts these values to either an empty list `[]` or an empty dictionary `{}` as appropriate.

The test cases that cover these two functions have been updated. This has also been tested manually by submitting pipelines.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/processor.py
+++ b/elyra/pipeline/processor.py
@@ -410,22 +410,24 @@ class RuntimePipelineProcessor(PipelineProcessor):
         envs['ELYRA_ENABLE_PIPELINE_INFO'] = str(self.enable_pipeline_info)
         return envs
 
-    def _process_dictionary_value(self, value: str) -> Union[Dict, str]:
+    def _process_dictionary_value(self, value: str) -> Optional[Union[Dict, str]]:
         """
         For component parameters of type dictionary, the user-entered string value given in the pipeline
         JSON should be converted to the appropriate Dict format, if possible. If a Dict cannot be formed,
         log and return stripped string value.
         """
-        converted_dict = None
         value = value.strip()
+        if not value:
+            return {}
+        if value == "None":
+            return None
+
+        converted_dict = None
         if value.startswith('{') and value.endswith('}'):
             try:
                 converted_dict = ast.literal_eval(value)
             except Exception:
                 pass
-
-        if not value or value == "None":
-            return {}
 
         # Value could not be successfully converted to dictionary
         if not isinstance(converted_dict, dict):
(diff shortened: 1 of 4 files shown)


elyra/pipeline/processor.py
```diff
@@ -424,6 +424,9 @@ def _process_dictionary_value(self, value: str) -> Union[Dict, str]:
             except Exception:
                 pass
 
+        if not value or value == "None":
+            return {}
+
```
lresende, line 429: Here is too late, you will get errors on line 420 when you try to strip to None, no? I would move this to top of the function.

lresende: btw, when do we get "None" and what is the difference from just using None?

kiersten-stokes: Ha we are on the exact same page based on the commit I just pushed. I have the same sort of question as to whether we should allow `"None"` to evaluate as `None` or keep the empty value. I've tested it with pipeline submission on Airflow so far and nothing has changed from the prior commit. As in, the pipeline executes successfully regardless.

kiersten-stokes: We would get `"None"` in the case that the user types None into the parameter field. And the description for some parameters mentions that None is an option, e.g. the BashOperator `env` parameter description says "If env is not None, ...". I feel like this may indicate to some users that they should/could type None into that field and expect the value of `None` to show up in the DAG.

kiersten-stokes: Hmm unfortunately this doesn't work very well with KFP. I'm going to just revert to only using the empty list or dictionary in both cases/ But keeping the new test to check against the "None" string

Issue #2152, opened 2021-09-16
VPE/palette doesn't auto-refresh after component registry entry was deleted


ptitzler:
**Describe the issue**
A clear and concise description of what the issue is.

The VPE component palette displays deleted entries after they have been removed from the registry. Open/close of the VPE does not resolve the issue.

![image](https://user-images.githubusercontent.com/13068832/133586224-28fa8831-58c6-45c1-8715-8bb4504a9a32.png)

It appears that the cache is not invalidated when a component registry entry is deleted.

**To Reproduce**
1. Create a component registry entry
2. Open VPE - the entry is displayed as expected
3. Delete the component registry entry
4. Open VPE - the entry is still displayed

**Expected behavior**
The palette should reflect the current state of the registry or provide the means to manually trigger a re-load.

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.1+

kiersten-stokes:
PR #2157 added the backend support for cache updates on registry modification. This still requires the user to reload or open a new pipeline, etc. in JL in order to see the changes on the frontend, so this will remain open until the corresponding frontend piece is added.

ptitzler:
This is resolved in version 3.7.

PR #2157, opened 2021-09-17
Fix caching to update immediately after component registry modifications

kiersten-stokes:
This provides the backend support for #2152. Frontend support that closes the issue to come in a followup PR.

This is a unique issue because metadata instances are updated via API calls to the `elyra/metadata` endpoint, but component registries are currently only parsed on API calls to the `elyra/pipeline/components` endpoint (if the cache is not expired). 

This PR adds `post_save` and `post_delete` methods onto the `Metadata` class. The new `ComponentRegistryMetadata` class then implements these methods to update the component registry cache using the runtime processor information present in the registry metadata instance.

`ComponentRegistryMetadata` and `RegistrySchemaFilter` are moved into their own `component_metadata.py` file to avoid a circular dependency in `import` statements (e.g. avoiding `processor.py` importing `ComponentRegistry` from `component_registry.py` and `component_registry.py` importing `PipelineProcessorRegistry` from `processor.py`). 

Additionally, the `CachedComponentRegistry` functionality has been moved into the `ComponentRegistry`, avoiding the need for both classes.

Another item for discussion is that this change in itself does not fix the issue. In order to see the updates on the UI, you have to either refresh lab or close and open a pipeline (i.e. anything that triggers a call to the `elyra/pipeline/components` endpoint) on the frontend.

After making the changes, I don't know that a new test case is required for this. The changes are fairly minor, and we would have to force the test cases to run the metadata 'post_save' and ' post_delete' asynchronously which sort of defeats the purpose of this behind-the-scenes cache update. 

Additionally, the old test cases ended up needing to be updated as a result of the latest change, so we do have existing test coverage of the 'Metadata' + 'ComponentRegistry' functionality. 
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_metadata.py
+++ b/elyra/pipeline/component_metadata.py
@@ -57,11 +57,13 @@ class ComponentRegistryMetadata(Metadata):
 
         # Get processor instance and update its cache
         processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
-        processor.component_registry.update_cache()
+        if processor.component_registry.caching_enabled:
+            processor.component_registry.update_cache()
 
     def post_delete(self, **kwargs: Any) -> None:
         processor_type = self.to_dict()['metadata']['runtime']
 
         # Get processor instance and update its cache
         processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
-        processor.component_registry.update_cache()
+        if processor.component_registry.caching_enabled:
+            processor.component_registry.update_cache()
--- a/elyra/pipeline/component_metadata.py
+++ b/elyra/pipeline/component_metadata.py
@@ -53,17 +53,23 @@ class ComponentRegistryMetadata(Metadata):
     """
 
     def post_save(self, **kwargs: Any) -> None:
-        processor_type = self.to_dict()['metadata']['runtime']
+        processor_type = self.metadata.get('runtime')
 
         # Get processor instance and update its cache
-        processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
-        if processor.component_registry.caching_enabled:
-            processor.component_registry.update_cache()
+        try:
+            processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
+            if processor.component_registry.caching_enabled:
+                processor.component_registry.update_cache()
+        except Exception:
+            pass
 
     def post_delete(self, **kwargs: Any) -> None:
-        processor_type = self.to_dict()['metadata']['runtime']
+        processor_type = self.metadata.get('runtime')
 
         # Get processor instance and update its cache
-        processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
-        if processor.component_registry.caching_enabled:
-            processor.component_registry.update_cache()
+        try:
+            processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
+            if processor.component_registry.caching_enabled:
+                processor.component_registry.update_cache()
+        except Exception:
+            pass
--- a/elyra/pipeline/component_metadata.py
+++ b/elyra/pipeline/component_metadata.py
@@ -1,0 +1,67 @@
+#
+# Copyright 2018-2021 Elyra Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from typing import Any
+from typing import Dict
+
+import entrypoints
+
+from elyra.metadata.metadata import Metadata
+from elyra.metadata.schema import SchemaFilter
+from elyra.pipeline.processor import PipelineProcessorRegistry
+
+
+class RegistrySchemaFilter(SchemaFilter):
+    """
+    This class exists to ensure that the component registry schema's runtime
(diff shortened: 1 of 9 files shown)


kevin-bates:
Thanks for opening this pull request and discussion.  

This strikes me as a good use case for setting [`metadata_class_name`](https://github.com/elyra-ai/elyra/blob/aa9fe98f2e7be6c9517857b12ffdc94109eae1ea/elyra/metadata/schemas/metadata-test.json#L7) in the `component-registry` schema where the implementation class would be something like: `elyra.pipeline.component_registry.ComponenRegistryMetadata`.  This class instance would be invoked on each metadata load, save, and delete of component registry instances, and would catch updates made using CLI tooling, not just UI.   I also don't think we need to do anything with the cache on loads (GETs).

We currently provide hooks for [`post_load`](https://github.com/elyra-ai/elyra/blob/aa9fe98f2e7be6c9517857b12ffdc94109eae1ea/elyra/metadata/metadata.py#L47), [`pre_save`](https://github.com/elyra-ai/elyra/blob/aa9fe98f2e7be6c9517857b12ffdc94109eae1ea/elyra/metadata/metadata.py#L54) and [`pre_delete`](https://github.com/elyra-ai/elyra/blob/aa9fe98f2e7be6c9517857b12ffdc94109eae1ea/elyra/metadata/metadata.py#L63) but would want to add `post_save` and `post_delete` hooks for this use-case since the cache should only be updated if the instance's save/delete was successful (and not prior to the update/delete). 

To address updating all registries of all runtimes, the code (wherever it is implemented) should pull the runtime name out of the component-registry instance being updated or deleted (as it's a required property) and ask the `PipelineProcessorRegistry` for the appropriate `RuntimePipelineProcessor` using its `get_processor()` method (e.g., `PipelineProcessorRegistry.instance().get_processor(<runtime value>)`).  We should then add an accessor method/property to `RuntimePipelineProcessor` (similar to [`component_parser`](https://github.com/elyra-ai/elyra/blob/aa9fe98f2e7be6c9517857b12ffdc94109eae1ea/elyra/pipeline/processor.py#L305)) that returns its `self._component_registry` instance and call `update_cache()` on that instance.  (I see that your PR adds this accessor to the RPP - so that's great.)

(Note: it looks like you used `PipelineProcessorManager` to deal with this.  To be honest, I don't really know the difference between `PipelneProcessorRegistry` and `PipelineProcessorManager`,  so if the latter is more appropriate then use that. Personally, I think PPR is more appropriate because It is a _registry_, and its `get_processor()` is "public" while PPM's is "private" (`_get_processor_for_runtime()`).)

Would that work and cover CLI and UI w/o updating all component registry caches?

Btw, I think we should move to one implementation of the `ComponentRegistry` and it is always cached.  Having the potential for two flavors, when the adjustment is not configurable, seems confusing to me - and we could more easily specify a config option to disable caching.  This would prevent checks like [ensuring the instance is the cached instance](https://github.com/elyra-ai/elyra/pull/2157/files#diff-25ca6854f58ce6bdb484f2625ce15f4467f5501a0c25f410b0fba05c04bd470bR105) and requiring callers to deal with the potential that the [instance they obtained doesn't support caching](https://github.com/elyra-ai/elyra/pull/2157/files#diff-25ca6854f58ce6bdb484f2625ce15f4467f5501a0c25f410b0fba05c04bd470bR318).

kiersten-stokes:
@kevin-bates this is great -- thank you so much for the ideas and feedback!

> (Note: it looks like you used `PipelineProcessorManager` to deal with this. To be honest, I don't really know the difference between `PipelneProcessorRegistry` and `PipelineProcessorManager`, so if the latter is more appropriate then use that. Personally, I think PPR is more appropriate because It is a _registry_, and its `get_processor()` is "public" while PPM's is "private" (`_get_processor_for_runtime()`).)

Yeah, I also am not completely clear on the difference. PPR seems to only deal with the the processor registry and ways to add/update or validate runtimes in it whereas the PPM seems to do more coordination of the 'work' related to the RPPs? I'll look into it further as I go through it. Probably worth adding a docstring to these classes too I guess.

> Would that work and cover CLI and UI w/o updating all component registry caches?

Sounds like it to me! I'll get started on the edits.

I do still think we'll have the issue with the UI not immediately updating without refresh, but I don't see a way around that with backend-only changes anyway.

> Btw, I think we should move to one implementation of the ComponentRegistry and it is always cached. Having the potential for two flavors, when the adjustment is not configurable, seems confusing to me

Yeah, it does seem like there's some potential for unexpected situations with the dual options. I'll make the main changes first and then go from there on merging the caching into `ComponentRegistry` itself

elyra/pipeline/component_metadata.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ComponentRegistryMetadata(Metadata):
+    """
+    This class contains methods to trigger cache updates on modification
+    and deletion of component registry metadata instances.
+    """
+
+    def post_save(self, **kwargs: Any) -> None:
+        processor_type = self.to_dict()['metadata']['runtime']
```
kevin-bates, line 56: I think its a little cleaner (and cheaper) to get after `runtime` using...
```suggestion
        processor_type = self.metadata.get('runtime')
```

elyra/pipeline/component_metadata.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Get processor instance and update its cache
+        processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
+        processor.component_registry.update_cache()
```
kevin-bates, line 60: Although I'll be adding rollback capabilities to the Manager for when the post-hooks throw exceptions, it may be good to ignore errors here since a failure to update the cache shouldn't roll back the "transaction".
```suggestion
        try:
            processor = PipelineProcessorRegistry.instance().get_processor(processor_type=processor_type)
            processor.component_registry.update_cache()
        except Exception:
            pass
```
Same comments or `post_delete`.

kevin-bates (approved):
This looks good to me.  If a complete cache load starts taking too long, we may want to use a [watchdog handler](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/storage.py#L149) so that the cache updates don't penalize the user operation that triggers the need for the update.

PR #2163, opened 2021-09-21
Add rollback to metadata persistence when post_save/delete hooks throw

kevin-bates:
With the recent addition of `post_save()` and `post_delete()` hooks to the `Metadata` subclass, data inconsistencies can arise if either of the "post" methods throw an exception.  In some cases, this may be the intention of the `Metadata` subclass and the expectation would be that the underlying _transaction_ (i.e., create, update, delete) be rolled back.  This change detects that an exception was raised from either post method and takes measures to appropriately roll back the operation via a compensating transaction.  The rollback varies depending on the operation:
1. For create, rollback _deletes_ the newly created instance.
2. For update, rollback _updates_ the instance to its pre-updated value.
3. For delete, rollback _creates_ the pre-deleted value.

It also adds the following note to the help strings for `post_save()` and `post_delete()`:

_Note: Since exceptions thrown from this method can adversely affect the operation, implementations are advised to trap all exceptions unless the operation's rollback is warranted._

The `metadata-test2` schema has been updated to reference a `metadata_class_name` that will be informed, via an environment variable, that it should raise an exception relative to one of create, update or delete operations.  In each case, a different exception is raised and the corresponding test asserts the expected exception was raised and the original transaction was rolled back (by searching for the target instance and/or confirming it was not updated, etc.).  The test then unsets the environment variable and performs the same operation confirming its success.

Note that this PR was originally applied to the branch for #2157 and shows some commits that are not actually included in this pull request.  I'm not sure why those show here but have confirmed that only changes relative to the description above are included.
 
           in the file; or

           it.


kiersten-stokes (approved):
Tested a few scenarios and all are working as intended!

PR #2166, opened 2021-09-22
Add new docker Makefile targets for local development builds

### LINKED ISSUE #2116: Elyra Docker Images to build directly from source
When developing new features, currently there is no way to directly build a Docker image (say for KF Notebooks or Elyra) directly from source. This makes building and testing Docker images a bit troublesome especially in a Kubernetes based environment. 

Would like to see a Docker file and pattern for which the core Pythaon distribution is built as part of the image build. I am thinking this can be done natively in the Docker root folder, however, context is important. Alternatively we can consider implementing this build as part of the makefile.

akchinSTC:
New target will allow users to build test docker images directly
from changes made to local source code.

Fixes #2116

- Dockerfiles accept `dev' as a TAG and builds using the provided wheel built using the local copy of elyra source
- Adds the `ORG` variable to the makefile allowing users to enter their own docker username name for container publishing

 
           in the file; or

           it.


etc/docker/dev/kubeflow/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Ubuntu 20.04.2 LTS (Focal Fossa)
+FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter:v1.3.0
+
+# install - elyra
+COPY elyra-*.whl ./
+RUN python3 -m pip install --quiet --no-cache-dir --use-deprecated legacy-resolver \
```
kevin-bates, line 23: Is this missing the `=` prior to `legacy-resolver`?  It might be helpful to make this identical to the `dev/elyra/Dockerile` and remove the line-break - or add one to the other.

etc/docker/dev/elyra/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    python -m pip install --ignore-installed --upgrade setuptools pandas && \
+    npm config set scripts-prepend-node-path auto && \
+    npm config get prefix /home/jovyan/.npm-global && \
+    mkdir -p /home/jovyan/.npm-global && \
+    npm install -g yarn && \
+    npm install -g npm
+
+RUN python3 -m pip install --quiet --no-cache-dir --use-deprecated=legacy-resolver elyra-*.whl
+
+RUN jupyter lab build
```
kevin-bates, line 41: Can/should these be combined into a single layer (RUN statement)?  (Same question applies for `dev/kubeflow/Dockerfile`.)

akchinSTC: I can combine or separate them. I just did the latter since thats what the kubeflow dockerfile did. Not 100%, I think we combined them with another step to perform cleanup (rm -rf blah*) as well in a single layer so we could keep things as neat as possible along the way. But since we aren't...having them separate should be ok
Tested and there was a 0.01 GB size difference between the two.

Makefile
```diff
@@ -17,14 +17,19 @@
 .PHONY: help purge install uninstall clean test-dependencies lint-server lint-ui lint yarn-install eslint-ui eslint-check-ui prettier-ui prettier-check-ui flake lint-server-dependencies dev-link dev-unlink
 .PHONY: build-ui build-server install-server watch install-extensions build-jupyterlab install-server-package check-install only-install-server
 .PHONY: test-server test-ui test-integration test-integration-debug test docs-dependencies docs dist-ui release pytest
-.PHONY: validate-runtime-images elyra-image publish-elyra-image kf-notebook-image
+.PHONY: validate-runtime-images elyra-image publish-elyra-image kf-notebook-image elyra-dev-images
```
lresende, line 20: Do we need to introduce another task? Can't we handle this based on the tag in use?

etc/docker/dev/elyra/Dockerfile
ptitzler: Combining them would perhaps be the better approach because the additional layer should not provide any benefits. (It's unlikely that Docker would re-use the "RUN python3 ..." layer but not the "RUN jupyterlab ..." during a build.)

Makefile
akchinSTC: Modified the existing dockerfiles to accept `local` as a tag and to builds using the local wheel dist.

etc/docker/dev/kubeflow/Dockerfile
akchinSTC: +1 on catch

etc/docker/dev/elyra/Dockerfile
akchinSTC: done

kevin-bates (approved):
Thanks Alan.

lresende (changes requested):
Launching the built dev images does not produce a runnable Elyra. Testing it here, but this might be the difference between `build-server` vs `release` make targets.

akchinSTC:
build isnt able to locate the ui packages with the current dev version e.g. 3.2.0-dev in npm central. will need to copy these into the docker build directory (thought we packaged the npm packages in the wheel?) and make sure that build uses them instead of looking online

Issue #2139, opened 2021-09-14
Add support for Github-based component registries


kiersten-stokes:
**Is your feature request related to a problem? Please describe.**
#2083 set the groundwork for supporting component registries that include multiple components from a single source type (e.g. a directory containing `.yaml` or `.py` files). We also would like to support registries that contain component definitions from a Github repo.

**Describe the solution you'd like**
Build out the support for searching through a Github repo for component definitions. 

**Considerations**
We will need to figure out how to discriminate between, e.g. files that are component definitions versus files (of the same type) that are not component definitions.

This [article](https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/) may be of use in designing a solution.

Design details to come.

ptitzler:
Since users can already specify public web resources as a source, non-bulk load scenarios are already somewhat supported. Therefore it might be better to defer until we have a better understanding how (KFP/AA) users currently manage component specifications.

Ark-kun:
Some relevant information:

KFP SDK has support for component search and listing: [kfp.components.ComponentStore.search](https://kubeflow-pipelines.readthedocs.io/en/latest/source/kfp.components.html#kfp.components.ComponentStore.search).

The Cloud Pipelines [Pipeline Editor](https://cloud-pipelines.net/pipeline-editor/) app has support for GitHub component search as well.

Pipeline Editor also supports syndicated component feeds: https://github.com/Ark-kun/pipeline_components/blob/pipeline_component_feed/pipeline_component_feed.yaml

thesuperzapper:
@kiersten-stokes @ptitzler @akchinSTC I think we can implement a GitHub component catalog in a very similar way to the `Artifactory Catalog Connector` that will be added in https://github.com/elyra-ai/examples/pull/99.

The gist of the idea is that you have a folder structure like the following:

```
component_1/
   __COMPONENT__
   component-1.0.9.yaml
   component-1.0.10.yaml
component_2/
   hidden_component/
      __COMPONENT__
      component-1.0.0.yaml
      component-1.1.0.yaml
   __COMPONENT__
   component-1.0.0.yaml
   component-1.1.0.yaml
component_3/
   component-1.0.0.yaml
   component-1.1.0.yaml
```

Where the presence of a `__COMPONENT__` marker file, tells the catalog connector that this folder contains components, and also stops further recursion (so in the above example `component_2/hidden_component/` is NOT matched).

The connector has the following configs which change how the connector traverses the folder structure:

[...]

PR #2169, opened 2021-09-22
Parallelize reading of component definitions in registry

kiersten-stokes:
Overall, this PR will speed up reading of component definition by parallelizing `read_component_definition` for each path in a registry.

A function, `read_component_definitions`, is added to the `ComponentReader` base class and called from `read_component_registries()`. For each component path in a registry, this new function creates and starts a thread that will execute `read_component_definition(self, location: str, location_to_def: Dict[str, str])` for its component location.

If the read succeeds, the `location_to_def` dictionary that is passed to `read_component_definition()` is updated with a new key-value pair, where the key is the component location and the value is the component content. Due to dictionary mutability and the fact that keys (component paths) will be unique, this is the means by which the component content for each thread's read is persisted. Once all threads have executed, this dictionary is returned to `read_component_registries()`, which will loop through each key-value pair in order to construct a `Component` object exactly as before.

For the `elyra-airflow-url-preconfig` registry, which includes 5 url-based component paths, the read has been sped up by a factor of ~7-8x (from ~1.5 seconds to read and parse the given components to ~0.2 seconds) in my tests. 

Other minor changes include:

- Moved the `get_absolute_locations()` logic into the new `read_component_definitions()` function and out of the registry  logic
- Removes the return of `None` from the component readers in favor of just not adding a new key-value pair to the `location_to_def` object (end-to-end functionality is equivalent in that an 'unreadable' component will be skipped and not have a `Component` object created for it)
- Separated out the code that retrieves all the registries for a given runtime into its own function (`get_registries_for_runtime()`) in order to clean up `read_component_registries()` 

Existing tests have been updated accordingly. I'll probably want to add comments or otherwise tweak those tests to clarify why variables are being accessed the way that they are.

Not sure if this warrants any new tests.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/tests/test_component_parser_airflow.py
+++ b/elyra/pipeline/airflow/tests/test_component_parser_airflow.py
@@ -49,6 +49,7 @@ def test_modify_component_registries():
     parser = AirflowComponentParser()
     component_registry = ComponentRegistry(parser, caching_enabled=False)
     initial_components = component_registry.get_all_components()
+    initial_components = sorted(initial_components, key=lambda component: component.id)
 
     metadata_manager = MetadataManager(namespace=MetadataManager.NAMESPACE_COMPONENT_REGISTRIES)
 
@@ -94,6 +95,7 @@ def test_modify_component_registries():
     # Delete the test registry
     metadata_manager.remove("new_registry")
     post_delete_components = component_registry.get_all_components()
+    post_delete_components = sorted(post_delete_components, key=lambda component: component.id)
     assert len(post_delete_components) == len(initial_components)
 
     # Check that the list of component ids is the same as before addition of the test registry
@@ -151,7 +153,7 @@ def test_parse_airflow_component_file():
 
     # Get path to component definition file and read contents
     path = _get_resource_path('airflow_test_operator.py')
-    component_definition = reader.read_component_definition(path)
+    component_definition = reader.read_component_definition(path, {})[path]
 
     # Build entry for parsing
     entry = {
@@ -217,7 +219,7 @@ def test_parse_airflow_component_url():
 
     # Get path to component definition file and read contents
     path = 'https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/bash_operator.py'  # noqa: E501
-    component_definition = reader.read_component_definition(path)
+    component_definition = reader.read_component_definition(path, {})[path]
 
     # Build entry for parsing
     entry = {
@@ -249,7 +251,7 @@ def test_parse_airflow_component_file_no_inputs():
 
     # Get path to component definition file and read contents
     path = _get_resource_path('airflow_test_operator_no_inputs.py')
-    component_definition = reader.read_component_definition(path)
+    component_definition = reader.read_component_definition(path, {})[path]
 
     # Build entry for parsing
     entry = {
@@ -285,8 +287,8 @@ async def test_parse_components_url_invalid_location():
 
     # Get path to an invalid component definition file and read contents
     invalid_path = 'https://nourl.py'
-    component_definition = reader.read_component_definition(invalid_path)
-    assert component_definition is None
+    component_definition = reader.read_component_definition(invalid_path, {})
+    assert component_definition == {}
 
     # Build entry for parsing
     entry = {
--- a/elyra/pipeline/airflow/tests/test_component_parser_airflow.py
+++ b/elyra/pipeline/airflow/tests/test_component_parser_airflow.py
@@ -17,6 +17,7 @@ import os
 from types import SimpleNamespace
 
 import jupyter_core.paths
+import pytest
 
 from elyra.metadata.manager import MetadataManager
 from elyra.metadata.metadata import Metadata
@@ -26,6 +27,11 @@ from elyra.pipeline.component import UrlComponentReader
 from elyra.pipeline.component_registry import ComponentRegistry
 
 COMPONENT_CATALOG_DIRECTORY = os.path.join(jupyter_core.paths.ENV_JUPYTER_PATH[0], 'components')
+
+
+@pytest.fixture
+def invalid_url(request):
+    return request.param
 
 
 def _get_resource_path(filename):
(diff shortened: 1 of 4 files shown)


ptitzler:
Perhaps we should also provide a timeout for the GET requests to prevent excessive waiting?

kiersten-stokes:
@ptitzler I was thinking of asking that but forgot to put it in the description -- thanks for the reminder!

The thread join() function accepts an optional timeout parameter (in seconds) that we could use for that. What do you think would be an appropriate amount of time?

ptitzler:
> The thread join() function accepts an optional timeout parameter (in seconds) that we could use for that. What do you think would be an appropriate amount of time?

I was thinking we should add it to the GET request itself https://github.com/elyra-ai/elyra/blob/6fca9f65c900abcc537098ed4c762f8707e767d5/elyra/pipeline/component.py#L389 and any other future network requests (https://github.com/elyra-ai/elyra/issues/2139 comes to mind) that are more likely to encounter delays. A one size fits all approach (timing out on the join) might not work well because different remote sources might have different response requirements. For example, should we support GitHub registries in the future it might take longer to pull the content from there.

Not sure what an appropriate setting would be for HTTP requests - it really depends on the network environment. We have to find a good balance between waiting long enough and waiting too long - the latter would lead to a meaningful degradation in user experience.  Is two seconds a threshold where people would start to get annoyed if the palette reload doesn't happen fast enough?

kevin-bates:
> Perhaps we should also provide a timeout for the GET requests to prevent excessive waiting?

Isn't this largely moot once the cache update logic is pushed to a separate thread?  

I agree that we should probably have a timeout in order to ensure the threads go away at some point but I'm inclined to deal with this at the thread-join point as @kiersten-stokes mentions.  The primary reason for that is it covers the various location types and, theoretically, we could run into issues if the registries are persisted on network volumes.  By moving this issue into the watchdog observer thread, I think this also buys us some time (pun intended).

Regarding race conditions that arise where the cache update threads complete after the request, I think that's more a function of _how_ the cache is updated and see at least two options:
1. By removing a registry's components from the cache prior to their refresh, we will get the behavior that should the _cache-updating request_ complete before the cache is updated **and** a subsequent request touches the cache, a cache-miss will ensue, triggering the in-request fetch of that single item.   This penalizes the subsequent request, but, assuming the item can be fetched, ensures they are up to date.
2. We could skip the item's removal so that, in the case of a long-running cache update, the stale item is returned and document that the component registry cache is not purely transactional behavior and  "good enough" (much like how NoSQL DBs handle distribution).  Users would then know that a refresh is required in those (hopefully rare) cases.

Another area we may want to apply some thought is thread pool exhaustion.  For example, if there are dozens (or hundreds) of URL entries, we should probably consider queuing the requests and using a fixed number of threads to parallelize across the queue entries.

elyra/pipeline/component.py
```diff
@@ -252,13 +254,46 @@ def resource_type(self):
...
+        # Start a reader thread for each location
+        threads = []
+        for location in self.get_absolute_locations(locations):
+            t = Thread(target=self.read_component_definition, args=(location, location_to_def))
+            threads.append(t)
+            t.start()
+
+        # Wait for all threads to finish
+        for t in threads:
+            t.join()
```
kevin-bates, line 293: I think it would be a good idea to go ahead and limit the number of threads to a configurable amount (thinking a default of 3).  Since `ComponentReader` is a `LoggingConfigurable` this could be easily expressed via traitlets, e.g., `max_reader_threads`.  (Here's an [example of a configured trait from EG](https://github.com/jupyter/enterprise_gateway/blob/d6d6ff59fb33ddb9916f1e396f5e335d5a3dc6b8/enterprise_gateway/mixins.py#L286-L295).)

To implement, we'd toss the locations into a queue and use a semaphore (initialized to `min(queueLen, max_threads)`) to gate the threads and let the threads drain the queue.

Of course, this won't have the overall performance increase as thread-per-location but it will avoid any pathological cases that arise with larger registries.

kevin-bates:
@ptitzler and I had a brief sidebar and it might be a good idea to add timeouts to both the request and the join (with the join's timeout being some fixed amount longer than the request's).  That said, I think we should add this behavior in the watchdog effort since it isn't directly related to the parallelling behavior.  We can look closer into that in that PR - if that's okay.

elyra/pipeline/component.py
```diff
@@ -271,8 +277,7 @@ def read_component_definition(self,
 
     def read_component_definitions(self, locations: List[str]) -> Dict[str, str]:
         """
-        This function starts a thread for each of the given locations in order to read
-        component definitions in parallel.
+        This function starts 3 or fewer threads that read component definitions in parallel.
```
kiersten-stokes, line 280: Probably want to change this from `3` to `max_readers`

kevin-bates (approved):
Looks good @kiersten-stokes - thank you!  Yeah, just touch up the help string it appears this is good - unless you have more.

elyra/pipeline/component.py
```diff
@@ -252,13 +260,57 @@ def resource_type(self):
...
+            loc_q.put_nowait(location)
+
+        def read_with_thread():
+            """Read the contents of a given location from the queue"""
+            while not loc_q.empty():
+                try:
+                    loc = loc_q.get(timeout=.1)
+                except Empty:
+                    continue
+                self.read_component_definition(loc, location_to_def)
```
kevin-bates, line 300: After thinking about this more and since we have a fixed set of threads, we should protect the thread from unwarranted termination by wrapping `self.read_component_definition()` in a `try/except`. 

We should also test this with a "bad edit" where the edit to the registry is for an unavailable or inaccessible URL.  Seems like we need to deal with "poison messages" in some manner.

elyra/pipeline/airflow/tests/test_component_parser_airflow.py
```diff
@@ -280,20 +294,24 @@ def test_parse_airflow_component_file_no_inputs():
     assert properties_json['current_parameters']['component_source'] == component_entry.location
 
 
-async def test_parse_components_url_invalid_location():
+@pytest.mark.parametrize('invalid_url', [
+    'https://nourl.py',  # test an invalid host
+    'https://raw.githubusercontent.com/elyra-ai/elyra/master/elyra/\
+     pipeline/tests/resources/components/invalid_file.py'  # test an invalid file
+], indirect=True)
```
kevin-bates, line 301: :+1:

PR #2170, opened 2021-09-23
Add 'properties' element to run as pipeline template 

### LINKED ISSUE #2168: the "run as pipeline" button is broken in 3.1.0
After PR https://github.com/elyra-ai/elyra/pull/2082, if you press the "run as pipeline" button on a Jupyter notebook you will get the following of errors:
1. the [`Node().label()`](https://github.com/elyra-ai/elyra/blob/c0a62b89556c2ed655482def009e009bb6d134db/elyra/pipeline/pipeline_definition.py#L208) function cannot handle `label` being unset in `self._node['app_data']['ui_data']` (which is always the case with "run as pipeline").
2. the [`PipelineDefinition().validate.()`](https://github.com/elyra-ai/elyra/blob/c0a62b89556c2ed655482def009e009bb6d134db/elyra/pipeline/pipeline_definition.py#L395-L398) function cannot handle `properties` being unset in `primary_pipeline['app_data']` (which is always the case with "run as pipeline").

This means uses cannot use the "run as pipeline" button in elyra `3.1.0`.

---

The first problem can be easily addressed by fixing the  [`Node().label()`](https://github.com/elyra-ai/elyra/blob/c0a62b89556c2ed655482def009e009bb6d134db/elyra/pipeline/pipeline_definition.py#L208) function:
```python
    @property
    def label(self) -> str:
        """
        The node label
        :return:  node label
        """
        # first, try getting label from "app_data"
        node_label = self._node['app_data'].get('label', None)

        # second, try getting label from "ui_data"
        if not node_label:
            node_label = self._node['app_data'].get('ui_data', {}).get('label', None)

        # fail, we haven't found a label
        if not node_label:
            raise ValueError("Could not find label for Node.")

        return node_label
```

But the second one either requires us to disable the `properties` check or we need to set `properties` in `primary_pipeline['app_data']` when using "run as pipeline".

lresende:
Fixes #2168

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/pipeline_definition.py
+++ b/elyra/pipeline/pipeline_definition.py
@@ -115,7 +115,7 @@ class Pipeline(AppDataBase):
                                     'Generic': 'generic'}
 
         if 'properties' in self._node['app_data']:
-            pipeline_type_description = self._node['app_data']['properties'].get('runtime')
+            pipeline_type_description = self._node['app_data']['properties'].get('runtime', 'Generic')
             if pipeline_type_description not in type_description_to_type.keys():
                 raise ValueError(f'Unsupported pipeline runtime: {pipeline_type_description}')
             return type_description_to_type[pipeline_type_description]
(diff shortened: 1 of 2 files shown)


kiersten-stokes (approved):
LGTM! Tested with Airflow and KFP examples

PR #2181, opened 2021-09-30
Fix pipeline name calculation algorithm

### LINKED ISSUE #2088: Conflicting algorithm is used to calculate pipeline name
**Describe the issue**
A conflicting algorithm is used to calculate pipeline name 

Some code/tests are looking for it at
```
pipeline_definitions['pipelines'][0]['app_data'].pop('name')
```

But pipelines are using it as:
```
pipeline_definitions['pipelines'][0]['app_data']['properties'].pop('name')
```

lresende:
Fixes #2088

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/pipeline_definition.py
+++ b/elyra/pipeline/pipeline_definition.py
@@ -128,7 +128,10 @@ class Pipeline(AppDataBase):
         The pipeline name
         :rtype: The pipeline name or `untitled`
         """
-        return self._node['app_data'].get('name') or 'untitled'
+        pipeline_name = 'untitled'
+        if 'properties' in self._node['app_data']:
+            pipeline_name = self._node['app_data']['properties'].get('name', pipeline_name)
+        return pipeline_name
 
     @property
     def source(self) -> str:
--- a/elyra/pipeline/pipeline_definition.py
+++ b/elyra/pipeline/pipeline_definition.py
@@ -128,10 +128,7 @@ class Pipeline(AppDataBase):
         The pipeline name
         :rtype: The pipeline name or `untitled`
         """
-        pipeline_name = 'untitled'
-        if 'properties' in self._node['app_data']:
-            pipeline_name = self._node['app_data']['properties'].get('name', pipeline_name)
-        return pipeline_name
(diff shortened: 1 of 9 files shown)


bourdakos1 (approved):

elyra/pipeline/pipeline_definition.py
```diff
@@ -128,7 +128,10 @@ def name(self) -> str:
         The pipeline name
         :rtype: The pipeline name or `untitled`
         """
-        return self._node['app_data'].get('name') or 'untitled'
+        pipeline_name = 'untitled'
+        if 'properties' in self._node['app_data']:
+            pipeline_name = self._node['app_data']['properties'].get('name', pipeline_name)
```
kevin-bates, line 133: I think the test pipelines need to be updated - pipeline_with_invalid_list_values.json and pipeline_valid.json - since they still use a name field at `'app_data'`, not `'app_data.properties'`, yet are also version 5.  Not sure if that's an indication that this fix is incomplete or not.

If we wanted, these lines could be simplified to:
```suggestion
        pipeline_name = self._node['app_data'].get('properties', {}).get('name', 'untitled')
```

kiersten-stokes: Yeah I think this is the case that the sample pipelines are a bit out of date. Only `pipeline_with_invalid_list_values.json` needs to be updated for these test failures (plus a minor change to one of the tests itself). 

There are 5 other sample pipelines that have this versioning issue, and fixing them does not have any negative effect on the current tests, so I'll go ahead and just do that in the same commit

kevin-bates: I see that `pipeline_valid.json` still has `name` in both `app_data` and `app_data.properties`: https://github.com/lresende/elyra/blob/dev-pipeline-title/elyra/tests/pipeline/resources/sample_pipelines/pipeline_valid.json#L32-L44

Seems like we might as well update that resource as well.

kiersten-stokes: Oh, weird, I could have sworn I made that change with these others. Commit incoming

Issue #779, opened 2020-07-25
Deprecation Warnings when opening Pipeline Editor


lresende:
We might need to check some of the `deprecation warnings`  below

```
The Delete16 component has been deprecated and will be removed in the next major version of @carbon/icons-react. 
Delete16@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:502602:18
t@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:175601:21510
div
div
div
t@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:175601:28634
button
Button@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:96621:18
li
ul
li
div
ul
div
ChildWrapper@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:265178:20
ResizeDetector@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:634322:20
t@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:175601:760659
injectIntl(t)
div
div
div
t@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:175601:503085
injectIntl(t)
IntlProvider@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:140560:9
div
PipelineEditor@http://localhost:8888/static/lab/vendors~main.7c5ddb2b6e3847a02153.js:282775:9 react_devtools_backend.js:2273:24
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: CollapsibleSection react_devtools_backend.js:2273:24
```

vabarbosa:
[...]

PR #2193, opened 2021-10-05
Update component registry schema to enforce length for categories

kiersten-stokes:
This PR is a workaround related to [elyra-ai/canvas #779](https://github.com/elyra-ai/canvas/issues/779) but does not directly fix.

This adds a `maxLength` field to the component-registry schema to limit category length to 18 characters. If an attempt is made to add a longer category, the following message is displayed (or, in the case of the CLI, the same wording is printed to the console):

Screen Shot 2021-10-05 at 2 07 57 PM

Other minor changes to the schema include:

- a typo fix in the word "runtimes"
- fixes the schema to remove the brackets around the `items` category details (this is what was preventing the validation from working in earlier tests of the `maxLength` field)
- ~~adds a `uniqueItems` field to categories to avoid repeat category names -- can remove if desired~~ removed because this case is already covered during the palette JSON render
- updates the `Pipeline components` topic in the docs to include length limit information

No test updates are required as core functionality really hasn't changed. Schema validation is already tested in the metadata test module, so I don't think new tests are required either. 

I manually tested CRUD operations via the UI and CLI and all are working as expected.
 
           in the file; or

           it.


elyra/metadata/schemas/component-registry.json
```diff
@@ -38,11 +38,11 @@
           "title": "Category Names",
           "description": "Category names associated with this Component Registry (the components defined in this registry will be organized in the component palette according to these categories)",
           "type": "array",
-          "items": [
```
kevin-bates, line 41: FWIW, the `items` issue is also addressed in the BYO Schemas PR.  This started raising a validation error once the version of `jsonschema` went to `4.0.1` (from `3.2.0`) and must enforce more stringent validators as it supports an updated draft of the JSON schema.

elyra/metadata/schemas/component-registry.json
```diff
@@ -38,11 +38,11 @@
...
-            {
-              "type": "string"
-            }
-          ],
+          "uniqueItems": true,
```
kevin-bates, line 41: This requirement will need to get conveyed to the tooling since this will only be enforced at the time of validation.  Not sure the messaging will be clear from that.

elyra/metadata/schemas/component-registry.json
```diff
@@ -38,11 +38,11 @@
...
+          "items": {
+            "type": "string",
+            "maxLength": 18
```
kevin-bates, line 44: Is this value based on anything definitive?  (I ask purely out of curiosity since it was made two characters shorter in a follow-on commit.)

kevin-bates (commented):
This looks good.  Just the one (meaningful) comment regarding the tooling's knowledge that the list of category names requires unique elements (set).  This might require a tooling change although I think the CLI is pretty good at conveying validation errors.

ptitzler:
Could you also update the documentation topic @ https://github.com/elyra-ai/elyra/blob/master/docs/source/user_guide/pipeline-components.md#category-categories

and indicate what the max length is and update the examples, since at leat one of them is too long (:

elyra/metadata/schemas/component-registry.json
kiersten-stokes: You know what, on further inspection I'm thinking of removing this. Looking through how we render the palette information, it looks like I've already accounted for the case of a repeat instance and it won't be a problem.

I was trying to get fancy for no reason 🥲

kevin-bates: Awesome - thanks for checking on that.  Yeah, this kind of stuff can be more problematic than it's worth.

kiersten-stokes:
@ptitzler yeah I had that same issue with all numbers. Hoping no one decides to go that route 😁

Issue #2208, opened 2021-10-07
Pipeline migration produces invalid container image reference in pipeline file


ptitzler:
**Describe the issue**

Currently we have two hooks touching the same value (runtime image) which are expected to occur sequentially, but are actually triggered separately by changes to the jupyterlab widget content and the canvas content respectively. In theory this should be the same, but is causing this race condition when only the widget detects changes but the canvas does not.

**To Reproduce**
1. Clone https://github.com/elyra-ai/examples
2. Open [`examples/pipelines/dax_noaa_weather_data/analyze_NOAA_weather_data.pipeline`](https://github.com/elyra-ai/examples/blob/master/pipelines/dax_noaa_weather_data/analyze_NOAA_weather_data.pipeline) in a text editor and review all `runtime_image` property values:
    ```
    "runtime_image": "docker.io/amancevice/pandas:1.1.1"
    ```
2. Open [`examples/pipelines/dax_noaa_weather_data/analyze_NOAA_weather_data.pipeline`](https://github.com/elyra-ai/examples/blob/master/pipelines/dax_noaa_weather_data/analyze_NOAA_weather_data.pipeline) in the VPE
3. The migration wizard opens. Migrate the pipeline.
4. Save the pipeline and close the pipeline editor.
5. Open the pipeline file again in a text editor and review the `runtime_image` property values:
   ```
   "runtime_image": "Pandas 1.1.1"
   ```

   Note that the updated values are not valid container image names.
6. Try to run the pipeline. It fails because there is no image with such a name.

**Expected behavior**

The two hooks content should either not be dependent on each other or we should make sure one hook triggers the next if that behavior is expected.

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.2/rc0

ajbozarth:
I've done some initial digging into the code and this seems to stem from  #1671 and possibly #1882 and has been in place since 3.0

IIUC this will still work if using only the VPE since it converts it to the correct format at submission. But it makes the persisted pipeline incompatible with the CLI.

I'm continuing to look into this and figure out the solution

ajbozarth:
Ok after some serious digging I've traced this down to a race condition with two of our onChange hooks.

 One hook is tied to changes to the content of the editor (ie any change via jupyterlab) and sets the value to the display_name so it renders in the dropdown correctly. The other hook changes it back to the actual image value and is tied to various common canvas on change hooks (ie any change via canvas). 

[...]

PR #2213, opened 2021-10-08
Validate runtime image format with regex

kiersten-stokes:
Provides a part of the solution for #2208

Adds a regex match to `_validate_container_image_name()` to catch incorrectly formatted container images. The regex follows the basic rules that we define [here](https://elyra.readthedocs.io/en/latest/user_guide/runtime-image-conf.html#image-name-image-name) in our docs (i.e. hostname is optional, tag name is required, etc.).

**NOTE:** I have 2 regex's currently in the `validation.py` file. The uncommented one is very basic and only checks for the lack of spaces and presence of forward slashes. The commented-out regex is much more complex and follows the [rules](https://docs.docker.com/engine/reference/commandline/tag/) that docker sets forth for their image names. I'm not sure we want to go the more complex route, though, because it's difficult to determine whether other platforms enforce the same rules that docker does. We may also want more of a middle ground where we allow only digits, word characters, underscores, dashes and periods. Let me know in the review, and I'll remove the unused one!

A case was added to an existing test to cover this situation. Also did some manual testing.

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/validation.py
+++ b/elyra/pipeline/validation.py
@@ -375,7 +375,8 @@ class PipelineValidationManager(SingletonConfigurable):
                                              "must conform to the format [registry/]owner/image:tag",
                                      data={"nodeID": node_id,
                                            "nodeName": node_label,
-                                           "propertyName": 'runtime_image'})
+                                           "propertyName": 'runtime_image',
+                                           "imageName": image_name})
 
     def _validate_resource_value(self, node_id: str, node_label: str, resource_name: str,
                                  resource_value: str, response: ValidationResponse) -> None:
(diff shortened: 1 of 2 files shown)


ptitzler:
I'm not sure either one (haven't had a chance to test) would accept something like this
```
quay.io/elyra/kfp-tutorial-count-rows-component@sha256:8b9f58b52ee26ea8aa48eb7b8b04da95cc9bb5b57ad11b554afb250235fcefb8
```
as a valid value - which it should be (ignoring for now that above example URL is a container  image that does not have the prerequisites installed to run a notebook.

Could you please build a dummy runtime container image, obtain the URL and try this? The last command below obtains the URL in the format specified above:
```
$ docker build -t $(CONTAINER_IMAGE_NAME) .
$ docker tag $(CONTAINER_IMAGE_NAME) quay.io/$(QUAY_ORG)/$(CONTAINER_IMAGE_NAME)
$ docker push quay.io/$(QUAY_ORG)/$(CONTAINER_IMAGE_NAME)
$ docker inspect --format="{{index .RepoDigests 0}}" quay.io/$(QUAY_ORG)/$(CONTAINER_IMAGE_NAME)

```

PR #2231, opened 2021-10-14
Fix issues in pipeline migration hook

### LINKED ISSUE #2208: Pipeline migration produces invalid container image reference in pipeline file

ajbozarth:
Fixes two bugs with the migration: multiple dialogs opening and
sometimes setting the runtime image property of the nodes to the
display name instead of the value.

The first bug was partial fixed in #2045 but the code setting the
flag to true was lost during revision.

The second bug described in #2208 was due to a mistake I made in #1860
by using  the transient copy of the pipeline used by the GUI instead of
the persisted content when running the migration.
This caused a race condition when the API call to get the runtime
images returned and triggered the changeHandler to update to the
display names before the user clicked the migrate button in the dialog.

Fixes #2208

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -308,6 +308,7 @@ const PipelineWrapper: React.FC<IProps> = ({
       if (isDialogAlreadyShowing.current) {
         return; // bail, we are already showing a dialog.
       }
+      isDialogAlreadyShowing.current = true;
       if (error instanceof PipelineOutOfDateError) {
         showDialog({
           title: 'Migrate pipeline?',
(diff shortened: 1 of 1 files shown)


lresende:
Can we make sure we have a more descriptive pr/commit description before merging.

PR #2244, opened 2021-10-20
Add support for data exchange between airflow operators

akchinSTC:
Testing Instructions:
1. Clone and pull the pipeline editor [PR](https://github.com/elyra-ai/pipeline-editor/pull/170) and run `make clean dev-link install` 
2. Pull this PR and run `make clean dev-link install`
3. Create a new pipeline, older pipelines will not work due to changes in schema.

Depends on PRs : 
[2263](https://github.com/elyra-ai/elyra/pull/2263)
[170](https://github.com/elyra-ai/pipeline-editor/pull/170)

To Do:
- [x] Modify/fix canvas properties for kfp to accept new component response
- [x] Update and add additional parser tests
- [x] Update labelling for dropdown selections when choosing between control IDs in properties
- [x] Update validation checks for new format of inputs

Small snippet of the slack operator changes
```
{
                "parameter_ref": "elyra_bash_command",
                "control": "custom",
                "custom_control_id": "OneOfControl",
                "label": {
                    "default": "bash_command"
                },
                "description": {
                    "default": "The command, set of commands or reference to a        bash script (must be '.sh') to be executed. (templated) (type: str)",
                    "placement": "on_panel"
                },
                "data": {
                    "controls": {
                        "StringControl": {
                            "label": "Please enter a string value :",
                            "format": "string"
                        },
                        "NestedEnumControl": {
                            "label": "Please select an output from a parent :",
                            "format": "inputpath",
                            "allownooptions": false
                        }
                    },
                    "required": true
                }
            },
            {
                "parameter_ref": "elyra_xcom_push",
                "control": "custom",
                "custom_control_id": "OneOfControl",
                "label": {
                    "default": "xcom_push"
                },
                "description": {
                    "default": "If xcom_push is True, the last line written to stdout        will also be pushed to an XCom when the bash command completes. (type: bool)",
                    "placement": "on_panel"
                },
                "data": {
                    "controls": {
                        "BooleanControl": {
                            "label": "Please select or deselect the checkbox :",
                            "format": "boolean"
                        },
                        "NestedEnumControl": {
                            "label": "Please select an output from a parent :",
                            "format": "inputpath",
                            "allownooptions": false
                        }
                    },
                    "required": false
                }
            },


```

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/component_parser_airflow.py
+++ b/elyra/pipeline/airflow/component_parser_airflow.py
@@ -23,6 +23,7 @@ from typing import Optional
 from elyra.pipeline.component import Component
 from elyra.pipeline.component import ComponentParameter
 from elyra.pipeline.component import ComponentParser
+from elyra.pipeline.component import ControllerMap
 from elyra.pipeline.runtime_type import RuntimeProcessorType
 
 
@@ -44,19 +45,18 @@ class AirflowComponentParser(ComponentParser):
             # Create a Component object for each class
             component_properties = self._parse_properties(component_content)
 
-            components.append(
-                Component(
-                    id=registry_entry.component_id,
-                    name=component_class,
-                    description='',
-                    catalog_type=registry_entry.catalog_type,
-                    source_identifier=registry_entry.component_identifier,
-                    definition=self.get_class_def_as_string(component_content),
-                    runtime_type=self.component_platform.name,
-                    categories=registry_entry.categories,
-                    properties=component_properties
-                )
-            )
+            new_component = Component(id=registry_entry.component_id,
+                                      name=component_class,
+                                      description='',
+                                      catalog_type=registry_entry.catalog_type,
+                                      source_identifier=registry_entry.component_identifier,
+                                      definition=self.get_class_def_as_string(component_content),
+                                      runtime_type=self.component_platform.name,
+                                      categories=registry_entry.categories,
+                                      properties=component_properties
+                                      )
+
+            components.append(new_component)
 
         return components
 
@@ -103,8 +103,11 @@ class AirflowComponentParser(ComponentParser):
             if arg in ['self', '*args', '**kwargs']:
                 continue
 
+            # Component parameter to be required to function correctly
+            required = True
             value = None
             if '=' in arg:
+                required = False  # We can use the default value if nothing is provided
                 arg, value = arg.split('=', 1)[:2]
                 value = ast.literal_eval(value)
                 if value and "\n" in str(value):
@@ -128,16 +131,28 @@ class AirflowComponentParser(ComponentParser):
                 self.log.debug(f"Data type from parsed data ('{data_type}') could not be determined. "
                                f"Proceeding as if 'string' was detected.")
 
-            properties.append(
-                ComponentParameter(
-                    id=arg,
-                    name=arg,
-                    data_type=data_type_info.data_type,
-                    value=(value or data_type_info.default_value),
-                    description=description,
-                    control_id=data_type_info.control_id
-                )
-            )
+            # Override control id since all input properties can take in an xcom
+            control_id = "OneOfControl"
+
+            # Set the default control type
+            default_control_type = data_type_info.control_id
+
+            # Create Dict of control ids that this property can use
+            one_of_control_types = [(default_control_type, data_type_info.data_type,
+                                    ControllerMap[default_control_type].value),
+                                    ("NestedEnumControl", "inputpath",
+                                    ControllerMap["NestedEnumControl"].value)]
+
+            component_params = ComponentParameter(id=arg,
+                                                  name=arg,
+                                                  data_type=data_type_info.data_type,
+                                                  value=(value or data_type_info.default_value),
+                                                  description=description,
+                                                  default_control_type=default_control_type,
+                                                  control_id=control_id,
+                                                  one_of_control_types=one_of_control_types,
+                                                  required=required)
+            properties.append(component_params)
 
         return properties
 
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -297,12 +297,13 @@ be fully qualified (i.e., prefixed with their package names).
                     self.log.debug(f"Processing component parameter '{component_property.name}' "
                                    f"of type '{component_property.data_type}'")
 
-                    if property_value and str(property_value)[0] == '{' and str(property_value)[-1] == '}':
-                        pv = json.loads(json.dumps(property_value))
-                        if isinstance(pv, dict) and set(pv.keys()) == {'value', 'option'}:
-                            parent_node_name = self._get_node_name(sorted_operations, pv['value'])
-                            processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
-                            operation.component_params[component_property.ref] = processed_value
+                    if property_value and str(property_value)[0] == '{' and str(property_value)[-1] == '}' and \
+                        isinstance(json.loads(json.dumps(property_value)), dict) and \
+                            set(json.loads(json.dumps(property_value)).keys()) == {'value', 'option'}:
+                        parent_node_name = self._get_node_name(sorted_operations,
+                                                               json.loads(json.dumps(property_value))['value'])
+                        processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
+                        operation.component_params[component_property.ref] = processed_value
                     elif component_property.data_type == "boolean":
                         operation.component_params[component_property.ref] = property_value
                     elif component_property.data_type == "string":
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -278,12 +278,34 @@ be fully qualified (i.e., prefixed with their package names).
                         continue
 
                     # Get corresponding property's value from parsed pipeline
-                    property_value = operation.component_params.get(component_property.ref)
+                    property_value_dict = operation.component_params.get(component_property.ref)
 
+                    # The type and value of this property can vary depending on what the user chooses
+                    # in the pipeline editor. So we get the current active parameter (e.g. StringControl)
+                    # from the activeControl value
+                    active_property_name = property_value_dict['activeControl']
+
+                    # One we have the value (e.g. StringControl) we use can retrieve the value
+                    # assigned to it
+                    property_value = property_value_dict.get(active_property_name, None)
+
+                    # If the value is not found, assign it the default value assigned in parser
+                    if not property_value:
+                        property_value = component_property.value
+
+                    self.log.debug(f"Active property name : {active_property_name}, value : {property_value}")
                     self.log.debug(f"Processing component parameter '{component_property.name}' "
                                    f"of type '{component_property.data_type}'")
 
-                    if component_property.data_type == "string":
+                    if property_value and str(property_value)[0] == '{' and str(property_value)[-1] == '}':
+                        pv = json.loads(json.dumps(property_value))
+                        if isinstance(pv, dict) and set(pv.keys()) == {'value', 'option'}:
+                            parent_node_name = self._get_node_name(sorted_operations, pv['value'])
+                            processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
+                            operation.component_params[component_property.ref] = processed_value
+                    elif component_property.data_type == "boolean":
+                        operation.component_params[component_property.ref] = property_value
+                    elif component_property.data_type == "string":
                         # Add surrounding quotation marks to string value for correct rendering
                         # in jinja DAG template
                         operation.component_params[component_property.ref] = json.dumps(property_value)
@@ -426,6 +448,11 @@ be fully qualified (i.e., prefixed with their package names).
         if isinstance(converted_value, str):
             converted_value = json.dumps(converted_value)
         return converted_value
+
+    def _get_node_name(self, operations_list: list, node_id: str) -> str:
+        for operation in operations_list:
+            if operation.id == node_id:
+                return operation.name
 
 
 class AirflowPipelineProcessorResponse(PipelineProcessorResponse):
--- a/elyra/pipeline/component.py
+++ b/elyra/pipeline/component.py
@@ -14,11 +14,13 @@
 # limitations under the License.
 #
 from abc import abstractmethod
+from enum import Enum
 from logging import Logger
 from types import SimpleNamespace
 from typing import Any
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 from traitlets.config import LoggingConfigurable
 
@@ -38,6 +40,8 @@ class ComponentParameter(object):
                  required: bool = False,
                  control: str = "custom",
                  control_id: str = "StringControl",
+                 one_of_control_types: Optional[List[Tuple[str, str, str]]] = None,
+                 default_control_type: str = "StringControl",
                  items: Optional[List[str]] = None):
         """
         :param id: Unique identifier for a property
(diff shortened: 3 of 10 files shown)


elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -280,12 +280,24 @@ def _cc_pipeline(self, pipeline, pipeline_name):
                         continue
 
                     # Get corresponding property's value from parsed pipeline
-                    property_value = operation.component_params.get(component_property.ref)
+                    property_value_dict = operation.component_params.get(component_property.ref)
+                    property_value = property_value_dict[property_value_dict['activeControl']]
```
kiersten-stokes, line 284: Is this supposed to be:

```suggestion
                    property_value = property_value_dict['activeControl']
```

kiersten-stokes: Oh wait sorry, I mis-read this. But for some reason I'm still getting an error thrown in this scenario on submit

ajbozarth: are you testing using a completely new pipeline? Because this would probably break if your using a pipeline not created on this PR branch (just guessing without an error message)

kiersten-stokes: That's a good question, so I had to try again. Unfortunately, the same issue happened with a new pipeline, new component/different property under the same circumstances as above (entered a value, then removed it)

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -127,16 +127,25 @@ def _parse_properties(self, content: Dict[str, List]) -> List[ComponentParameter
...
-                    id=arg,
-                    name=arg,
-                    data_type=data_type_info.data_type,
-                    value=(value or data_type_info.default_value),
-                    description=description,
-                    control_id=data_type_info.control_id
-                )
-            )
+            # Override control id since all input properties can take in an xcom
+            control_id = "OneOfControl"
```
kevin-bates, line 131: Could you please expand on the comment of what "OneOfControl" does or implies?  Based on our scrum conversations, `OneOfControl` is kind of a _meta-control_ that "encapsulates" other controls - one of which is chosen by the user (and thus its name: "OneOfControl").  I believe it's the "NestedEnumControl", unconditionally added below, which adds the "xcom" support while the "default_control_type" is more for handling the native type's input.  (If my understanding is correct.)

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -127,16 +127,25 @@ def _parse_properties(self, content: Dict[str, List]) -> List[ComponentParameter
...
+
+            # Set the default control type
+            default_control_type = data_type_info.control_id
+
+            # Create Dict of control ids that this property can use
+            control_types = {default_control_type: data_type_info.data_type,
```
kevin-bates, line 137: Perhaps a name like `one_of_control_types` might clarify things here a bit.

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -280,12 +280,24 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
+
+                    self.log.debug(f"Property value : {property_value}")
```
kevin-bates, line 286: I find this in need of some comments.  Reverse engineering this a bit, `property_value_dict` is a dictionary of name/value pairs where the name (key) is a property name.  In the case of the "activeControl" property, its value is another property name.  Its value is, I'm assuming, the name of the control chosen by the user based on the behavior associated with the "OneOfConrol".

I think some kind of comment explaining this would be helpful along with an update to the debug statement to include the "discovered" property name.  Sans the comment, perhaps something like this:
```suggestion
                    property_value_dict = operation.component_params.get(component_property.ref)
                    property_name = property_value_dict['activeControl']
                    property_value = property_value_dict[property_name]

                    self.log.debug(f"Property name : {property_name}, value : {property_value}")
```

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -280,12 +280,24 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
 
-                    if component_property.data_type == "string":
+                    if property_value and not type(property_value) == bool and \
+                            str(property_value)[0] == '{' and str(property_value)[-1] == '}':
+                        pv = json.loads(json.dumps(property_value))
+                        if isinstance(pv, dict) and set(pv.keys()) == {'value', 'option'}:
+                            parent_node_name = self._get_node_name(sorted_operations, pv['value'])
+                            processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
+                            operation.component_params[component_property.ref] = processed_value
+                    elif component_property.data_type == "boolean":
```
kevin-bates, line 298: is `type(property_value) == bool` always true when `component_property.data_type == "boolean"`?  If so, I would move this to the primary `if` statement so that `not type(property_value) == bool` can be removed from the check for the xcom stuff.

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -280,12 +280,24 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
+                    property_value = property_value_dict[property_value_dict['activeControl']]
+
+                    self.log.debug(f"Property value : {property_value}")
 
                     self.log.debug(f"Processing component parameter '{component_property.name}' "
                                    f"of type '{component_property.data_type}'")
 
-                    if component_property.data_type == "string":
+                    if property_value and not type(property_value) == bool and \
+                            str(property_value)[0] == '{' and str(property_value)[-1] == '}':
```
kevin-bates, line 292: Can `property_value` be `None` or the empty string?  If so, is there a way to prevent that case so this condition (the `property_value` has a value portion) can be removed?

kevin-bates: It just occurred to me that even if `property_value` was of `type` `bool`, that portion of the clause still wouldn't have any impact on the checks for the begin/end curly braces, so is that clause necessary?  If the value was a boolean, it would be an incantation of True or False (1 or 0, etc.) none of which would include curly braces.

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -389,6 +396,11 @@ def _process_list_value(self, value: str) -> Union[List, str]:
             converted_value = json.dumps(converted_value)
         return converted_value
 
+    def _get_node_name(self, operations_list: list, node_id: str) -> str:
+        for operation in operations_list:
+            if operation.id == node_id:
+                return operation.name
```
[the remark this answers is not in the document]
kevin-bates: Is this feasible?

elyra/pipeline/component.py
```diff
@@ -101,6 +108,14 @@ def control(self) -> str:
     def control_id(self) -> str:
         return self._control_id
 
+    @property
+    def control_types(self) -> Dict[str, str]:
```
kevin-bates, line 112: Could you please add help-strings, particularly for these control-oriented properties, to help clarify things?  What you have for the parameter descriptions is probably sufficient.

elyra/pipeline/kfp/component_parser_kfp.py
```diff
@@ -87,9 +87,10 @@ def _parse_properties(self, component_yaml: Dict[str, Any]) -> List[ComponentPar
                 # Change type to reflect the parameter type (inputValue vs inputPath vs outputPath)
                 data_type_adjusted = data_type_parsed
                 if self._is_path_based_parameter(param.get('name'), component_yaml):
-                    data_type_adjusted = f"{param_type[:-1]}Path"
+                    data_type_adjusted = f"{param_type[:-1]}Path"  # i.e. inputpath to inputPath
```
kevin-bates, line 90: I don't understand how the comment relates to the statement.  I assume, from the comment that the value of `{param_type[:-1]}` must be the string `input`, but I don't understand the `inputPath to inputPath` - which implies some kind of transfer or conversion.

kevin-bates (commented):
Thanks Alan.  It would be helpful to add some comments to help clarify things a bit.

elyra/pipeline/airflow/component_parser_airflow.py
akchinSTC: > `OneOfControl` is kind of a _meta-control_ that "encapsulates" other controls - one of which is chosen by the user (and thus its name: "OneOfControl"). I believe it's the "NestedEnumControl", unconditionally added below, which adds the "xcom" support while the "default_control_type" is more for handling the native type's input. (If my understanding is correct.)

Yes! this is pretty much the gist of it.

elyra/pipeline/airflow/component_parser_airflow.py
akchinSTC: done

elyra/pipeline/airflow/processor_airflow.py
akchinSTC: @ajbozarth - I can confirm, after entering a value and deleting it, the resulting json payload sent to the backend is missing the default controller and value. 
```
'bash_command': {'activeControl': 'StringControl'},
```
it should be 
```
'bash_command': {'activeControl': 'StringControl',
                 'StringControl': ''
},
```

elyra/pipeline/airflow/processor_airflow.py
akchinSTC: updated with changes

elyra/pipeline/airflow/processor_airflow.py
akchinSTC: removed, I think this was older remnant from a commit I forgot to remove, thanks

elyra/pipeline/airflow/processor_airflow.py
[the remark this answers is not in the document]
akchinSTC: It probably is, I just need to get back on this horse when I have time

elyra/pipeline/component.py
akchinSTC: done

elyra/pipeline/airflow/processor_airflow.py
ajbozarth: I thought that was odd since it should behave the same as it does currently, so I tested out the same scenario in 3.2.2 and that how it behave there too. If you enter a value and then delete it the parameter will no longer be in the json (ie it's set to undefined). So in this case we should treat the key found in activeControl not being found the same as if the property key is not found. (OR we can discuss changing how we handle falsy values)

elyra/pipeline/kfp/component_parser_kfp.py
kiersten-stokes: After re-familiarizing myself with what's happening here, maybe the following comment makes more sense:

`# define adjusted type as either inputPath or outputPath`

akchinSTC: updated

elyra/pipeline/airflow/processor_airflow.py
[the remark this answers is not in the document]
kevin-bates: No worries - we could tackle this another time as well.

elyra/templates/components/canvas_properties_template.jinja2
```diff
@@ -3,14 +3,23 @@
...
+        "activeControl": "{{ property.default_control_type }}",
+        "{{ property.default_control_type }}":
+    {% endif %}
     {% if property.data_type|lower == "bool" or property.data_type|lower == "boolean" %}
-      "elyra_{{ property.ref }}": {{ property.value|lower }}
+        {{ property.value|lower }}
     {% elif property.data_type|lower == "int" or property.data_type|lower == "integer" or property.data_type|lower == "number" %}
-      "elyra_{{ property.ref }}": {{ property.value }}
+        {{ property.value }}
     {% elif property.type|lower == "dict" %}
```
kevin-bates, line 16: Should this be:
```suggestion
    {% elif property.data_type|lower == "dict" %}
```

elyra/pipeline/component.py
```diff
@@ -101,6 +109,29 @@ def control(self) -> str:
...
+        The `OneOfControl` controller is an encapsulating control ID that allows users to select
+        between multiple input types when configuring the component. For instance, in Airflow, a
+        component parameter can take in, as input, both a value as well as an output from a parent
+        node.
+        When using the `OneOfControl` as the primary control ID for the component parameter,
+        `one_of_control_types` provides canvas a list of control IDs that will be used by the
+        `OneOfControl` controller. These control IDs are what allow the user to select different
+        types of inputs.
+        :return: A list of 3-tuples containing the default_control_type, data_type, label associated with controller
+        """
```
kevin-bates, line 124: Awesome - thank you!

kiersten-stokes (commented):
Sort of a follow-up to the empty/falsy value issue I had last time, it looks like validation does check for this now, and I get the below error details. But there's nothing I (slash the user) can really do to fix this besides entering a value back into those fields, which seems a bit weird if I don't want a value there.

```JSON
{
  "severity": 1,
  "source": "Elyra Pipeline Validation Service",
  "type": "invalidNodeProperty",
  "message": "Node has invalid component parameter information",
  "data": {
    "nodeID": "adb88630-2a20-47ea-ba6c-479ebf6c5955",
    "nodeName": "SimpleHttpOperator",
    "nodeParameters": {
      "activeControl": "StringControl"
    }
  }
}
{
  "severity": 1,
  "source": "Elyra Pipeline Validation Service",
  "type": "invalidNodeProperty",
  "message": "Node has invalid component parameter information",
  "data": {
    "nodeID": "adb88630-2a20-47ea-ba6c-479ebf6c5955",
    "nodeName": "SimpleHttpOperator",
    "nodeParameters": {
      "activeControl": "StringControl"
    }
  }
}
```

Other than this, my data exchange pipelines have been working nicely!

ajbozarth:
> which seems a bit weird if I don't want a value there.

@akchinSTC @kiersten-stokes  my understanding was that the validation is only run on required params

akchinSTC:
> > which seems a bit weird if I don't want a value there.
> 
> @akchinSTC @kiersten-stokes my understanding was that the validation is only run on required params

I dont believe we have the any parameter as required in airflow unless `required` is parsed from the description. So should we disable validation checks on airflow components? 
I think we should at the very least, as @kiersten-stokes states, include the original default parameter value if no value is provided. As for what is required, perhaps we can designate required fields as ones that do not have any default values assigned in the operator constructor?

kiersten-stokes (approved):
Looks awesome! I tested a few different scenarios, including my earlier scenario of adding then deleting the value of a property, and all are working as expected!

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -278,12 +278,34 @@ def _cc_pipeline(self, pipeline, pipeline_name):
...
                     self.log.debug(f"Processing component parameter '{component_property.name}' "
                                    f"of type '{component_property.data_type}'")
 
-                    if component_property.data_type == "string":
+                    if property_value and str(property_value)[0] == '{' and str(property_value)[-1] == '}':
+                        pv = json.loads(json.dumps(property_value))
+                        if isinstance(pv, dict) and set(pv.keys()) == {'value', 'option'}:
+                            parent_node_name = self._get_node_name(sorted_operations, pv['value'])
+                            processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
+                            operation.component_params[component_property.ref] = processed_value
```
kiersten-stokes, line 305: While working on the tutorial, I found a bug with processing of dictionary types. I entered a dictionary-value into the `data` field of the SimpleHTTPOperator (using the following format `{'key': 'value'}`). That property is then being processed in this block rather than in the `elif component_property.data_type == 'dictionary'` block as it should, and as a result it's getting assigned the value of the control dictionary. 

Here's a screenshot of the produced DAG:

Screen Shot 2021-11-11 at 6 38 57 PM

akchinSTC: fixed! 
# Operator source: {'url': 'https://raw.githubusercontent.com/apache/airflow/1.10.15/airflow/operators/http_operator.py'}
```
op_379d0a63_873d_4f0e_b1d1_0a59eda6fe7d = SimpleHttpOperator(
    task_id="SimpleHttpOperator",
    endpoint="api/test",
    method="POST",
    data={"test": "test_value"},
    headers={},
    response_check="",
    extra_options={},
    xcom_push=False,
    http_conn_id="http_default",
    log_response=False,
    dag=dag,
)
```

akchinSTC: also small minor fix for the dict in default values not adding the curly braces

PR #2257, opened 2021-10-27
Improve KF authentication handling

### LINKED ISSUE #2107: create "Authentication Type" selector in Kubeflow Pipelines Runtime
Currently, Elyra assumes that all Kubeflow Pipelines APIs are either unsecured or secured by dex (which also needs to be generalized: https://github.com/elyra-ai/elyra/issues/2106). 

There are effectively infinite ways that a Kubeflow Pipelines API could be secured (Istio IngressGateways can be secured in many ways), but we should am to support the most common ones:
- unsecured (no istio auth)
- dex
- `KF_PIPELINES_SA_TOKEN_PATH`
    - https://github.com/kubeflow/pipelines/issues/5138#issuecomment-797007907
    - https://github.com/kubeflow/website/pull/2905

We should allow the user to EXPLICITLY specify HOW they want to authenticate with the Kubeflow Pipelines API. The best way to do this is by extending the "Runtimes" editing screen to include an `Authentication Type` dropdown.

### Picture of current Kubeflow Pipelines "Runtime Configs":
Screen Shot 2021-09-07 at 11 58 29 am

### LINKED ISSUE #2108: support Kubeflow Pipelines `KF_PIPELINES_SA_TOKEN_PATH` authentication method
As of [`kfp` SDK `1.7.0`](https://pypi.org/project/kfp/#history), there is a new authentication method involving setting an environment variable called `KF_PIPELINES_SA_TOKEN_PATH`, we should support this authentication method for Elyra running inside Kubeflow Notebooks.

__You can read more about `KF_PIPELINES_SA_TOKEN_PATH` in the following Kubeflow Issues/PRs:__
  - https://github.com/kubeflow/pipelines/issues/5138#issuecomment-797007907
  - https://github.com/kubeflow/website/pull/2905

_NOTE: I think we should resolve https://github.com/elyra-ai/elyra/issues/2107 before adding support for `KF_PIPELINES_SA_TOKEN_PATH`_

### LINKED ISSUE #2240: Istio error kubeflow 1.4
**Describe the issue**

After upgrading to kubeflow manifests 1.4.0 - with kfp 1.7.0, elyra 3.2.1 fails with the logs below. 

From the sdk, the pipelines are created/updated and runned without any issues.

**To Reproduce**
Run a simple pipeline and point it to the ml-pipeline endpoint. 

**Screenshots or log output**
<pre>
  File "/opt/conda/lib/python3.7/site-packages/elyra/pipeline/kfp/processor_kfp.py", line 113, in process
    password=api_password
  File "/opt/conda/lib/python3.7/site-packages/elyra/pipeline/kfp/processor_kfp.py", line 761, in _get_istio_auth_session
    f"HTTP status code '{resp.status_code}' for GET against: {url}"
RuntimeError: HTTP status code '404' for GET against: http://ml-pipeline.kubeflow.svc.cluster.local:8888

The above exception was the direct cause of the following exception:

  File "/opt/conda/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/opt/conda/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 118, in post
  File "/opt/conda/lib/python3.7/site-packages/elyra/pipeline/processor.py", line 114, in process
    res = await asyncio.get_event_loop().run_in_executor(None, processor.process, pipeline)
  File "/opt/conda/lib/python3.7/concurrent/futures/thread.py", line 57, in run
  File "/opt/conda/lib/python3.7/site-packages/elyra/pipeline/kfp/processor_kfp.py", line 119, in process
    ) from ex
RuntimeError: Failed to create istio auth session for Kubeflow endpoint: 'http://ml-pipeline.kubeflow.svc.cluster.local:8888' - Check Kubeflow Pipelines runtime configuration: 'kfp_pipelines'
</pre>
</details>

**Expected behavior**
The update and run of the pipeline is successful. 

**Deployment information**
 - Elyra version: v3.2.1
 - Installation source: PyPI
 - Deployment type: Kubeflow manifests 1.4.0
 - Operating system: linux

Thanks

ptitzler:
This PR 
- enables the user to choose an authentication type for Kubeflow Runtime configurations
- provides support for an initial set of authentication types:
  - no authentication
  - DEX with static password authentication
  - DEX with LDAP authentication
  - DEX (legacy/generic)
  - Kubernetes service account token
- updates  the `Runtime Configuration` topic in the user guide (the help link on the runtime configuration screen links to this topic)
- adds code that augments legacy runtime configurations (Elyra < 3.3) with authentication type information based on existing configuration settings
- sets the authentication type for new runtime configurations to "No authentication"

image

Not that this implementation does not provide BYO authentication type support (no plug-and-play like we provide in https://github.com/elyra-ai/elyra/pull/2241 for component catalog types)

Closes #2240 
Closes #2107 
Closes #2108

 - Manual testing using
    - KF secured using DEX/static pwd
    - KF secured using DEX/LDAP
    - KF secured using DEX/static pwd + DEX/LDAP
 - Reviewed output of `make docs`

Known limitations:
 - Irrespective of how many authentication inputs an authenticator requires (e.g. 0 for `No Authentication`, or 2 for `DEX/LDAP`) the UI currently always displays two input fields with static labels (`username` and `password`). The authenticators ignore any values a user might enter if they are not applicable for the selected authentication type.

Additional work is required (and outside the scope of this PR) to improve the user experience:
 - display user-friendly authentication type (https://github.com/elyra-ai/elyra/pull/2257#discussion_r741094757)

- Manual testing
- reviewed the output of `make docs`

 
           in the file; or

           it.


ptitzler:
Starting with the least controversial one: For `None` "authentication" implementation the intention was to 
- Confirm that the URL is valid. 
- Confirm that KF is unsecured. This behavior is/will be consistent with the behavior of other providers (with exception of `Auto`). If a user specifies one authentication approach (e.g. LDAP) but that approach is not enabled/configured in KF, it should be treated as an error.

ptitzler:
Re: `Auto` - generic authentication. I believe this is a must have for the reasons @portellaa provided. While there might be implicitly some overlap (`Auto` might work for static password authentication and so might `Static password`) there is a chance that a designated `Static password` authentication does not work whereas `Auto` does.

ptitzler:
The latest code drop
- Fixes a bug in the NONE/not secured authenticator
- Removes the authenticator that was introduced in 3.2
- Includes static password and ldap authenticators based on @thesuperzapper's  code (with small modifications). Both authenticators raise an error if KF is not secured or secured using a provider other than the one the user selected.

I haven't had a chance yet to look at the token scenario.

ptitzler:
> the PR right now would be very confusing to users, the auth providers all overlap in features, and don't actually tell you what kind of authentication they will use.

Agreed and resolved. Wherever possible error messages will indicate if a mismatch between selected auth provider and configured auth provider was detected.

ptitzler:
The main motivation for introducing "`Auto`" was to achieve backward compatibility (users can continue to use their existing KF runtime configurations as is) and to support authentication types that are not explicitly supported otherwise (LDAP and static password) or might not work as expected (for the supported types).

I believe there is a solution that avoids the need for breaking changes and aligns the behavior of "`Auto`" (perhaps a better name -> "`Generic`") with the behavior of the other authenticators:
- Legacy KF runtime configurations will be implicitly converted to "No authentication" (if no credentials are stored in the configuration) or "Generic" (if id/password are stored in the configuration) using schema filters (need to confirm with @kevin-bates).
- Users can still explicitly choose any other authentication type in the runtime configuration as desired. Once they do the selection is persisted and no more implicit auth type conversion as outlined above is performed.
- The "Generic" auth type raises an auth error if the KF endpoint is not secured (same behavior as other auth types) and tries to authenticate using the provided credentials without making any changes to the auth flow that KF initiates. The latter is an important differentiator to the LDAP/static password approach: if, for example, DEX responds with a request to choose an authentication type, "Generic" authentication won't pick one.

kevin-bates:
> Legacy KF runtime configurations will be implicitly converted to "No authentication" (if no credentials are stored in the configuration) or "Generic" (if id/password are stored in the configuration) using schema filters (need to confirm with @kevin-bates).

Since "KF runtime configurations" are _instances_ of the `kfp` schema, these kinds of changes would need to be applied using the `metadata_class_name` reference such that when an instance of the kfp-schema is used, that instance will actually be an instance of `metadata_class_name` where we could probably use the `on_load()` method to make these adjustments and, essentially, _migrate on the fly_.

Although not immediately necessary at this time, it would be good to introduce a class hierarchy that introduces a base class from which all Runtimes schemas would at least reference.  Then, for cases like this, the kfp subclass of that base would be used to apply kfp-specific changes.

Something like this this:
```python
class Metadata(object):  # today's base instance
    pass

class RuntimesMetadata(Metadata):  # schemaspace-specific metadata class
    pass

class KfpMetadata(RuntimesMetadata):  # applies changes specific to kfp schema
    def on_load(self, **kwargs: Any) -> None:
        pass
```
Then the kfp schema would add:
```JSON
"metadata_class_name": "elyra.pipeline.kfp.kfp_metadata.KfpMetadata",
```
Examples of this will be available once changes for BYO catalogs and RuntimeProcessorType are merged, as we're doing this kind of thing there.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+    STATIC_PASSWORD = 'Static password'
+    # KF is secured using KF_PIPELINES_SA_TOKEN_PATH
+    # (See SATokenPathKFPAuthenticator implementation)
+    KF_PIPELINES_SA_TOKEN_PATH = 'KF_PIPELINES_SA_TOKEN_PATH'
+    # Supports multiple authentication mechanisms
+    # (See AutoKFPAuthenticator implementation)
+    AUTO = 'Auto'
+    # Supports LDAP authentication
+    # (See LDAPKFPAuthenticator implementation)
+    LDAP = 'LDAP'
```
thesuperzapper, line 56: @ptitzler @portellaa @kevin-bates This is NOT a generic LDAP authenticator, it is a dex-specific one.
(It won't work in any situation except if you are using dex with its LDAP connector)

Given that, this should be called `DEX_LDAP_CONNECTOR` to avoid confusion.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+    "auth_type" property in elyra/kfp/metadata/schemas/kfp.json
+    to enable users to select the provider from the list.
+
+    """
+    # KF is not secured
+    # (See UnsecuredKFPAuthenticator)
+    NO_AUTHENTICATION = 'No authentication'
+    # KF is secured using static id/password
+    # (See StaticPasswordKFPAuthenticator implementation)
+    STATIC_PASSWORD = 'Static password'
```
thesuperzapper, line 47: @ptitzler @portellaa @kevin-bates This is NOT a generic "username/password" authenticator, it is a dex-specific one.
(It won't work in any situation except if you are using dex with StaticPasswords)

Given that, this should be called `DEX_STATIC_PASSWORDS` to avoid confusion.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+            if re.search(r"/auth$", redirect_url_obj.path):
+                redirect_url_obj = redirect_url_obj._replace(
+                    path=re.sub(r"/auth$", "/auth/ldap", redirect_url_obj.path)
+                )
+            else:
+                # verify that KF is secured by LDAP
+                m = re.search(r"/auth/([^/]*)/?", redirect_url_obj.path)
+                if m and m.group(1) != 'ldap':
+                    raise AuthenticationError(
+                        f'The Kubeflow server at {kf_endpoint} is secured using \'{m.group(1)}\'. '
```
thesuperzapper, line 445: This error message should indicate you MIGHT be using `Dex - {m.group(1)}`, there is nothing dex-specific about having `/auth/{local,ldap}` in your path.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+                                      f'Update runtime configuration \'{runtime_config_name}\' and try again.',
+                                      provider=self._type,
+                                      request_history=request_history)
+
+        # Running in a Kubernetes pod, kfp.Client can utilize environment
+        # variable KF_PIPELINES_SA_TOKEN_PATH for authentication. Verify
+        # that the variable is defined in the current environment.
+        env_var_name = 'KF_PIPELINES_SA_TOKEN_PATH'
+        if os.environ.get(env_var_name) is None:
+            raise AuthenticationError(f'Environment variable {env_var_name} is undefined.',
```
thesuperzapper, line 518: There is no requirement to set `KF_PIPELINES_SA_TOKEN_PATH` when using this type of authentication.

If `KF_PIPELINES_SA_TOKEN_PATH` is unspecified, the `kfp.Client()` will use `/var/run/secrets/kubeflow/pipelines/token`.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+    NO_AUTHENTICATION = 'No authentication'
+    # KF is secured using static id/password
+    # (See StaticPasswordKFPAuthenticator implementation)
+    STATIC_PASSWORD = 'Static password'
+    # KF is secured using KF_PIPELINES_SA_TOKEN_PATH
+    # (See SATokenPathKFPAuthenticator implementation)
+    KF_PIPELINES_SA_TOKEN_PATH = 'KF_PIPELINES_SA_TOKEN_PATH'
+    # Supports multiple authentication mechanisms
+    # (See AutoKFPAuthenticator implementation)
+    AUTO = 'Auto'
```
thesuperzapper, line 53: I strongly think this should NOT be called `AUTO`, as this implies we are doing some kind of authentication-type detection. 

If you really want to keep the old authentication code in Elyra, the only sensible name would be `DEX_LEGACY`.

__NOTE:__ the code for `AUTO` right now is actually just a less sophisticated version of the `DEX_STATIC_PASSWORDS` auth (It won't work if multiple Dex providers are enabled).

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+    """
+    # KF is not secured
+    # (See UnsecuredKFPAuthenticator)
+    NO_AUTHENTICATION = 'No authentication'
+    # KF is secured using static id/password
+    # (See StaticPasswordKFPAuthenticator implementation)
+    STATIC_PASSWORD = 'Static password'
+    # KF is secured using KF_PIPELINES_SA_TOKEN_PATH
+    # (See SATokenPathKFPAuthenticator implementation)
+    KF_PIPELINES_SA_TOKEN_PATH = 'KF_PIPELINES_SA_TOKEN_PATH'
```
thesuperzapper, line 50: I think it would be more clear to name this authentication-type based on how it works, rather than a Kubeflow environment variable.

I think a more clear name would be `K8S_SERVICEACCOUNT_TOKEN` because:
1. It highlights this authentication type will ONLY work on Kubernetes (not if Elyra is running on your laptop)
2. `KF_PIPELINES_SA_TOKEN_PATH` doesn't have to be defined to use this auth (Note, `kfp.Client()` defaults to `/var/run/secrets/kubeflow/pipelines/token` when unset)
2. It describes what method is actually being used (a ServiceAccount token mounted to the pod by Kubernetes)

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+                redirect_url_obj = redirect_url_obj._replace(
+                    path=re.sub(r"/auth$", "/auth/local", redirect_url_obj.path)
+                )
+            else:
+                # verify that KF is secured by LDAP
+                m = re.search(r"/auth/([^/]*)/?", redirect_url_obj.path)
+                if m and m.group(1) != 'local':
+                    raise AuthenticationError(
+                        f'The Kubeflow server at {kf_endpoint} is secured using \'{m.group(1)}\'. '
+                        f'Update runtime configuration \'{runtime_config_name}\' and try again.',
```
thesuperzapper, line 329: This error message should indicate you MIGHT be using `Dex - {m.group(1)}`, there is nothing dex-specific about having `/auth/{local,ldap}` in your path.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,605 @@
...
+    def authenticate(self,
+                     kf_endpoint: str,
+                     runtime_config_name: str) -> Optional[str]:
+
+        request_history = []
+
+        # Verify the API endpoint
+        resp = requests.get(kf_endpoint)
+        request_history.append({'request_url': kf_endpoint, 'response': resp})
+        if resp.status_code != HTTPStatus.OK:
```
thesuperzapper, line 506: Why are you checking for HTTP 200 here, there are many other successful HTTP codes?

thesuperzapper (changes requested):
@ptitzler @portellaa @kevin-bates I have put my comments inline to help discuss each one individually.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: changed! (name, value and related class names)

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: changed! (name, value and related class names)

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: I changed the display label to `DEX (generic)` and added information to the documentation when to choose this type.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: I agree and have updated the code + user-facing information. If you have good reference URLs that we should include please let me know.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: Not sure I understand your concern. The error message text does not include "DEX" for that exact reason. Whether or not the extracted character sequence (e.g. `xyz`) is meaningful to the user I cannot comment on.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: You are right. The main purpose of this test was to validate that the URL is valid. I've updated the code to follow redirects, which should now yield status code 200.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: Fixed! This should now work as expected based on your earlier input.

ptitzler:
Still working on a solution for Python DSL export.

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,693 @@
...
+class KFPAuthenticator():
+    """
+    Use this class to authenticate with Kubeflow Pipelines. The authenticate
+    method delegates the actual authentication to an implementation of the
+    AbstractAuthenticator class.
+    """
+
+    def authenticate(self,
+                     api_endpoint: str,
+                     auth_type: str,
```
kevin-bates, line 143: I think we should use the enumerated type here: 
```suggestion
                     auth_type: SupportedAuthProviders,
```

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,693 @@
...
+        auth_info = {
+            'api_endpoint': kf_url,   # KF API endpoint, source: runtime config
+            'auth_type': None,        # Authentication type, source: runtime config
+            'kf_secured': False,      # Indicates whether KF API is secured
+            'cookies': None,          # passed to KFP SDK client as "cookies" param value
+            'existing_token': None    # passed to KFP SDK client as "existing_token" param value
+        }
+
+        try:
+            auth_info['auth_type'] = auth_type
```
kevin-bates, line 178: I suggest using the name property of the Enum so that we preserve the level of indirection to the named value.  This would reference the string `'NO_AUTHENTICATION'` for example.
```suggestion
            auth_info['auth_type'] = auth_type.name
```

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,693 @@
...
+
+            # Process the authentication request using the appropriate authenticator
+            # implementation. Refer to the class definitions for information how
+            # the request is processed
+
+            if auth_type == SupportedAuthProviders.NO_AUTHENTICATION.value:
```
kevin-bates, line 184: Then compare the types.
```suggestion
            if auth_type == SupportedAuthProviders.NO_AUTHENTICATION:
```

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,693 @@
...
+                                      provider=self._type)
+        return None
+
+
+class DEXStaticPasswordAuthenticator(AbstractAuthenticator):
+    """
+    Authenticator for DEX/static passwords
+    """
+
+    _type = SupportedAuthProviders.DEX_STATIC_PASSWORDS.value
```
kevin-bates, line 310:
```suggestion
    _type = SupportedAuthProviders.DEX_STATIC_PASSWORDS
```

elyra/pipeline/kfp/kfp_metadata.py
```diff
@@ -0,0 +1,42 @@
...
+            # - api_username and api_password present -> use DEX Legacy
+            # - otherwise -> use no authentication type
+            if self.metadata.get('api_username') is None or\
+               len(self.metadata.get('api_username').strip()) == 0 or\
+               self.metadata.get('api_password') is None or\
+               len(self.metadata.get('api_password').strip()) == 0:
+                self.metadata['auth_type'] = SupportedAuthProviders.NO_AUTHENTICATION.value
+            else:
+                self.metadata['auth_type'] = SupportedAuthProviders.DEX_LEGACY.value
+
```
kevin-bates, line 41: You need to write out the updated instance, else these changes will occur every time.
```python
MetadataManager(schemaspace="runtimes").update(self.name, self, for_migration=True)
```
(Note: this requires the merge of #2241)

elyra/metadata/schemas/kfp.json
```diff
@@ -53,6 +54,34 @@
...
+          "title": "Authentication Type",
+          "description": "Authentication type Elyra uses to authenticate with Kubeflow",
+          "type": "string",
+          "enum": [
+            "DEX (generic)",
+            "DEX with LDAP",
+            "DEX with static passwords",
+            "Kubernetes service account token",
+            "No authentication"
+          ],
```
kevin-bates, line 78: Hmm - these will contradict my other review comment about preserving the level of indirection to the displayable value.  If we use the enum _name_ fields here, we are free to change the displayable values.  Of course, I'm not sure where _values_ will be displayed other than in the metadata editor.  Not sure how best to handle enums in a metadata-driven app.

One approach might be to _build_ a `uihint`via the Runtimes schemaspace implementation that looks for auth_type and essentially puts a mapping from the enum's `name` to its `value`.  The metadata editor persists the field using the `name` but displays the `value` when such a hint is present (if that makes sense).

cc: @marthacryan

docs/source/user_guide/runtime-conf.md
```diff
@@ -115,6 +116,7 @@ elyra-metadata install runtimes \
        --name="my_kubeflow_pipelines_runtime" \
        --display_name="My Kubeflow Pipelines Runtime" \
        --api_endpoint=https://kubernetes-service.ibm.com/pipeline \
+       --auth_type="DEX with static password" \
```
thesuperzapper, line 119: Is `--auth_type="DEX with static password"` actually how you refer to `DEX_STATIC_PASSWORDS`, that seems a bit crazy.

Why not `--auth_type=DEX_STATIC_PASSWORDS`?

elyra/pipeline/kfp/kfp_authentication.py
thesuperzapper: @ptitzler What I am trying to say is, this error should describe what actually went wrong.

In this case, an error like `Expected "/auth/ldap" HTTP path for DEX_LDAP_CONNECTOR, but gox "/auth/XXXX"` would be much more useful to end users.

docs/source/user_guide/runtime-conf.md
thesuperzapper, line 119: @ptitzler Is `--auth_type="DEX with static password"` actually how you refer to `DEX_STATIC_PASSWORDS`, because that seems a bit crazy.

Why not `--auth_type=DEX_STATIC_PASSWORDS`?

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,693 @@
...
+    DEX_STATIC_PASSWORDS = 'DEX with static passwords'
+    # KF is secured using K8S_SERVICEACCOUNT_TOKEN
+    # (See K8sServiceAccountTokenAuthenticator implementation)
+    K8S_SERVICEACCOUNT_TOKEN = 'Kubernetes service account token'
+    # Supports multiple authentication mechanisms
+    # (See DEXLegacyAuthenticator implementation)
+    DEX_LEGACY = 'DEX (generic)'
+    # Supports DEX with LDAP authentication
+    # (See DEXLDAPAuthenticator implementation)
+    DEX_LDAP_CONNECTOR = 'DEX with LDAP'
```
thesuperzapper, line 74: nit: I think a more natural ordering for these is:

- `NO_AUTHENTICATION`
- `K8S_SERVICEACCOUNT_TOKEN`
- `DEX_LDAP_CONNECTOR`
- `DEX_STATIC_PASSWORDS`
- `DEX_LEGACY`

elyra/metadata/schemas/kfp.json
thesuperzapper: @ptitzler similar to my [comments here](https://github.com/elyra-ai/elyra/pull/2257/files#r740680262), I think the enum should be strings without whitespace in them:
- `NO_AUTHENTICATION`
- `K8S_SERVICEACCOUNT_TOKEN`
- `DEX_LDAP_CONNECTOR`
- `DEX_STATIC_PASSWORDS`
- `DEX_LEGACY`

kevin-bates: Hi Mathew - this is an area where being metadata-driven can be a bit thorny.  On the one hand, we'd like to present users with nice displayable (potentially localizable) values, while, on the other, introduce a level of indirection for the _persisted_ values that essentially remain fixed and are not privy to the whims of others inherent in human display values.  This is exactly what is driving my previous _`name_value_map`_ comment.

To elaborate more on the proposed `uihints`, I think this schema stanza would resemble something like the following:
```JSON
        "auth_type": {
          "title": "Authentication Type",
          "description": "Authentication type Elyra uses to authenticate with Kubeflow",
          "type": "string",
          "enum": [
            "DEX_LEGACY",
            "DEX_LDAP_CONNECTOR",
            "DEX_STATIC_PASSWORDS",
            "K8S_SERVICEACCOUNT_TOKEN",
            "NO_AUTHENTICATION"
          ],
          "default": "No authentication",
          "uihints": {
            "field_type": "dropdown",
            "category": "Kubeflow Pipelines",
            "name_value_map": {
                "NO_AUTHENTICATION": "No authentication",
                "DEX_STATIC_PASSWORDS": "DEX with static passwords",
                "K8S_SERVICEACCOUNT_TOKEN": "Kubernetes service account token",
                "DEX_LEGACY": "DEX (generic)",
                "DEX_LDAP_CONNECTOR": "DEX with LDAP"
            }
          }
        },
```
The metadata-editor would be responsible for converting schema `enum` values to their displayable value found in the map **and** setting the field to the selected _name_ (constant-looking string) value in order to adhere to the constraints imposed by the _enum_ (so schema validation succeeds).

If we find we can't implement a `uihints.name_value_map` in this release, then we let the constant-looking values ride.

docs/source/user_guide/runtime-conf.md
kevin-bates: I agree that we should use the `name` property for the CLI (and enum values in general - per my previous comment).

elyra/metadata/schemas/kfp.json
ptitzler: The latest updates:
 - use the `SupportedAuthProviders` enum to automatically populate the list of supported authentication types in the GUI
 - display the enum member name instead of the enum member value
 - use enum members wherever possible
 - should address all enum related feedback

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: Got it. I updated this error message and it's DEX/static pwd equivalent to include the path and instructions on what to do.

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: Done.

elyra/pipeline/kfp/kfp_metadata.py
ptitzler: Done! (merge of #2241 and added ^^

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+    KUBERNETES_SERVICEACCOUNT_TOKEN = 'Kubernetes service account token'
+    # KF is secured using DEX with static id/password
+    # (See StaticPasswordKFPAuthenticator implementation)
+    DEX_STATIC_PASSWORDS = 'DEX with static passwords'
+    # Supports DEX with LDAP authentication
+    # (See DEXLDAPAuthenticator implementation)
+    DEX_LDAP = 'DEX with LDAP'
+    # Supports multiple authentication mechanisms
+    # (See DEXLegacyAuthenticator implementation)
+    DEX_LEGACY = 'DEX (generic)'
```
thesuperzapper, line 71: We should probably update the "text version" of `DEX_LEGACY` to match its internal name:
```python
DEX_LEGACY = 'DEX (Legacy)'
```

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+    in this module. Each entry in this list must be associated
+    with an implementation of AbstractAuthenticator.
+    """
+
+    # KF is not secured
+    # (See NoAuthenticationAuthenticator)
+    NO_AUTHENTICATION = 'No authentication'
+    # KF is secured using KUBERNETES_SERVICEACCOUNT_TOKEN
+    # (See K8sServiceAccountTokenAuthenticator implementation)
+    KUBERNETES_SERVICEACCOUNT_TOKEN = 'Kubernetes service account token'
```
thesuperzapper, line 62: nit: "ServiceAccount" is one word (in the Kubernetes world), making this:
```python
Kubernetes ServiceAccount token
```

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+                raise AuthenticationError(f'Error detecting whether Kubeflow server at {kf_endpoint} is secured: '
+                                          f'HTTP status code {resp.status_code}'
+                                          f'Update runtime configuration \'{runtime_config_name}\' and try again.',
+                                          provider=self._type,
+                                          request_history=request_history)
+
+            if len(resp.history) == 0:
+                # if we were NOT redirected, then the endpoint is UNSECURED
+                # treat this as an error.
+                raise AuthenticationError(f'The Kubeflow server at {kf_endpoint} is not secured using DEX with LDAP. '
```
thesuperzapper, line 403: This is the `DEXStaticPasswordAuthenticator(`, you probably did not mean to say "DEX with LDAP"

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+            redirect_url_obj = urlsplit(resp.url)
+
+            # if we are at `/auth?=xxxx` path, we need to select the
+            # static password auth type
+            if re.search(r"/auth$", redirect_url_obj.path):
+                redirect_url_obj = redirect_url_obj._replace(
+                    path=re.sub(r"/auth$", "/auth/local", redirect_url_obj.path)
+                )
+            else:
+                # verify that KF is secured by LDAP
```
thesuperzapper, line 420: This is the `DEXStaticPasswordAuthenticator(` you probably did not mean to say "verify that KF is secured by LDAP"

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+                        f'The Kubeflow server at {kf_endpoint} redirected to an unexpected HTTP path '
+                        f'(\'{redirect_url_obj.path}\'). Verify that Kubeflow is secured using \'{self._type.name}\''
+                        f' and, if necessary, update the authentication type in runtime configuration '
+                        f'\'{runtime_config_name}\'.',
+                        provider=self._type,
+                        request_history=request_history)
+
+            # if we are at `/auth/xxxx/login` path, then no further action is needed
+            # (we can use it for login POST)
+            if re.search(r"/auth/.*/login$", redirect_url_obj.path):
```
thesuperzapper, line 433: This check should not be for arbitrary `/auth/xxxx/login` but for `/auth/local/login`

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+                        f'The Kubeflow server at {kf_endpoint} redirected to an unexpected HTTP path '
+                        f'(\'{redirect_url_obj.path}\'). Verify that Kubeflow is configured for \'{self._type.name}\''
+                        f' and, if necessary, update the authentication type in runtime configuration '
+                        f'\'{runtime_config_name}\'.',
+                        provider=self._type,
+                        request_history=request_history)
+
+            # if we are at `/auth/xxxx/login` path, then no further action is needed
+            # (we can use it for login POST)
+            if re.search(r"/auth/.*/login$", redirect_url_obj.path):
```
thesuperzapper, line 561: This check should not be for arbitrary `/auth/xxxx/login` but for `/auth/ldap/login`

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+        likely cause a KFP client failure.
+        """
+
+        request_history = []
+
+        # Verify the API endpoint
+        resp = requests.get(kf_endpoint, allow_redirects=True)
+        request_history.append((kf_endpoint, resp))
+        if resp.status_code != HTTPStatus.OK:
+            raise AuthenticationError(f'Error detecting whether Kubeflow server at {kf_endpoint} is secured: '
```
thesuperzapper, line 631: We should be checking that the redirect history is empty because for `KUBERNETES_SERVICEACCOUNT_TOKEN` to work, the pipelines endpoint must be the internal `ml-pipeline-ui.kubeflow.svc.cluster.local` address (or equivalent), not through the `istio-ingressgateway`.

ptitzler: Fixed!

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,739 @@
...
+        :param runtime_config_name: Runtime configuration name where kf_endpoint is specified
+        :type runtime_config_name: str
+        :raises AuthenticationError: A potential issue was detected that will
+        likely cause a KFP client failure.
+        """
+
+        request_history = []
+
+        # Verify the API endpoint
+        resp = requests.get(kf_endpoint, allow_redirects=True)
```
portellaa, line 628: this will continue to fail if we, as users specify an internal endpoint resolved by the kubedns, giving kubeflow as an example, ml-pipeline.kubeflow.svc:8888 doesn't serve an GET request on the base path, neither any other http method.

that will wrongly fail saying that the url is invalid on unreachable and that's not true, when sent to `kfp.Client` it handles it properly.
i don't get why is this check 🤷 

another thing, if this is a secure authenticator, where are you doing any kind of authentication when performing the request? If that endpoint is secure, it is more then acceptable that it returns not found if you don't do any kind of authentication.

And, in a standard way of doing this, the ideal method to do this would be OPTIONS, which is used to describe possible methods that are accepted at the given endpoint.

ptitzler: For `K8sServiceAccountTokenAuthenticator` the underlying idea was to determine whether there are any missing prerequisites that would cause the kfp sdk client to fail, not to actually perform authentication. For example, if the resolved service account token path is invalid, we know that the client won't be able to connect.

I don't think adding more options to cover _every possible configuration scenario_ is the way to go. If my assumption was wrong that the provided URL can always be resolved then I'll remove the connectivity test.

ptitzler:
Latest update
 - incorporates review feedback (no connectivity check for service account tokens, etc)
 - adds convenience method `to_dict` to `SupportedAuthProviders` enum in support of https://github.com/elyra-ai/elyra/pull/2257#discussion_r741094757
 - standardizes spelling of  "service account" (not "serviceaccount") to align with Kubernetes naming convention (enum member name is now `KUBERNETES_SERVICE_ACCOUNT_TOKEN` and all messages and comments should now be consistent)
 - adds defensive code that properly handles the scenario where a developer adds a new provider to `SupportedAuthProviders` but forgets to implement it

docs/source/user_guide/runtime-conf.md
```diff
@@ -181,16 +183,23 @@ The namespace used to run your pipeline in Kubeflow Pipelines. This setting is r
 
 Example: `anonymous`
 
+##### Kubeflow authentication type (auth_type)
+Authentication type Elyra uses to gain access to Kubeflow Pipelines. This setting is required. Supported types are:
+- No authentication (`NO_AUTHENTICATION`).
+- Kubernetes service account token (`KUBERNETES_SERVICEACCOUNT_TOKEN`). This authentication type is only supported if Elyra runs as a pod in Kubernetes, e.g. as a Kubeflow notebook server. You must configure a service account token in Kubernetes, as outlined [here](https://www.kubeflow.org/docs/components/pipelines/sdk/connect-api/#multi-user-mode).
```
thesuperzapper, line 189: This needs to be updated to `KUBERNETES_SERVICE_ACCOUNT_TOKEN`

docs/source/user_guide/runtime-conf.md
```diff
@@ -181,16 +183,23 @@ The namespace used to run your pipeline in Kubeflow Pipelines. This setting is r
...
+- DEX configured for static password authentication (`DEX_STATIC_PASSWORDS`). This authentication requires a username and a password.
+- DEX configured for LDAP authentication (`DEX_LDAP`). This authentication requires a username and a  password.
+- DEX (`DEX_LEGACY`). Use this type only if none of the other authentication types applies or if your Kubeflow deployment is not configured for any other listed type. This authentication requires a username and a password.
```
thesuperzapper, line 192: I think we should state that `DEX_LEGACY` is the same code as Elyra `3.1.1` and earlier.

(Also, that `DEX_LEGACY` is deprecated, and will be removed in a future release)

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -0,0 +1,772 @@
...
+    NO_AUTHENTICATION = 'No authentication'
+    # KF is secured using KUBERNETES_SERVICE_ACCOUNT_TOKEN
+    # (See K8sServiceAccountTokenAuthenticator implementation)
+    KUBERNETES_SERVICE_ACCOUNT_TOKEN = 'Kubernetes service account token'
+    # KF is secured using DEX with static id/password
+    # (See StaticPasswordKFPAuthenticator implementation)
+    DEX_STATIC_PASSWORDS = 'DEX with static passwords'
+    # Supports DEX with LDAP authentication
+    # (See DEXLDAPAuthenticator implementation)
+    DEX_LDAP = 'DEX with LDAP'
```
thesuperzapper, line 68: For consistency with `DEX_LEGACY`, I think the "description" of these should be:

```python
DEX_STATIC_PASSWORDS = 'DEX (static passwords)'

DEX_LDAP = 'DEX (ldap)'
```

elyra/templates/kfp/kfp_template.jinja2
```diff
@@ -66,128 +60,10 @@ def create_pipeline():
...
-            password=api_password
-        )
-    except Exception as ex:
-        raise RuntimeError(
-            f"Failed to create istio auth session for Kubeflow endpoint: '{api_endpoint}'"
-        ) from ex
-
-    auth_cookie = auth_session['session_cookie']
-{% else %}
+    # TODO authentication
```
thesuperzapper, line 65: What is this line?

docs/source/user_guide/runtime-conf.md
ptitzler: While I agree that it would be desirable, I'm not sure removal will be possible unless we can make the other (auth specific) providers work for everyone (See @portellaa's problem).

elyra/pipeline/kfp/kfp_authentication.py
ptitzler: Changed. Note that the enum member names will be displayed instead of the values until https://github.com/elyra-ai/elyra/issues/2268 is supported.

elyra/templates/kfp/kfp_template.jinja2
ptitzler: Pipeline export to Python DSL is currently not supported (and thus disabled). We have not seen any evidence that anybody actually used that feature in a long time. I didn't want to invest time adding jinja templates for the authentication code if it is decided to permanently remove this feature.  The TODO is just a reminder that the template does not generate any authentication code.

Users that want to export pipelines can still use the YAML format as an alternative.

docs/source/user_guide/runtime-conf.md
thesuperzapper: What kind of auth are they using, do you know how their cluster is set up? 

In my mind, the idea is to only have single-kind auth-types in the future, that will allow us to do advanced error handling and provide useful messages in logs.

ptitzler:
@thesuperzapper @portellaa @andre-lx Thank you so much for your feedback. Much appreciated! We will soon be cutting a first 3.3 release candidate (waiting for other PRs to be approved). If there's anything that should not/cannot be deferred to a subsequent [patch] release please let me know.

akchinSTC (approved):
tested dex auth types internal test clusters, looks good im ok to proceed.

PR #2277, opened 2021-11-04
Sort palette nodes by component label

### LINKED ISSUE #2171: Consider sorting components in palette for consistent display
With threading now implemented to read and parse component definitions, the order in which the components are rendered may change depending on the order in which threads execute.

Discussion is needed to determine where the sorting should be done. 

We can also consider just leaving it as-is as a visual for when components are updating.

kiersten-stokes:
Resolves #2171 (again)

After the merge of #2241, the component `op`s are no longer in alphabetical order according to the label (name) of the component. 

Note that `localeCompare` is already case insensitive.

This PR gets the label from the palette nodes from the `app_data.ui_data.label` field. The top-level `label` field (not within `app_data.ui_data`) appears to be available on each node object, but when trying to directly access it directly, build fails with the following error: 

```
@elyra/pipeline-editor-extension: src/pipeline-hooks.ts(142,9): error TS2339: Property 'label' does not exist on type '{ op: string; id: string; type: "execution_node"; inputs: { app_data: any; }[]; outputs: { app_data: any; }[]; app_data: any; }'.
```

I **think** this might have something to do with Canvas and how they enforce attributes on a node? I'm guessing this is what @bourdakos1 was alluding to with his comment [here](https://github.com/elyra-ai/elyra/blob/5eceb1753e45801bfc356a91582d045cb551daf1/packages/pipeline-editor/src/pipeline-hooks.ts#L124).

All this to say, this PR seems to work, but I don't know if it's the best way to get it done. Will definitely need some input from the frontend experts.

Tested manually to ensure that components are sorted by label.
 
Screen Shot 2021-11-04 at 5 37 47 PM

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/pipeline-hooks.spec.ts
+++ b/packages/pipeline-editor/src/pipeline-hooks.spec.ts
@@ -45,7 +45,8 @@ const createMockCategory = (
 
 const createMockComponent = (name: string): Component => {
   return {
-    op: name,
+    op: 'string',
+    label: name,
     id: 'string',
     type: 'execution_node',
     inputs: [],
--- a/packages/pipeline-editor/src/pipeline-hooks.ts
+++ b/packages/pipeline-editor/src/pipeline-hooks.ts
@@ -85,6 +85,7 @@ export interface IRuntimeComponent {
   node_types: {
     op: string;
     id: string;
+    label: string;
     type: 'execution_node';
     inputs: { app_data: any }[];
     outputs: { app_data: any }[];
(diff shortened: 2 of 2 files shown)


bourdakos1:
@kiersten-stokes I think we manually created the type definition, if there is a label at the top level we can update it and typescript will stop complaining

bourdakos1:
I can push some changes to this tomorrow morning

kiersten-stokes:
> @kiersten-stokes I think we manually created the type definition, if there is a label at the top level we can update it and typescript will stop complaining

Ah ok good, that's easier than I thought it would be! (well, for someone besides me, that is 😄). But yeah `label` is there at the top level when I checked out each object so we're good to go on that front

bourdakos1:
pushed changes, but did not do any testing haha 🤞

bourdakos1:
Tested, seems to be working

PR #2287, opened 2021-11-11
Refactor Pipeline Editor UI to better support Runtime Types

bourdakos1:

This PR updates the `pipeline-editor-extension` to only use and persist `app_data.runtime_type`. At submission/export time a runtime config is chosen. This `app_data.runtime_config` field is the only value the server should use and can use it to determine platform/processor if needed.

In addition to these changes, a few things have been consolidated:
- a new `RuntimeConfigSelect` has been created to be used across all dialogs (run, export, run script as pipeline, run notebook as pipeline)
- `handleRunPipeline` and `handleExportPipeline` have been merged into one function
- `SubmitNotebookButtonExtension` and `SubmitScriptButtonExtension` have been merged into one extension

These changes almost cut the amount of code in half and makes submission logic unified and easier to update.

           in the file; or

           it.


kevin-bates:
Hi @bourdakos1 - this is looking great.  

It looks like you're noticing the _pipeline_ type vs, _runtime_ type phenomenon introduced by "generic" pipelines.  There are also some potential issues relative to "generic" pipeline types in that, today, it assumed they run on all (2) supported runtime types, but that won't always be true since "generic" is really about runtime _implementation_ (i.e., processor). So as soon as someone brings their own Kubeflow or Airflow processors (or introduces a third type (with implementation)) that don't support "generic" _components_ (Notebook, Python, R) then "generic" pipelines break.

lresende:
> Hi @bourdakos1 - this is looking great.  
> 
> 
> 
> It looks like you're noticing the _pipeline_ type vs, _runtime_ type phenomenon introduced by "generic" pipelines.  There are also some potential issues relative to "generic" pipeline types in that, today, it assumed they run on all (2) supported runtime types, but that won't always be true since "generic" is really about runtime _implementation_ (i.e., processor). So as soon as someone brings their own Kubeflow or Airflow processors (or introduces a third type (with implementation)) that don't support "generic" _components_ (Notebook, Python, R) then "generic" pipelines break.

The ability to run generic components is an assumption we have adopted and should hold true until its not true anymore.


Also, not sure I am too late here, but I noticed 'runtime_type' vs 'runtime-config'. Could we be consistent on the pipeline property name and use dash.

kevin-bates:
> Also, not sure I am too late here, but I noticed 'runtime_type' vs 'runtime-config'. Could we be consistent on the pipeline property name and use dash.

I agree with consistency, but it seems like an underscore would be more consistent as all other "separated" names in the pipeline use underscores - only `runtime-config` uses a dash.  Since we require a migration for this release, this is a good opportunity to switch to `runtime_config`.

bourdakos1:
Changing runtime-config to runtime_config wouldn’t require a migration since it’s not persisted

karlaspuldaro (approved):
LGTM!

PR #2305, opened 2021-11-16
Add support for runtime type resources

kevin-bates:
This pull request sets up the framework for retrieving resources corresponding to _active/configured_ runtime type (Kubeflow Pipelines, Apache Airflow, Argo, etc.).

The primary location for this information is a `RuntimeTypeResources` subclass which consists of static values pointing at the platform's icon, its export file types (and their displayable value) along with the runtime-type's displayable name.

A new endpoint `pipeline/runtimes/types` has been added that, when accessed, returns a list of resources corresponding to each _active runtime-type_.  

The Elyra server extension has been updated to expose a `static` directory with an `icons` sub-directory in which the icons are located and to which the icon information within the `RuntimeTypeResources` instance points.

`GET` against `http://127.0.0.1:8888/elyra/pipeline/runtimes/types` returns:
```JSON
{
  "runtime_types": [{
    "id": "APACHE_AIRFLOW",
    "display_name": "Apache Airflow",
    "icon": "static/elyra/airflow.svg",
    "export_file_types": [{
      "id": "py",
      "display_name": "Airflow domain-specific language Python code"
    }]
  }, {
    "id": "LOCAL",
    "display_name": "Local",
    "icon": "static/elyra/pipeline-flow.svg",
    "export_file_types": []
  }, {
    "id": "KUBEFLOW_PIPELINES",
    "display_name": "Kubeflow Pipelines",
    "icon": "static/elyra/kubeflow.svg",
    "export_file_types": [{
      "id": "py",
      "display_name": "KFP domain-specific language Python code"
    }, {
      "id": "yaml",
      "display_name": "KFP static configuration file (YAML formatted)"
    }]
  }]
}
```

`GET` against `http://127.0.0.1:8888/static/elyra/kubeflow.svg` returns the Kubeflow icon.

 
           in the file; or

           it.


lresende:
We just introduce a runtime type, and we call this Runtimes everywhere, should we just call this `/runtimes`?

Also, could you clarify `file_types`? Are those actually related to export and thus should be called something like `export_file_types`?  

Now you see how we get to so many names :) for similar things :)


NOT A CONTRIBUTION

kevin-bates:
> We just introduce a runtime type, and we call this Runtimes everywhere, should we just call this /runtimes?

Yes.  And we're _still_ in the process of introducing `runtime type`.  In my opinion,  _runtime type_ identifies a _platform_.  Yes, after implementing `runtime_type` I was feeling more and more, this should be named `platform_type` but felt it was too close to the release cycle to change it.

I can move the endpoint back to `pipeline/runtimes`, which is what I had before, but felt it was confusing with the OTHER `runtimes` term already in use.  The one that maps `Runtimes` schemas to Runtime processors - which are implementations.

> Also, could you clarify file_types? Are those actually related to export and thus should be called something like export_file_types?

`file_types` is what is used on the front end, I could care less what its called so long as the front end knows what field name to use.  I will update to `export_file_types` - that sounds really great.  Thank you.

kevin-bates:
The following outlines various terms used throughout Elyra in an attempt to come to a better understanding for what names should be applied down the road.

(These definitions are based solely on _my understanding_ and may be incorrect.)

`runtime`: This refers to the environment in which the pipeline is _processed_ and identifies a `PipelineProcessor` _implementation_.  A requirement for _Bring Your Own Runtime_ is that the runtime implementations entrypoint name match the metadata schema within the `Runtimes` schemaspace.  Instances of these schemas consist of properties used by the runtime implementation to perform its work.  Examples of runtimes today are `kfp` and `airflow` and `local` (although there is no corresponding schema for `local` runtimes because its a special case (which is wrong IMHO)).

`runtime_type`: Since `runtime` was already defined, `runtime_type` was introduced to convey the _type_ of runtime.  In Elyra, examples of `runtime_type` are "Kubeflow Pipelines" or "Apache Airflow" (we also added "Local" for consistency).  `runtime_type` is necessary because users may implement their own `runtime` for "Kubeflow Pipelines" and the application need to be able to differentiate between the two implementations.  In addition, various resources, like components, should be available across `runtime` implementations of the same `runtime_type`.  "Platform" is a commonly-used synonym for `runtime_type`.

`RuntimeProcessorType`: The class (enum) used to convey the `runtime_type` value.  It has both a symbolic name form: `KUBEFLOW_PIPELINES` as well as a value that is more of a display name: `Kubeflow Pipelines`.  Any persisted forms of `runtime_type` should use the symbolic name form since the display name value can be changed and introduces an appropriate level of indirection to accomplish that.

In order to better support BYO Runtimes, there are a number of resources that can be associated with a `runtime_type` that should also be derived from the server.  By doing this, there (theoretically) should be ZERO front-end changes necessary to introduce a new `runtime_type` or an implementation (i.e., `runtime`) of an existing `runtime_type`.  These resources must be associated with the `runtime_type` and not the `runtime`.  Examples of resources are the icon associated with the runtime-type and the set of file types supported for export (although these _could_ be associated with the implementation or `runtime` which may also include things like "support for the generic components" - so there's probably a similar need for that, but that affects all implementations and requires more thought).

The original plan was to alter the `RuntimeProcessorType` enum to be a different class.  I wanted to honor the release cycle and hesitated to introduce that change, so I implemented another class with the idea that `RuntimeProcessorType` could be subsumed in another release.  The name I chose for the other class was `RuntimePlatformInfo`.  I understand its a new term and perhaps that was a mistake.  Perhaps we just make the change to `RuntimeProcessorType` now? 

Whatever we do, we should just do the right thing now and not introduce a release that is only going to have to change in the next revision.

kiersten-stokes (commented):
Haven't had a chance to review in full yet, but this solved the error I was having in 3.3.0 rc0 testing with exporting a Generic pipeline when no runtimes were configured. 

Before:
Screen Shot 2021-11-18 at 4 49 27 PM

After:
Screen Shot 2021-11-18 at 5 04 34 PM

kiersten-stokes:
Looks like the test failure is actually related to the comment I made above. While this behavior in the second screenshot above is expected for that case, it looks like it **also** pops up with that message when a runtime config is defined.

I think the test fix may eventually involve adding the following to the `pipeline.ts` file at `it('should run pipeline after adding runtime image', () => {` (line 361) depending on how this issue is fixed.

```
// Create runtime configuration
cy.createRuntimeConfig({ type: 'kfp' });
```

kiersten-stokes (changes requested):
See my comment above

bourdakos1:
Found the issue! New generic pipelines were being created with a `runtime_type` of `LOCAL` instead of undefined. I was using a pipeline not created in this PR to test locally so didn't notice the issue. Should be a quick fix on the UI side

kiersten-stokes (commented):
One remaining edge case that I found is that I get the following when I try to export a generic pipeline when no runtime configurations are defined:

Screen Shot 2021-11-19 at 12 42 42 PM

Submit works as expected with no runtimes defined:
Screen Shot 2021-11-19 at 12 42 59 PM

And all cases work as expected when 1 or more runtime configs are defined!

That being said, I'm having some weird environment issues, so maybe that's not reproduceable?

bourdakos1:
> One remaining edge case that I found is that I get the following when I try to export a generic pipeline when no runtime configurations are defined

I think this bug was introduced in the other PR. Maybe we would open a separate PR to fix?

kiersten-stokes (approved):
> > One remaining edge case that I found is that I get the following when I try to export a generic pipeline when no runtime configurations are defined
> 
> I think this bug was introduced in the other PR. Maybe we would open a separate PR to fix?

That sounds right. In that case, LGTM!

bourdakos1:
Opened PR fixes the bug here https://github.com/elyra-ai/elyra/pull/2314

PR #2313, opened 2021-11-19
Set the default xcom_push for Airflow operators

akchinSTC:

Update `xcom_push` boolean to default to True so downstream operators will be able to read in values in case the user forgets to explicitly set it.

Existing unit tests were updated to reflect the change
 
           in the file; or

           it.


elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -127,6 +127,10 @@ def _parse_properties(self, content: Dict[str, List]) -> List[ComponentParameter
             description = self._format_description(description=description, data_type=data_type)
 
             data_type_info = self.determine_type_information(data_type)
+
+            if 'xcom' in arg.lower() and data_type_info.data_type == 'boolean':
+                data_type_info.default_value = True
+
             if data_type_info.undetermined:
                 self.log.debug(f"Data type from parsed data ('{data_type}') could not be determined. "
                                f"Proceeding as if 'string' was detected.")
```
kevin-bates, line 136: This is mostly a nit - so up to you, but I did a double-take that the `default_value = True` could be applied to an `undetermined` type and had to go look at `determine_type_information()` to find that isn't the case.  I would suggest swapping the two statements and not assume that `undetermined` types will always be strings (although, so long as they remain a usable component, 'string' is the right choice - this is more about visibility).  Perhaps something like the following?
```suggestion
            data_type_info = self.determine_type_information(data_type)
            if data_type_info.undetermined:
                self.log.debug(f"Data type from parsed data ('{data_type}') could not be determined. "
                               f"Proceeding as if '{data_type_info.data_type}' was detected.")
            elif 'xcom' in arg.lower() and data_type_info.data_type == 'boolean':
                data_type_info.default_value = True
```

akchinSTC: thx!!! committed

elyra/pipeline/validation.py
```diff
@@ -422,23 +419,38 @@ async def _validate_custom_component_node_properties(self,
...
+                    # TODO: Update this hardcoded check for xcom_push. This parameter is specific to a runtime
+                    # (Airflow). i.e. abstraction for byo validation?
+                    node_param_value = node_param['NestedEnumControl'].get('value')
+                    xcom_node = pipeline_definition.get_node(node_param_value)
+                    xcom_param = xcom_node.get_component_parameter('xcom_push')
+                    xcom_value = xcom_param.get('BooleanControl')
+                    if not xcom_value:
+                        response.add_message(severity=ValidationSeverity.Error,
+                                             message_type="invalidNodeProperty",
+                                             message="Node contains an invalid inputpath reference. The parent "
```
kiersten-stokes, line 449: NIT: I think referencing `inputpath` might be confusing to people that only use Airflow. May be worth changing it from `inputpath` to `input`?

I'm happy to make the change if we intend to try to get this merged today and others agree

kiersten-stokes (approved):
Tried a few scenarios, and all are working as expected! I have one NIT below

elyra/pipeline/validation.py
ptitzler: How about `input source reference`?

ptitzler (changes requested):
There does seem to be a problem. Even though my test pipeline file contains
```
            "xcom_push": {
                "activeControl": "BooleanControl",
                "BooleanControl": false
              },
```
the generated DAG has the  `xcom_push` property set:

image

ptitzler:
> There does seem to be a problem. 

@kiersten-stokes' commit appears to have resolved this. Thank you!

elyra/pipeline/validation.py
```diff
@@ -422,23 +419,38 @@ async def _validate_custom_component_node_properties(self,
...
+                    node_param_value = node_param.get('value')
+                    if node_param_value not in parent_list:
                         response.add_message(severity=ValidationSeverity.Error,
                                              message_type="invalidNodeProperty",
                                              message="Node contains an invalid inputpath reference. Please "
                                                      "check your node-to-node connections",
                                              data={"nodeID": node.id,
                                                    "nodeName": node.label})
+                elif isinstance(node_param, dict) and node_param.get('activeControl') == 'NestedEnumControl':
+                    # TODO: Update this hardcoded check for xcom_push. This parameter is specific to a runtime
```
ptitzler, line 440: The code below appears to assume that `NestedEnumControl` is only applicable to Airflow because line 446 (`if not xcom_value:`) will trigger an error if no `xcom_value` is present. For runtime types other than Airflow, lines 443 onwards would all yield unexpected results. Instead of deferring a resolution, the code below should already tolerate the absence of `xcom_*` in the pipeline file.

kiersten-stokes: I can push a basic fix for this (i.e. just check for the existence of the `xcom_param` before moving on). But it looks like we might still want something more robust for a long-term solution

ptitzler: Yes.

ptitzler:
The checking doesn't seem to work anymore:

Child node:

image

Parent node:

image

Pipeline passes validation:

image

ptitzler (approved):
LGTM

PR #2316, opened 2021-11-22
support operators with type hints

wenjun93:
This fix pipeline component parser failure for component `__init__` with type hints, default parameters assignment with empty str alongside `=`  can be fixed as well.

           in the file; or

           it.

### DIFF
--- a/etc/generic/requirements-elyra.txt
+++ b/etc/generic/requirements-elyra.txt
@@ -15,7 +15,6 @@ requests==2.25.1
 tornado==6.0.4
 traitlets==4.3.3
 urllib3==1.26.5
-astunparse==1.6.3
 #
 # These excluded are transitive dependencies of the included python packages.
 #ansiwrap==0.8.4
--- a/etc/generic/requirements-elyra.txt
+++ b/etc/generic/requirements-elyra.txt
@@ -15,6 +15,7 @@ requests==2.25.1
 tornado==6.0.4
 traitlets==4.3.3
 urllib3==1.26.5
+astunparse==1.6.3
 #
 # These excluded are transitive dependencies of the included python packages.
 #ansiwrap==0.8.4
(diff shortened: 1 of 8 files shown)


kiersten-stokes:
Hi @wenjun93! Thank you so much for your contribution - this is an area of the code that we have been wanting to clean up.

I see that this PR is still a WIP, but I have a few suggestions that may or may not help when it's ready for review. I think the area that you want to be looking in is the `_parse_properties()` function around line 109. Here we check for the presence of default value assignment (`=`), but we don't make any checks for `:` and type hints like we should.

I would also recommend adding a test for this particular case since parsing can result in so many edge cases (as the existence of this PR itself proves 😄 ). Our current tests for this functionality are located in `elyra/tests/pipeline/airflow/test_component_parser_airflow.py:172-290`. These 3 tests check against a dummy operator files located in `elyra/tests/pipeline/resource/components` (`airflow_test_operator.py` and `airflow_test_operator_no_inputs.py`). If you're familiar with pytest and feel comfortable doing so, I think it would be worthwhile to create a new resource file - maybe something like `airflow_test_operator_type_hints.py` and add a test to `test_component_parser_airflow.py` as well. 

Feel free to reach out with any questions!

wenjun93:
@kiersten-stokes thanks for the detailed hints! updated.

etc/generic/requirements-elyra.txt
```diff
@@ -15,6 +15,7 @@ requests==2.25.1
 tornado==6.0.4
 traitlets==4.3.3
 urllib3==1.26.5
+astunparse==1.6.3
```
akchinSTC, line 18: Probably dont need to change this requirements file since this is strictly for generic operators. 
Remove this dependency and adding it to the root `setup.py` should resolve the CI issue

wenjun93: @akchinSTC  oh, I did put the requirement mistakenly, moved now

kiersten-stokes (approved):
LGTM! Very nice test case

PR #2335, opened 2021-12-02
Add support for AWS IRSA

thesuperzapper:

The goal of this PR is to improve support for using S3 as the Cloud Object Store. Specifically it enables support for [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) authentication, in addition to the existing hard-coded `AWS_ACCESS_KEY_ID` (`cos_username`) and `AWS_SECRET_ACCESS_KEY` (`cos_password`) configs.

__Other Changes:__
- Make `cos_username` and `cos_password` NOT required fields for new runtimes
    - (IRSA requires that no access id/key is set) 
- Updates `minio` python client library from `6.X.X` to `7.X.X`
   - (Minio client `7.0.0` was the first version to support IRSA)
- removes `test_fail_bad_endpoint_main_method()` test
   - (Minio `7.X.X` no longer can throws an `InvalidEndpointError` error for invalid `cos-endpoint`)

- Running `make test`
- In an AWS EKS cluster with IRSA enabled

[Edit] For testing set environment variables as follows prior to launching JupyterLab:
 - `ELYRA_GITHUB_ORG` to `thesuperzapper`
 - `ELYRA_GITHUB_BRANCH` to `cos-support-irsa`

 
----

           in the file; or

           it.


ptitzler:
The suggested change
- requires updates to the content in user_guide/runtime-conf.md
- should introduce an 'authentication type' selector for COS in the runtime schema (similar to what we already have for KFP authentication) 

I've already associated this PR with https://github.com/elyra-ai/elyra/issues/2338. Once delivered, 2338 should enable us to resolve the usability issue (only show/enforce properties that are applicable)

thesuperzapper:
> The suggested change
> 
> * requires updates to the content in user_guide/runtime-conf.md

I am happy to do this change

> * should introduce an 'authentication type' selector for COS in the runtime schema (similar to what we already have for KFP authentication)

I am not sure we need to wait for this larger change to enable this feature, no existing feature should break with this change, it just allows you to NOT specify `cos_username` and `cos_password`.

Secondarily, it's probably a good idea to get this merged first, so we can update the `minio` python client library to `7.X.X` without also introducing lots of other things (keeps PRs small).

---

@ptitzler Do you think we can get this merged if I update the docs you asked?

elyra/airflow/bootstrapper.py
```diff
@@ -74,10 +75,28 @@ def __init__(self, **kwargs: Any) -> None:
         # Infer secure from the endpoint's scheme.
         self.secure = self.cos_endpoint.scheme == 'https'
 
-        self.cos_client = minio.Minio(self.cos_endpoint.netloc,
-                                      access_key=os.getenv('AWS_ACCESS_KEY_ID'),
-                                      secret_key=os.getenv('AWS_SECRET_ACCESS_KEY'),
-                                      secure=self.secure)
+        # get minio credentials provider
+        if "cos-user" in self.input_params and "cos-password" in self.input_params:
```
lresende, line 79: Are we still passing this? I believe we removed it from input_params for security issues (leaking these details) and only pass them as ENV VARS?

elyra/airflow/bootstrapper.py
```diff
@@ -74,10 +75,28 @@ def __init__(self, **kwargs: Any) -> None:
...
+            cred_provider = providers.StaticProvider(
+                access_key=self.input_params.get("cos-user"),
+                secret_key=self.input_params.get("cos-password"),
+            )
+        elif "AWS_ACCESS_KEY_ID" in os.environ and "AWS_SECRET_ACCESS_KEY" in os.environ:
+            cred_provider = providers.EnvAWSProvider()
+        elif "AWS_ROLE_ARN" in os.environ and "AWS_WEB_IDENTITY_TOKEN_FILE" in os.environ:
```
lresende, line 86: How is the role being propagated to the container? Are you assuming that the container would have the info and use the same for all users?

elyra/airflow/bootstrapper.py
thesuperzapper: This is for the unit tests, it's not used in production.

Here is the [`_get_operation_instance()`](https://github.com/elyra-ai/elyra/blob/67779c3917a4587bd8a9f2e589addda0fc27a64d/elyra/tests/kfp/test_bootstrapper.py#L155-L156) function which sets `cos-user` and `cos-password` used in unit tests.

elyra/airflow/bootstrapper.py
thesuperzapper: This is how IRSA works, it injects `AWS_ROLE_ARN` and  `AWS_WEB_IDENTITY_TOKEN_FILE` based of the Kubernetes Pod's ServiceAccount.

If the user has chosen to not pass `cos_user` and `cos_password` we check if they have set up IRSA (out of band, with respect to Elyra).

Does that answer your question?

ptitzler:
> I am not sure we need to wait for this larger change to enable this feature, no existing feature should break with this change, it just allows you to NOT specify `cos_username` and `cos_password`.
> 
> Secondarily, it's probably a good idea to get this merged first, so we can update the `minio` python client library to `7.X.X` without also introducing lots of other things (keeps PRs small).
> 
> @ptitzler Do you think we can get this merged if I update the docs you asked?

Yes certainly, but a follow-up PR (to be included in the same Elyra release) would be required to update the schema to allow for the selection of an "auth type"  (similar to what we added for [KFP authentication](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/kfp.json#L77-L86)).

The follow-up would 
 - query a new COS auth runtime config property to determine which auth type to use
 - use the new runtime config property value in the code instead of `cos_username` and `cos_secret` to determine how to manipulate the environment variables
 - add migration code for the existing runtime configuration schema ([similar to this](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/kfp/kfp_metadata.py) that sets the COS auth runtime config property
 
If you don't think you'll have the bandwidth to make those follow-up changes let me know and I'll take care of it.

elyra/pipeline/kfp/processor_kfp.py
```diff
@@ -510,15 +510,15 @@ def _cc_pipeline(self,
         runtime_configuration = self._get_metadata_configuration(schemaspace=Runtimes.RUNTIMES_SCHEMASPACE_ID,
                                                                  name=pipeline.runtime_config)
 
-        cos_endpoint = runtime_configuration.metadata['cos_endpoint']
-        cos_username = runtime_configuration.metadata['cos_username']
-        cos_password = runtime_configuration.metadata['cos_password']
+        cos_endpoint = runtime_configuration.metadata.get('cos_endpoint')
```
ptitzler, line 513: Did you mean to treat this as an optional property? The KFP and Airflow schema updates you've made still list `cos_endpoint` as required and the existing code `cos_endpoint = runtime_configuration.metadata['cos_endpoint']` would therefore be more appropriate.

thesuperzapper: I was making this consistent with the [`processor_airflow.py`](https://github.com/elyra-ai/elyra/blob/243dab7a22761ef71c1f34cce0bf4828cdd8e130/elyra/pipeline/airflow/processor_airflow.py#L94).

NOTE: this does not change the fact that `cos_endpoint` is still required, as it is specified under [`metadata/schemas/kfp.json`](https://github.com/elyra-ai/elyra/blob/25ffbe35965426b89bd91bbf69660959bcec6846/elyra/metadata/schemas/kfp.json#L162)

thesuperzapper:
@ptitzler I have updated the PR with docs inside `docs/source/user_guide/runtime-conf.md`.

I believe this should be ready to merge now, unless there are other comments. 

---

Regarding any follow-up PR to add an auth mechanism like we have for Kubeflow, I would greatly appreciate if you could take this up.

thesuperzapper:
Actually, I have found a bug, please don't merge yet.

elyra/pipeline/kfp/processor_kfp.py
ptitzler: Got it! I am all for consistency. However, I confirmed with Alan that the referenced Airflow code should have used `['..']` instead of `.get('')` to access required properties. Since the metadata service enforces the required constraint the existing code should never fail. Can you please revert the change to avoid the impression that this particular property is optional?

ptitzler:
doc updates look good

thesuperzapper:
@ptitzler I am still investigating a possible issue, please give me a couple of hours.

elyra/pipeline/kfp/processor_kfp.py
thesuperzapper: Done in https://github.com/elyra-ai/elyra/pull/2335/commits/4f057142900ccfd0c4502ac493b5549ef230dd7d

thesuperzapper:
@ptitzler I just tested this in a real-world EKS cluster (with IRSA enabled), and after the fix in https://github.com/elyra-ai/elyra/pull/2335/commits/9347b616b20b585b3b44dcd197a9d1a098a936eb, this seems to be working properly now.

From my end, this looks good to merge.

thesuperzapper:
@ptitzler is there anything else we need to do for this to be merged?

I have tested it in a real-world EKS cluster, and it works with IRSA if `cos_username` and `cos_password` are not set.

ptitzler (commented):
LGTM - verified that kfp/aa generic pipeline execution yields the expected results for the legacy use cases:
 - uid + password (no Kubernetes secret defined)
 - uid + password (with Kubernetes secret defined)

No access to AWS - hence didn't test that part.

ptitzler (approved):
LGTM

thesuperzapper:
@ptitzler I assume you also tested it with `ELYRA_GITHUB_ORG` and `ELYRA_GITHUB_BRANCH` env-vars set? Because this PR updates `elyra/airflow/bootstrapper.py`, `elyra/kfp/bootstrapper.py`, `etc/generic/requirements-elyra.txt`, and Elyra will download these from the main `elyra-ai/elyra` repo by default.
- `export ELYRA_GITHUB_ORG=thesuperzapper`
- `export ELYRA_GITHUB_BRANCH=cos-support-irsa`

___PS:__ We should probably ensure our unit tests have these variables set 
(assuming we are actually doing end-to-end Kubeflow pipelines testes in the CI/CD)_

thesuperzapper:
@ptitzler I am happy for this to be merged if you are.

PR #2286, opened 2021-11-09
Make component examples an optional Elyra feature

ptitzler:
This PR
 - ~requires merge of https://github.com/elyra-ai/examples/pull/79~ **RESOLVED**
 - ~requires merge of https://github.com/elyra-ai/elyra/pull/2282~ **RESOLVED**
 - ~requires merge of https://github.com/elyra-ai/elyra/pull/2263~ **RESOLVED**
 - closes https://github.com/elyra-ai/elyra/pull/2272
 - closes https://github.com/elyra-ai/elyra/issues/2102
 - closes https://github.com/elyra-ai/elyra/issues/2164

**New user experience**

- after running `pip install elyra` (`make install`) in a clean environment
   - the component catalog list is empty
    - the KFP VPE palette contains  only generic components (in the `Elyra` category) 
    - the AA VPE palette contains  generic components (in the `Elyra` category) 
   - the user can manually install the example components at a later time, as outlined in  the documentation provided in https://github.com/elyra-ai/examples/pull/79
  
- after running `pip install elyra[all]` (`make install-all`) in a clean environment
    - the component catalog list is empty
    - the KFP VPE palette contains  only generic components (in the `Elyra` category) 
    - the AA VPE palette contains  generic components (in the `Elyra` category) 
    - the user can enable the Kubeflow Pipelines examples by creating a new "Kubeflow Pipelines example catalog" (via the GUI or CLI)
    - the user can enable the Apache Airflow examples by creating a new "Apache Airflow example catalog" (via the GUI or CLI)
    - the examples can be removed from the catalog list (and thus the VPE palette) or re-added after removal

- after launching the pre-built `elyra` container image
    - the component catalog lists two entries - one for KFP examples and one for AA examples
    - the KFP VPE palette contains  generic components (in the `Elyra` category) and the KFP examples (in the `examples` category)
    - the AA VPE palette contains  generic components (in the `Elyra` category) and the AA examples (in the `examples` category)
    - users can customize the container image `Dockerfile` as desired (with respect to making examples available or not)

- after launching the pre-built `kf-notebook` container image
    - the component catalog lists one entry for KFP examples 
    - the KFP VPE palette contains  generic components (in the `Elyra` category) and the KFP examples (in the `examples` category)
    - the examples can be removed from the catalog list (and thus the VPE palette) or re-added after removal
    - users can customize the container image `Dockerfile` as desired (with respect to making examples available or not)

### What changes were proposed in this pull request?

This PR
[...]

PR #2340, opened 2021-12-03
Re-add integration test for runtime specific components in the palette

### LINKED ISSUE #2298: Re-add integration test for discovery of runtime specific components in the palette
As of PR #2286, built-in components are no longer included by default for KFP and Airflow. Instead, they are provided as component catalog connectors (see [examples #79](https://github.com/elyra-ai/examples/pull/79)). As a result, the following integration tests have been commented out:

- `kfp pipeline should display custom components`
- `airflow pipeline should display custom components`

We will want to revisit including a similar test for this case when time allows.

kiersten-stokes:
Fixes #2298 

PR #2286 removed built-in components being included by default in the palette, and the corresponding integration tests (`kfp pipeline should display custom components` and `airflow pipeline should display custom components`) were commented out. This PR creates a function to add an example component catalog for each runtime type, which allows the custom components to be displayed in the palette again.

This update also necessitates the examples packages be installed, so the `install-examples` Makefile target has been added to the existing integration test targets (`test-integration` and `test-integration-debug`).

The component id's also have been updated to reflect the latest values.

Uncommented tests are now passing (locally and CI)

 
           in the file; or

           it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,6 +142,7 @@ jobs:
         run: |
           make yarn-install
           make only-install-server
+          make install-examples
       - name: Build
         run: make build-ui
       - name: Setup JupyterLab
--- a/Makefile
+++ b/Makefile
@@ -177,10 +177,10 @@ test-ui: lint-ui test-ui-unit test-integration # Run frontend tests
 test-ui-unit: # Run frontend jest unit tests
 	yarn test:unit
 
-test-integration: # Run frontend cypress integration tests
+test-integration: install-examples # Run frontend cypress integration tests
 	yarn test:integration
 
-test-integration-debug: # Open cypress integration test debugger
+test-integration-debug: install-examples # Open cypress integration test debugger
 	yarn test:integration:debug
 
 test: test-server test-ui ## Run all tests (backend, frontend and cypress integration tests)
--- a/Makefile
+++ b/Makefile
@@ -177,10 +177,10 @@ test-ui: lint-ui test-ui-unit test-integration # Run frontend tests
 test-ui-unit: # Run frontend jest unit tests
 	yarn test:unit
 
-test-integration: install-examples # Run frontend cypress integration tests
+test-integration: # Run frontend cypress integration tests
 	yarn test:integration
 
-test-integration-debug: install-examples # Open cypress integration test debugger
+test-integration-debug: # Open cypress integration test debugger
 	yarn test:integration:debug
 
 test: test-server test-ui ## Run all tests (backend, frontend and cypress integration tests)
--- a/tests/support/commands.ts
+++ b/tests/support/commands.ts
@@ -60,8 +60,19 @@ Cypress.Commands.add('createRuntimeConfig', ({ type } = {}): void => {
 });
 
 Cypress.Commands.add('createExampleComponentCatalog', ({ type } = {}): void => {
+  cy.on('fail', e => {
+    console.error(
+      `Example catalog connectors do not appear to be installed.\n${e}`
+    );
+    throw new Error(
+      `Example catalog connectors do not appear to be installed.\n${e}`
+    );
+  });
+
   cy.findByRole('tab', { name: /component catalogs/i }).click();
-  cy.findByRole('button', { name: /create new component catalog/i }).click();
+  cy.findByRole('button', { name: /create new component catalog/i })
+    .click()
+    .PinCustomMessage();
 
   if (type === 'kfp') {
     cy.findByRole('menuitem', { name: /kubeflow pipelines/i }).click();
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -177,10 +177,10 @@ test-ui: lint-ui test-ui-unit test-integration # Run frontend tests
 test-ui-unit: # Run frontend jest unit tests
 	yarn test:unit
 
-test-integration: # Run frontend cypress integration tests
+test-integration: install-examples # Run frontend cypress integration tests
 	yarn test:integration
 
-test-integration-debug: # Open cypress integration test debugger
```
kiersten-stokes, line 183: Feedback wanted here: not sure if this is the best solution, but it's a relatively easy fix

marthacryan: Not sure if I'm who you're looking for feedback from, but it seems fine to me? Maybe add a comment to explain why though

marthacryan (approved):
Code changes LGTM - haven't tried locally but looks like CI is working so should be good there. I would be interested to see what other frontend people think about the changes to the Makefile though (@ajbozarth @karlaspuldaro thoughts?)

akchinSTC (approved):

karlaspuldaro (commented):
Looks good @kiersten-stokes !
I don't mind the installation command in the makefile, but maybe keeping it inside pipeline tests file could make the debug mode a bit more efficient, that way examples wouldn't be installed when testing specific files other than pipeline. 
That alternative would be removing the makefile commands and adding to `pipeline.ts` something like
  ```
before(() => {
    cy.exec('make install-examples',
      {failOnNonZeroExit: false}
    );
  });
```

**Comment by ajbozarth** [comment]
I dont think we should be adding build targets to test targets in the Makefile. The test targets should assume that the build steps have already been run. If we need to update the CI build steps to handle this then we should do that.

I also am unsure if we should be referring to make calls from within the tests either like @karlaspuldaro suggested, those tests are also able to be run without the Makefile and should not be dependent on it.

**Comment by karlaspuldaro** [comment]
@ajbozarth good point, what about calling the commands directly instead of depending on make like
`cy.exec('pip install --upgrade elyra-examples-kfp-catalog')`

**Comment by ajbozarth** [comment]
@karlaspuldaro If we decided to include it in the tests then that would be how we would want to do it, but I'm still not sure we should. Don't our tests assume that the code is already build before they're run? If so, then they should assume that the examples are already installed and set up correctly. Though it is probably worth adding a check for it in the tests so they fail with a message as to why (ie that the examples aren't installed)

**Comment by kiersten-stokes** [comment]
@karlaspuldaro and @ajbozarth great points and I agree! The integration tests start with a `make install`, though, rather than a `make install-all`, so the example components are not included and the 2 tests in question fail (and are basically obsolete anyway at that point). I like the idea of doing the `make install-all` for the integration tests instead, but I don't know where that would have to be changed/updated.

**Comment by kiersten-stokes** [comment]
As discussed in the SCRUM, I reverted all changes to the Makefile and instead updated the CI workflow 'Install' command to include `make install-examples`.

I also added some custom error messaging to make it more obvious during troubleshooting that the error is likely with the missing example catalog connector packages.

File: .github/workflows/build.yml [hunk 2]
```diff
@@ -142,6 +142,7 @@ jobs:
         run: |
           make yarn-install
           make only-install-server
+          make install-examples
```
ajbozarth, line 145: in order to match our normal build order I would put this in the Setup JupyterLab section below

tests/support/commands.ts
```diff
@@ -59,6 +59,31 @@ Cypress.Commands.add('createRuntimeConfig', ({ type } = {}): void => {
...
+  cy.findByRole('tab', { name: /component catalogs/i }).click();
+  cy.findByRole('button', { name: /create new component catalog/i }).click();
+
+  if (type === 'kfp') {
+    cy.findByRole('menuitem', { name: /kubeflow pipelines/i }).click();
+  } else {
+    cy.findByRole('menuitem', { name: /apache airflow/i }).click();
+  }
+
+  cy.findByLabelText(/^name/i).type('Example Components');
```
ajbozarth, line 81: we should probably make this step check if one already exists first in case we end up using this in multiple tests in the future

kiersten-stokes: I'm currently removing the example catalog at the very end of the `afterEach` block for the pipeline tests (similar to how test runtime configs are handled). I see what you're saying, but I wonder if we should just leave this to be dealt with when the time comes? Idk, I could go either way

Regardless, I am going to change the name of the catalog to be something like "Test Example Components" to reduce the likelihood of collisions with existing catalogs when running locally.

ajbozarth: ah I see, I though that code was run once at the end, not per test. In that case ignore my comment, this way is good

PR #2346, opened 2021-12-06
Fix code snippet language check

### LINKED ISSUE #2018: Adding a code snippet into a script editor does not trigger language check
**Describe the issue**
Regression - Adding a code snippet into a script editor does not trigger language check. 
It still works as expected on notebooks.

After a quick investigation, looks like the widget class names are being mapped to something else during runtime, so [this statement](https://github.com/elyra-ai/elyra/blob/db8f0db786b580cfa2080bd788fc1df11dd8ad3e/packages/code-snippet/src/CodeSnippetWidget.tsx#L126) that checks for script editor is never true.

**To Reproduce**
1. Open a Python or R Editor
2. Open the code snippets tab
4. Click the insert button on the snippet item (Make sure to pick a code snippet of a language different from the open editor)
5. Notice the snippet is automatically added to the editor without displaying a warning dialog

**Expected behavior**
Language check should be triggered and display a warning when there is a language mismatch when inserting a code snippet into a script editor

**Context information**:
 - Elyra version: 3.0.0
 - Operating system: macos
 - Browser: Chrome

karlaspuldaro:
Fix code snippet language check when inserting into script editors using the `insert` button.
Fixes #2018 

When the package is published - eg. not installed from source - widget names are mapped to some kind of id. 
This is (probably) why when installing elyra/code-snippets extension from `pip` or `conda` the condition 
`if (widget.constructor.name === 'ScriptEditor')` fails, and language check is never triggered.Although it works as expected when installing from source code.

This PR removes the reference to ScriptEditor widget, which means the code-snippet extension will verify language match against any file editor which has kernel language support (such as Elyra Python/R editors).

**NOTE**: It won't change the behavior from source installation, proper fix would only be verified on a new publishing including these changes (or is there a way to mock that environment?)

New integration tests added.
Includes small refactor/cleanup of existing relevant tests.
 
           in the file; or

           it.

### DIFF
--- a/packages/code-snippet/src/CodeSnippetWidget.tsx
+++ b/packages/code-snippet/src/CodeSnippetWidget.tsx
@@ -107,7 +107,8 @@ class CodeSnippetDisplay extends MetadataDisplay<
   // Handle code snippet insertion into an editor
   private insertCodeSnippet = async (snippet: IMetadata): Promise<void> => {
     const widget = this.props.getCurrentWidget();
-    const snippetStr = snippet.metadata.code.join('\n');
+    const codeSnippet = snippet.metadata.code.join('\n');
+    const snippetLanguage = snippet.metadata.language;
 
     if (widget === null) {
       return;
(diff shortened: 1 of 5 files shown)


packages/code-snippet/src/CodeSnippetWidget.tsx
```diff
@@ -178,6 +179,15 @@ class CodeSnippetDisplay extends MetadataDisplay<
     return (widget as DocumentWidget).content instanceof FileEditor;
   };
 
+  // Return the language of the editor or empty string
+  private getEditorLanguage = (
+    widget: DocumentWidget<FileEditor>
+  ): string | undefined => {
+    const editorLanguage =
+      widget.context.sessionContext.kernelPreference.language;
+    return editorLanguage === 'null' ? '' : editorLanguage;
```
ajbozarth, line 188: is there a reason you're checking for the string containing null here now instead of checking for undefined?

karlaspuldaro: Yes. Plain text editors will have this field set to null as a string.

marthacryan (approved):
Tried locally and it seems to be working well! I didn't do too thorough of a code review but it looks good to me. Thanks!

ajbozarth (approved):
As mentioned in todays dev meeting, the code LGTM, but I have not had time to test it locally

PR #2349, opened 2021-12-07
Enable async component cache updates using SingletonConfigurable

### LINKED ISSUE #2162: Enable async cache updates for component registry
Currently the component cache is prompted to update 1. on cache expiry (5 minutes by default) and 2. any time a registry metadata instance is modified. The cache update in the former situation includes parsing of every component of every registry for that runtime type. As the number and size of the registries grow, this will become an issue.

~~Use a watchdog handler so that the cache updates don't penalize the user operation that triggers the need for the update.~~ After investigating, it looks like a Watchdog cannot be used because it would conflict with the existing watchdog for Metadata linked below.

An alternative approach involves decoupling the component parser from the runtime processor. This would mean creating a registry cache updater thread that waits for update requests to be dropped into its queue by the metadata post-update hooks. This cache updater would be a Singleton class, modifying the existing `ComponentCatalog` class.

The cache updater should also make it possible to only update the components in the component registry instance that is being modified. This necessitates changing the structure of the component cache dictionary from a dictionary of `Component` instances indexed by their ids to a include the runtime type at the top level, followed by catalog name, followed by the dictionaries of `Component` instances indexed by their ids one level down.

This avoids the creation of a MetadataManager instance every time we need to figure out the catalog instances for a given runtime. It also avoids the need to get a processor from the `PipelineProcessorRegistry` when catalog metadata instances are touched.

This could allow us to get rid of the cache TTL, as the cache should always be up to date.

**Additional context**
~~See [here](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/storage.py#L149) for an example of an existing watchdog handler.~~

### LINKED ISSUE #2278: Component files w/i subfolders of a directory-type catalog linger after catalog edit
**Describe the issue**
This is an corner case resulting from #2253 

Components defined within subdirectories of a directory-type catalog do not fall off the palette when the "Includes Subdirectories" checkbox is unchecked.

Note that this will be solved by the solution laid out for #2162 to enable async cache updates

**To Reproduce**
1. Create a local-directory-catalog type and assign the base directory to a folder that contains component definition files both in that directory and in a subdirectory
2. Select "Include Subdirectories" and save
2. Open a pipeline and see the available components
3. Edit the catalog you created and uncheck the "Include Subdirectories" checkbox, then save
4. The components in the subdirectory of the specified folder will still be present in the palette (until the cache refreshes 5 minutes after saving a catalog entry)

**Expected behavior**
The component within a subdirectory should drop off the palette on save.

**Deployment information**
Elyra latest master

kiersten-stokes:
Fixes #2162
Fixes #2278

The major change proposed here is making the `ComponentCatalog` class a `SingletonConfigurable`. This has 2 benefits:

1. decouples the processors from the component cache stored in `ComponentCatalog` and from the `ComponentParser` classes as well (increases start-up speed)
2. enables async component cache updates and removes the need for the cache TTL

All interfaces remain the same, so no updates to, e.g. the catalog connectors or APIs, are required. Further details below.

With the new singleton `ComponentCatalog`, initialization of the class instance is moved to the `PipelineProcessorManager`, similar to how the `PipelineProcessorRegistry` singleton class is initialized. This removes the need for `RuntimePipelineProcessor`s to perform their own runtime-type-specific catalog initialization in their init function. Subsequent component cache accesses are also much faster, simply requiring accessing the existing instance of `ComponentCatalog` and its methods hence (I say! 🧐 ).

Similarly, the `post_save` and `post_delete` hooks in `ComponentMetadata` no longer require processor access in order to call for an update of the component cache. Instead, the post-hooks call the catalog instance methods directly. 

On initialization of the `ComponentCatalog` instance, all existing catalogs are read and parsed, which is similar to how this is currently done. This PR adds the creation of a cache update manager thread that is responsible for the creation and monitoring of child threads -- one per cache update task that exists in a queue. This allows for reads to occur asynchronously and to update the cache as necessary when complete. No lock is required, as Python built-in types are already thread-safe. A `wait_for_all_cache_updates` method is also added to `ComponentCatalog`. The purpose of this is to allow tests that require complete updates to the component cache the ability to call to wait for the cache update thread to finish (i.e. `join()`) _before_ checking against assertions. 

The combination of all the above factors allows for the removal of the cache TTL.

The `Processor` classes are also decoupled from the `ComponentParser`s and allow for the removal of the `component_parser` property. Parsers are, however, still runtime-type-specific by necessity. Assignment is taken care of using the `ComponentParser` `create_instance` method, which is passed the runtime platform type given in the catalog Metadata instance (e.g. 'APACHE_AIRFLOW', 'KUBEFLOW_PIPELINES'). The runtime platform type argument matches the `component_platform` class variable name on the relevant `ComponentParser` subclass (`AirflowComponentParser` and `KfpComponentParser`) and returns that subclass as the assigned parser.

As proposed in #2162, the component_cache dictionary is restructured to be of the following general format:

```
{
  "APACHE_AIRFLOW": {    # runtime type
    "aa_examples": {     # catalog instance name
      "elyra-airflow-examples-catalog:3a55d015ea96": <Component object>,
      "elyra-airflow-examples-catalog:a043648d3897": <Component object>,
      ...
    },
    "my_airflow_components": {    # catalog instance name
      "elyra-airflow-examples-catalog:b94cd49692e2": <Component object>
    }, 
    ...
  },
  "KUBEFLOW_PIPELINES": {    # runtime type
    "kfp_examples": {     # catalog instance name
      "elyra-kfp-examples-catalog:61e6f4141f65": <Component object>,
      "elyra-kfp-examples-catalog:737915b826e9": <Component object>
    },
    ...
  },
  ... (eventually, additional runtime types)
}
```

This enables faster per-runtime type access (i.e., in `get_components()` and `get_component()`) and simpler per-catalog updates to the cache (e.g. the modification of a catalog only requires update of the given catalog's sub-dictionary rather than all catalogs for a runtime).

Finally, the `ComponentCatalog` class has been renamed to `ComponentCache`. There's been a lot of renaming recently, but I think this new name is more indicative of the purpose of the class (to maintain the component cache). 

Existing test cases have been updated and are passing locally and in the CI.

A new test resource has been added (`download_data.yaml`) in order to simplify the pipeline processor tests that construct a true pipeline. This is intended to replace the `filter_text.yaml` resource, which requires more complicated set of properties (those using `inputpath` and requiring upstream connections). I plan to remove `filter_text.yaml` and replace any uses with the new download test component in a future commit (or future PR since this will cause some test failures at first - see `test_processing_url_runtime_specific_component` in `test_processor_kfp`).

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component.py
+++ b/elyra/pipeline/component.py
@@ -15,9 +15,11 @@
 #
 from abc import abstractmethod
 from enum import Enum
+from importlib import import_module
 from logging import Logger
 from types import SimpleNamespace
 from typing import Any
+from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -279,11 +281,22 @@ class Component(object):
 class ComponentParser(LoggingConfigurable):  # ABC
     component_platform: RuntimeProcessorType = None
     _file_types: List[str] = None
+    _parser_class_map: Dict[str, str] = {
+        'APACHE_AIRFLOW': 'elyra.pipeline.airflow.component_parser_airflow:AirflowComponentParser',
+        'KUBEFLOW_PIPELINES': 'elyra.pipeline.kfp.component_parser_kfp:KfpComponentParser'
+    }
 
-    def __new__(cls, platform_type: str):
-        parser_class_map = {subclass.component_platform.name: subclass for subclass in cls.__subclasses__()}
-        parser_subclass = parser_class_map[platform_type]
-        return super(ComponentParser, parser_subclass).__new__(parser_subclass)
+    @classmethod
+    def create_instance(cls, platform: RuntimeProcessorType) -> 'ComponentParser':
+        """
+        Class method that creates the appropriate instance of ComponentParser based on platform type name.
+        """
+        try:
+            module_name, class_name = cls._parser_class_map[platform.name].split(':')
+            module = import_module(module_name)
+            return getattr(module, class_name)()
+        except Exception as e:
+            raise RuntimeError(f"Could not get appropriate ComponentParser class: {e}")
 
     @property
     def file_types(self) -> List[str]:
--- a/elyra/pipeline/component.py
+++ b/elyra/pipeline/component.py
@@ -277,12 +277,13 @@ class Component(object):
 
 
 class ComponentParser(LoggingConfigurable):  # ABC
-    _component_platform: RuntimeProcessorType = None
+    component_platform: RuntimeProcessorType = None
     _file_types: List[str] = None
 
-    @property
-    def component_platform(self) -> RuntimeProcessorType:
-        return self._component_platform
+    def __new__(cls, platform_type: str):
+        parser_class_map = {subclass.component_platform.name: subclass for subclass in cls.__subclasses__()}
+        parser_subclass = parser_class_map[platform_type]
+        return super(ComponentParser, parser_subclass).__new__(parser_subclass)
 
     @property
     def file_types(self) -> List[str]:
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -95,8 +95,8 @@ class CacheUpdateManagerThread(Thread):
                 duration = int(time.time() - thread.last_warn_time)
                 if duration > CATALOG_UPDATE_TIMEOUT:
                     thread.last_warn_time = time.time()
-                    self.log.warning(f"Cache update for catalog '{thread.name}' is still "
-                                     f"processing after {duration} seconds ...")
+                    self.log.warning(f"Cache update for catalog '{thread.name}' is still processing after "
+                                     f"{int(thread.last_warn_time - thread.task_start_time)} seconds ...")
 
             else:
                 # Thread has been joined and can be removed from the list
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -86,15 +86,17 @@ class CacheUpdateManagerThread(Thread):
         """
         outstanding_threads = False
         for thread in self._threads:
-            try:
-                # Attempt to join thread within the given amount of time
-                thread.join(timeout=NONBLOCKING_TIMEOUT)
+            # Attempt to join thread within the given amount of time
+            thread.join(timeout=NONBLOCKING_TIMEOUT)
 
-            except Exception:
+            if thread.is_alive():
                 # Thread is still running (thread join timed out)
                 outstanding_threads = True
-                if time.time() - thread.task_start_time > CATALOG_UPDATE_TIMEOUT:
-                    self.log.warning(f"Thread {thread.name} is taking too long...")  # TODO Fix this message
+                duration = int(time.time() - thread.last_warn_time)
+                if duration > CATALOG_UPDATE_TIMEOUT:
+                    thread.last_warn_time = time.time()
+                    self.log.warning(f"Cache update for catalog '{thread.name}' is still "
+                                     f"processing after {duration} seconds ...")
 
             else:
                 # Thread has been joined and can be removed from the list
@@ -119,6 +121,7 @@ class CacheUpdateThread(Thread):
         super().__init__()
 
         self.setDaemon(True)
+        self.name = catalog.name
 
         self._component_cache = component_cache
         self._queue = queue
@@ -129,6 +132,7 @@ class CacheUpdateThread(Thread):
 
         # Thread metadata
         self.task_start_time = time.time()
+        self.last_warn_time = self.task_start_time
 
     def run(self):
         # Add sub-dictionary for this platform type if not present
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -89,7 +89,7 @@ class CacheUpdateManagerThread(Thread):
             # Attempt to join thread within the given amount of time
             thread.join(timeout=NONBLOCKING_TIMEOUT)
 
-            cumulative_run_time = int(thread.last_warn_time - thread.task_start_time)
+            cumulative_run_time = int(time.time() - thread.task_start_time)
             if thread.is_alive():
                 # Thread is still running (thread join timed out)
                 outstanding_threads = True
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -72,6 +72,11 @@ class ComponentCatalog(SingletonConfigurable):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+        # Indicate whether this is a test environment. This is relevant in
+        # update_cache_for_catalog() - we want to wait for the cache updater
+        # thread to finish in test envs
+        self._for_test = bool(kwargs.get('for_test'))
+
         if not self._component_cache:
             self._build_cache()
 
@@ -88,27 +93,29 @@ class ComponentCatalog(SingletonConfigurable):
         """
         Updates the component cache for the given catalog Metadata instance using a thread.
         """
+        platform_type = catalog.metadata['runtime_type']
+
+        # Add sub-dictionary for this platform type if not present
+        if not self._component_cache.get(platform_type):
+            self._component_cache[platform_type] = {}
 
         def update_cache_with_thread():
-            platform_type = catalog.metadata['runtime_type']
-
-            # Add sub-dictionary for this platform type if not present
-            if not self._component_cache.get(platform_type):
-                self._component_cache[platform_type] = {}
-            catalog_components = self._read_component_catalog(catalog, platform_type)
-
+            # TODO add protections against race conditions
+            catalog_components = self._read_component_catalog(catalog)
             if operation == 'delete':
                 # Remove only the components from this catalog
-                self._component_cache[platform_type].pop(catalog.name)
+                self._component_cache[platform_type].pop(catalog.name, None)
             else:
                 # Replace all components for the given catalog
                 self._component_cache[platform_type][catalog.name] = catalog_components
 
-        # TODO should anything else use a thread?
+        # Start thread to perform the cache update
         updater_thread = Thread(target=update_cache_with_thread)
-
-        updater_thread.setDaemon(True)  # daemon thread will not block program exit
         updater_thread.start()
+
+        # Wait for tests to finish if test environment was indicated
+        if self._for_test:
+            updater_thread.join()
 
     def get_all_components(self, platform_type: str) -> List[Component]:
         """
@@ -116,7 +123,7 @@ class ComponentCatalog(SingletonConfigurable):
         """
         components: List[Component] = []
 
-        platform_components_dict = ComponentCatalog._component_cache.get(platform_type, {})
+        platform_components_dict = self._component_cache.get(platform_type, {})
         for catalog_name, component_dict in platform_components_dict.items():
             components.extend(list(component_dict.values()))
 
@@ -131,7 +138,7 @@ class ComponentCatalog(SingletonConfigurable):
         """
         component: Optional[Component] = None
 
-        platform_components_dict = ComponentCatalog._component_cache.get(platform_type, {})
+        platform_components_dict = self._component_cache.get(platform_type, {})
         for catalog_name, component_dict in platform_components_dict.items():
             component = component_dict.get(component_id)
             if component:
@@ -142,7 +149,7 @@ class ComponentCatalog(SingletonConfigurable):
 
         return component
 
-    def _read_component_catalog(self, catalog: Metadata, platform_type) -> Dict[str, Component]:
+    def _read_component_catalog(self, catalog: Metadata) -> Dict[str, Component]:
         """
         Read a component catalog and return a dictionary of components indexed by component_id.
 
@@ -154,6 +161,7 @@ class ComponentCatalog(SingletonConfigurable):
 
         # Assign reader based on the type of the catalog (the 'schema_name')
         try:
+            platform_type = catalog.metadata['runtime_type']
             parser = ComponentParser(platform_type=platform_type)
 
             catalog_reader = entrypoints.get_group_named('elyra.component.catalog_types').get(catalog.schema_name)
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -27,17 +27,22 @@ from jinja2 import Template
 from traitlets.config import SingletonConfigurable
 
 from elyra.metadata.manager import MetadataManager
-from elyra.metadata.metadata import Metadata
 from elyra.metadata.schemaspaces import ComponentCatalogs
 from elyra.pipeline.airflow.component_parser_airflow import AirflowComponentParser  # noqa F401 - TODO
 from elyra.pipeline.catalog_connector import ComponentCatalogConnector
 from elyra.pipeline.component import Component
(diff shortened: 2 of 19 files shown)


elyra/pipeline/component.py
```diff
@@ -277,12 +277,13 @@ def _log_warning(msg: str, logger: Optional[Logger] = None):
...
+    component_platform: RuntimeProcessorType = None
     _file_types: List[str] = None
 
-    @property
-    def component_platform(self) -> RuntimeProcessorType:
-        return self._component_platform
+    def __new__(cls, platform_type: str):
+        parser_class_map = {subclass.component_platform.name: subclass for subclass in cls.__subclasses__()}
+        parser_subclass = parser_class_map[platform_type]
+        return super(ComponentParser, parser_subclass).__new__(parser_subclass)
```
kiersten-stokes, line 287: This probably warrants a comment

elyra/pipeline/processor.py
```diff
@@ -116,6 +103,7 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.root_dir = get_expanded_path(kwargs.get('root_dir'))
         self._registry = PipelineProcessorRegistry.instance()
+        self._component_catalog = ComponentCatalog.instance(parent=self.parent)
```
kiersten-stokes, line 106: We don't really need the variable here since it isn't currently accessed anywhere else in the class - I just followed the pattern that the `PipelineProcessorRegistry` took

elyra/pipeline/component.py
kevin-bates: Although this may be the "python" way (I really don't know), I would recommend a class factory pattern where you have a `get_instance()` `@classmethod`.  The content of this method would be similar, although I think the creation of the `parser_class_map` should occur at the class level (and be treated as a class variable).  That way it's only initialized once.

The code in `_get_component_parser_class`  would then be `ComponentParser.get_instance(platform)` - which is more clear that its acting as a class factory rather than just constructing a base `ComponentParser` instance.  (And you probably don't need `_get_component_parser_class`, since `get_instance` could perform the error handling.  Hmm, I think it might be better to raise an error rather than return `None` since the caller would always need to check for `None` and this seems fairly critical - but I'm also a bit ignorant to the bigger picture right now.)

elyra/pipeline/component_catalog.py
```diff
@@ -72,90 +70,165 @@ class ComponentCatalog(LoggingConfigurable):
...
 
-    def _is_cache_expired(self) -> bool:
-        is_expired = True
-        if self._cache_last_updated:
-            now = time.time()
-            elapsed = int(now - self._cache_last_updated)
-            if elapsed < self.cache_ttl:
-                is_expired = False
+        # Assign component parser based on the runtime platform type
+        parser = self._get_component_parser_class(platform_type=catalog.metadata['runtime_type'])
```
kevin-bates, line 202: If we're going to use `platform_type` as the full type, we should extend `ComponentCatalogMetadata` with a `runtime_type` property - something like:
```python
@property
def runtime_type() -> RuntimeProcessorType:
    return RuntimeProcessorType.get_instance_by_name(self.metadata['runtime_type'])
```

elyra/pipeline/kfp/component_parser_kfp.py
```diff
@@ -28,9 +28,10 @@
 
 
 class KfpComponentParser(ComponentParser):
```
kevin-bates, line 30: Because parsers are _contributed_ and not _byo_'d, we should move the KFP and AA parsers in the parent directory.

elyra/pipeline/component_catalog.py
```diff
@@ -25,25 +24,24 @@
...
 from jinja2 import Template
-from traitlets import default
-from traitlets import Integer
-from traitlets.config import LoggingConfigurable  # noqa: H306 (alphabetical order catch-22)
+from traitlets.config import SingletonConfigurable
 
 from elyra.metadata.manager import MetadataManager
 from elyra.metadata.metadata import Metadata
 from elyra.metadata.schemaspaces import ComponentCatalogs
+from elyra.pipeline.airflow.component_parser_airflow import AirflowComponentParser  # noqa F401 - TODO
```
kevin-bates, line 32: Since you're going thru `__subclasses__`, I'm surprised you need these import statements.  Do things break at runtime otherwise?  (I suppose they do since the "get_instance" method needs to create their instance.)

Since the `parse_class_map` is essentially static and (will be) initialized once, it might be best to just reference the subclasses directly when building the map and add that map's update to the list of things folks need to do when _contributing_ a new runtime type (along with a complete component parser :smile: ).

kevin-bates (commented):
Hi @kiersten-stokes - I've completed an initial pass.  This looks really good.  I would also like to thank you for writing such a helpful description!!

Some comments were a tad global in scope (e.g, use of the actual runtime-type instance, avoidance of types in variable names, and update of `Metadata` to `ComponentCatalogMetadata`) so I didn't bother to call them out everywhere but may have included a reminder or two.

Regarding the thread join issue, is there a race condition in non-test scenarios or will the caller just see the older instance until the cache has been updated?  If the latter, we might want to push things "down a thread" by perhaps having a long-running "manager type" thread hand-off work for the catalog reader threads, where the manager thread could perform (timed-out) join operations and perhaps report on long-running sub-threads.  This way, we'd be able to determine rogue/pathological behaviors and report (log) the conditions when they've exceeded some threshold.

elyra/pipeline/kfp/component_parser_kfp.py
kiersten-stokes: After discussion with @kevin-bates, we're going to leave these files where they are for now

elyra/pipeline/component_catalog.py
```diff
@@ -72,90 +70,165 @@ class ComponentCatalog(LoggingConfigurable):
...
-                self.update_cache()
-            return list(self._cached_components.values())
 
-        return list(self._read_component_catalogs().values())
+        all_catalogs = MetadataManager(schemaspace=ComponentCatalogs.COMPONENT_CATALOGS_SCHEMASPACE_ID).get_all()
+        for catalog in all_catalogs:
+            self.update_cache_for_catalog(catalog)
 
-    def get_component(self, component_id: str) -> Optional[Component]:
+    def update_cache_for_catalog(self, catalog: Metadata, operation: Optional[str] = None):
```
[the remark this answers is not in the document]
kiersten-stokes: Hmm I ran into 2 issues when trying to change this

1. since the call to `MetadataManager(...).get_all()` in line 91 returns `Metadata`, we get typehint errors when calling `update_cache_for_catalog` -- "Expected type 'ComponentCatalogMetadata', got 'Metadata' instead" 
2. there's a circular import reference (importing `ComponentCatalogMetadata` in `component_catalog` and importing `ComponentCatalog` in `component_metadata`) that we'd have to work around

Thoughts on the best way forward?

elyra/pipeline/component_catalog.py
```diff
@@ -72,90 +70,165 @@ class ComponentCatalog(LoggingConfigurable):
...
-        else:
-            component = self._read_component_catalogs().get(component_id)
+        platform_type = catalog.metadata['runtime_type']
+
+        # Add sub-dictionary for this platform type if not present
+        if not self._component_cache.get(platform_type):
+            self._component_cache[platform_type] = {}
+
+        def update_cache_with_thread():
+            catalog_components = self._read_component_catalog(catalog)
```
kiersten-stokes, line 105: This will change when the threading aspect is re-worked, but note that we shouldn't bother to read the catalog if it's a 'delete' operation

elyra/pipeline/component_catalog.py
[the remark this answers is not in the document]
kevin-bates: Regarding 1, this would be a bug.  I can see a raw `Metadata` instance returned in error situations, but the `from_dict()` method should leverage any existing `metadata_class_name` value.  Are you able to step through `get_all()` to see why these aren't `ComponentCatalogMetadata` instances?  Since the cache update is obviously getting triggered, we know the schema is pointing at the right class name.

For 2, these kinds of things can be difficult.  Since `ComponentCatalogMetadata` is just a metadata representation of a `ComponentCatalog`, one could argue that its definition also be included adjacent to `ComponentCatalog`'s definition (i.e., in the same file).  This may be worth a try, although I may not resolve the circular reference.  There are other approaches we could take (I believe), but let's try this first since it makes sense from a definition-location standpoint.

elyra/pipeline/component_catalog.py
```diff
@@ -72,90 +70,165 @@ class ComponentCatalog(LoggingConfigurable):
...
+                    # Replace all components for the given catalog
+                    self._component_cache[platform_type][catalog.name] = catalog_components
+
+        # Start thread to perform the cache update
+        updater_thread = threading.Thread(target=update_cache_with_thread)
+        updater_thread.start()
+
+        # Wait for tests to finish if test environment was indicated
+        if self._for_test:
+            updater_thread.join()
```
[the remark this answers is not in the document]
kiersten-stokes: As decided in a mini-design discussion with @kevin-bates, the manager thread will take care of running periodic timeout joins

elyra/pipeline/component_catalog.py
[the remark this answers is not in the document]
kiersten-stokes: In order to avoid the need to update the example catalog schema files, we'll have to address the circular dependency issue by importing the entire `component_catalog` module in `component_metadata.py` and calling the `ComponentCache` instance using `component_catalog.ComponentCache.instance()`

elyra/pipeline/component.py
```diff
@@ -277,12 +279,24 @@ def _log_warning(msg: str, logger: Optional[Logger] = None):
...
+    def create_instance(cls, platform: RuntimeProcessorType) -> 'ComponentParser':
+        """
+        Class method that creates the appropriate instance of ComponentParser based on platform type name.
+        """
+        try:
+            module_name, class_name = cls._parser_class_map[platform.name].split(':')
+            module = import_module(module_name)
+            return getattr(module, class_name)()
+        except Exception as e:
+            raise RuntimeError(f"Could not get appropriate ComponentParser class: {e}")
```
kiersten-stokes, line 299: This is currently mimicking the way that entrypoints loads their class instances. An alternative would be something like the below (including removal of the class-level `_parser_class_map` structure).

```
@classmethod
def create_instance(cls, platform: RuntimeProcessorType) -> 'ComponentParser':
    parser_class_map = {subclass.component_platform.name: subclass for subclass in cls.__subclasses__()}
    return parser_class_map[platform]
```

elyra/pipeline/component.py
kiersten-stokes: Updated in the latest set of commits!

elyra/pipeline/processor.py
kiersten-stokes: Removed in the latest set of commits

elyra/pipeline/component_catalog.py
```diff
@@ -25,25 +29,129 @@
...
+        for thread in self._threads:
+            try:
+                # Attempt to join thread within the given amount of time
+                thread.join(timeout=NONBLOCKING_TIMEOUT)
+
+            except Exception:
+                # Thread is still running (thread join timed out)
+                outstanding_threads = True
+                if time.time() - thread.task_start_time > CATALOG_UPDATE_TIMEOUT:
+                    self.log.warning(f"Thread {thread.name} is taking too long...")  # TODO Fix this message
```
kiersten-stokes, line 97: Fix this TODO (suggestions welcome here)

kevin-bates: I would suggest setting the thread's name to the catalog associated with the update by adding `name=catalog.name` when initializing the superclass, then use a message something like...
```python
self.log.warning(f"Cache update for catalog '{thread.name}' is still processing after {duration} seconds ...")
```
where `durantion = int(time.time() - thread.task_start_time)` converted to seconds.  

More importantly though, once the `CATALOG_UPDATE_TIMEOUT` has been initially exceeded, this warning will spam the log at sub-seconds intervals, so this should be prevented.  Ideally, it would be best to warn only at `CATALOG_UPDATE_TIMEOUT` intervals, so one approach would be to add a `last_warn_time` field that is initialized to `task_start_time` when the thread is created, then use that field to determine the warning and reset that field on each logged warning.  

Since catalog readers tend to just execute on a finite set of items, I'm not sure we need to plumb a "give up" mechanism - or can probably hold off on that.  If we did need that, we'd probably also need a counter to track the times we log a warning - which is roughly every `CATALOG_UPDATE_TIMEOUT` seconds.  We'd then pick a value at which _that_ multiple of `CATALOG_UPDATE_TIMEOUT` is the "give up" point.  For example, since `CATALOG_UPDATE_TIMEOUT` is 15 seconds, a count of 20 would be roughly 5 minutes of waiting for that thread to complete - but I don't think we need this right now.

kiersten-stokes: Great point and idea! As I was going through to test these updates after making them, I realized that `thread.join` doesn't actually throw an exception, so that has also been fixed in the latest commit and is now working

akchinSTC (approved):
LG!

elyra/pipeline/component_catalog.py
```diff
@@ -25,25 +29,133 @@
...
+            thread.join(timeout=NONBLOCKING_TIMEOUT)
+
+            if thread.is_alive():
+                # Thread is still running (thread join timed out)
+                outstanding_threads = True
+                duration = int(time.time() - thread.last_warn_time)
+                if duration > CATALOG_UPDATE_TIMEOUT:
+                    thread.last_warn_time = time.time()
+                    self.log.warning(f"Cache update for catalog '{thread.name}' is still "
+                                     f"processing after {duration} seconds ...")
```
kevin-bates, line 99: I think the duration here should be cumulative...
```suggestion
                                     f"processing after {int(thread.last_warn_time - thread.task_start_time)} seconds ...")
```

kiersten-stokes: Good catch!

elyra/pipeline/component_catalog.py
```diff
@@ -25,25 +29,133 @@
...
+
+            else:
+                # Thread has been joined and can be removed from the list
```
kevin-bates, line 102: I think it would be a good idea to also log an informational message that indicates the completion of the cache update (including its overall duration), particularly if a warning message was logged.   As a result, the cumulative time could be computed unconditionally (immediately following the `join`).

kiersten-stokes: True! I added an `info` log message for threads that have been previously warned

elyra/pipeline/component_catalog.py
```diff
@@ -89,19 +89,28 @@ def check_outstanding_threads(self) -> bool:
             # Attempt to join thread within the given amount of time
             thread.join(timeout=NONBLOCKING_TIMEOUT)
 
+            cumulative_run_time = int(thread.last_warn_time - thread.task_start_time)
```
kevin-bates, line 92: Sorry, but I think this needs to be: `int(time.time() - thread.task_start_time)` otherwise it could be close to `CATALOG_UPDATE_TIMEOUT` seconds short (worst case).

kiersten-stokes: Ugh yes, of course! I'm having a true Monday

kevin-bates (approved):
LGTM - thank you @kiersten-stokes.

PR #2350, opened 2021-12-07
Allow Kubeflow node inputValues to consume outputPath outputs

### LINKED ISSUE #2197: KFP: Support data exchange for component inputs passed as `inputValue`

In version 3.2 users can use the output of a (custom component) node as input for another (custom component node) if, and only if, the input is declared as `inputPath` in the specification but not as `inputValue`. 

If a component input is declared as `inputValue` the user is given the option in the pipeline editor to
- manually enter a value (e.g. using a text box)
- select the output of an upstream node

akchinSTC:
Fixes #2197 
Requires https://github.com/elyra-ai/pipeline-editor/pull/175

- [x] Add additional tests for KFP meta typing 
- [x] Migrate existing test resource pipelines to new version

Updating KFP pipeline construction and behavior. Prior to this change any component parameter deemed an inputvalue could only accept either default values provided in the component spec/definition or a value provided by the user. This change adds a third option where it is able to consume a outputpath value provided by a parent component in the pipeline 

Existing unit tests were modified to check the new component property structure.
Additional kfp operator and checks were added for kfp component meta typing

 
           in the file; or

           it.


elyra/pipeline/kfp/processor_kfp.py
```diff
@@ -629,12 +629,22 @@ def _cc_pipeline(self,
...
+                            operation.component_params[component_property.ref] = \
+                                target_ops[output_node_id].outputs[output_node_parameter_key]
+                        elif component_property.default_data_type == 'dictionary':
+                            processed_value = self._process_dictionary_value(active_property_value)
+                            operation.component_params[component_property.ref] = processed_value
+                        elif component_property.default_data_type == 'list':
+                            processed_value = self._process_list_value(active_property_value)
+                            operation.component_params[component_property.ref] = processed_value
+                        else:
+                            operation.component_params[component_property.ref] = active_property_value
```
kevin-bates, line 647: Are each of these statements over-indented or does only the last `else` statement need to be duplicated for the outer `if`-block?  If neither, then perhaps the `elif` on L632 should be an `else`.

akchinSTC: I should add an outer if block `else` to clarify that it would catch the last case which would be `outputPath` parameters. We don't do anything with these parameters right now. 
The statements should all be indented properly in that they should only apply to any field designated an `inputvalue`

kiersten-stokes (approved):
Looking good! Tested several pipelines and all are working as expected

ptitzler:
In this scenario (upstream node has no output) it should not be possible to select the parent as source:

image

Component sources: 
- https://raw.githubusercontent.com/ptitzler/examples/how-to-byo-cr/component-catalog-connectors/connector-template/todo_catalog_connector/dummy-component.yaml
- https://raw.githubusercontent.com/elyra-ai/examples/master/component-catalog-connectors/kfp-example-components-connector/kfp_examples_connector/resources/download_data.yaml

ptitzler:
A pipeline configured as follows 

image

triggers this error:

```
  File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/elyra/pipeline/handlers.py", line 112, in post
  File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 172, in validate
    response=response)
  File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 334, in _validate_node_properties
    pipeline_definition=pipeline_definition)
  File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 442, in _validate_custom_component_node_properties
    node_param_value = node_param['NestedEnumControl'].get('value')
KeyError: 'NestedEnumControl'
```

ptitzler (changes requested):
Some manual tests failed. See comments for problematic test scenarios.

akchinSTC:
>   File "/opt/anaconda3/envs/akchin-value/lib/python3.7/site-packages/elyra/pipeline/validation.py", line 442, in _validate_custom_component_node_properties
>     node_param_value = node_param['NestedEnumControl'].get('value')
> KeyError: 'NestedEnumControl'
> ```
> 
> Edit: it's sufficient to create a one node pipeline and not select an input source to trigger this.

Pushed a fix for this. 

Looking into the parsing no output bug. That should be handled by the parsing and UI i think

akchinSTC:
@ptitzler - rebased and both scenarios should be resolved now.

elyra/pipeline/component.py
```diff
@@ -136,6 +140,18 @@ def default_control_type(self) -> str:
...
+    def default_data_type(self) -> str:
+        """
+            The `default_data_type` is the first data type that is assigned to this specific parameter
+            after parsing the component specification.
+        """
+        return self._default_data_type
+
+    @property
+    def allow_no_options(self) -> str:
+        return "true" if self._allow_no_options else "false"
```
kevin-bates, line 153: Having a member variable property masquerade as a different data type just doesn't strike me as a good practice.

Can the boolean value be coerced to the appropriate string within the template - since that's the only time it's needed as a string?  If not, then I would recommend renaming the property method to something like `allow_no_options_to_jinja`, or pass a boolean to the property method indicating you want the lowercase string form and make the return type of the `allow_no_options` property a `Union[bool, str]` with the default behavior that it returns its native data type.

akchinSTC: :nods in agreement: 
Updated!

ptitzler (approved):
LGTM. Confirmed that the issues I've reported earlier are no longer reproducible. Thanks Alan!

PR #2354, opened 2021-12-09
Add explicit support for COS authentication types to runtime configurations

ptitzler:
This PR is a follow up to https://github.com/elyra-ai/elyra/pull/2335, which introduced implicit support for COS authentication types in runtime configurations.

With this change the runtime configuration requires a new authentication type property:

image

The property takes the following values:
 - `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS` (applicable to AWS only)
 - `KUBERNETES_SECRET`
 - `USER_CREDENTIALS`

Behavior:
- For new runtime configurations the default selection is `USER_CREDENTIALS` (the most likely choice users are making today).
- For existing runtime configurations the following heuristic is used to determine which type to set during on-the-fly migration, which is performed when configurations are loaded:
   - if user id and password are set, use `USER_CREDENTIALS` type
   - if user id,  password, and secret are set, use `KUBERNETES_SECRET` type
- Notes:
   -  Migration only covers scenarios for runtime configurations that were created using 3.3.0 or earlier, but not the master branch that includes https://github.com/elyra-ai/elyra/pull/2335 but not this PR
   - Custom code enforces the following conditional schema constraints:
      - `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS` -> no values must be set for the following properties: `cos_secret`, `cos_username`, and `cos_password` 
      - `KUBERNETES_SECRET` -> values must be present for the following properties: `cos_secret`, `cos_username`, and `cos_password` 
      - `USER_CREDENTIALS` -> values must be present in the following properties: `cos_username`, `cos_password`, but not `cos_secret`

 - Add new required `cos_auth_type` property to Kubeflow Pipelines schema
 - Re-organize order of COS properties in the Kubeflow Pipelines schema
 - Add on-load migration code to Kubeflow Pipelines metadata class that adds `cos_auth_type` to existing runtime configurations
 - Add new required `cos_auth_type` property to Airflow schema
 - Re-organize order of COS properties in the Kubeflow Pipelines schema
 - Add new Airflow metadata class that contains on-load migration code that adds `cos_auth_type` to existing runtime configurations
 - Improve COS connectivity error handling and messages. (Identify what kind of connection attempt is failing, why it is failing, and add reference to problematic RTC) Example:
   ```
   Error connecting to object storage: SignatureDoesNotMatch. Verify the password in runtime configuration 'aa with secret' and try again.
   ```
 - Update _Runtime Configuration_ topic in the User Guide

Reviewed the output of `make docs`

Tested default COS auth type assignments:
- [x] New KFP RTC
- [x] New Airflow RTC
- [x] Migrated KFP RTC with kubernetes secret
- [x] Migrated KFP RTC without kubernetes secret
- [x] Migrated Airflow RTC with kubernetes secret
- [x] Migrated Airflow RTC without kubernetes secret

Tested connectivity and error handling for KFP:
- [x] USER_CREDENTIALS: constraints fulfilled
- [x] USER_CREDENTIALS: invalid credentials (username or password)
- [x] USER_CREDENTIALS: constraints violated
- [x] KUBERNETES_SECRET: constraints fulfilled
- [x] KUBERNETES_SECRET: invalid credentials (username or password)
- [x] KUBERNETES_SECRET: constraints violated
- [x] AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS: constraints fulfilled
- [x] AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS: constraints violated

Tested connectivity and error handling for Airflow:
- [x] USER_CREDENTIALS: constraints fulfilled
- [x] USER_CREDENTIALS: invalid credentials (username or password)
- [x] USER_CREDENTIALS: constraints violated
- [x] KUBERNETES_SECRET: constraints fulfilled
- [x] KUBERNETES_SECRET: invalid credentials (username or password)
- [x] KUBERNETES_SECRET: constraints violated
- [x] AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS: constraints fulfilled
- [x] AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS: constraints violated

For testing set environment variables as follows prior to launching JupyterLab:

- `export ELYRA_GITHUB_ORG=ptitzler`
- `export ELYRA_GITHUB_BRANCH=cos-auth`

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/kfp/kfp_metadata.py
+++ b/elyra/pipeline/kfp/kfp_metadata.py
@@ -54,3 +54,37 @@ class KfpMetadata(RuntimesMetadata):
             MetadataManager(schemaspace="runtimes").update(self.name, self, for_migration=True)
 
         return None
+
+    def pre_save(self, **kwargs: Any) -> None:
+        """
+        This method enforces conditional constraints related to
+        COS authentication type properties.
+        TODO: Remove after https://github.com/elyra-ai/elyra/issues/2338
+        was resolved.
+        """
+        super().pre_save(**kwargs)
+
+        if self.metadata.get('cos_auth_type') is None:
+            # nothing to do
+            return
+
+        if self.metadata['cos_auth_type'] == 'USER_CREDENTIALS':
+            if len(self.metadata.get('cos_username', '').strip()) == 0 or\
+               len(self.metadata.get('cos_password', '').strip()) == 0:
+                raise ValueError('A username and password are required '
+                                 'for the selected Object Storage authentication type.')
+            if len(self.metadata.get('cos_secret', '').strip()) > 0:
+                raise ValueError('Kubernetes secrets are not supported '
+                                 'for the selected Object Storage authentication type.')
+        elif self.metadata['cos_auth_type'] == 'KUBERNETES_SECRET':
+            if len(self.metadata.get('cos_username', '').strip()) == 0 or\
+               len(self.metadata.get('cos_password', '').strip()) == 0 or\
+               len(self.metadata.get('cos_secret', '').strip()) == 0:
+                raise ValueError('Username, password, and Kubernetes secret are required '
+                                 'for the selected Object Storage authentication type.')
+        elif self.metadata['cos_auth_type'] == 'AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS':
+            if len(self.metadata.get('cos_username', '').strip()) > 0 or\
+               len(self.metadata.get('cos_password', '').strip()) > 0 or\
+               len(self.metadata.get('cos_secret', '').strip()) > 0:
+                raise ValueError('Username, password, and Kubernetes secret are not supported '
+                                 'for the selected Object Storage authentication type.')
--- a/elyra/util/cos.py
+++ b/elyra/util/cos.py
@@ -32,12 +32,23 @@ class CosClient(LoggingConfigurable):
         cred_provider = None
         if config is None:
             # The client was invoked by an entity that does not utilize
-            # runtime configurations. Use the explicitly provided parameter
-            # values instead.
-            cred_provider = providers.StaticProvider(
-                access_key=access_key,
-                secret_key=secret_key,
-            )
+            # runtime configurations.
+            if access_key is None or secret_key is None:
+                # use env variables for authentication
+                if len(os.environ.get('AWS_ACCESS_KEY_ID', '').strip()) == 0 or\
+                   len(os.environ.get('AWS_SECRET_ACCESS_KEY', '').strip()) == 0:
+                    raise RuntimeError('Cannot connect to object storage. No credentials '
+                                       ' were provided and environment variables '
+                                       ' AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not '
+                                       ' properly defined.')
+                else:
+                    cred_provider = providers.EnvAWSProvider()
+            else:
+                # use provided username and password for authentication
+                cred_provider = providers.StaticProvider(
+                    access_key=access_key,
+                    secret_key=secret_key,
+                )
             self.endpoint = endpoint
             self.bucket = bucket
         else:
@@ -49,13 +60,6 @@ class CosClient(LoggingConfigurable):
                     access_key=config.metadata['cos_username'],
                     secret_key=config.metadata['cos_password'],
                 )
-            elif auth_type == 'AWS_ENV_VARIABLES':
-                if os.environ.get('AWS_ACCESS_KEY_ID') is None or\
-                   os.environ.get('AWS_SECRET_ACCESS_KEY') is None:
-                    raise RuntimeError('Cannot connect to object storage. '
-                                       f'Authentication provider \'{auth_type}\' requires '
-                                       'environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.')
-                cred_provider = providers.EnvAWSProvider()
             elif auth_type == 'AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS':
                 if os.environ.get('AWS_ROLE_ARN') is None or\
                    os.environ.get('AWS_WEB_IDENTITY_TOKEN_FILE') is None:
--- a/elyra/util/cos.py
+++ b/elyra/util/cos.py
@@ -45,26 +45,6 @@ class CosClient(LoggingConfigurable):
             self.endpoint = urlparse(config.metadata['cos_endpoint'])
             self.bucket = config.metadata['cos_bucket']
             if auth_type in ['USER_CREDENTIALS', 'KUBERNETES_SECRET']:
-                # TODO remove temporary validation checks after
-                # metadata can properly enforce constraints
-                if len(config.metadata.get('cos_username', '').strip()) == 0 or\
-                   len(config.metadata.get('cos_password', '').strip()) == 0:
-                    raise RuntimeError('Cannot connect to object storage. '
-                                       f'Authentication provider \'{auth_type}\' requires '
-                                       'a username and password. Update runtime configuration '
-                                       f'\'{config.display_name}\' and try again.')
-                if auth_type == 'USER_CREDENTIALS' and\
-                   len(config.metadata.get('cos_secret', '').strip()) > 0:
-                    raise RuntimeError('Cannot connect to object storage. '
-                                       f'Authentication provider \'{auth_type}\' does not '
-                                       'support Kubernetes secrets. Update runtime configuration '
-                                       f'\'{config.display_name}\' and try again.')
-                if auth_type == 'KUBERNETES_SECRET' and\
-                   len(config.metadata.get('cos_secret', '').strip()) == 0:
-                    raise RuntimeError('Cannot connect to object storage. '
-                                       f'Authentication provider \'{auth_type}\' requires '
-                                       'a Kubernetes secret name. Update runtime configuration '
-                                       f'\'{config.display_name}\' and try again.')
                 cred_provider = providers.StaticProvider(
                     access_key=config.metadata['cos_username'],
                     secret_key=config.metadata['cos_password'],
@@ -87,17 +67,6 @@ class CosClient(LoggingConfigurable):
                     raise RuntimeError('Cannot connect to object storage. The value of environment '
                                        'variable AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS references '
                                        f"'{os.environ['AWS_WEB_IDENTITY_TOKEN_FILE']}', which is not a valid file.")
-
-                # TODO remove temporary validation checks after
-                # metadata can properly enforce constraints
-                if len(config.metadata.get('cos_username', '').strip()) > 0 or\
-                   len(config.metadata.get('cos_password', '').strip()) > 0 or\
-                   len(config.metadata.get('cos_secret', '').strip()) > 0:
-                    raise RuntimeError('Cannot connect to object storage. '
-                                       f'Authentication provider \'{auth_type}\' does not '
-                                       'support credentials or Kubernetes secrets. Update runtime configuration '
-                                       f'\'{config.display_name}\' and try again.')
-
                 cred_provider = providers.IamAwsProvider()
             else:
                 raise RuntimeError('Cannot connect to object storage. '
@@ -134,7 +103,7 @@ class CosClient(LoggingConfigurable):
             raise ex from ex
 
         except ValueError as ex:
-            # providers.IamAwsProvider raises this if if something bad happened
+            # providers.IamAwsProvider raises this if something bad happened
             if isinstance(cred_provider, providers.IamAwsProvider):
                 raise RuntimeError(f'Cannot connect to object storage: {ex}. Verify that '
                                    f'environment variable AWS_WEB_IDENTITY_TOKEN_FILE contains a valid value.')
--- a/elyra/util/cos.py
+++ b/elyra/util/cos.py
@@ -46,16 +46,23 @@ class CosClient(LoggingConfigurable):
             self.bucket = config.metadata['cos_bucket']
             if auth_type in ['USER_CREDENTIALS', 'KUBERNETES_SECRET']:
                 # TODO remove temporary validation checks after
-                # the 'cos_username' and 'cos_password' schema properties
-                # are marked as required.
+                # metadata can properly enforce constraints
                 if len(config.metadata.get('cos_username', '').strip()) == 0 or\
                    len(config.metadata.get('cos_password', '').strip()) == 0:
-                    raise RuntimeError(f'Authentication provider \'{auth_type}\' requires '
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' requires '
                                        'a username and password. Update runtime configuration '
+                                       f'\'{config.display_name}\' and try again.')
+                if auth_type == 'USER_CREDENTIALS' and\
+                   len(config.metadata.get('cos_secret', '').strip()) > 0:
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' does not '
+                                       'support Kubernetes secrets. Update runtime configuration '
                                        f'\'{config.display_name}\' and try again.')
                 if auth_type == 'KUBERNETES_SECRET' and\
                    len(config.metadata.get('cos_secret', '').strip()) == 0:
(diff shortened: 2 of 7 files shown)


elyra/pipeline/kfp/kfp_metadata.py
```diff
@@ -33,14 +33,23 @@ def on_load(self, **kwargs: Any) -> None:
             # Inject auth_type property for metadata persisted using Elyra < 3.3:
             # - api_username and api_password present -> use DEX Legacy
             # - otherwise -> use no authentication type
-            if self.metadata.get('api_username') is None or\
-               len(self.metadata.get('api_username').strip()) == 0 or\
-               self.metadata.get('api_password') is None or\
-               len(self.metadata.get('api_password').strip()) == 0:
+            if len(self.metadata.get('api_username', '').strip()) == 0 or\
```
ptitzler, line 36: Note: this simplifying change (this line and next) is unrelated to the goal of this PR but was included since the file is touched by other changes.

kevin-bates:
> Due to current UI/CLI limitations, _required properties are not_ enforced using metadata

I think the wording of this is too loose since this _could_ be accomplished via the `pre_save()` or `post_save()` hooks as part of metadata persistence.  Perhaps something like this?

"Due to current UI/CLI limitations, _conditionally-required properties are not_ enforced via metadata schema validation."

elyra/util/cos.py
```diff
@@ -46,16 +46,23 @@ def __init__(self, config=None, endpoint=None, access_key=None, secret_key=None,
...
             if auth_type in ['USER_CREDENTIALS', 'KUBERNETES_SECRET']:
                 # TODO remove temporary validation checks after
-                # the 'cos_username' and 'cos_password' schema properties
-                # are marked as required.
+                # metadata can properly enforce constraints
                 if len(config.metadata.get('cos_username', '').strip()) == 0 or\
                    len(config.metadata.get('cos_password', '').strip()) == 0:
-                    raise RuntimeError(f'Authentication provider \'{auth_type}\' requires '
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' requires '
```
kevin-bates, line 53: I would suggest this be moved into the respective metadata class and let metadata properly enforce the constraints via the available "save hooks".

elyra/util/cos.py
```diff
@@ -46,16 +46,23 @@ def __init__(self, config=None, endpoint=None, access_key=None, secret_key=None,
...
                                        'a username and password. Update runtime configuration '
                                        f'\'{config.display_name}\' and try again.')
+                if auth_type == 'USER_CREDENTIALS' and\
+                   len(config.metadata.get('cos_secret', '').strip()) > 0:
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' does not '
+                                       'support Kubernetes secrets. Update runtime configuration '
+                                       f'\'{config.display_name}\' and try again.')
```
kevin-bates, line 61: Do we need to validate items that will be ignored anyway?  This will make it frustrating for users to switch between authentication types if they're forced to unset various fields depending on the type.  If this is required, I would suggest it too move in the appropriate metadata class hook method.

elyra/util/cos.py
```diff
@@ -28,36 +28,84 @@ class CosClient(LoggingConfigurable):
...
+                   len(config.metadata.get('cos_secret', '').strip()) == 0:
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' requires '
+                                       'a Kubernetes secret name. Update runtime configuration '
+                                       f'\'{config.display_name}\' and try again.')
+                cred_provider = providers.StaticProvider(
+                    access_key=config.metadata['cos_username'],
+                    secret_key=config.metadata['cos_password'],
+                )
+            elif auth_type == 'AWS_ENV_VARIABLES':
```
kevin-bates, line 72: Is this a valid `cos_auth_type`?  It's not listed in the description.  It too could be embedded in the save hook.

elyra/util/cos.py
```diff
@@ -28,36 +28,84 @@ class CosClient(LoggingConfigurable):
...
+                if os.environ.get('AWS_ACCESS_KEY_ID') is None or\
+                   os.environ.get('AWS_SECRET_ACCESS_KEY') is None:
+                    raise RuntimeError('Cannot connect to object storage. '
+                                       f'Authentication provider \'{auth_type}\' requires '
+                                       'environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.')
+                cred_provider = providers.EnvAWSProvider()
+            elif auth_type == 'AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS':
+                if os.environ.get('AWS_ROLE_ARN') is None or\
```
kevin-bates, line 80: Ditto here - could be part of a save hook.

kevin-bates (commented):
I think it would be reasonable to move the validation into the metadata class.  Then we'll be better positioned within the pipeline code once `oneOf` support is available.

ptitzler:
TODO (KFP): error handling in existing code disguises root cause if invalid credentials are specified.

```
RuntimeError: Failed to compile pipeline 'hello' with engine 'Argo' to: '/var/folders/h8/l3qh8v192rv_vvggc6wyphzc0000gn/T/tmpa1d19202/hello.tar.gz'
```

elyra/util/cos.py
ptitzler: It's not, but this is a side effect of the fact that the PR was and not ready for review and still a work in progress.

elyra/util/cos.py
ptitzler: I did consider validating the existence of those environment variables but ultimately decided against it:
- it's a valid scenario for the variables to not exist until they are needed (an analogy here would be if we were to prevent saving of a KFP runtime configuration if the specified kubeflow server cannot be "pinged" - e.g. because one forgot to connect to a VPN)
- the metadata service doesn't seem to support yet warnings that alert the user to the fact that something in the configuration might be wrong but is not treated as a fatal error (if there is please let me know!)

elyra/util/cos.py
kevin-bates: I guess I don't agree here.  If I have `auth_type` == `KUBERNETES_SECRET` working but want to switch to `USER_CREDENTIALS` how do I unset `cos_secret` assuming the UI will hide that field when I've switched to `USER_CREDENTIALS`.   Must I unset the field _prior_ to switching `auth_type`?  If `cos_secret` is ignored for `USER_CREDENTIALS`, why enforce that it must have no value?

I think the same could be said about `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS`.  When set, ignore the others.

Once we have support for `oneOf` and object-valued properties this form of validation won't be applicable since the fields will be a function of the `auth_type`.

elyra/util/cos.py
kevin-bates: No worries.  Just had some responses to your points.

> it's a valid scenario for the variables to not exist until they are needed

Since these env vars are in the environment of the Elyra process, how are they going to come into existence _after_ the process has started?  Although Python does allow a running process's environment to be modified (Java does not), we don't provide any mechanisms for setting these variables after startup (nor should we IMO).

> the metadata service doesn't seem to support yet warnings that alert the user to the fact that something in the configuration might be wrong but is not treated as a fatal error 

Correct, this is by design.  The metadata service is agnostic to the context of its instances other than what can be expressed in JSON schema.  However, schema validation is binary (success or failure), so warnings are not something that will be produced by the Metadata Service itself.  However, this is one of the reasons we support custom metadata classes since they DO have contextual awareness and could produce similar "warnings" (which appear to be `RuntimeError`s) relative to a given instance.

I'm curious how we handle this with anything other than an exception from the server since the request has to fail.  What is your notion of a _fatal error_ vs. a _warning_ in this particular instance (where an `auth_type` doesn't have its required properties)?

ptitzler: > Since these env vars are in the environment of the Elyra process, how are they going to come into existence after the process has started?

I was thinking of container-based deployments, where metadata is persisted across invocations, yet the environment settings can change.

> What is your notion of a fatal error vs. a warning in this particular instance (where an auth_type doesn't have its required properties)?

I was thinking that a warning could alert the user that some action needs to be taken for this configuration to be ready for use. In this specific scenario a message could indicate that environment variables XYZ1 and XYZ2 are required but not set. Since we currently don't explicitly expose prerequisites in the UI ("requires environment variables 'XYZ1' and 'XYZ2' to be set" ) users probably won't have them set in their Elyra environment the first time they try to use an RTC with this authentication type.

elyra/pipeline/kfp/kfp_metadata.py
```diff
@@ -33,15 +33,58 @@ def on_load(self, **kwargs: Any) -> None:
...
+            return
+
+        if self.metadata['cos_auth_type'] == 'USER_CREDENTIALS':
+            if len(self.metadata.get('cos_username', '').strip()) == 0 or\
+               len(self.metadata.get('cos_password', '').strip()) == 0:
+                raise ValueError('A username and password are required '
+                                 'for the selected Object Storage authentication type.')
+            if len(self.metadata.get('cos_secret', '').strip()) > 0:
+                raise ValueError('Kubernetes secrets are not supported '
+                                 'for the selected Object Storage authentication type.')
```
akchinSTC, line 78:
```suggestion
                raise ValueError('Kubernetes secrets are not supported '
                                 'for the selected Object Storage authentication type.'
                                 'Please clear the "Cloud Object Storage Credentials Secret" field.' )
```
Can we add an extra sentence here telling the user to clear the cos secret parameter in the config?
image

elyra/pipeline/kfp/kfp_metadata.py
```diff
@@ -33,15 +33,58 @@ def on_load(self, **kwargs: Any) -> None:
...
+               len(self.metadata.get('cos_password', '').strip()) == 0 or\
+               len(self.metadata.get('cos_secret', '').strip()) == 0:
+                raise ValueError('Username, password, and Kubernetes secret are required '
+                                 'for the selected Object Storage authentication type.')
+        elif self.metadata['cos_auth_type'] == 'AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS':
+            if len(self.metadata.get('cos_username', '').strip()) > 0 or\
+               len(self.metadata.get('cos_password', '').strip()) > 0 or\
+               len(self.metadata.get('cos_secret', '').strip()) > 0:
+                raise ValueError('Username, password, and Kubernetes secret are not supported '
+                                 'for the selected Object Storage authentication type.')
```
akchinSTC, line 90: ditto for this one?

elyra/pipeline/kfp/kfp_metadata.py
akchinSTC: just makes it quicker for someone to figure out what to do if they keep clicking and it wont save. either this or have the UI clear the field automatically after selecting a different cos auth type, but thats def more work

elyra/util/cos.py
kevin-bates: > I was thinking of container-based deployments, where metadata is persisted across invocations, yet the environment settings can change.

Ok, but since the code that is checking for the environment variable _is_ the server _process_, if the envs are not present, the container would require a redeployment (where the variables are set via the container's CLI or helm chart, etc.) - just like you'd need to restart the process (with the envs present) in non-container configurations.

> I was thinking that a warning could alert the user that some action needs to be taken for this configuration to be ready for use.  ...

I see.  The request would still fail, but the message returned with that failure would be phrased as a warning - am I understanding this correctly?

elyra/pipeline/kfp/kfp_metadata.py
kevin-bates: This is exactly what I was getting at with this https://github.com/elyra-ai/elyra/pull/2354#discussion_r767062791.  If `cos_secret` isn't used by the `auth_type` why enforce it to not be set?  (Same goes for `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS` (below) by also including `username` and `password` in the set of tolerated values.)

elyra/util/cos.py
ptitzler: > > I was thinking of container-based deployments, where metadata is persisted across invocations, yet the environment settings can change.
> 
> Ok, but since the code that is checking for the environment variable _is_ the server _process_, if the envs are not present, the container would require a redeployment (where the variables are set via the container's CLI or helm chart, etc.) - just like you'd need to restart the process (with the envs present) in non-container configurations.

If you mean by 'redeployment' restart, then I believe the answer is yes. One can, for example, pod defaults that would take effect when a container comes up. 


> > I was thinking that a warning could alert the user that some action needs to be taken for this configuration to be ready for use.  ...
> 
> I see. The request would still fail, but the message returned with that failure would be phrased as a warning - am I understanding this correctly?
Yes!

elyra/pipeline/kfp/kfp_metadata.py
ptitzler: Sorry, somehow the referenced comments doesn't open up for me. I get tis preview but can see the remainder of your comment:
image

> If cos_secret isn't used by the auth_type why enforce it to not be set?

My motivation for taking this approach was as follows:
 - Ensure compatibility and consistency - the saved schema instance must not include any properties that are not applicable. This approach should mimic the behavior one will observe (I am assuming), once conditional properties are supported: the stored schema instance file only includes relevant properties to avoid the need for interpretation.
 - I initially considered removing conditional properties that are not applicable in `pre_save()` to accomplish the above. However, doing so would lead to two potential usability issues: 
   1. When the user enters problematic values and no error is raised they might infer that these values are applicable, even though they are not. With the current limitation the UI already leads to confusion and this would add to it.
   1. When the user re-opens the RTC  those non-applicable property values would be "gone". A user _might_ come to believe that this behavior is a bug because they entered values and the save operation completed without an error.

ptitzler: I do agree that changing the message wording to include an action would be helpful.

nanaones:
I'm looking forward to this feature.

elyra/pipeline/kfp/kfp_metadata.py
ptitzler: A potential mitigation option for "[...] those non-applicable property values would be 'gone'. A user might come to believe that this behavior is a bug" might be to update the property description to explicitly state something like

```
This property is required for authentication type 'XYZ'. It does not apply to other authentication types. 
```

kevin-bates: > Sorry, somehow the referenced comments doesn't open up for me. I get tis preview but can see the remainder of your comment:

No worries - it's in one of the resolved comments above.

Yeah, it dawned on me that the UI for this is probably an issue since probably doesn't "hide" non-applicable properties.  I think this is going to be a usability problem until we have better support for this kind of thing - so whatever you want to do is okay with me.  I just figured ignoring unused properties is easier, allows users the ability to switch between types, and is more in line with how the server code will behave when `oneOf` support is present.

Will there be any hint to the user which fields are required per authentication type?

ptitzler: > ignoring unused properties is easier, 

True, but it also introduces unpredictability with respect to which properties might be present in the persisted runtime metadata file. I am trying to avoid that. For example, if we were to take this approach, there are several property permutations one could find in those files for  the `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS` type, depending on what the user decided to enter:
 - no username, no password, no secret
 - username, no password, no secret, ...
 - ...
 - username, password, secret
 
Of all the permutations only one would be semantically valid.

> allows users the ability to switch between types, 

While in edit mode that will still be possible with the delivery of this PR, but I don't know to what extend the UI will support something like that in the future. For illustrative purposes (I'm not trying to document any expectations here)  if a user chooses `USER_CREDENTIALS` and enters uid + password, then switches to the `AWS_IAM_ROLES_FOR_SERVICE_ACCOUNTS` auth type - which has no properties, and then switches to the `KUBERNETES_SECRET` auth type. Are the uid + password values supposed to be pre-populated?

> Will there be any hint to the user which fields are required per authentication type?

We've updated the RTC documentation

image

and I am about to extend the property descriptions that are exposed as tool tips:

image

kiersten-stokes (approved):
LGTM! I didn't run into an unexpected failures (or unexpected successes) in my various tests

elyra/pipeline/kfp/kfp_metadata.py
kevin-bates: Understood, thanks.

kevin-bates (approved):

PR #2358, opened 2021-12-13
Fix KUBERNETES_SERVICE_ACCOUNT_TOKEN auth

thesuperzapper:

This PR fixes the `KUBERNETES_SERVICE_ACCOUNT_TOKEN` auth type, as it currently will never work.

The current problem is caused by a VERY bad assumption being made by Kubeflow itself. Inside the `kfp.Client()` initialization code, if a `host` is specified it will never enable ServiceAccount auth. ([Relevant code in `kfp.Client()`](https://github.com/kubeflow/pipelines/blob/1e032f550ce23cd40bfb6827b995248537b07d08/sdk/python/kfp/_client.py#L269-L271))

Because we require that `api_endpoint` is set and [pass it to the `kfp.Client()` constructor](https://github.com/elyra-ai/elyra/blob/741440086b7179c1fb14daf1f7a6a3a13777ac22/elyra/pipeline/kfp/processor_kfp.py#L135), we always specify a `host`.

Luckily the solution is relatively simple, we can explicitly tell the `kfp.Client()` constructor that we want to use ServiceAccount auth, by passing a `kfp.auth.ServiceAccountTokenVolumeCredentials()` to the `credentials` argument of `kfp.Client()`.

- Running Elyra in a Kubeflow cluster with ServiceAcount auth

----
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/kfp/kfp_authentication.py
+++ b/elyra/pipeline/kfp/kfp_authentication.py
@@ -29,6 +29,8 @@ from urllib.parse import urlsplit
 
 from kfp.auth import KF_PIPELINES_SA_TOKEN_ENV
 from kfp.auth import KF_PIPELINES_SA_TOKEN_PATH
+from kfp.auth import ServiceAccountTokenVolumeCredentials
+from kfp.auth import TokenCredentialsBase
 import requests
 
 
@@ -226,6 +228,7 @@ class KFPAuthenticator():
             'auth_type': None,        # Authentication type, source: runtime config
             'kf_secured': False,      # Indicates whether KF API is secured
             'cookies': None,          # passed to KFP SDK client as "cookies" param value
+            'credentials': None,      # passed to KFP SDK client as "credentials" param value
             'existing_token': None    # passed to KFP SDK client as "existing_token" param value
         }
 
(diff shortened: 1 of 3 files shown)


ptitzler:
@alexisc67 sounds like the problem you've brought up on gitter!

elyra/pipeline/kfp/kfp_authentication.py
```diff
@@ -628,7 +633,7 @@ class K8sServiceAccountTokenAuthenticator(AbstractAuthenticator):
 
     def authenticate(self,
                      kf_endpoint: str,
-                     runtime_config_name: str) -> None:
+                     runtime_config_name: str) -> TokenCredentialsBase:
```
ptitzler, line 636: Can you please also update the doc string (`:return: ...`) to reflect the change?

ptitzler (approved):
LGTM

elyra/pipeline/kfp/kfp_authentication.py
thesuperzapper: Fixed in https://github.com/elyra-ai/elyra/pull/2358/commits/0c354d5ba8fe7c31026f9a11f6a0a3d0a6ab42a0

thesuperzapper:
@ptitzler this PR should be ready to merge now.

PR #2363, opened 2021-12-14
Fix runtime pipeline property

### LINKED ISSUE #2347: Pipeline properties don't display runtime type information
**Describe the issue**

The displayed pipeline runtime property value is empty for all pipelines.

image

**To Reproduce**
1. Create a pipeline
2. Open the pipeline properties view

**Screenshots or log output**
<pre>
Paste the log output here.
</pre>
</details>

**Expected behavior**

**Deployment information**
 - Elyra version: 3.3.0

marthacryan:
Fixes #2347. Manually writes the runtime display name into the pipeline properties runtime property. 
image
image
image

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineEditorWidget.tsx
+++ b/packages/pipeline-editor/src/PipelineEditorWidget.tsx
@@ -284,6 +284,7 @@ const PipelineWrapper: React.FC<IProps> = ({
           PathExt.extname(pipeline_path)
         );
         pipelineJson.pipelines[0].app_data.properties.name = pipeline_name;
+        pipelineJson.pipelines[0].app_data.properties.runtime = runtimeDisplayName;
       }
       setPipeline(pipelineJson);
       setLoading(false);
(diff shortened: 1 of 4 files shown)


karlaspuldaro:
LGTM.
Only requires updating the integration tests snapshots to have `app_data.properties.runtime` field I guess.

kevin-bates:
I personally think this field (and the corresponding pipeline editor field) is mislabeled and should really be something like  "Pipeline Type".  This is because "Generic" is not describing a _runtime_ (or, um, _platform_) at all, and the runtime, in the case of "Generic pipelines", is not determined until there's a runtime configuration associated with the pipeline.  IMO I think it's more correct that both instances of "Runtime" labels be renamed to "Type" (on the canvas border - or whatever that's called) and "Pipeline Type" (in the properties page).  "That is, pipelines of _type_ Generic can run in Airflow or Kubeflow."

I'm also not thrilled with the re-introduction of `app_data.properties.runtime` when (IIRC) we also have `app_data.runtime` and it was very confusing when the two existed previously.  (I'm not sure `app_data.runtime` still exists, but, if so, I think one of the two deserves a different name.)

If this is something that needs to be in 3.4 in some form, perhaps we can circle back on this another time to further cleanup naming.

ajbozarth (approved):
> If this is something that needs to be in 3.4 in some form, perhaps we can circle back on this another time to further cleanup naming.

I agree with this

> I'm also not thrilled with the re-introduction of app_data.properties.runtime

We never removed this value, we just don't set it consistently. Also it's consistently used as a display name so the I do believe we should rename the value (or just switch to using a differently named value so as to avoid needing migration)

PR #2371, opened 2021-12-22
support customized schedule interval

### LINKED ISSUE #471: Enable recurring schedule of Notebooks and Pipelines
Looks like Kubeflow has enabled recurring schedule of experiments, and we could use to enable notebook and pipelines to be scheduled 

image

and the detailed configuration

image

wenjun93:
dag schedule interval is set '@once' by default, we should expose it as an user-defined option

Fixes #471

### DIFF
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -400,6 +400,9 @@ be fully qualified (i.e., prefixed with their package names).
             if pipeline_description is None:
                 pipeline_description = f"Created with Elyra {__version__} pipeline editor using `{pipeline.source}`."
 
+            schedule_interval = pipeline.schedule_interval
+            if schedule_interval is None:
+                schedule_interval = '@once'
             python_output = template.render(operations_list=target_ops,
                                             pipeline_name=pipeline_name,
                                             namespace=user_namespace,
@@ -407,6 +410,7 @@ be fully qualified (i.e., prefixed with their package names).
                                             kube_config_path=None,
                                             is_paused_upon_creation='False',
                                             in_cluster='True',
+                                            schedule_interval=schedule_interval,
                                             pipeline_description=pipeline_description)
 
             # Write to python file and fix formatting
--- a/packages/pipeline-editor/src/pipelineProperties.ts
+++ b/packages/pipeline-editor/src/pipelineProperties.ts
@@ -18,9 +18,15 @@ const pipelineProperties = {
   current_parameters: {
     name: '',
     runtime: '',
+    schedule_interval: '',
     description: ''
   },
-  parameters: [{ id: 'name' }, { id: 'runtime' }, { id: 'description' }],
+  parameters: [
+    { id: 'name' },
+    { id: 'runtime' },
+    { id: 'schedule_interval' },
+    { id: 'description' }
+  ],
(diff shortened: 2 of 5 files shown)


elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -400,13 +400,17 @@ def create_pipeline_file(self, pipeline, pipeline_export_format, pipeline_export
             if pipeline_description is None:
                 pipeline_description = f"Created with Elyra {__version__} pipeline editor using `{pipeline.source}`."
 
+            schedule_interval = pipeline.schedule_interval
```
lresende, line 403: Would it be better to have this in pipeline definition or pipeline value object directly?

packages/pipeline-editor/src/pipelineProperties.ts
```diff
@@ -36,6 +42,19 @@ const pipelineProperties = {
         parameter_ref: 'runtime',
         label: { default: 'Pipeline Runtime' }
       },
+      {
+        control: 'custom',
+        custom_control_id: 'StringControl',
+        parameter_ref: 'schedule_interval',
+        label: {
+          default:
```
lresende, line 50: Could you work with @marthacryan or @ajbozarth and see if they could help get a better user experience.

lresende:
Overall LGTM, but I would suggest collaborating with others on the community to come up with a better UX.

packages/pipeline-editor/src/pipelineProperties.ts
wenjun93: @marthacryan @ajbozarth could please take a look and see how to improve the UX

ptitzler:
Perhaps a better/cleaner approach would be to introduce this as a global pipeline property that's specific to Apache Airflow pipelines.

packages/pipeline-editor/src/pipelineProperties.ts
```diff
@@ -36,6 +42,19 @@ const pipelineProperties = {
...
+            'Pipeline Schedule Interval(CRON expression, once if left empty)'
+        },
+        data: {
+          placeholder: '0 0 * * *',
```
akchinSTC, line 54: If possible (albeit more work), maybe we can make this more intuitive to use by include the airflow presets as an option via dropdown in addition to supporting the cron job notation e.g. left column here:
image

akchinSTC:
See related if work scope is to expand : https://github.com/elyra-ai/elyra/issues/471

lresende:
I have updated #471 with some more details on the requirements for this. It would be good if we could handle both Airflow and KFP runtimes for this feature, but if you are not completely familiar with KFP I can help. Also, if you need help on the ui field, please let me know and I will push a commit here but should be simple with the options from the airflow pattern.

packages/pipeline-editor/src/pipelineProperties.ts
ajbozarth: sorry I didn't comment earlier, I actually though this through last week. I agree with Alan's comment below about leveraging the airflow presets. If we do that then this property can be an enum using the preset values and then the backend can interpret those values into the correct string for the runtimes that dont support the notation

ptitzler:
Closing PR due to lack of activity. Please re-open if you still intend to work in this.

Issue #2376, opened 2022-01-06
[COMMUNITY REVIEW] Deprecate metadata parameters in favor of json for metadata cli


lresende:
**Is your feature request related to a problem? Please describe.**
We have been discussing improvements on the Metadata CLI and one of the options would be to start using files as the input for the CLI instead of using parameters for each field which is not so flexible and flat. Using files also would help in importing/exporting existing metadata configuration from the filesystem (e.g. moving to new envs, etc)

The idea would be to deprecate the current functionality on 3.x and remove it on 4.0

Please express your opinion in the comments.

kevin-bates:
Thanks for opening this @lresende. 

I think the richness inherent in JSON schema introduces a sustainability issue with attempting to interpret, then convert, each field of a schema property to a CLI option.  Since we now allow folks to bring their own schemas, we have less control over what constructs are used in the schema.  As a result, I think, at a minimum, we need a means of allowing a direct bypass of the schema-to-options processing to occur.

This will, however, affect users.  

- No longer will there be a rich "usage statement" that lists the schema properties as parameters, including their description, type, and required-ness.
- Schema validation errors are cryptic and we'd be at the mercy of these for most things that are caught today prior to submission to the metadata service.
- Instance creation requires knowledge of the JSON syntax, where quoted strings, lists, object-valued properties will be difficult for users.

We need to make sure there's also a way to create instances directly (not using files), so I would propose both a `--file` and a `--json` option with the latter treating the CLI value as a JSON string (and using `json.loads()` to apply the JSON syntax interpretation).

Regarding import and export, I believe these should be separate sub-commands for the tooling where import can read from an export "sink" (most likely a directory).

One of the potential benefits of using a more "raw" approach, is that we could entertain the idea of using the REST API instead (I suppose we don't preclude doing this today, although it requires a running instance).  This would then allow for admins to more easily replicate configurations by not having to be local to each system.

For example, the following would replicate instances of `my_schema` from the system running at `foo` to that running at `bar`:
```
elyra-metadata export --schema_name=my_schema --endpoint=http://foo:port --directory=my_exported_data
elyra-metadata import --schema_name=my_schema --endpoint=http://bar:port --directory=my_exported_data
```

[...]

Issue #1066, opened 2020-11-12
Import/export functionality for runtime configurations


romeokienzler:
In runtime configurations, it would be helpful to import/export runtime configurations.
In addition, importing / exporting a list or selection of configurations would be a nice additional feature

![image](https://user-images.githubusercontent.com/5694071/98913978-5ef36280-24c8-11eb-8eae-ce8beae52165.png)

ptitzler:
We should probably support this for all types of metadata (code snippets, runtime images, ...), not just runtime configurations.

kevin-bates:
We should also discuss:
- if this is to be performed from UI or CLI or both?
- what is the output format - which becomes the input format?  A combined JSON or tar file, etc.? (JSON is more user friendly)
- how are desired export sets expressed?  List of specific names?  Everything for a given schema?  namespace?
- how to deal with differing schema versions on import?

lresende:
Also, it's good to remember that these are pretty much json files on a local folder, so in the mean time, they can be manually copied. 

```
$ elyra-metadata list runtimes
Available metadata instances for runtimes (includes invalid):

Schema   Instance    Resource                                                           
------   --------    --------                                                           
kfp      kfp-shared  /Users/lresende/Library/Jupyter/metadata/runtimes/kfp-shared.json  
```

ptitzler:
Metadata import using CLI will be delivered by the PR for https://github.com/elyra-ai/elyra/issues/2500. 

A related feature (duplicate metadata) was added to the GUI to make it easier to create similar metadata instances quickly. ([Documentation](https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html#duplicating-a-runtime-configuration), [Demo](https://github.com/elyra-ai/elyra/releases/tag/v3.6.0))

Import/export from/to a remote location (browser on machine A, JupyterLab running on machine B) is not supported.

PR #2378, opened 2022-01-08
Add support for CLI metadata and OVPs using JSON or file contents

kevin-bates:
Per [this comment](https://github.com/elyra-ai/elyra/issues/2376#issuecomment-1006798310) we have a sustainability issue with attempting to interpret complex schemas (nested object-valued properties, schema composition, references, etc.). This pull request attempts to alleviate those concerns while also better-positioning us with the ability to replicate metadata configurations by enabling the following...

- Mutually exclusive options `--file` and `--json` have been added where:
    -  `--file`: will have a value of a file name containing the metadata for the instance being created or updated.  The contents must be valid JSON and can contain the complete metadata (with a `metadata` stanza and top-order properties like `schema_name` and `display_name`) or, if `metadata` is not present, treated as _the_ `metadata` stanza (thereby allowing external applications to _inject_ their JSON into the metadata service without knowledge of the system-owned properties as they can also be specified on the command-line).
    - `--json`: will have a value of JSON string, containing the metadata for the instance being created or updated.  As with `--file` the contents of the string could be the complete metadata or only the `metadata` stanza itself.
- Overrides of values _within_ the contents of the `--file` or `--json` references via the command-line are also supported.  As a result, one can specify a different `--display_name` value than is in the referenced content, or a different `--name` value than what is inferred by the filename, enabling replication.  
- `--replace` functionality has been updated to apply individual property updates and no longer requires the complete instance definition.  (I could see this having utility, both in terms of being able to update complex schema types, but also in terms of security since, in theory, the contents can be protected by filesystem permissions, etc.)
- General object-valued properties are also supported, such that a given _property_ could be overridden via a file or JSON reference.  (Note that this is also true for _any_ top-level property (metadata or above)).  For example, if the schema defines an object-valued property (e.g., `authentication`) and a user wishes to replace one instance's authentication with a new value, the following command could be issued:
```
elyra-metadata install runtimes --schema_name=kfp --name=my_kubeflow_config --replace --authentication=/path/to/my/authentication_values.json
```
and replace instance `my_kubeflow_config`'s current authentication OVP with the contents of `/path/to/my/authentication_values.json`.
~(Note: As of this moment the above example doesn't work because we make _replacement_ (i.e., update), fully specify the object.  We should make `--replace` only allow the specification of the _changed_ data - along with the necessary data to identify the existing instance.  This PR should address that as well. See 3rd item.)~

Still to do:
- [x] Auto-detect complex schemas and warn that `--file` or `--json` options should be used
- [x] Allow for `--replace` to only specify the properties that will be changed - along with necessary identifying options.
- [x] Add tests for the new functionality
- [x] Update docs to include new options

This PR will remain in **draft** until all items have been completed.
 
           in the file; or

           it.

### DIFF
--- a/docs/source/user_guide/command-line-interface.md
+++ b/docs/source/user_guide/command-line-interface.md
@@ -82,47 +82,58 @@ The `elyra-metadata` application derives its command-line options (aside from a 
 
 Application-level properties within a schema reside as top-level proeprties within the schema's `metadata` stanza.  For example, here's the `code-snippet` schema for the `code-snippets` schemaspace:
 
-```JSON
-    "metadata": {
-      "description": "Additional data specific to this Code Snippet",
-      "type": "object",
-      "properties": {
-        "description": {
-          "title": "Description",
-          "description": "Code snippet description",
-          "type": "string"
-        },
-        "tags": {
-          "title": "Tags",
-          "description": "Tags for categorizing snippets",
-          "type": "array",
-          "uihints": {
-            "field_type": "tags"
-          }
-        },
-        "language": {
-          "title": "Language",
-          "description": "Code snippet implementation language",
-          "type": "string",
-          "uihints": {
-            "field_type": "dropdown",
-            "default_choices": ["Python", "Java", "R", "Scala", "Markdown"],
-            "category": "Source"
-          },
-          "minLength": 1
-        },
-        "code": {
-          "title": "Code",
-          "description": "Code snippet code lines",
-          "type": "array",
-          "uihints": {
-            "field_type": "code",
-            "category": "Source"
-          }
+```json
+{
+  "metadata": {
+    "description": "Additional data specific to this Code Snippet",
+    "type": "object",
+    "properties": {
+      "description": {
+        "title": "Description",
+        "description": "Code snippet description",
+        "type": "string"
+      },
+      "tags": {
+        "title": "Tags",
+        "description": "Tags for categorizing snippets",
+        "type": "array",
+        "uihints": {
+          "field_type": "tags"
         }
       },
-      "required": ["language", "code"]
-    }
+      "language": {
+        "title": "Language",
+        "description": "Code snippet implementation language",
+        "type": "string",
+        "uihints": {
+          "field_type": "dropdown",
+          "default_choices": [
+            "Python",
+            "Java",
+            "R",
+            "Scala",
+            "Markdown"
+          ],
+          "category": "Source"
+        },
+        "minLength": 1
+      },
+      "code": {
+        "title": "Code",
+        "description": "Code snippet code lines",
+        "type": "array",
+        "uihints": {
+          "field_type": "code",
+          "category": "Source"
+        }
+      }
+    },
+    "required": [
+      "language",
+      "code"
+    ]
+  }
+}
 ```
 and `elyra-metadata` generates options corresponding to each of the `metadata` properties and including helpful tips like whether the property is required and a hint as to how its value should be entered:
 ```
--- a/docs/source/user_guide/command-line-interface.md
+++ b/docs/source/user_guide/command-line-interface.md
@@ -77,6 +77,79 @@ Refer to the topics below for detailed information on how to use `elyra-metadata
  - [Manage runtime images](runtime-image-conf.html#managing-runtime-images-with-the-command-line-interface)
  - [Manage pipeline components](pipeline-components.html#managing-custom-components-using-the-elyra-cli)
 
+#### Creating and updating metadata with complex properties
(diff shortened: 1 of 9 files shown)


ptitzler:
A common [usability] issue with files is that it's not trivial to create them from scratch. Let's say, for example, a user wants to create a runtime configuration from a file. Is there something Elyra could do to  provide a "skeleton" that a user can use to create a file?

kevin-bates:
> Is there something Elyra could do to provide a "skeleton" that a user can use to create a file?

Not really since the point of enabling `--file`, `--json`, or _JSON-rich object-valued properties_ is because the schema has become too complex - preventing Elyra from programmitcally producing such a template.  I think the common approach (i.e., 90% case) would be that users create instances using the UI and leverage those instances as a "template".  Any previously-created instances are guaranteed to meet minimum requirements (i.e., have the necessary required values).

I suppose some of the factory-based schemas could document a template, but that erodes when considering BYO schemas (in which the primary use of complex schemas exists - e.g., kfp and airflow authentication).

akchinSTC:
possibly related : Export of metadata/configurations https://github.com/elyra-ai/elyra/issues/1066

ptitzler:
> replace functionality has been updated to apply individual property updates and no longer requires the complete instance definition. (I could see this having utility, both in terms of being able to update complex schema types, but also in terms of security since, in theory, the contents can be protected by filesystem permissions, etc.)

Can you please also update the CLI examples/description in https://elyra.readthedocs.io/en/latest/user_guide/runtime-conf.html#managing-runtime-configurations-using-the-elyra-cli?

kevin-bates:
Thanks @ptitzler - This exercise actually uncovered a bug where display-name was being required on updates.  This has been addressed (in the code and tests).  Here's a screenshot for the doc update:
image

docs/source/user_guide/command-line-interface.md
```diff
@@ -77,6 +77,90 @@ Refer to the topics below for detailed information on how to use `elyra-metadata
  - [Manage runtime images](runtime-image-conf.html#managing-runtime-images-with-the-command-line-interface)
  - [Manage pipeline components](pipeline-components.html#managing-custom-components-using-the-elyra-cli)
 
+#### Creating and updating metadata with complex properties
+The `elyra-metadata` application derives its command-line options (aside from a handful of system options) directly from the schema associated with the referenced schemaspace. In most cases, the schema properties are straightforward and easily determined.  However, JSON schemas can also contain complex properties and references that are not within the scope of `elyra-metadata`.  This section presents ways to create and update instances built from complex schemas.
+
+Application-level properties within a schema reside as top-level proeprties within the schema's `metadata` stanza.  For example, here's the `code-snippet` schema for the `code-snippets` schemaspace:
```
kiersten-stokes, line 83:
```suggestion
Application-level properties within a schema reside as top-level properties within the schema's `metadata` stanza. For example, here's the `code-snippet` schema for the `code-snippets` schemaspace:
```

kiersten-stokes (approved):
Just a small typo as noted below. Other than that, all the scenarios that I tested are working as expected!

docs/source/user_guide/runtime-conf.md
```diff
@@ -90,6 +90,7 @@ To create a runtime configuration for a Kubeflow Pipelines deployment:
 
 ```bash
 elyra-metadata install runtimes \
+       --schema_name=kfp
```
ptitzler, line 93: There's a continuation character missing:

```suggestion
       --schema_name=kfp \
```

docs/source/user_guide/runtime-conf.md
```diff
@@ -101,36 +102,25 @@ elyra-metadata install runtimes \
...
        --api_password=mynewpassword \
-       --engine=Argo \
-       --cos_endpoint=http://minio-service.kubeflow:9000 \
-       --cos_auth_type="USER_CREDENTIALS" \
-       --cos_username=minio \
-       --cos_password=minio123 \
-       --cos_bucket=test-bucket \
-       --tags="['kfp', 'v1.1']" \
-       --schema_name=kfp
+       --tags="['kfp', 'v1.1']"
```
ptitzler, line 120: As is, the command fails:
```
$ elyra-metadata install runtimes \
       --replace \
       --name="my_kubeflow_pipelines_runtime" \
       --schema_name=kfp \
       --api_password=mynewpassword \
       --tags="['kfp', 'v1.1']"
[I 2022-01-19 12:44:38,517.517] Upgrading runtime kfp instance 'my_kubeflow_pipelines_runtime' to include runtime_type 'KUBEFLOW_PIPELINES'...
The following exception occurred saving metadata instance for schema 'kfp': Username and password are not supported for the selected Kubeflow authentication type.
```

kevin-bates: Thanks for catching this.  I can reproduce this and believe it is related to default value handling.  I need to take a closer look and submit a fix.

kevin-bates: This issue is all in the CLI processing where options are built from the schema and properties with default values that were NOT also specified on the command line during replace operations were being included in the update payload.  Fix (and test update) forthcoming.

akchinSTC (approved):

ptitzler (approved):
LGTM

PR #2388, opened 2022-01-12
Add Script Editor language-specific subclasses

### LINKED ISSUE #1517: Create language-specific subclasses of ScriptEditor
Currently, the new [ScriptEditor class](https://github.com/elyra-ai/elyra/blob/master/packages/script-editor/src/ScriptEditor.tsx) handles its own editor language (either Python or R) based on the file type passed to its widget factory upon new widget creation.

As suggested by [this review comment](https://github.com/elyra-ai/elyra/pull/1435#discussion_r600885327), this could be better encapsulated in language-specific subclasses of `ScriptEditor` (`PythonEditor` and `REditor`),  making the `ScriptEditor` class language agnostic. The subclasses would be then free to better handle any other language-specific properties, such as the editor icon or certain kernel specs that may differ.

karlaspuldaro:
Fixes #1517 
This PR introduces new `PythonEditor` and `REditor` sublasses, making the `ScriptEditor` class language agnostic. 
The existing functionality is maintained.

Ran existing Script Editor integration tests.

           in the file; or

           it.


packages/python-editor/src/PythonEditor.tsx
```diff
@@ -0,0 +1,101 @@
...
+    options: DocumentWidget.IOptions<FileEditor, DocumentRegistry.ICodeModel>
+  ) {
+    super(options);
+    this.editorLanguage = 'python';
+
+    // Add icon to main tab
+    this.title.icon = pyIcon;
+
+    this.context.ready.then(() => {
+      this.initializeKernelSpecs();
```
kevin-bates, line 42: I suspect both constructors (`PythonEditor` and `REditor`) could be removed simply by treating `ScriptEditor` as an abstract base class that defines two abstract methods: `getLanguage()` and `getIcon()` that `PythonEditor` and `REditor` implement.  Then, wherever there's `this.editorLanguage` in `ScriptEditor`, it is replaced with `this.getLanguage()` and same goes for `this.title.icon` -> `this.getIcon()`.  The intialize Kernelspecs just moves to the `ScriptEditor` constructor as well.

You probably don't need the fields `this.editorLanguage` or `this.title.icon`, although, if they're a hassle to remove, the ScriptEditor constructor could just be like this...

```javascript
    this.editorLanguage = this.getLanguage();

    // Add icon to main tab
    this.title.icon = this.getIcon();

    this.context.ready.then(() => {
      this.initializeKernelSpecs();
```
Since the subclasses know what they are, the fields could probably be made `private`.

I also suspect the factory stuff could be simplified just based on java/python experiences. Typically, the factory for something like this would be a class method on `ScriptEditor` like `ScriptEditor.createInstance()` where `createInstance()` takes an "identifying characteristic" corresponding to the class to create. Language would probably be a good "id".  

Take this with a grain of salt since I don't know how much applies in the JS world.

packages/script-editor/src/ScriptEditor.tsx
```diff
@@ -80,7 +72,8 @@ export class ScriptEditor extends DocumentWidget<
   private runDisabled: boolean;
   private kernelSelectorRef: RefObject<ISelect> | null;
   private controller: ScriptEditorController;
-  private editorLanguage: string;
+  abstract getLanguage(): string;
+  abstract getIcon(): LabIcon | string;
```
lresende, line 76: why are we passing `labIcon` or `string`? we do have an rIcon defined in ui-components and thus I was expecting to use that in favor of consistency.

lresende (changes requested):
See comments above about rIcon

packages/script-editor/src/ScriptEditor.tsx
karlaspuldaro: I believe our `rIcon` is created differently, that's why we define it with a string, I didn't change the original implementation for it. 
When I import it from `ui-components` library the same way we add python icons, the r icon isn't loaded properly in the launcher, file browser or editor tab:
image

@ajbozarth might be able to elaborate on why/how that works?

ajbozarth: I can't remember why this fix works, but there's an issue with LabIcons where depending on the DOM within a svg added multiple copies of it causes problems. The R icon is the icon we found that has this issue.

packages/script-editor/src/ScriptEditorWidgetFactory.tsx
```diff
@@ -0,0 +1,91 @@
...
+  DocumentWidget
+} from '@jupyterlab/docregistry';
+import { FileEditor } from '@jupyterlab/fileeditor';
+
+import { ScriptEditor } from './ScriptEditor';
+
+/**
+ * A widget factory for script editors.
+ */
+export class ScriptEditorWidgetFactory extends ABCWidgetFactory<
```
ajbozarth, line 30: two comments on this (both non-blocking):

1. Why the name change from `ScriptEditorFactory` to `ScriptEditorWidgetFactory`?

2. Why move this into a separate file?

ajbozarth (approved):
A couple clarification questions for @karlaspuldaro otherwise LGTM

packages/script-editor/src/ScriptEditorWidgetFactory.tsx
karlaspuldaro: 1. The name has changed to make it more clear that it is a widget factory class.
2. It has moved to its own file because the the editor is independent of its factory (imo this should have been done before). Makes it easier to maintain (the `ScriptEditor` class will become more complex with the debugger), follows common conventions and doesn't expose to `ScriptEditor` things it doesn't need to know about from certain libraries.

PR #2399, opened 2022-01-18
Resolve naming conflict in case of multiple Airflow operators defined in single file

kiersten-stokes:
After several recent PRs that refactored the component catalog and parser functionality, a bug was introduced that does not persist a `Component` object in the case where multiple operators are present in the same file.

When parsing Airflow operators, a `Component` object is created for each operator defined in that file, and a component id is assigned. This component id is passed through from the relevant component catalog connector (recall that this class 1.) reads in the full python file, and 2.) builds a unique id based on the catalog connector type and the relevant unique-ifying keys as defined by the connector). This means that if multiple operators exist in a file, they are all getting assigned the same component id on init.

`Component` objects are added to the component cache keyed by their id. This results in the most recently parsed operator in a given file overwriting all previous operators also parsed from that file. 

This PR appends the operator classname string to the id in the case of multiple operators defined in the same file. The id for a `Component` that is one of several in a single file would now have the format: `<catalog-connector-type>:<hash-of-relevant-keys>:<class_name_of_operator>`. Note that this would not require migration, as all our existing supported example operators have a 1:1 class:file ratio.

A related change is also included that prevents helper classes from being parsed as operators. Classes are only parsed if the class from which they derive ends in the phrase `'Operator'`. See lines 96-102 in this PR.

- [x] Add a test case for this scenario

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/component_parser_airflow.py
+++ b/elyra/pipeline/airflow/component_parser_airflow.py
@@ -83,11 +83,9 @@ class AirflowComponentParser(ComponentParser):
 
     def _get_all_classes(self, component_definition: str) -> Dict[str, Dict]:
         # Organize lines and arguments according to the class to which they belong
-        class_to_content = {
-            "no_class": {"lines": [], "args": []}
-        }
+        class_to_content = {}
 
-        class_name = "no_class"
+        class_name = None
         class_regex = re.compile(r"class ([\w]+[.\w]*)\(\w*\):")
         for line in component_definition.split('\n'):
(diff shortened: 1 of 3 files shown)


kiersten-stokes:
Test `test_airflow_component_parser:test_modify_component_catalogs` is expected to fail right now because it access a resource via url that has been modified in this PR

ptitzler:
I'm wondering if it's worthwhile to fix the old code base if we are going to switch to AST-based parsing in 3.6 ...

kiersten-stokes:
Closing as this functionality is now covered by #2418

PR #2409, opened 2022-01-21
[HOLD] Add airflow package catalog connector

ptitzler:
This PR adds a catalog connector for Apache Airflow packages. Connector instances (of which there should typically be only one) require the user to configure a download URL for the Apache Airflow package that is used in the cluster.

Requires https://github.com/elyra-ai/elyra/pull/2418

- Add new `catalog-connectors` directory to the repository root, containing in the `airflow` subdirectory the newly introduced connector
- Update `Makefile` to include new `lint-connectors` target, which was also added as a dependency to the `lint` task
- Add new `lint-connectors` task to Github's `build.yaml`
- Update installation topic in documentation
- Add connector to `extras` install dependency in `setup.py` (The connector is also installed if one runs `pip install elyra[all]`)
 
Notes:
  - The connector is not included in the Elyra release process and needs to be published independently, as necessary.
  - The connector declares file name (e.g. `operators/bash_operator.py`) as _hash keys_, which are used to _internally_ identify operators in the palette. The archive name, e.g.`apache_airflow-1.10.15-py2.py3-none-any.whl`, is currently not part of the key to avoid potential versioning issues. For example, assume user A adds operators from archive `apache_airflow-1.10.15-py2.py3-none-any.whl` to the Elyra deployment and creates a pipeline using some of the operators. User B adds operators from an older archive, such as  `apache_airflow-1.10.12-py2.py3-none-any.whl `. If we were to include the archive name _as is_ as a key, user B would not be able to run pipelines that user A created (and vice versa) because (pseudo code) 
    ```
    "apache_airflow-1.10.15-py2.py3-none-any.whl:operators/bash_operator.py:BashOperator" != "apache_airflow-1.10.12-py2.py3-none-any.whl:operators/bash_operator.py:BashOperator"
    ```

    We need to decide whether the implemented behavior is sufficient (archive version numbers are completely ignored, even though this might lead to issues if the loaded operator signatures in Elyra's pipeline editor are _significantly different_ from those of the operators that are installed  in the Airflow cluster) or if semver support is required. The latter could be accomplished by using only parts of the archive name as key, e.g. by omitting/masking minor and patch version numbers. It does require though that archive names follow a constant naming pattern to allow for the extraction of version strings.

- Install connector from source, as documented in the connector's README
- Enable the connector as documented in the connector's README
  - [Download URL for Airflow 1.10 wheel](https://archive.apache.org/dist/airflow/1.10.15/apache_airflow-1.10.15-py2.py3-none-any.whl)
  - Open VPE for Airflow
  - Expand palette (core Airflow operators should be displayed) 
- Review the installation topic in the 'getting started' guide

Notes:
 - There are unresolved Elyra Airflow component parser issues that need to be addressed before the Airflow 1.10.15 package can be used.
 - The connector should already support Airflow 2.x packages but they have not been tested because Elyra does not support Airflow 2.x.

 
           in the file; or

           it.


ptitzler:
> We need to decide whether the implemented behavior is sufficient [...] or if semver support is required. The latter could be accomplished by using only parts of the archive name as key

Wheel package names appear to be consistent (random selection of releases):
- apache_airflow-1.10.12-py2.py3-none-any.whl 
- apache_airflow-1.10.15-py2.py3-none-any.whl 
- apache_airflow-2.0.2-py3-none-any.whl 
- apache_airflow-2.1.2-py3-none-any.whl 
- apache_airflow-2.2.2-py3-none-any.whl 
- apache_airflow-2.2.3-py3-none-any.whl

ptitzler:
Clarification based on today's discussion in the dev meeting. The proposed  PR implementation is based on the following assumptions:
 - the connector requires it's on life-cycle, meaning updates can be published without the need for a new Elyra release
 - the connector is treated as an optional Elyra component that is only installed when users deploy Elyra with Apache Airflow pipeline support

ptitzler:
Will redo.

Issue #2414, opened 2022-01-24
Address shortcomings in the Airflow parser code


kiersten-stokes:
**Describe the issue**
The current Airflow parser is not robust enough to support various operator specification files properly and/or without error. A few of those issues that cause partial or total failure of component parsing include (but are not limited to):

- [regression] multiple operators defined in same file only render the last operator parsed (addressed by #2399) 
- encountering `Optional` in the argument type hint
- operators that use the `*`parameter in `__init__` to signal a set of keyword-only args

Other features that should potentially be supported:

- ability to specify values for `*args` and `**kwargs` 
- a way to support operators that derive from other operator classes (e.g. not derived directly from the `BaseOperator`)

Note that support for the above features will also require changes to the Airflow processor as well.

**To Reproduce**
1. Add a component catalog that contains an operator that includes any of the above cases in it's specification file.
2. Note that the new operator may either not show up at all (parsing completely failed) or shows up and, once dragged onto the canvas, does not properly catch all properties or their types, etc.

**Expected behavior**
The Airflow parser should be able to handle most issues related to the above and should display proper log messages otherwise.

ptitzler:
MVP: parser should be able to expose the Airflow 1.10.15 core operators in the palette that were "imported" using https://github.com/elyra-ai/elyra/pull/2409

PR #2418, opened 2022-01-25
Improve Airflow parser functionality

kiersten-stokes:
Fixes portions of #2414 
Supports #2437  & #2438

Functionality Summary:

- The new parser logic covers all cases that the previous parser logic covered
- The new parser logic is much more robust against handling additional and edge cases, such as gathering data type information from type hints
- The only 'regression' is that it does not gather any default values in the case of a dictionary or list-typed parameter (this should not be an issue because default values shouldn't be given in arguments for Python mutable types)

This PR refactors the `AirflowComponentParser` to use `ast.parse` to cover additional parsing cases cleanly. General flow is the same as before: the parser takes the contents of a file, parses it for all its classes, then parses each class for its properties (i.e. init arguments).

What is different is the method by which is does those things. Rather than using regexes, `ast.parse` is used to get the abstract syntax tree (AST) of the file. This provides the class definitions and their corresponding init function (including argument objects) and docstring for further parsing. Annoyingly, `ast.parse` produces slightly different data structures for different python versions (Python 3.7 and lower vs Python 3.8+) - the relevant cases are currently handled to the best of my knowledge. The class definitions are filtered to ensure that they extend (directly or indirectly) from the `BaseOperator`. For operators that extend from a class that is not defined in the same file, it is sufficient to check the import statement for operators defined in either `airflow.providers.*.operators.` or `airflow.operators.*` (see this [link](https://airflow.apache.org/docs/apache-airflow-providers/#how-to-create-your-own-provider) for details). As before, classes are looped through to provide a list of properties for each (represented by the init argument objects and any additional information gleaned from the operator class docstring parsed earlier). Note that regexes are still used to search the class docstring for 1. a description of the property, and 2. any type information available in the docstring (which is often more verbose and therefore more helpful than the type information parsed from AST). 

This PR doesn't (yet) touch any of the processor classes.

Some discussion points:

- [x] ~~Do we want to/how should we handle cases where arguments are of 'non-standard' data types, e.g. the `PythonOperator` specifies an argument of the type 'python callable' (which will default to string)~~ This will be addressed in a follow-up PR as it falls under the umbrella of 'new functionality' and is already tracked in #2414
- [x] ~~Similar to the above, are there cases where an operator may not have an init function but we still want to include them as a Component? (again see some of the classes defined in `python_operator.py`) (different from the case where init does not have any arguments)~~ I propose that we handle this issue in another PR; this is essentially the same case as determining how to support operators that derive from other operator classes (which is already tracked in #2414)

Left to do:

- [x] Add log messages for parsing failure cases
- [x] Add a try/except around the `ast.parse` (and/or an even larger section)

- [x] Backend tests to be revised once functionality here is confirmed
- [x] Manual testing must occur to ensure that we are accurately covering as many cases as possible
  - This includes testing in Python 3.6, 3.7, 3.8, and 3.9 environments since `ast.parse` behavior can vary slightly between versions

One additional change made to the tests includes a new fixture that tears down the test component catalog instance created in certain tests (`test_modify_component_catalogs`, e.g.). In this way, if a test failure occurs before the test catalog can be removed, there are no issues in tests that occur later down the line.

 
           in the file; or

           it.


elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -20,24 +20,17 @@
...
-    def visit_FunctionDef(self, node):
-        # remove the return type defintion
-        node.returns = None
-        # remove all argument annotations
-        if node.args.args:
-            for arg in node.args.args:
-                arg.annotation = None
-        return node
+CONTROL_ID = 'OneOfControl'
+DEFAULT_DATA_TYPE = str.__name__
```
kiersten-stokes, line 30: Can probably just change this to `DEFAULT_DATA_TYPE = 'str'`

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
-                                      categories=registry_entry.categories,
-                                      properties=component_properties
-                                      )
+            new_component = Component(
+                id=component_id,
+                name=component_class,
+                description=DEFAULT_DESCRIPTION,
+                catalog_type=registry_entry.catalog_type,
+                source_identifier=registry_entry.component_identifier,
+                definition=component_definition,  # TODO use full file contents or only the class contents?
```
kiersten-stokes, line 93: Related to the to-do: not sure it really matters either way since this field isn't currently used for Airflow (it is for KFP)

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
+        if match:
+            return match.group(1).strip()
+        return DEFAULT_DATA_TYPE
 
     def get_runtime_specific_properties(self) -> List[ComponentParameter]:
         """
         Define properties that are common to the Airflow runtime.
         """
+
+        # TODO add args and kwargs as RTS-properties here?
```
kiersten-stokes, line 351: Can remove this TODO as this should probably be addressed in a follow-up PR

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
+        ]
+        operator_bases = ['BaseOperator']
+        for module in import_from_nodes:
+            if any(regex.match(module.module) for regex in regex_patterns):
+                operator_bases.extend([name.name for name in module.names])
+
+        # Determine whether each class directly extends the BaseOperator or whether it
+        # must be further analyzed for indirect extension
+        for node in class_def_nodes:
+            if len(node.bases) == 0:
```
kevin-bates, line 153: Can `node.bases` be `None`?  If so, it might be safer to ask `if not node.bases:`

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
+            for node in classes_to_analyze:
+                if any(base.id in [op_class.name for op_class in operator_classes] for base in node.bases):
+                    # This class directly extends an Operator class defined in this file
+                    operator_classes.append(node)
+                    classes_to_analyze.remove(node)
+
+                    # The classes still present in classes_to_analyze must be
+                    # re-analyzed with the addition of the new Operator class
+                    analysis_incomplete = True
+                    continue
```
kevin-bates, line 179: Is the intention here to stop the `for` loop and go back to the `while` so that the update to the list (i.e., removal of the node) can occur?  If so, then I think this should be a `break`, otherwise the `continue` applies to the `for`.  (At least that's what I found in a simple side script.)

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
+            description = self._format_description(
+                description=description,
+                data_type=(data_type_parsed or data_type_from_ast)
+            )
+
+            # Standardize data type information
+            data_type_info = self.determine_type_information(data_type_parsed or data_type_from_ast)
             if data_type_info.undetermined:
-                self.log.debug(f"Data type from parsed data ('{data_type}') could not be determined. "
+                self.log.debug(f"Data type from parsed data ('{data_type_parsed}') could not be determined. "
```
kevin-bates, line 231: Since `data_type_info` _could_ come from `data_type_from_ast`, this should probably be: 
```python
('{data_type_parsed} or {data_type_from_ast}')
```

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +39,316 @@ class AirflowComponentParser(ComponentParser):
...
+                one_of_control_types=one_of_control_types,
+                allow_no_options=True,
+                required=arg_attributes.get('required')
+            )
+            properties.append(component_param)
 
         return properties
 
-    def get_class_def_as_string(self, content: Dict[str, List[str]]) -> str:
+    def _get_init_arguments(self, init_function: ast.FunctionDef) -> Dict[str, Dict]:
```
kevin-bates, line 260: This method (and perhaps others) is geared completely to `python` (IIUC) - really cool.  Although not necessary now, we may want to consider introducing a base class to this one like `PythonParser` where these kinds of functions can live.  This can definitely wait until there's another platform that uses python component definitions, but something to think about in the background. 

I noticed "pieces" like `xcom` that are specific to Airflow, but perhaps those kinds of things could be abstracted into an override like `specific_keywords()` that the leaf implementations provide - assuming we can generalize those uses in a meaningful way (if that makes sense?).  This would then enable some of the other methods to be moved to a python-specific base class, for example.

elyra/pipeline/airflow/component_parser_airflow.py
kiersten-stokes: Good point. I haven't seen a scenario where it's `None` yet, but better safe than sorry!

elyra/pipeline/airflow/component_parser_airflow.py
kiersten-stokes: That is a good point and good to start thinking about it now. Especially with BYO runtimes and the open issue we have to support Kubeflow python components (in addition to just YAML-based)

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -332,24 +339,38 @@ def _get_init_arguments(self, init_function: ast.FunctionDef) -> Dict[str, Dict]
...
                     elif (
-                            isinstance(arg.annotation.slice, ast.Tuple) and
-                            isinstance(arg.annotation.value, ast.Name)
+                        isinstance(arg.annotation.slice, (ast.Tuple, ast.Index)) and
+                        isinstance(arg.annotation.value, ast.Name) and
+                        arg.annotation.value.id != 'Optional'
                     ):
                         # arg is of the form `<arg>: <multi-valued_type>`
-                        # e.g. `env: Dict[str, str]` or `env: List[bool]`
+                        # e.g. `env: Dict[str, str]` or `env: List[bool]`**
```
kiersten-stokes, line 347: typo here

```suggestion
                        # e.g. `env: Dict[str, str]` or `env: List[bool]`
```

ptitzler:
> Do we want to/how should we handle cases where arguments are of 'non-standard' data types, e.g. the PythonOperator specifies an argument of the type 'python callable' (which will default to string)

A few thoughts about the `PythonOperator`:
- In a limited way the existing support for running Python scripts approximates the `PythonOperator` functionality. One exception is the lack of support for `xcom` data exchange.
- It _would_ be possible to provide  _rudimentary support_ if the VPE accepts a string as text input that is treated as if it defines a valid Python function, which is then embedded in the DAG. Usability would be poor though because 1) the input would have to be copied from somewhere (like a Python editor) and 2) "business logic" would be embedded in the pipeline file, which makes it harder to re-use.
- The `PythonOperator` is (roughly) equivalent to KFP Python-function based components, which Elyra currently does not support.

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +40,368 @@ class AirflowComponentParser(ComponentParser):
...
+                           f"could not be parsed: {e}. Skipping...")
+            return None
 
-        for component_class, component_content in component_classes.items():
-            # Create a Component object for each class
-            component_properties = self._parse_properties(component_content)
+        for component_class, content in parsed_class_nodes.items():
+            if not content.get('init_function'):
+                # Without the init function, class can't be parsed for properties
+                # TODO is this the correct functionality? see classes that derive from PythonOperator
```
kiersten-stokes, line 61: This comment can probably be removed. This case will be covered in a follow-up PR as noted in #2414

ptitzler (approved):
LGTM

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -46,149 +40,367 @@ class AirflowComponentParser(ComponentParser):
...
+        :returns: a filtered list of ast.ClassDef objects that can be considered 'Operators'
+        """
+        operator_classes = []
+        classes_to_analyze = []
+
+        # Get class names for package imports that match one of the following patterns,
+        # indicating that this class does match a known Operator class as defined in
+        # a provider package or core Airflow package
+        regex_patterns = [
+            re.compile(r'airflow\.providers\.[a-z_]+\.operators'),  # airflow.providers.*.operators. (provider package)
```
kevin-bates, line 137: Should this regex include upper-case as well as digits or does airflow restrict provider names to lower-case alpha?

elyra/tests/pipeline/airflow/test_component_parser_airflow.py
```diff
@@ -52,7 +53,8 @@ def test_component_catalog_can_load_components_from_registries(component_cache_i
     assert len(components) > 0
 
 
-def test_modify_component_catalogs():
+@pytest.mark.usefixtures('teardown_test_catalog')
+def test_modify_component_catalogs(teardown_test_catalog):
```
kevin-bates, line 57: This seems redundant.  We usually only use the parameter form for "declaring" a fixture.

conftest.py
```diff
@@ -68,3 +71,22 @@ def component_cache_instance(request):
...
+
+    # Run test
+    yield
+
+    # Remove test catalog
+    try:
+        if metadata_manager.get(TEST_CATALOG_NAME):
+            metadata_manager.remove(TEST_CATALOG_NAME)
+    except Exception:
+        pass
```
kevin-bates, line 92: This seems a little unusual in that it leads to this mysterious/missing declaration of `metadata_manager` in the tests.  Instead, I think naming this fixture something like `metadata_manager_with_teardown` and adding `metadata_manager` to the `yield` statement would then help clarify things.

The tests would then look something like:
```python
def test_modify_component_catalogs(metadata_manager_with_teardown):
   # do some stuff, build the metadata instance
   metadata_manager_with_teardown.create(TEST_CATALOG_NAME, registry_instance)
   # finish the test
```
Then at least there's _less_ mystery as to where the metadata manager instance came from.

elyra/pipeline/airflow/component_parser_airflow.py
```diff
@@ -201,3 +413,16 @@ def get_runtime_specific_properties(self) -> List[ComponentParameter]:
                 required=True,
             )
         ]
+
+    def _get_content_between_lines(self, start_line: int, end_line: int, content: str) -> str:
```
kevin-bates, line 417: Ah, reading between the lines I see.  That's good. :wink:

kevin-bates (commented):
This looks really good.  There's a general question/comment about how the "teardown" fixture is declared (and used), but everything else is fairly specific (and minor).

Note, I don't really have time to fully test this out but hoping others are able to spend time on this in that respect.

elyra/pipeline/airflow/component_parser_airflow.py
kiersten-stokes: Hmm good question. I'll look into it

conftest.py
kiersten-stokes: That makes sense! I'll re-work those

elyra/pipeline/airflow/component_parser_airflow.py
kiersten-stokes: I think we're safe sticking to all lower case since we're looking for the imported modules in this case ([justification](https://www.python.org/dev/peps/pep-0008/#package-and-module-names))

akchinSTC (approved):

elyra/pipeline/airflow/component_parser_airflow.py
kevin-bates: That's a fair point.  

I think digits should be included though.  Here are some of the packages in my current elyra env with digits: argon2, flake8, httplib2, jinja2, json5, s3contents,  urllib3

Hmm.  I also see two packages with uppercased characters: `Cryptodome` and `IPython` and here's an[ import statement referencing the latter](https://github.com/ipython/ipykernel/blob/6ea9018f9f7e6554181c52ef9610787c2ca05548/ipykernel/kernelbase.py#L37).  I'm sure this is rare, not to mention unconventional, but it seems like we'd be safer with:
```suggestion
            re.compile(r'airflow\.providers\.[a-zA-Z0-9_]+\.operators'),  # airflow.providers.*.operators. (provider package)
```

kiersten-stokes: Fair enough!

kevin-bates (approved):
Thank Kiersten - these changes look good.

PR #2429, opened 2022-02-01
Unpin click dependency

kiersten-stokes:
Our recent test failures seem to be due to a conflicting pin of `click` (we pin <8 whereas the latest `black` release pins `click` at >= 8). This PR removes the pin on `click` (imposed in #2010) as further research revealed that the KFP `click` dependency is only for use in the KFP CLI.

Tests are passing. Also ran some pipelines and all are passing.

 
           in the file; or

           it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ setup_args = dict(
     packages=find_packages(),
     install_requires=[
         'autopep8>=1.5.0,<1.5.6',
-        'click>=7.1.1,<8',  # Required by kfp 1.6.3
+        'click',
         'colorama',
         'entrypoints>=0.3',
         'jinja2>=2.11',
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ setup_args = dict(
     packages=find_packages(),
     install_requires=[
         'autopep8>=1.5.0,<1.5.6',
-        'click',
+        'click>=7.1.1',
         'colorama',
         'entrypoints>=0.3',
         'jinja2>=2.11',


ptitzler (approved):
LGTM

setup.py
```diff
@@ -72,7 +72,7 @@
     packages=find_packages(),
     install_requires=[
         'autopep8>=1.5.0,<1.5.6',
-        'click>=7.1.1,<8',  # Required by kfp 1.6.3
+        'click',
```
kevin-bates, line 75: I think we should leave the floor in.  Floors are much safer than caps. :smile:

kevin-bates (commented):
The cap on `jupyter_core` (5.0) should be removed as well.  We're close enough to that repo (I personally watch it) to be aware of any breaking major releases.

Other caps that reference non-existent releases are:
- `minio` (8.0) This cap was updated from 7 when a contribution required version 7, implying there was no reason for it to be capped at 7 in the first place.  The contributor added 8, I'm sure, due to precedence.  Btw, the same [PR](https://github.com/elyra-ai/elyra/pull/2335) that moved the `minio` cap to 8.0 also updated the `kfp` base to `1.7.0`, so the comment regarding `click` and `kfp==1.6.3` is moot anyway.
- `pyyaml` (6.0) - This cap looks like it [might be legit](https://github.com/yaml/pyyaml/issues/576) as 6.0 does (now) exist.  However, if we use `yaml.safe_load()` - which appears to be the case, then `safe_load()` calls `load()` with the `SafeLoader` loader - so all is good.
- `requests` (3.0) - Non-existent release.  However, their [docs](https://docs.python-requests.org/en/latest/community/release-process/#release-process-and-rules) indicate that major releases will contain breaking changes.  As a result, I think they view releases differently in that new features that are backward compatible will not be a major release. 

I understand not wanting to remove other caps because as soon as we do, it will probably bite us. But, on the other hand, I think this is the kind of thing that should be due to a real issue, not fear-based.  If folks run into issues due to new major releases, then a) the community is probably having the same issues, and b) it's a matter of downgrading the impacted package. (My two cents)

kevin-bates:
I'm okay only addressing the `click` issue right now.  I would like to see the floor restored since we have to assume there was some reason for it being there in the first place. 

The lack of consistency with caps (or not) is what really bugs me.  It seems completely arbitrary for those cases that use a cap.

ptitzler:
> I'm okay only addressing the `click` issue right now. I would like to see the floor restored since we have to assume there was some reason for it being there in the first place.

I agree. The purpose of this PR should only be to resolve the failing tests. This unblocks other PRs that used to pass validation just fine.

kevin-bates:
I'm going ahead with this merge so that tests will start passing (or not be side-affected by this).

PR #2433, opened 2022-02-02
Add GUI support for metadata instance cloning

ptitzler:
Closes #2311 

### What changes were proposed in this pull request?

Add a new clone/copy button to the metadata editor, which duplicates the selected metadata instance and saves it using display name `Copy of ...` and file name `copy_of_<existing_instance>`

Once upon a time there was a component catalog  ...
image

First clone ...
image

Many clones later ...
image

If an attempt is made to clone a metadata instance and the clone name is already taken, an error is raised. For example

   image

yields

image

The new feature description was added to the relevant documentation topics.

### How was this pull request tested?

- Reviewed the output of `make docs`
- Cloned an existing
  - runtime configuration entry
  - component catalog entry
  - runtime image entry
  - code snippet entry
- Compared the original metadata file with a cloned metadata file, e.g.
   ```
   $ diff metadata/component-catalogs/copy_of_copy_of_copy_of_copy_of_hash.json metadata/component-catalogs/copy_of_copy_of_hash.json
   2c2
   <   "display_name": "Copy of Copy of Copy of Copy of hash",
   ---
   >   "display_name": "Copy of Copy of hash",
 
   ```
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #2436, opened 2022-02-02
Ad support for metadata duplication

### LINKED ISSUE #2311: Support cloning of metadata instances in the GUI
Originally suggested by @karlaspuldaro. Currently users cannot clone existing metadata instances in the GUI. 

Example scenario:
 - user has an existing runtime configuration
 - user wants to create a similar second runtime configuration without the need to manually re-enter _all_ information (host, port, ...)

- User invokes "clone" task in the metadata side panel (showing list of all instances).
- An identical copy of the source metadata is created, differing only in name, e.g. "`Copy of <original-name>`".
- User opens "Copy of <original-name>" (just like today) and changes properties as desired.

ptitzler:
Closes #2311 

Add a new duplicate button to the metadata editor, which duplicates the selected metadata instance and saves it using display name `Copy of ...` and file name `copy_of_<existing_instance>`

Once upon a time there was a component catalog  ...
image

First duplicate...
image

Many duplicates later ...
image

If an attempt is made to duplicate a metadata instance and the duplicate name is already taken, an error is raised. For example

   image

yields

image

The new feature description was added to the relevant documentation topics.

- Reviewed the output of `make docs`
- Duplicated an existing
  - runtime configuration entry
  - component catalog entry
  - runtime image entry
  - code snippet entry
- Compared the original metadata file with a duplicated metadata file, e.g.
   ```
   $ diff metadata/component-catalogs/copy_of_copy_of_copy_of_copy_of_hash.json metadata/component-catalogs/copy_of_copy_of_hash.json
   2c2
   <   "display_name": "Copy of Copy of Copy of Copy of hash",
   ---
   >   "display_name": "Copy of Copy of hash",
 
   ```
 
           in the file; or

           it.


ptitzler:
This PR replaces https://github.com/elyra-ai/elyra/pull/2433, which had issues.

kevin-bates:
Could we instead adopt Lab's duplicate-name functionality where a `-CopyN` is appended to the source name?
image

They also have logic such that an attempt to duplicate a "copy" results in the next integer on the copy sequence.  For example `Untitled10-Copy4.ipynb` was produced via a duplicate action on `Untitled10-Copy1.ipynb`, while the other 3 copies were duplicates of `Untitled10.ipynb`.

This would resolve the `Copy of Copy of ...` chain, as well as the attempt to make multiple copies from the same source - which I would view as a requirement.

ptitzler:
> Could we instead adopt Lab's duplicate-name functionality where a `-CopyN` is appended to the source name? image

The suggested approach would result in more work because the code would have to retrieve all instances and analyze their metadata to identify the next available identifier. Unless, of course, it would keep trying to save a duplicate until the backend operation succeeds.  I'm already a bit weary that the current code has built-in awareness of how the metadata backend chooses the 'file' name... plenty of opportunities to break in the future if changes are made.

> This would resolve the `Copy of Copy of ...` chain, as well as the attempt to make multiple copies from the same source - which I would view as a requirement.

I don't believe a user would want to retain the generated name or use the duplicated copies as is. Therefore, after duplication, it's likely that  the cloned metadata properties are customized and the 'Copy of' goes away.

ptitzler:
> Could we instead adopt Lab's duplicate-name functionality where a -CopyN is appended to the source name?

Done!

akchinSTC (approved):
Tested locally and with snippets, runtime images, runtimes and catalogs. ✔️

packages/code-snippet/src/CodeSnippetService.ts
```diff
@@ -70,4 +70,44 @@ export class CodeSnippetService {
...
+
+  /**
+   *  Duplicates an existing code snippet, which is named
+   *  'Copy of <original-name>'.
+   *
+   * @param codeSnippet: code snippet to be duplicated
+   *
+   * @returns A promise
+   */
+  static duplicateCodeSnippet(
```
ajbozarth, line 82: Given this function is exactly the same as the the code in `MetadataWidget.tsx`, I think it would make more sense to put the util function in `metadata-common` and call it from both places like you called it here in code snippets

ajbozarth (changes requested):
Overall the code looks good, just one opinion on handling duplicate code better. I didn't look through the doc changes though.

Tested locally and behavior works as expected as well.

packages/code-snippet/src/CodeSnippetService.ts
ptitzler: Done.

ajbozarth (approved):
The update to the code LGTM, didn't re-test them locally though

karlaspuldaro (commented):
This looks great, thx @ptitzler !
A non-blocker suggestion that can be done in a follow up, it would be helpful to add integration test for this new functionality.

PR #2449, opened 2022-02-04
Sort runtime types alphabetically, to enforce consistent order is displayed in launcher

### LINKED ISSUE #2423: Airflow | Kubeflow Pipelines tiles are displayed in unpredictable order in the launcher
**Describe the issue**

The order in which the two tiles is displayed appears to be somewhat random.

image

image

Sometimes KFP is displayed first, sometimes AA. Because the behavior is not predictable I sometimes click on the "wrong" tile. For example, 9/10 of the times the AA  VPE tile is third one in the list.

**To Reproduce**
1. Open JL
2. Inspect the 'Elyra' group in the launcher window

**Deployment information**
 - Elyra version: 3.5/master

samali0996:
fixes #2423 
Since the order of the runtimes returned by the api call `elyra/pipeline/runtimes/types` can be unpredictable, I sort the array `res.runtime_types` alphabetically by the id of each runtime.

This will enforce order of the runtime types, and local will still be listed as the first type in the launcher since it is given a rank of 1. But now the remaining runtime types will always be in the same order (apache, kube).

This ensures that the list of returned runtime types will always be in the same order (local, apache, kube)

Testing by seeing if the resulting code I added did change the order, if I could reverse the order with the code (I was able to ie Kube first then Apache). 

Created testing that will test Pipeline Service component and validate that the expected runtimes were returned and in the correct order

Clicking on the icon will then create the appropriate workflow with the intended runtime chosen

 
           in the file; or

           it.


samali0996:
I created a new test file in `pipeline-editor` package, if this is unnecessary I can remove it and just keep the sorting of the runtime types to enforce consistent order

lresende:
How is this influencing the launcher order which is coming [from this code](https://github.com/elyra-ai/elyra/blob/master/packages/theme/src/launcher.tsx)?

samali0996:
> How is this influencing the launcher order which is coming [from this code](https://github.com/elyra-ai/elyra/blob/master/packages/theme/src/launcher.tsx)?

So from what I understood, in [pipeline-editor/src/index.ts](https://github.com/elyra-ai/elyra/blob/a0cc77c5c5121b98147e6d301d349244be826a79/packages/pipeline-editor/src/index.ts#L259) , this is where the resolved runtime types are added to the launcher as commands

```
 PipelineService.getRuntimeTypes()
      .then(async types => {
        const promises = types.map(async t => {
          return {
            ...t,
            icon: await createRemoteIcon({
              name: `elyra:platform:${t.id}`,
              url: t.icon
            })
          };
        });

        const resolvedTypes = await Promise.all(promises);

        // Add the command to the launcher
        if (launcher) {
          for (const t of resolvedTypes as any) {
            launcher.add({
              command: openPipelineEditorCommand,
              category: 'Elyra',
              args: { runtimeType: t },
              rank: t.id === 'LOCAL' ? 1 : 2
            });
```
The `resolvedTypes` object is returned from the `PipelineService.getRuntimeTypes()` function call which is what makes the API call to the backend.

The order of the array that `resolvedTypes` returns matters, the runtime object with id `Local` will be given a rank of 1 so it will display first, but the others (apache, kube) are given a rank of 2 so the way they will be displayed relates to the order that they are added to the launder.

This is why by enforcing the order that `PipelineService.getRuntimeTypes()` returns affects the order that the runtimes are displayed on the launcher page

samali0996:
And just to show that the order does actually affect how it will show up on the launcher page,

`return res.runtime_types.sort((a: any, b: any) => a.id.localeCompare(b.id));`

image

`return res.runtime_types.sort((a: any, b: any) => b.id.localeCompare(a.id));`

image

karlaspuldaro (changes requested):
Thank you @samali0996 !
Tested locally and it works as expected. Since `getRuntimeTypes` is also called on pipeline submission/export, I also checked the order displayed in the dialog dropdowns are not impacted as [this code](https://github.com/elyra-ai/elyra/blob/11bc6811c213bab6d3e32ddf90d8dd13a7af58e8/packages/pipeline-editor/src/RuntimeConfigSelect.tsx#L36) adds `local` config to the beginning of the array.

My only suggestion is to call the new test file `pipeline-service.spec.ts`, @ajbozarth do you agree?

lresende (approved):

Issue #2269, opened 2021-11-03
Add support for const types in metadata editor


kevin-bates:
The runtime types PR (#2263) introduces a `runtime_type` property to each schema within the Runtimes schemaspace.  This property denotes the type of runtime to which instances of the schema correspond and uses a `const` meta-property to hold this value - since the hosting schema is one-to-one with a runtime processor implementation (and, therefore, a type).  Here's the schema entry relative to the `kfp` schema:
```JSON
       "runtime_type": {
          "title": "Runtime Type",
          "description": "The runtime associated with this instance",
          "type": "string",
          "const": "KUBEFLOW_PIPELINES"
        },
```

Left unaddressed, this will present a `Runtime Type` property with an empty edit control.  If any value other than the constant value is entered, the validation for that instance will fail.   Since `runtime_type` is not required, the only "working" approach would be to never enter any value (or know the constant value to enter). Because this is a constant, the user should not need to enter a value or update previously persisted constant values.

For `const` types, the metadata editor should unconditionally set the value of the property to the value stored in the `const` meta-property treat the property as read-only.

The idea that constant-based values should perhaps be hidden is also applicable here.  As a result, it might be good to offer a `uihints` "style" that denotes a particular property be completely hidden from the end-user.

ptitzler:
How about introducing a UI property that identifies special attributes, such as 'hidden' (immutable and not rendered) or 'read-only' (immutable)?

kevin-bates:
I think this a lot of sense and is used in other metadata-based systems I've worked on in the past.

lresende:
Based on the conversation above, do we still need this?

kevin-bates:
Yes, the discussion above is essentially regarding the two forms for how we want to _display_ constant-based properties.  We should still transfer the value reflected in the `const` field to the metadata instance.  That transfer isn't really necessary if the `uihint` were to indicate a 'hidden attribute, but is for 'read-only' and 'visible'.

akchinSTC:
See: https://github.com/elyra-ai/elyra/pull/2263#pullrequestreview-803006967

akchinSTC:
@marthacryan - check to see if this has been resolved earlier release

Issue #2268, opened 2021-11-03
Add support for enum UI hints in the metadata editor


ptitzler:
**Is your feature request related to a problem? Please describe.**

**TL;DR**

The metadata editor currently renders enums as HTML select tags, using enum member names as "OPTION_VALUE" and "OPTION_LABEL" in the HTML example below

```
<select name="dogs" id="dogs>
  <option value="OPTION_VALUE">OPTION_LABEL</option>
</select>
```

Since enum member names are typically not user friendly, it should be possible to define in the schema what to use as "OPTION_VALUE" for each member.

```
<select name="dogs" id="dogs">
  <option value="ENUM-MEMBER-NAME">ENUM-MEMBER-VALUE</option>
</select>
```

**Description**

This is a follow-up from a discussion in https://github.com/elyra-ai/elyra/pull/2257.

> On the one hand, we'd like to present users with nice displayable (potentially localizable) values, while, on the other, introduce a level of indirection for the _persisted_ values that essentially remain fixed and are not privy to the whims of others inherent in human display values.  This is exactly what is driving my previous _`name_value_map`_ comment.

> To elaborate more on the proposed `uihints`, I think this schema stanza would resemble something like the following:
```JSON
        "auth_type": {
          "title": "Authentication Type",
          "description": "Authentication type Elyra uses to authenticate with Kubeflow",
          "type": "string",
          "enum": [
            "DEX_LEGACY",
            "DEX_LDAP_CONNECTOR",
            "DEX_STATIC_PASSWORDS",
            "K8S_SERVICEACCOUNT_TOKEN",
            "NO_AUTHENTICATION"
          ],
          "default": "No authentication",
          "uihints": {
            "field_type": "dropdown",
            "category": "Kubeflow Pipelines",
            "name_value_map": {
                "NO_AUTHENTICATION": "No authentication",
                "DEX_STATIC_PASSWORDS": "DEX with static passwords",
                "K8S_SERVICEACCOUNT_TOKEN": "Kubernetes service account token",
                "DEX_LEGACY": "DEX (generic)",
                "DEX_LDAP_CONNECTOR": "DEX with LDAP"
            }
          }
        },
```
> The metadata-editor would be responsible for converting schema `enum` values to their displayable value found in the map **and** setting the field to the selected _name_ (constant-looking string) value in order to adhere to the constraints imposed by the _enum_ (so schema validation succeeds).

[...]

PR #2464, opened 2022-02-08
Update metadata editor to use JupyterLab's new form editor

marthacryan:
This PR adds the package [rjsf](https://react-jsonschema-form.readthedocs.io/)(react JSON schema form - a renderer for showing JSON schema as a form) to the metadata editor for consistency with JupyterLab (moved to using this package in the most recent release with jupyterlab/jupyterlab#11079. Going forward, this will also handle many of the more complex cases for rendering the metadata editor and provide built-in features that will solves issues such as #2269 and #2268.

Screenshots / gifs to come

**Todo:**
- [x] Still need to address validation failing on some categories and preventing save and close
~~- [ ] Update documentation~~ Follow up PR to come
- [x] Clean up look and feel with CSS

           in the file; or

           it.


packages/metadata-common/src/MetadataEditor.tsx
```diff
@@ -31,306 +24,434 @@ import {
...
     if (this.dirty) {
       showDialog({
         title: 'Close without saving?',
-        body: (
-          <p>
-            {' '}
-            {`"${this.displayName}" has unsaved changes, close without saving?`}{' '}
-          </p>
-        ),
+        body: <p>Metadata has unsaved changes, close without saving?</p>,
```
karlaspuldaro, line 454: Maybe we should keep the reference to the metadata display name here

karlaspuldaro (commented):
This looks great! Only have a few comments/questions.

- Not introduced by this PR, but we should probably take a look at the warnings from Code QL when using `setState` in `MetadataEditorTags.tsx`:
```
React component state updates using setState may asynchronously update this.props and this.state, thus it is not safe to use either of the two when calculating the new state passed to setState.

Recommendation
Use the callback-based variant of setState: instead of calculating the new state directly and passing it to setState, pass a callback function that calculates the new state when the update is about to be performed.
```
- `MetadataEditor.tsx` file is pretty long, I wonder if it would make sense to move the widget code into its own file?

- In schemas updates I'm not sure why sometimes we completely remove `field_type: dropdown` even when it is an enum or a string array of more than one item, wouldn't those need to be updated to `ui:field: dropdown` ?

karlaspuldaro:
A few comments on the UI, design can be further discussed with the team
image

marthacryan:
Thanks for the review @karlaspuldaro!

> * Not introduced by this PR, but we should probably take a look at the warnings from Code QL when using `setState` in `MetadataEditorTags.tsx`:
> 
> ```
> React component state updates using setState may asynchronously update this.props and this.state, thus it is not safe to use either of the two when calculating the new state passed to setState.
> 
> Recommendation
> Use the callback-based variant of setState: instead of calculating the new state directly and passing it to setState, pass a callback function that calculates the new state when the update is about to be performed.
> ```

I'll try to fix this in this PR but we could also make an issue if it ends up being complicated. 

> * `MetadataEditor.tsx` file is pretty long, I wonder if it would make sense to move the widget code into its own file?

Good idea! It is a huge file, I'll move the widget into its own file. 

> * In schemas updates I'm not sure why sometimes we completely remove `field_type: dropdown` even when it is an enum or a string array of more than one item, wouldn't those need to be updated to `ui:field: dropdown` ?

The form editor automatically detects if there is an `enum` field defined in the schema and uses the dropdown component if so - so we no longer needed that UI hint because it's already figuring it out for us.

packages/ui-components/src/FormComponents/CodeBlock.tsx
```diff
@@ -0,0 +1,78 @@
...
+      editorRef.current?.model.value.changed.connect(handleChange);
+    }
+
+    return (): void => {
+      editorRef.current?.model.value.changed.disconnect(handleChange);
+    };
+    // NOTE: The parent component is unstable so props change frequently causing
+    // new editors to be created unnecessarily. This effect on mount should only
+    // run on mount. Keep in mind this could have side effects, for example if
+    // the `onChange` callback actually does change.
```
ajbozarth, line 57: out of curiosity would it be worth adding a partial list of dependencies here for fields that we don't expect to change but we'd want to trigger an update? (ie fields not touched by the unstable parent component)

packages/ui-components/src/FormComponents/MetadataEditorTags.tsx
```diff
@@ -197,63 +198,79 @@ export class MetadataEditorTags extends React.Component<
         </button>
       );
     return (
```
ajbozarth, line 200: We should address the codeQL warnings above, I think they can be addressed by some refactoring to use `useEffect` or similar

packages/ui-components/src/icons.tsx
```diff
@@ -99,6 +101,14 @@ export const pyIcon = new LabIcon({
...
   svgstr: pyIconSvg
 });
+export const viewIcon = new LabIcon({
+  name: 'elyra:view',
+  svgstr: viewIconSvg
+});
+export const viewOffIcon = new LabIcon({
+  name: 'elyra:viewOff',
+  svgstr: viewOffIconSvg
+});
```
ajbozarth, line 111: what lib did we get these icon from? do we need to add a license reference for them?

packages/metadata-common/package.json
```diff
@@ -43,6 +43,7 @@
     "@lumino/messaging": "^1.4.3",
     "@lumino/widgets": "^1.19.0",
     "@material-ui/core": "^4.12.1",
+    "@rjsf/core": "^3.1.0",
```
ajbozarth, line 46: this is only added here but should also be in ui-components since it is referenced by the components as well. Though if we move the FormEditor to ui-components it would only need to be referenced there

ajbozarth (commented):
Finally found time for a full code review. Mostly just nits and questions, nothing truly blocking so no requests for changes or approval for now.

I'll be testing it locally on another day, though on my initial start I seem to not be able to open the editor, with jupyterlab hanging with the Loading text and no errors in the console or log

elyra/metadata/schemas/kfp.json
```diff
@@ -153,7 +150,8 @@
           "description": "Kubernetes secret that's defined in the specified user namespace, containing the Cloud Object Storage username and password. This property is required for authentication type KUBERNETES_SECRET.",
           "type": "string",
           "uihints": {
-            "secure": true,
+            "ui:field": "password",
+            "minLength": 8,
```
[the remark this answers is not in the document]
marthacryan: Oh weird

packages/metadata-common/src/FormEditor.tsx
```diff
@@ -0,0 +1,189 @@
...
+    </div>
+  );
+};
+
+/**
+ * React component that wraps the RJSF form editor component.
+ * Creates a uiSchema from given uihints and passes relevant information
+ * to the custom renderers.
+ */
+export const FormEditor: React.FC<IFormEditorProps> = ({
```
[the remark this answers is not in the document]
marthacryan: Good point, then I can move all rjsf code to ui-components. Moved it but I have to move the css over now

packages/ui-components/src/icons.tsx
marthacryan: It comes from carbon components (I think that's where we've gotten our other icons so it should be fine?)

packages/metadata-common/src/MetadataEditorWidget.tsx
```diff
@@ -0,0 +1,337 @@
...
+          title: ' ',
+          properties: {
+            display_name: {
+              title: 'Display Name',
+              description: 'Name used to identify an instance of metadata.',
+              type: 'string'
+            }
+          },
+          required: ['display_name']
+        }
```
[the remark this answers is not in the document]
marthacryan: I actually added translation for this title / description since it's hard coded - does that work?

packages/metadata/src/index.ts
```diff
@@ -44,17 +55,30 @@ const commandIDs = {
...
+    ITranslator
+  ],
   optional: [IThemeManager],
   activate: async (
     app: JupyterFrontEnd,
     palette: ICommandPalette,
     editorServices: IEditorServices,
     status: ILabStatus,
+    componentRegistry: IFormComponentRegistry,
+    translator: ITranslator,
```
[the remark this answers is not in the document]
marthacryan: Added some more - not sure if it's covering every possible case but it's mostly there for the editor part

akchinSTC:
@marthacryan - played around with the combining the regex patterns in the kfp.json schema. 
Let me know if this works.
```
"pattern": "^[^ ]+( +[^ ]+)*$"
```

packages/ui-components/package.json
```diff
@@ -40,7 +40,8 @@
     "@material-ui/lab": "^4.0.0-alpha.18",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "react-json-tree": "^0.11.2"
+    "react-json-tree": "^0.11.2",
+    "@rjsf/core": "^3.1.0"
```
ajbozarth, line 44: I believe this should be above react (the @ is sorted first)

packages/ui-components/src/FormComponents/MetadataEditorTags.tsx
```diff
@@ -137,130 +106,121 @@ export class MetadataEditorTags extends React.Component<
...
+      <li className={METADATA_EDITOR_TAG_LIST}>
+        {hasTags
+          ? tags.map((tag: string, index: number) =>
+              ((): JSX.Element => {
+                if (!selectedTags) {
+                  return (
+                    <button
+                      onClick={handleClick}
+                      className={`${METADATA_EDITOR_TAG} tag unapplied-tag`}
+                      id={'editor' + '-' + tag + '-' + index}
```
ajbozarth, line 177: why not use formatted strings for the strings in the return block (instead of using +)

packages/ui-components/style/formeditor.css
```diff
@@ -0,0 +1,284 @@
...
+.elyra-editor-tag.tag.unapplied-tag input {
+  border: none;
+}
+
+.elyra-formEditor .elyra-metadataEditor-arrayInput li {
+  padding-left: 0;
+  padding-bottom: 0;
+}
+
+.elyra-metadataEditor
```
ajbozarth, line 195: we shouldn't have css specific to metadataEditior in ui-components, so either this should move back, or if it's actually generic to the form editor we should update the css class name (this applies to all the references in this file)

ajbozarth (changes requested):
a few comments on the code changes since my last review, and I'm building this to test locally atm

ajbozarth:
UI/UX Review as mentioned above, all tested in Safari:

- The mouseover description shows when mousing over both the title and input area for a filed as well as if you mouse over the description itself. This hides the title for the field below. Unless there's a reason to read it while filling out the field, or to copy/paste the descriptions contents I think a better UX would be to only show when mousing over the title. (@ptitzler option on this?) (it can be seen in the video in bullet 3)
- Show hide icon css not quite right
Screen Shot 2022-05-23 at 4 09 49 PM
- Form spacing changes on initial mouseover, see video below:

https://user-images.githubusercontent.com/13952758/169907212-566a1a35-5647-4cdb-b899-ebe4c24042d8.mov

- some escaping is going wrong here:
Screen Shot 2022-05-23 at 4 19 03 PM
- description tooltips are misaligned for code snippet fields, also there is no description for tags (dont know if that's intentional) 
Screen Shot 2022-05-23 at 4 20 43 PM
Screen Shot 2022-05-23 at 4 20 47 PM
- css for this field looks off 
Screen Shot 2022-05-23 at 4 22 13 PM
- there's no longer button highlighting when mousing over the save button like before 

That's all I found on an initial pass, I only played with it for less than an hour

marthacryan:
> * The mouseover description shows when mousing over both the title and input area for a filed as well as if you mouse over the description itself. This hides the title for the field below. Unless there's a reason to read it while filling out the field, or to copy/paste the descriptions contents I think a better UX would be to only show when mousing over the title. (@ptitzler option on this?) (it can be seen in the video in bullet 3)

Yeah this is the current behavior so I figured it was the UX we wanted, but I could see that being a little annoying. If @ptitzler has an opinion let us know!

> * Show hide icon css not quite right

Fixed

> * Form spacing changes on initial mouseover, see video below:

I'm not able to reproduce this? Maybe we can do a call to debug

> * some escaping is going wrong here:

Fixed

> * description tooltips are misaligned for code snippet fields, also there is no description for tags (dont know if that's intentional)

Fixed misalignment of tooltips and added tooltip to code field. I don't think tags already had that so not sure if I should add that in? 

> * css for this field looks off

Fixed the spacing up a little, let me know if you have more input on the style

> * there's no longer button highlighting when mousing over the save button like before

Fixed

ptitzler:
It appears that the required annotation (*) isn't displayed for all required properties, as seen in this code snippet example. Note the discrepancy between the validation messages and the property annotations:

image

ptitzler:
In various forms the label styling is inconsistent. For example (small, regular font vs bigger, bold font): 
image

ptitzler:
It appears that for list-valued properties there are two new buttons displayed (move up, move down) that appear to give the user the ability to reorder list items.

image

Is this something that could be disabled? While there might be general use cases where order matters, I can't think of one for any of our properties. 

Part of the problem is that the list-specific action buttons are rendered in a way that feels too attention grabbing, considering that they don't serve a practical purpose.

image

The old/existing rendering (shown here for a list-valued pipeline property) of list-specific blends in nicely and doesn't require additional screen real estate:

image

ptitzler:
Ageneral observation that the current styling appears to use almost always bold fonts (section headers, property labels, ...) as shown here for runtime configurations:

image

As a result everything seems to blend in, eliminating the benefits, for example, of section headers that were meant to help the user put labels into proper context. For comparison, here's the current rendering, which is a lot less attention grabbing:

image

ajbozarth (approved):
This LGTM, I gave @marthacryan a couple UI tweaks to follow up on over a Webex that I would consider non-blocking

PR #2468, opened 2022-02-09
Replace invalid characters in airflow parent node references

### LINKED ISSUE #2459: Airflow: Data exchange does not work if node labels contain special characters
**Describe the issue**

Pipeline:
image

Generated DAG
image

**To Reproduce**
1. Create an Airflow pipeline
2. Add a custom operator op1  to the pipeline, specifying a node label that contains a space (or other reserved character)
3. Add a second custom operator op2 to the pipeline
4. Connect the two nodes and configure data exchange between the nodes, using op1's output as input source
5. Export the DAG
6. Review the DAG
7. Note the `invalid task_id`

**Expected behavior**
Referenced task id is valid

**Deployment information**
 - Elyra version: 3.6/rc0

akchinSTC:
Fixes #2459

Removes and replaces all special characters in node labels so downstream components can reference them correctly.

- Unit tests updated with lists of valid and invalid characters

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -220,7 +220,12 @@ be fully qualified (i.e., prefixed with their package names).
 
         PipelineProcessor._propagate_operation_inputs_outputs(pipeline, sorted_operations)
 
-        for operation in sorted_operations:
+        # Scrub all node labels of invalid characters
+        scrubbed_operations = self._scrub_invalid_characters_from_list(sorted_operations)
+        # Generate unique names for all operations
+        unique_operations = self._create_unique_node_names(scrubbed_operations, None)
+
+        for operation in unique_operations:
 
             if isinstance(operation, GenericOperation):
                 operation_artifact_archive = self._get_dependency_archive_name(operation)
@@ -264,10 +269,7 @@ be fully qualified (i.e., prefixed with their package names).
                                                inputs=operation.inputs,
                                                outputs=operation.outputs)
 
-                unique_operation_name = self._get_unique_operation_name(operation_name=operation.name,
-                                                                        operation_list=target_ops)
-
-                target_op = {'notebook': unique_operation_name,
+                target_op = {'notebook': operation.name,
                              'id': operation.id,
                              'argument_list': bootscript.container_cmd,
                              'runtime_image': operation.runtime_image,
@@ -330,8 +332,7 @@ be fully qualified (i.e., prefixed with their package names).
                             set(json.loads(json.dumps(property_value)).keys()) == {'value', 'option'}:
                         parent_node_name = self._get_node_name(target_ops,
                                                                json.loads(json.dumps(property_value))['value'])
-                        processed_parent_node_name = self._scrub_invalid_characters(parent_node_name)
-                        processed_value = "\"{{ ti.xcom_pull(task_ids='" + processed_parent_node_name + "') }}\""
+                        processed_value = "\"{{ ti.xcom_pull(task_ids='" + parent_node_name + "') }}\""
                         operation.component_params[component_property.ref] = processed_value
                     elif component_property.data_type == "boolean":
                         operation.component_params[component_property.ref] = property_value
@@ -349,9 +350,6 @@ be fully qualified (i.e., prefixed with their package names).
                 # Remove inputs and outputs from params dict until support for data exchange is provided
                 operation.component_params_as_dict.pop("inputs")
                 operation.component_params_as_dict.pop("outputs")
-
-                unique_operation_name = self._get_unique_operation_name(operation_name=operation.name,
-                                                                        operation_list=target_ops)
 
                 # Locate the import statement. If not found raise...
                 import_stmts = []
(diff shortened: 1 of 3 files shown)


elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -483,6 +483,10 @@ def _get_node_name(self, operations_list: list, node_id: str) -> str:
             if operation['id'] == node_id:
                 return operation['notebook']
 
+    def _scrub_invalid_characters(self, node_name) -> str:
```
ptitzler, line 486:
```suggestion
    def _scrub_invalid_characters(self, node_name: str) -> str:
```

ptitzler (changes requested):
The current approach does not appear to take into account that the node id can be different from the label, as shown in the example below:

image

Note the `task_id` for the two nodes labeled `this!bash`. During pipeline processing, the ambiguity is resolved (to `this_bash_2`), yet the parent reference does not use that resolved task_id.

ptitzler:
Pipeline + exported DAG: [conflict.tar.gz](https://github.com/elyra-ai/elyra/files/8036920/conflict.tar.gz)

ptitzler:
Another issue - unrelated to the fix is that we need to generate unique task ids that take the n:1 mapping to `_` into account:
- task 1: `this bash` -> `this_bash`
- task 2: `this!bash` -> `this_bash_2`
- task 3: `this!bash` -> `this_bash_3`
- task n: `this?bash` -> `this_bash_n`

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -442,19 +440,38 @@ def create_pipeline_file(self, pipeline, pipeline_export_format, pipeline_export
...
         return pipeline_export_path
 
-    def _get_unique_operation_name(self, operation_name: str, operation_list: list) -> str:
-        unique_name_counter = 1
-        unique_operation_name = operation_name
-        names = [op['notebook'] for op in operation_list]
-        while unique_operation_name in names:
-            unique_name_counter += 1
-            unique_operation_name = ''.join([operation_name, '_', str(unique_name_counter)])
+    def _create_unique_node_names(self, operation_list, parent_unique_names=None):
```
kevin-bates, line 448: I think type hints would be useful here:
```suggestion
    def _create_unique_node_names(self, operation_list: List[Operation], parent_unique_names: Optional[Dict[str, int]] = None):
```

I'm not clear what should be done about the return value.  The way things stand is the return value is only used by the original caller and not recursive call - which is odd.

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -442,19 +440,38 @@ def create_pipeline_file(self, pipeline, pipeline_export_format, pipeline_export
...
+        for operation in operation_list:
+            # if a duplicate name is found
+            if operation.name in unique_names:
+                unique_names[operation.name] += 1
+                new_name = f"{operation.name}_{unique_names[operation.name]}"
+                if new_name in unique_names:  # look-ahead
+                    self._create_unique_node_names([operation], unique_names)
+                else:
+                    operation.name = new_name
+                    unique_names[operation.name] = 1
```
kevin-bates, line 459: - The recursion doesn't feel right here - both in the return value behavior (previous comment), and also due to the fact that it will always be called with a single-item list.
- The "look-ahead" comment is more like "look-behind" because it only considers past names
- Perhaps a `while not unique` mechanism may be easier to follow...

```python
            if operation.name in unique_names:
                new_name = operation.name
                while new_name in unique_names:
                    new_name = f"{operation.name}_{unique_names[operation.name]}"
                    unique_names[operation.name] += 1
                operation.name = new_name
```
- I find the implicit `_1` suffix a bit counterintuitive - should the seed value be `0`? (The above example will use `_1` for the first duplicate w/o having to change the seed value from `1`.)
- I think this statement `unique_names[operation.name] = 1` could be unconditional since it must be called regardless of the route.
- Eliminating the recursion eliminates the need for the `parent_unique_names` parameter and `unique_names` could start out as an empty dict.

kevin-bates (commented):
Thanks Alan - had a few comments, mostly surrounding maintainability.

elyra/pipeline/airflow/processor_airflow.py
akchinSTC: Thx for the review!
- Updated to use non recursive method
- duplicate name seeding to begin at index 1
- type hinting updated and fixed for some others
- tests updated

elyra/pipeline/airflow/processor_airflow.py
```diff
@@ -445,30 +446,28 @@ def create_pipeline_file(self, pipeline, pipeline_export_format, pipeline_export
 
         return pipeline_export_path
 
-    def _create_unique_node_names(self, operation_list, parent_unique_names=None):
-        unique_names = parent_unique_names or {}
+    def _create_unique_node_names(self, operation_list: List[Operation]) -> List[Operation]:
+        unique_names = {}
         for operation in operation_list:
             # if a duplicate name is found
             if operation.name in unique_names:
```
kevin-bates, line 453: With the last simplification, this `if` could be removed.  This would save a "contains" check on each iteration, but probably not a big deal.  The comment could probably be updated to something like `# Ensure operation name is unique` which would then apply to the block.

akchinSTC: done!

kiersten-stokes (approved):
LGTM! Tested a few scenarios (AKA stole Patrick's ideas 😄) and everything is checking out.

ptitzler (approved):
LGTM

kevin-bates (approved):
Sorry for the delay - thanks Alan!

PR #2492, opened 2022-02-18
Support well-defined ComponentDefinition return values for catalog connectors

### LINKED ISSUE #2460: Update the catalog connector's 'read_catalog_entry' signature to accept optional metadata

The catalog connector's interface for the [`read_catalog_entry` method](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/catalog_connector.py#L100) currently expects the connector to return a string that contains one or more  component definitions:

```
    def read_catalog_entry(self,
                           catalog_entry_data: Dict[str, Any],
                           catalog_metadata: Dict[str, Any]) -> Optional[str]:
```

For Apache Airflow (and perhaps other runtimes in the future) it should be possible for the connector to also return well-defined metadata, which aids with the processing of the component definitions. For example, we've discovered that for Airflow the connector needs to also return metadata that identifies the package that the operator class is defined in. This avoids the need for the user to manually annotate each operator class.

kiersten-stokes:
Fixes #2460
Supports #2488

Defines a new `EntryData` object class that includes the definition string as well as any runtime-specific attributes (e.g. `package_name` for `AirflowEntryData`). A new function, `get_entry_data` is defined, which takes the place of `read_catalog_entry`, and returns an `EntryData` object. If no `get_entry_data` function is present for a given CCC, the return value of `read_catalog_entry` is manually coerced into a `EntryData` object. 

A new `CatalogEntry` object is also defined, which holds in its bag of properties the `EntryData` object above, a unique id, information on the source of the component, and some catalog information (such as the relevant categories). This object is constructed after the `EntryData` object is returned by the CCC author and is passed directly to the parse methods, eliminating the need to define a `SimpleNamespace` object as a middleman.

There have also been some minor changes to the `Component` object, namely the inclusion of an optional `package_name` attribute and changing the same of the `source_identifier` attribute to `component_reference`.

During processing of Airflow components, the `package_name` for that component is now accessed directly (using the `import_statement` property) if it is available. This eliminates the need to manually change the `available_airflow_operators`  configurable trait to include import statements for all custom components. 

Changes are backwards compatible, so any implementations of CCC's using the previous version that doesn't return the well-defined object will still work properly (e.g. the url catalog). Built-in CCC's will have to be updated (in a separate PR) to use the `get_entry_data`.

For now, here is what you will need to do (@ptitzler) to implement the new metadata return values:

- Change the name of the `read_catalog_entry` function to `get_entry_data` (parameters are the same)
- The new return value is an `EntryData` object, and more specifically for the Airflow package connector use-case, an `AirflowEntryData` object, so:
  - Add the import statement (duh) `from elyra.pipeline.catalog_connector import AirflowEntryData` and `from elyra.pipeline.catalog_connector import EntryData` 
  - Change type hint return value to `Optional[EntryData]`
  - The package name string should not include the class name (which should be the natural behavior)
  - Finally, instead of returning a string value, you'll now want to initialize an `AirflowEntryData` and return that instead:
```python
return AirflowEntryData(definition=<definition string>, package_name=<package name string>)
```

- Convert `get_hash_keys` to a `classmethod` by adding the decorator and changing the parameter from `self` to `cls`

Initial manual tests following the above structure are working.

Tests have been updated.
Successful and error scenarios have been manually tested.

           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/catalog_connector.py
+++ b/elyra/pipeline/catalog_connector.py
@@ -41,7 +41,7 @@ class ComponentDefinition(object):
     An object corresponding to a single entry of a component catalog, which has a
     unique id, a string definition, and a dict of identifying key-value pairs
     """
-    entry_id: str = None
+    id: str = None
     definition: str = None
     identifier: Dict = None
 
@@ -70,7 +70,7 @@ class ComponentDefinition(object):
 
         # Use only the first 12 characters of the resulting hash
         hash_digest = f"{hashlib.sha256(hash_str.encode()).hexdigest()[:12]}"
-        self.entry_id = f"{catalog_type}:{hash_digest}"
+        self.id = f"{catalog_type}:{hash_digest}"
 
     @property
     def identifier_as_string(self) -> str:
@@ -154,7 +154,7 @@ class ComponentCatalogConnector(LoggingConfigurable):
                            catalog_entry_data: Dict[str, Any],
                            catalog_metadata: Dict[str, Any]) -> Optional[str]:
         """
-        DEPRECATED. get_component_definition() must be implemented instead.
+        DEPRECATED. Will be removed in 4.0. get_component_definition() must be implemented instead.
 
         Reads a component definition for a single catalog entry using the catalog_entry_data returned
         from get_catalog_entries() and, if needed, the catalog metadata.
@@ -329,7 +329,7 @@ class ComponentCatalogConnector(LoggingConfigurable):
 
                     try:
                         # Attempt to get a ComponentDefinition object from get_component_definition
-                        catalog_entry = self.get_component_definition(
+                        component_definition = self.get_component_definition(
                             catalog_entry_data=catalog_entry_data,
                             catalog_metadata=catalog_metadata
                         )
@@ -342,22 +342,22 @@ class ComponentCatalogConnector(LoggingConfigurable):
                             catalog_metadata=catalog_metadata
                         )
 
-                        catalog_entry = ComponentDefinition(definition=definition, identifier=catalog_entry_data)
+                        component_definition = ComponentDefinition(definition=definition, identifier=catalog_entry_data)
 
                     # Ignore this entry if no definition content is returned
-                    if not catalog_entry or not catalog_entry.definition:
+                    if not component_definition or not component_definition.definition:
                         self.log.warning(f"No definition content found for catalog entry with identifying information: "
-                                         f"{catalog_entry.identifier_as_string}. Skipping...")
+                                         f"{component_definition.identifier_as_string}. Skipping...")
                         catalog_entry_q.task_done()
                         continue
 
                     # Generate hash for this catalog entry and set as ComponentDefinition id
-                    catalog_entry.set_entry_id(
+                    component_definition.set_entry_id(
                         catalog_type=catalog_instance.schema_name,
                         hash_keys=keys_to_hash
                     )
 
-                    catalog_entries.append(catalog_entry)
+                    catalog_entries.append(component_definition)
 
                 except NotImplementedError as e:
                     msg = f"{self.__class__.__name__} does not meet the requirements of a catalog connector class: {e}."
--- a/elyra/pipeline/catalog_connector.py
+++ b/elyra/pipeline/catalog_connector.py
@@ -351,7 +351,6 @@ class ComponentCatalogConnector(LoggingConfigurable):
                 keys_to_hash = ComponentCatalogConnector.get_hash_keys()
 
             except Exception:
-                print('hello!')
                 keys_to_hash = self.get_hash_keys()
 
             # Add display_name attribute to the metadata dictionary
--- a/elyra/pipeline/catalog_connector.py
+++ b/elyra/pipeline/catalog_connector.py
@@ -123,7 +123,7 @@ class ComponentCatalogConnector(LoggingConfigurable):
 
         For example, the FilesystemCatalogConnector includes both a base directory ('base_dir') key-value
         pair and a relative path ('path') key-value pair in its 'catalog_entry_data' dict. Both fields
-        are needed in order to access the corresponding component definition in read_catalog_entry().
+        are needed in order to access the corresponding component definition in get_component_definition().
 
         Every catalog_entry_data should contain each of the keys returned in get_hash_keys() to ensure
         uniqueness and portability among entries. For the same reason, no two catalog entries should have
@@ -143,7 +143,7 @@ class ComponentCatalogConnector(LoggingConfigurable):
                     }
 
         :returns: a list of catalog entry dictionaries, each of which contains the information
-                  needed to access a component definition in read_catalog_entry()
+                  needed to access a component definition in get_component_definition()
         """
         raise NotImplementedError(
             "abstract method 'get_catalog_entries()' must be implemented"
@@ -221,8 +221,8 @@ class ComponentCatalogConnector(LoggingConfigurable):
         also enables pipeline portability across installations when the keys returned here are
         chosen strategically. For example, the FilesystemCatalogConnector includes both a base
         directory key-value pair and a relative path key-value pair in its 'catalog_entry_data' dict.
-        Both fields are required to access the component definition in read_catalog_entry(), but
-        only the relative path field is used to create the unique hash. This allows a component
+        Both fields are required to access the component definition in get_component_definition(),
+        but only the relative path field is used to create the unique hash. This allows a component
         that has the same relative path defined in two separate a catalogs in two separate
         installations to resolve to the same unique id in each, and therefore to be portable across
         pipelines in these installations.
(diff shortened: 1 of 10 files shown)


kevin-bates:
I think `get_component_definition()` should Return a `ComponentDefinition` instance, assuming there isn't such a class defined already. 

We should also note the `read_catalog_entry()` is deprecated and will be removed in 4.0 and existing component connectors will continue to work until then.

kevin-bates:
> Could we make it optional? Seems more natural ...

We can make the return value optional.  However, it does introduce some ambiguity for backward-compatibility handling since we'll need to call the old method if the new method returns `None` because we're unable to determine if the new method is the user's implementation or the base class's (because we can't use an abstract method approach).  And we'll need to test the existence of the old method prior to calling it as new implementations won't have implemented it.   This is necessary if we want to release these changes within a major release.

elyra/pipeline/catalog_connector.py
```diff
@@ -101,6 +154,8 @@ def read_catalog_entry(self,
                            catalog_entry_data: Dict[str, Any],
                            catalog_metadata: Dict[str, Any]) -> Optional[str]:
         """
+        DEPRECATED. Will be removed in 4.0. get_component_definition() must be implemented instead.
```
ptitzler, line 157: Can we use the [`deprecated` decorator](https://deprecation.readthedocs.io/en/latest/) so this shows up in builds?

elyra/pipeline/catalog_connector.py
```diff
@@ -90,7 +143,7 @@ def get_catalog_entries(self, catalog_metadata: Dict[str, Any]) -> List[Dict[str
                     }
 
         :returns: a list of catalog entry dictionaries, each of which contains the information
-                  needed to access a component definition in read_catalog_entry()
+                  needed to access a component definition in get_component_definition()
         """
         raise NotImplementedError(
             "abstract method 'get_catalog_entries()' must be implemented"
```
ptitzler, line 149: If we leave the `@abstractmethod` decorator in the code then all implementing classes must provide an implemention. To avoid that we should remove the decorator.

kevin-bates: Was this comment intended for `read_catalog_entry()` below?  We still want `get_catalog_entries()` to be an abstract method.

FWIW, the behavior you describe only occurs when the class derives from the ABC meta-class.  Since that "thunking" is not happening here, these aren't true ABCs - where enforcement requires implementation.  Instead, the enforcement happens at the time of the call.  (True abstract methods don't require a `NotImplementedError`, they can even be a `pass` statement.) Kiersten and I talked about that and there was some concern that some unwitting future maintainer will see that and convert these to true abstract classes at which time, we would require implementations - so taking the approach of `@deprecated` is definitely preferred.  Unfortunately, abstract methods for publicly consumable classes only work for the initial implementation.

(Our [SchemasProvider class](https://github.com/elyra-ai/elyra/blob/647351101ed36c2c2ad9210c3bfb66a1112cbd90/elyra/metadata/schema.py#L379-L390) is an example that sets up the ABC in a manner that enforces implementation at instantiation.)

kevin-bates (commented):
Sorry - got a few comments mostly around naming but I'm hoping we can make things easier for connector authors as well.  I'd be happy to meet via Webex to go over things if that is easier.

elyra/pipeline/catalog_connector.py
```diff
@@ -27,15 +27,70 @@
...
+        # Use only the first 12 characters of the resulting hash
+        hash_digest = f"{hashlib.sha256(hash_str.encode()).hexdigest()[:12]}"
+        self.id = f"{catalog_type}:{hash_digest}"
+
+    @property
+    def identifier_as_string(self) -> str:
+        return str(self.identifier)
+
+
+class AirflowCatalogEntry(ComponentDefinition):
```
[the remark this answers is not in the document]
kiersten-stokes: Wow, yes. Not sure how I missed this

elyra/pipeline/catalog_connector.py
```diff
@@ -27,15 +27,70 @@
...
+    """
+    An Airflow-specific ComponentDefinition that includes the fully qualified package
+    name (excluding class name) that represents the definition file.
+    """
+    package_name: str = None
+
+    def __init__(self, definition: str, identifier: Dict, package_name: str):
+        super(AirflowCatalogEntry, self).__init__(definition, identifier)
```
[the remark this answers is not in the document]
kiersten-stokes: Yeah, I think my IDE auto-completes it this way. I'll remove

kiersten-stokes:
Resolving several review comments above after offline design discussion. Will update PR description to reflect new design once changes are pushed

elyra/pipeline/catalog_connector.py
```diff
@@ -27,13 +27,116 @@
...
+    A KFP-specific EntryData object
+    """
+    pass
+
+
+class CatalogEntry(object):
+    """
+    An object corresponding to a single entry of a component catalog, which has a
+    unique id, a string definition, a dict of identifying key-value pairs, and
+    other associated metadata.
```
kiersten-stokes, line 78: I'm going to amend this slightly

elyra/pipeline/catalog_connector.py
```diff
@@ -70,7 +173,7 @@ def get_catalog_entries(self, catalog_metadata: Dict[str, Any]) -> List[Dict[str
 
         For example, the FilesystemCatalogConnector includes both a base directory ('base_dir') key-value
         pair and a relative path ('path') key-value pair in its 'catalog_entry_data' dict. Both fields
-        are needed in order to access the corresponding component definition in read_catalog_entry().
+        are needed in order to access the corresponding component definition in get_component_definition().
```
kiersten-stokes, line 176: Missed updating this to `get_entry_data`

elyra/pipeline/catalog_connector.py
```diff
@@ -90,17 +193,21 @@ def get_catalog_entries(self, catalog_metadata: Dict[str, Any]) -> List[Dict[str
                     }
 
         :returns: a list of catalog entry dictionaries, each of which contains the information
-                  needed to access a component definition in read_catalog_entry()
+                  needed to access a component definition in get_component_definition()
```
kiersten-stokes, line 196: Missed updating this to `get_entry_data`

elyra/pipeline/catalog_connector.py
```diff
@@ -126,18 +233,50 @@ def read_catalog_entry(self,
...
+        :param catalog_metadata: the metadata associated with the catalog in which this catalog entry is
+            stored; this is the same dictionary that is passed into get_catalog_entries(); in addition to
+            catalog_entry_data, catalog_metadata may also be needed to read the component definition for
+            certain types of catalogs
+
+        :returns: an EntryData object representing the definition (and other identifying info) for a single
+            catalog entry; if None is returned, this catalog entry is skipped and a warning message logged
+        """
+        raise NotImplementedError(
+            "method 'get_component_definition()' must be overridden"
```
kiersten-stokes, line 262: Missed updating this to `get_entry_data`

elyra/pipeline/catalog_connector.py
```diff
@@ -126,18 +233,50 @@ def read_catalog_entry(self,
...
+        This function has been changed to a class method as of version 3.7. Connectors that still
+        implement this function as an abstract method will be supported in a fallback scenario.
+
         Besides being a means to uniquely identify a single component (catalog entry), the hash id
         also enables pipeline portability across installations when the keys returned here are
         chosen strategically. For example, the FilesystemCatalogConnector includes both a base
         directory key-value pair and a relative path key-value pair in its 'catalog_entry_data' dict.
-        Both fields are required to access the component definition in read_catalog_entry(), but
-        only the relative path field is used to create the unique hash. This allows a component
+        Both fields are required to access the component definition in get_component_definition(),
```
kiersten-stokes, line 278: Missed updating this to `get_entry_data`

elyra/pipeline/catalog_connector.py
```diff
@@ -27,13 +27,116 @@
...
+                # Catalog entry does not have key - build hash without it
+                continue
+            hash_str = hash_str + str(self.entry_reference[key]) + ":"
+        hash_str = hash_str[:-1]
+
+        # Use only the first 12 characters of the resulting hash
+        hash_digest = f"{hashlib.sha256(hash_str.encode()).hexdigest()[:12]}"
+        return f"{self.catalog_type}:{hash_digest}"
+
+    def get_component(self, id: str, name: str, description: str, properties: List[ComponentParameter]) -> Component:
```
kevin-bates, line 123: We should probably have a help-string for this helper method.

kiersten-stokes: ah yes good catch. I had one here once upon a time 😆

kevin-bates (approved):
This is really great @kiersten-stokes - thank you.

ptitzler (approved):
LGTM

PR #2498, opened 2022-02-22
adding integration test

Hema-Mathiyazhagan:

 
           in the file; or

           it.


karlaspuldaro:
Hi Hema,
thank you for contributing to the integration tests :)

While this PR is still in draft, I'd like to add some early feedback.
1. I think these new tests should be moved to `scripteditor.ts` file since we group them by extension, and that would avoid the overhead of creating a new test environment for the script runner.
2. I ran them locally and it fails at `cy.get('button[title="Run"]').click()` step because at that point it's still at the main launcher page.
Before that you would need to open a python file with something like `cy.createNewScriptFile('Python')`

Issue #2515, opened 2022-03-01
Components installed via Elyra CLI do not appear in KFP Editor


lehrig:
**Describe the issue**
Successfully added components using the Elyra CLI (`elyra-metadata install component-catalogs ...`)  **do not appear** in the Kubeflow Pipeline Editor.

**To Reproduce**
1. Go to 'Component Catalogs' view
<img width="431" alt="Bildschirmfoto 2022-03-01 um 19 34 27" src="https://user-images.githubusercontent.com/9555494/156228216-23a9f4f2-a978-47c7-865b-d414fec78a0b.png">

2. Confirm that the catalog is empty.
3. Inside Jupyter Lab, open a terminal and add any component catalog using the Elyra CLI:

```
elyra-metadata install component-catalogs \
       --display_name="filter components" \
       --description="filter text in files" \
       --runtime_type=KUBEFLOW_PIPELINES \
       --schema_name="url-catalog"\
       --paths="['https://raw.githubusercontent.com/elyra-ai/elyra/master/etc/config/components/kfp/filter_text_using_shell_and_grep.yaml']" \
       --categories='["filter content"]' 
```
<img width="958" alt="Bildschirmfoto 2022-03-01 um 19 38 28" src="https://user-images.githubusercontent.com/9555494/156228818-67459d34-770e-4745-bf1b-7d7b05553e03.png">

4. Confirm that the catalog got applied successfully
`elyra-metadata list component-catalogs`
<img width="902" alt="Bildschirmfoto 2022-03-01 um 19 39 45" src="https://user-images.githubusercontent.com/9555494/156228992-65c356d2-bbd7-4b55-a398-aa3fa0e2e2b3.png">

5. On the side bar menu, go to another view and then back to the 'Component Catalogs' view to get this update:
<img width="435" alt="Bildschirmfoto 2022-03-01 um 19 42 11" src="https://user-images.githubusercontent.com/9555494/156229321-44ea29c5-77fa-4bb0-be6e-2ebf75072a9e.png">

6. Open a new Kubeflow Pipeline Editor and note that the new component does not appear (this is the referred bug!):
<img width="991" alt="Bildschirmfoto 2022-03-01 um 19 44 10" src="https://user-images.githubusercontent.com/9555494/156229605-35547039-c532-4bdf-af56-6b941a186f03.png">

[...]

PR #2522, opened 2022-03-02
Make component cache event-driven to allow out-of-proc updates

kevin-bates:
When component catalog metadata instances are created from a process outside of Jupyter Lab (e.g., via the `elyra-metadata` CLI tool), the new metadata instance is detected in Lab (when forcing a refresh of instances) but the components associated with (or produced by) the catalog will not appear in the visual pipeline editor until the Jupyter Lab instance is **restarted**.  Similar problems can also occur when the catalog instances are updated or deleted out-of-process.

This pull request addresses these issues by altering the approach by which the catalog components cache (which contains the component definitions associated with each catalog) is maintained to an _event-driven_ model by monitoring updates to a _component cache manifest_ file.  This file is essentially a means of inter-process communication so that, when a Lab instance is also running, we can detect out-of-process updates that impact the component cache's consistency.

Here's the general algorithm:
1. When Elyra (Lab) starts, it loads() the component catalog cache.  This registers a watchdog _observer_ to monitor the Jupyter runtime directory, filtering on files prefixed with `component-cache-manifest`.  Note that this observer will NOT be started via any work performed by the `elyra-metadata` CLI tool.  As a result, we now distinguish whether the process is a _server process_ or not (i.e., a CLI utility).
1. When a component catalog metadata instance is added, updated, or deleted, the metadata instance's post-save or post-delete hooks fire.  When the process is a _server process_, the component cache queue is updated directly.  However, when the process is not a server process, an entry is added to a manifest file maintained by that process (i.e., it writes to add an entry to the file `component-cache-manifest-<pid>.json`) containing the component catalog name and its corresponding action (`modify` or `delete`).  If no Lab is running, the manifest file is ignored, but when an observer is running (starting by the _server process_) the observer will file a modification event and its hook will post the applicable entry into the component cache queue (just as if the metadata instance was updated by the server process).  It will then clear the manifest file. (We will want to determine if we should introduce a file locking mechanism, but do not see this being a highly contentious area for now.  Perhaps if we find Component Catalog imports into running servers a common thing, we may have more reason to handle this.)
1. No significant changes have taken place within the general component cache management.  We have extended behaviors to track errors that might occur when fetching component definitions from the catalog connectors, but the general cache management is the same as before.

A visual representation of the algorithm is given in the diagram below. Boxes with a black ⭐  indicate an item new to this PR.

ComponentCache structure with single refresh

Important notes:
1. A manifest file will be written by any CLI process that updates the instance of a Component Catalog.  If Lab (i.e., "the server") is not running, this file will be ignored, otherwise, it will be monitored and process its entries.
1. All manifest files are removed when the Lab process is started.
1. A given manifest file can be read/written by different processes.  We have not introduced a file locking mechanism at this time.  With the ability to refresh both Component Catalog instances and the Component Definition cache, users will always have a workaround should the lack of file locking prove problematic.

Other changes:
- Updates metadata persistence to one of "save with refresh" as the updated entry is now returned via the load/get plumbing - which will trigger the `on_load()` hook.  This results in the instance being validated twice, once before save and another following `on_load()` since `on_load()` _can_ alter the content of the instance but needs to adhere to its schema, but is probably more correct behavior relative to the hooks.
- Applicable tests have been updated to perform two passes, the first performing the metadata instance update/creation in-process (which is the case today), while the second is to perform the metadata instance update out-of-process via the `elyra-metadata` CLI.

Co-authored and co-designed by @kiersten-stokes 

Fixes: #2515
 
           in the file; or

           it.


kevin-bates:
The scrum item was regarding the out-of-process _update_ of a component catalog metadata instance.  This will not result in an update of the component cache because the catalog is already represented in the component cache and there's no mechanism for invalidating the cache due to an out-of-process update.  As a result, support for out-of-process updates will need to be handled by the "refresh button" changes.

I have edited this description to be specific that we're addressing out-of-process catalog _creation_ with this PR.

kevin-bates:
Also note that besides out-of-process _updates_ being an issue (resolved by "refresh"), out-of-process _deletes_ are NOT resolved by "refresh".  As a result, components associated with the out-of-process deleted catalog will not be removed from the cache.  Since the catalog metadata instance will be removed in the Lab process, I'm not sure if that will prevent the orphaned (cached) components from being used, but they will remain in the cache until the process is restarted or we introduce the ability to flush the cache.

kevin-bates:
Placing back into draft to prevent accidental merge while we further discuss the out-of-proc update and delete issues.

elyra/pipeline/component_catalog.py
```diff
@@ -195,36 +315,75 @@ class ComponentCache(SingletonConfigurable):
...
+            catalogs = MetadataManager(schemaspace=ComponentCatalogs.COMPONENT_CATALOGS_SCHEMASPACE_ID).get_all()
+            for catalog in catalogs:
+                self.update_manifest(catalog, 'modify')
+
+    def set_up_manifest_file(self):
         """
+        Build an empty manifest and overwrite or create the file. After this point,
+        only the CacheUpdateManagerThread is able to touch this file.
+        """
+        empty_manifest = {
```
kevin-bates, line 362: This seems like a good area to describe the manifest. Let's add as many type hints into this as we can.

elyra/pipeline/component_catalog.py
```diff
@@ -195,36 +315,75 @@ class ComponentCache(SingletonConfigurable):
...
+        }
 
-        all_catalogs = MetadataManager(schemaspace=ComponentCatalogs.COMPONENT_CATALOGS_SCHEMASPACE_ID).get_all()
-        for catalog in all_catalogs:
-            self.update_cache_for_catalog(catalog)
+        with open(self.manifest_filename, 'w') as f:
+            f.write(json.dumps(empty_manifest, indent=2))
 
-    def update_cache_for_catalog(self, catalog: ComponentCatalogMetadata, operation: Optional[str] = None):
+    def update_manifest(self, catalog: Optional[ComponentCatalogMetadata] = None, action: Optional[str] = None):
```
kevin-bates, line 370: Are these parameters optional?  I don't see any `None` handling in the related methods.

elyra/pipeline/component_catalog.py
```diff
@@ -44,40 +48,144 @@
...
+                                'errors': errors
+                            }
+
+                            # Fetch the catalog instance associated with this action
+                            catalog_instance = MetadataManager(
+                                schemaspace=ComponentCatalogs.COMPONENT_CATALOGS_SCHEMASPACE_ID
+                            ).get(name=catalog_name)
+
+                        # Add action to the cache queue
+                        self._cache_queue.put((catalog_instance, cache_action))
```
kevin-bates, line 118: What does it mean to place a `None` catalog_instance into the cache queue?  If this shouldn't happen perhaps the `elif` could be `else`.

elyra/pipeline/component_catalog.py
```diff
@@ -44,40 +48,144 @@
...
+                    # Any action that is not a 'cache' action means that it is a manifest action
+                    # Format necessary action and include in manifest only if not already present
+                    manifest_action = {catalog.name: action}
+                    if manifest_action not in manifest['actions']:
+                        manifest['actions'].append(manifest_action)
+                        write_required = True
+
+                if write_required:
+                    with open(self._manifest_filename, 'w') as f:
+                        f.write(json.dumps(manifest, indent=2))
```
kevin-bates, line 135: Per previous comment, we should use `dump()`:
```suggestion
                        json.dump(manifest, indent=2, f)
```

elyra/pipeline/component_catalog.py
kiersten-stokes: Per offline discussion, `action` is never `None`, but catalog in `None` in the case where the watchdog is trigger an action. Will move `action` to the first parameter position and remove the 'Optional' hint.

elyra/pipeline/component_catalog.py
```diff
@@ -44,40 +48,144 @@
...
-                # Create and start a thread for the task
-                updater_thread = CacheUpdateThread(self._component_cache, self._cache_task_queue, catalog, operation)
-                updater_thread.start()
+                # Make the required update to the manifest file
+                if os.path.isfile(self._manifest_filename):
+                    with open(self._manifest_filename, 'r') as f:
+                        manifest = json.loads(f.read())
+
+                write_required = False
+                if action == 'cache':
```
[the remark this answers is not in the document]
kiersten-stokes: Per offline discussion, we'll keep this variable named `action` because the subsequent `if` block acts on `cache_action`s and the `else` block acts on `manifest_action`s

elyra/pipeline/component_catalog.py
```diff
@@ -44,40 +48,144 @@
...
+                    # A 'cache' action means that updates to the component cache are required
+
+                    # Handle all pending actions in one batch
+                    for pending_action in manifest.get('actions'):
+                        catalog_name = list(pending_action.keys())[0]
+                        cache_action = list(pending_action.values())[0]
```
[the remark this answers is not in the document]
kiersten-stokes: Re: this comment and the one below, I've restructured the manifest to look like the following:

```json
{
  "actions": {
    "<catalog_name>": "<action>",
    "<catalog_name>": "<action>"
  },
  "status": {
    "<catalog_name>": {
      "action": "<action>",
      "errors": ["<error_as_str>"]
    },
    "<catalog_name>": {
      "action": "<action>",
      "errors": ["<error_as_str>"]
    },
  }
}
```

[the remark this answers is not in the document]
kevin-bates: This looks great.  I forget what the "action" field contains in the "status" entries, but I want to say its the actual status of that catalog entry - "current", "updating", "error".  If so, would it work to change the field to `"state"`?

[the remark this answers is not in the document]
kiersten-stokes: Yeah that's what I'm thinking as well, and I agree that we should change the name!

elyra/pipeline/component_catalog.py
```diff
@@ -44,40 +50,168 @@
...
+
+    def build_manifest_file(self):
+        """
+        Build an empty manifest and overwrite or create the file. Only the
+        CacheUpdateManagerThread is able to touch this file.
+        """
+        empty_manifest: Dict[str, str] = {}
+
+        with open(self._manifest_filename, 'w') as f:
+            f.write(json.dumps(empty_manifest, indent=2))
```
kevin-bates, line 165: Let's convert this to:
```suggestion
            json.dump(empty_manifest, f, indent=2)
```

conftest.py
```diff
@@ -43,12 +44,26 @@
...
 
 
+def wait_for_cache_updates(component_catalog):
+    """
+    TODO
+    """
+    t0 = time.time()
+    elapsed_time = t0 - time.time()
+    component_catalog.wait_for_all_manifest_updates()
+    while component_catalog.cache_manager.manifest_update_pending and elapsed_time < 60:
```
kevin-bates, line 54: Are you finding this `while` loop to be necessary?  I figured `join`ing the threads would be sufficient - just curious.

kevin-bates (commented):
Looking good so far - thank you.

elyra/tests/pipeline/airflow/test_component_parser_airflow.py
```diff
@@ -53,16 +55,20 @@ def test_component_catalog_can_load_components_from_registries(component_cache_i
...
-    # Initialize a ComponentCache instance and wait for all worker threads to compete
+@pytest.mark.parametrize('create_inprocess', [True, False])
+async def test_modify_component_catalogs(jp_environ, metadata_manager_with_teardown, create_inprocess):
+    # Clear any ComponentCache instance so next parametrized test starts with clean instance
+    ComponentCache.clear_instance()
+
+    # Initialize a ComponentCache instance
     component_catalog = ComponentCache.instance()
-    component_catalog.wait_for_all_cache_updates()
+    component_catalog.load()
```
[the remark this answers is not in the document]
kiersten-stokes: I think that's a good idea! I'm going to tweak some things too (noted in another comment below) and I think this will go nicely with those changes

conftest.py
kiersten-stokes: Yeah the more I thought about this loop, the more I don't like it. The way things are currently set up, `join`ing doesn't work, but I think I'm going to shift around when various threads mark their tasks as `done` so that we can avoid this

elyra/tests/pipeline/kfp/test_component_parser_kfp.py
```diff
@@ -43,20 +46,26 @@ def _get_resource_path(filename):
 
 @pytest.mark.parametrize('component_cache_instance', [KFP_COMPONENT_CACHE_INSTANCE], indirect=True)
 def test_component_catalog_can_load_components_from_registries(component_cache_instance):
-    components = ComponentCache.instance().get_all_components(RUNTIME_PROCESSOR)
+    component_cache = ComponentCache.instance()
```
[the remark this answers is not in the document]
kiersten-stokes: The `load` is actually done in the fixture, but I think I'm going to change that around to go with the new `component_catalog` fixture below. See if we can't speed some things up in the process.

elyra/pipeline/component_catalog.py
```diff
@@ -195,46 +292,122 @@ class ComponentCache(SingletonConfigurable):
...
+    def wait_for_all_cache_tasks(self):
         """
         Block execution and wait for all tasks in the cache task update queue to complete.
         Primarily used for testing.
         """
-        self._cache_task_queue.join()
+        if self.is_server_process:
+            self.cache_queue.join()
 
     def get_all_components(self, platform: RuntimeProcessorType) -> List[Component]:
```
akchinSTC, line 402:
```suggestion
    def get_all_components(self, platform: RuntimeProcessorType) -> List[Union[str, List[str]]]:
```

elyra/pipeline/component_catalog.py
```diff
@@ -195,46 +292,122 @@ class ComponentCache(SingletonConfigurable):
...
         """
         Retrieve all components from component catalog cache
         """
         components: List[Component] = []
```
akchinSTC, line 406:
```suggestion
        components: List[Union[str, List[str]]] = []
```

elyra/pipeline/component_metadata.py
```diff
@@ -38,14 +38,15 @@ def runtime_type(self) -> RuntimeProcessorType:
...
+        except Exception as ex:
+            print(ex)
             pass
 
     def post_delete(self, **kwargs: Any) -> None:
-        try:
-            component_catalog.ComponentCache.instance().update_cache_for_catalog(catalog=self, operation='delete')
+        try:  # Remove components associated with this catalog on deletes.
+            component_catalog.ComponentCache.instance().update(catalog=self, action='delete')
         except Exception:
```
akchinSTC, line 50:
```suggestion
        except Exception as ex:
            print(ex)
```

Guess do the same as post_save and print out the exception?

akchinSTC (approved):
nits, otherwise lgtm
some other small complaints mostly coming from the IDE but Ill leave that for later

elyra/pipeline/component_metadata.py
kiersten-stokes: Nice catch. I'm wondering if this isn't just leftover debugging stuff. I'll probably just remove both - thanks for catching that!

elyra/pipeline/component_catalog.py
kiersten-stokes: Huh my IDE doesn't yell at me for this (or the above case). Looking into it now. It should be returning `List[Component]` but maybe something is wrong with the `ComponentCacheType` typehint or something

kiersten-stokes: I think the type hint should actually be: `ComponentCacheType = Dict[str, Dict[str, Dict[str, Dict[str, Union[Component, str, List[str]]]]]]`. Making a note of it here so that I can change that after @kevin-bates pushes a few of his local commits.

elyra/pipeline/component_metadata.py
kevin-bates: I'd prefer we not use `print` (anywhere).  Since the `Metadata` base class isn't _plumbed_ for a logger (which we should do, but not in this PR), can the CodeQL stuff be resolved with a comment (I believe I saw something to that effect when I've seen exceptions ignored before)?

elyra/pipeline/component_catalog.py
```diff
@@ -195,46 +352,154 @@ class ComponentCache(SingletonConfigurable):
...
+    def _determine_server_process(**kwargs) -> bool:
+        """Determines if this process is a server (extension) process."""
+        app_names = ['ServerApp', 'ElyraApp']
+        is_server_process = False
+        if 'parent' in kwargs and kwargs['parent'].__class__.__name__ in app_names:
+            is_server_process = True
+        elif 'emulate_server_app' in kwargs and kwargs['emulate_server_app']:  # Used in unittests
+            is_server_process = True
+
+        return is_server_process
```
kevin-bates, line 391: This is proving to be a difficult thing to get working with tests, namely those using the `fp_fetch` fixture used in handler tests.  I'm finding that for the "servers" started (indirectly) by that fixture, the `ComponentCache` instance does not reflect its a "server process" when, in fact, it can ONLY be a server process.

When tests are run individually, it behaves correctly, but I'm suspicious of the fixture's teardown is not adequately cleaning things up.  I will continue looking into this, but, for now, the `test_double_refresh` test in the pipeline's `test_handlers` tests is failing.  Things work as expected in "real life".

kevin-bates: A single test has been found to side-affect the `test_double_refresh`.  That test is `elyra/tests/cli/test_pipeline_app.py::test_validate_with_missing_kfp_component` and there's no rhyme or reason why that particular test would do this (that I've found).  This test was reworked in #2582 so perhaps it resolves this.

For the time being, `test_double_refresh` will be marked with `xfail` so that the overall results will pass.  If the test is no longer side-affected, it will produce an `xpass` indication, at which time we can remove the marking.

kevin-bates: Looks like the changes in #2582 have resolved the need to mark `test_double_refresh` with `xfail` and removed via https://github.com/elyra-ai/elyra/pull/2522/commits/dbe585e976fd7624f8a64b65f9a9fa7cccabc6ca.

kiersten-stokes: After merging with latest master, the `test_double_refresh` test passes with `xpass`, so the `xfail` marker has now been removed.

elyra/pipeline/component_metadata.py
kiersten-stokes: I added a comment in the latest commit, but that didn't seem to stop the warning, unfortunately. I've dismissed it for now

kevin-bates: I wonder what is meant by "explanatory comment"?  Not really finding anything other than examples that include logging something, which would suck.

ptitzler (approved):
LGTM

PR #2532, opened 2022-03-04
Create Component Catalogs Widget

### LINKED ISSUE #2238: Allow for manual refresh of the VPE component palette view
Elyra will support remote component catalogs. Since the content of these catalogs can change at any time and Elyra can't monitor those catalogs for changes, a user needs the ability to re-fetch the content.

Option 1: Add a button to the VPE, which would re-populate the entire palette by querying ALL configured component catalogs.
      image

 The drawback is that this could be a time consuming operation, depending on how many catalogs are defined, whether they are remote catalogs, their response times, etc.

Probably better

Option 2: Add a refresh button for  each component catalog entry:

This allows for selective refresh of individual catalogs:

image

ajbozarth:
Add `ComponentCatalogsWidget` that includes a refresh button for
each instance that will trigger a cache refresh. Also updates the
header refresh to refresh the cache for all components prior to
updating the metadata list.

Fixes #2238

- [x] Blocked by https://github.com/elyra-ai/elyra/pull/2522
   - This PR has been "rebased" on that PR for testing, all code outside `packages/*` can be ignored as it comes from that PR

 
Screen Shot 2022-03-23 at 5 56 41 PM

           in the file; or

           it.


packages/pipeline-editor/src/ComponentCatalogsWidget.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A widget for displaying runtime images.
+ */
+export class ComponentCatalogsWidget extends MetadataWidget {
+  constructor(props: IMetadataWidgetProps) {
+    super(props);
+  }
+
+  renderDisplay(metadata: IMetadata[]): React.ReactElement {
```
ptitzler, line 73: Could you please forward fit the changes from https://github.com/elyra-ai/elyra/pull/2537, e.g.

```suggestion
  renderDisplay(metadata: IMetadata[]): React.ReactElement {
      if (Array.isArray(metadata) && !metadata.length) {
       // Empty metadata
       return (
         <div>
           <br />
           <h6 className="elyra-no-metadata-msg">
             Click the + button to add {this.props.display_name}
           </h6>
         </div>
       );
     }
```

lresende:
Please make sure we add the component to tracker/restorer so that page refreshes goes back to same state

packages/pipeline-editor/src/ComponentCatalogsWidget.tsx
ajbozarth: done

kiersten-stokes:
The new endpoints delivered by #2522 are as follows:

- to refresh all catalog components: `elyra/pipeline/components/cache`
- to refresh a single catalog's components: `elyra/pipeline/components/cache/{catalog_name}` 

For the latter, `catalog_name` is the metadata instance name for that catalog (i.e. not the display name)

ajbozarth:
This is now ready for review. I've updated the top description to match the current state. I will be leaving this in draft until https://github.com/elyra-ai/elyra/pull/2522 is merged

ajbozarth:
I've rebased off https://github.com/elyra-ai/elyra/pull/2522 again (which is where the test failures and conflicts come from). With the latest updates I've made `body` required for put calls again and add the body `{ action: 'refresh' }` to the API call. I've also wrapped the error handling to not surface `status 409` to the user.

ptitzler:
To avoid potential ambiguity, perhaps we should use in the tool tip a bit more specific language, such as "reload components from catalog"?

image

ajbozarth:
> To avoid potential ambiguity, perhaps we should use in the tool tip a bit more specific language, such as "reload components from catalog"?

I can switch to this text instead

> Just to confirm, a reload of the catalog does/does not not trigger a near-real-time refresh of the palette when the VPE is currently open?

I am unsure if how we leverage SWR handles this, and I do not know off the top of my head how to create a scenario to check it. Either way this PR does not make any intentional changes to affect that behavior.

> Since we are changing the behavior from refresh list (with no side effects) to refresh list (with side effects = reload components from all valid component catalogs) we should also update the tool tip to set expectations properly:

This is actually not so simple, that button is generic part of the MetadataEditor and is not accessible when extending it. I could surface that `title` value as a prop and then make that prop an extendable value if it's a big issue to change it.

ajbozarth:
I just pushed a big commit that does the following to address @ptitzler review:

- Adds more detailed tooltips for both individual and header refresh buttons (the header tooltip required adding new plumbing)

- Triggers a refresh of the palette in open pipeline editors when the header refresh or any of the individual action buttons are used.

   -  This required adding a new top level signal to the pipeline editor for triggering a palette refresh and tying it to a new command that will emit the signal. A function that executes that command is then passed to the Component Catalog Widget to trigger as stated above. In the pipeline editor the new signal is then tied to the `mutate` function returned by the SWR call to get the palette. Calling the `mutate` function causes the SWR to immediately refresh which in turn updates the palette

ajbozarth:
Now that all the other PRs are merged to master I've updated this PR onto latest master. The git log is a mess, but the diff should look correct now.

A quick note here on the following:

> Triggers a refresh of the palette in open pipeline editors when the header refresh or any of the individual action buttons are used.

There is a race condition we found when creating or editing an instance. On save sometimes the palette update occurs before the cache can be updating, meaning the changes are not reflected in the palette. This does not occur constantly though and from a UX perspective it is not an issue since a users first response to not seeing it update would be to click the refresh button right next to the edit button, which would then fix it. @kevin-bates and @kiersten-stokes said they would look into understanding the race condition from a server perspective next week, but it is non-blocking to this PR.

ptitzler (approved):
Latest UI changes LGTM. We should probably add a section between https://elyra.readthedocs.io/en/latest/user_guide/pipeline-components.html#adding-a-component-catalog and https://elyra.readthedocs.io/en/latest/user_guide/pipeline-components.html#modifying-a-component-catalog-entry that covers how to reload components from a catalog.

lresende (approved):

marthacryan (approved):

karlaspuldaro (approved):

PR #2533, opened 2022-03-04
Sanitize KFP input/output params in kwargs

thesuperzapper:

This PR prevents Elyra from throwing an error when it tries to pass invalid `kwargs` to a KFP component-factory function.

```
  File "/opt/conda/lib/python3.8/site-packages/elyra/pipeline/kfp/processor_kfp.py", line 622, in _cc_pipeline
    container_op = factory_function(**operation.component_params_as_dict)
TypeError: My Component - 0.1.0() got an unexpected keyword argument 'my-input'
```

A Kubeflow component can name its inputs/outputs things that are not valid python `kwargs` names, in these cases, KFP silently renames these inputs by replacing the invalid characters with `_`.
- EXAMPLES: 
    - `my-input` --> `my_input`
- Kubeflow Code:
    - https://github.com/kubeflow/pipelines/blob/1.8.1/sdk/python/kfp/components/_components.py#L359-L363
    - https://github.com/kubeflow/pipelines/blob/1.8.1/sdk/python/kfp/components/_naming.py#L32-L42
    - https://github.com/kubeflow/pipelines/blob/1.8.1/sdk/python/kfp/components/_naming.py#L49-L50

It is also possible to have a case where multiple different parameters map to the same "sanitized value". In these cases, KFP appends a number to the end of the ones which were transformed (starting at `2`). But because elyra stores parameters in an unordered dictionary, we can't possibly know which one maps to which index. (But it's probably not a huge issue if we don't support this extreme edge case)
- EXAMPLES: 
    - `my-input` and `my_input` --> `my_input_2` and `my_input`
- Kubeflow Code:
    - https://github.com/kubeflow/pipelines/blob/1.8.1/sdk/python/kfp/components/_naming.py#L97-L122

I have tested this in a Kubeflow cluster with a component that has parameters named things like `my-parameter-name`, and they were correctly compiled and run. 

---
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/kfp/processor_kfp.py
+++ b/elyra/pipeline/kfp/processor_kfp.py
@@ -619,7 +619,11 @@ class KfpPipelineProcessor(RuntimePipelineProcessor):
                         operation.component_params_as_dict.pop(parameter, None)
 
                     # Create ContainerOp instance and assign appropriate user-provided name
-                    container_op = factory_function(**operation.component_params_as_dict)
+                    sanitized_component_params = {
+                        self._sanitize_param_name(name): value
+                        for name, value in operation.component_params_as_dict.items()
+                    }
+                    container_op = factory_function(**sanitized_component_params)
                     container_op.set_display_name(operation.name)
 
                     if operation.doc:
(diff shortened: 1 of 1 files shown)


elyra/pipeline/kfp/processor_kfp.py
```diff
@@ -692,6 +696,14 @@ def _sanitize_operation_name(name: str) -> str:
         """
         return re.sub('-+', '-', re.sub('[^-_0-9A-Za-z ]+', '-', name)).lstrip('-').rstrip('-')
 
+    @staticmethod
+    def _sanitize_param_name(name: str) -> str:
+        """
+        Sanitize a component parameter name.
+        (this is how KFP ensures component inputs/outputs are valid python kwarg names)
+        """
+        return re.sub('[^_0-9a-z]+', '_', name.lower()).lstrip('_').rstrip('_')
```
ptitzler, line 705: Can you also take care of names that start with a numeric? As is, the regex still produces an invalid Python identifier. Example: `123-or-more`

thesuperzapper: @ptitzler thanks for raising this, but the issue was not with the regex, it was me being stupid and using the sanitization from Kubeflow's V2 API.

I have updated the PR to use the EXACT same sanitization as Kubeflow 1.X now.

ptitzler (approved):
LGTM

akchinSTC (approved):

PR #2536, opened 2022-03-04
Update Makefile python and pip calls to use python3

akchinSTC:

Invoke the python using python3 to avoid issues where python 2.x is being used

 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,10 @@
 .PHONY: container-images publish-container-images validate-runtime-images
 SHELL:=/bin/bash
 
+# Python execs
+PYTHON=python3
+PYTHON_PIP=$(PYTHON) -m pip
+
 TAG:=dev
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,6 +25,10 @@
 .PHONY: container-images publish-container-images validate-runtime-images
 SHELL:=/bin/bash
 
+# Python execs
+PYTHON=python3
```
kevin-bates, line 29: I think we should allow for overrides on the CLI...
```suggestion
PYTHON?=python3
```

ptitzler (commented):
Should https://github.com/elyra-ai/elyra/blob/1b2e30e7a113e873f75fe8984226fb4a17cd4bcf/Makefile#L37 also use the variable to remain consistent?

Issue #2019, opened 2021-08-05
Metadata Editor and CLI should block whitespace only tags


karlaspuldaro:
**Describe the issue**
Metadata Editor allows whitespace only strings as a valid tag
![image](https://user-images.githubusercontent.com/25207344/128399261-c6f1d7c9-9328-4b87-913b-bae4159b44d4.png)

Similarly, when [creating a metadata using CLI ](https://elyra.readthedocs.io/en/latest/user_guide/code-snippets.html#id1), tags can are also accepted as whitespace only.

Apparently leading/trailing whitespaces are already cleaned up in the UI.

**To Reproduce**
Steps to reproduce the behavior in the UI:
1. Create new or edit a metadata (code snippets, runtime, runtime image)
2. Add a tag containing only white spaces
3. Save & close
4. Check the metadata widget tag filter

Steps to reproduce the behavior in CLI:
1. Create new metadata (eg. code-snippet) and add a whitespace or empty string tag
```
elyra-metadata install code-snippets \
	--display_name="Preview DataFrame" \
	--description="Preview Pandas DataFrame" \
	--tags="['Python', 'Pandas', '  ', '']" \
	--language="Python" \
	--code="['df.head(5)']"
```
2. Check the json file just created containing whitespace/empty string tags

**Expected behavior**
UI - The editor should discard/flag whitespace only tags.
CLI - Metadata service should validate whitespace/empty string tags.

**Context information**
 - Elyra version: 3.0.0
 - Operating system: macos
 - Browser: Chrome

samali0996:
I'm interested in working on this issue!

JoshuaAlter:
Im also interested!

kevin-bates:
Hello @samali0996 and @JoshuaAlter - thank you for your interest.  I just saw this issue and would like to point out that the majority of this issue should be addressed via JSON schema and not in the `CLI` tooling itself since the CLI is completely schema-driven.  I suspect the same might be true for the front-end, but I'm not familiar with that (cc: @marthacryan).

I just don't want folks spending time trying to implement this in a manner that we've made an architectural decision not to do and that may seem counter-intuitive as to how one might expect to approach this kind of change.  I have a schema snippet that would resolve this, but given the interest in helping, I'd be happy to let others have a go at it and will be happy to assist in any manner.

When working with JSON schema, I find the following tool invaluable: https://www.jsonschemavalidator.net/

[...]

PR #2538, opened 2022-03-08
Metadata tag whitespace proposed solution

JoshuaAlter:
According to the issue referenced above there are problems with `whitespace only` and `empty` tags. In fact, upon looking into this issue there are some other concerns that we should look to address. The proposed solution in the schema will touch upon a couple of items:

1. Whitespace at the beginning of the line. A [regex](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html#regular-expressions) was added following the allowed pattern options in a json schema to stop any number of whitespace characters after the beginning of a tag - `"^[^ \t]+"` - this can be read as "after the beginning of line, only allow non-whitespace characters for any length." This stop only whitespace as well because a first character whitespace will not be a non-whitespace character.  
2. Whitespace at the end of the line. Very similar to the above, `"[^ \t]+$"` was added to stop whitespace after a tag. This can be read as "Only allow non-whitespace characters of any length before ending the tag". Great! This will stop that extra space after a tag being allowed - something we do not want. 
3. [Minimum length tags](http://json-schema.org/understanding-json-schema/reference/string.html?highlight=minlength#id5) Following the linked item, we added an `item` parameter to stop 0 length tags. This is aimed to stop the empty tag situation.
4. [Unique items](https://json-schema.org/understanding-json-schema/reference/array.html?highlight=unique#uniqueness) While we're here, we should think. This seems to make sense logically, why should there be duplicated tags? Well, we stop this with a uniqueness boolean, but it is something to be discussed. I can imagine a situation where somebody is scripting tag creation and they are tagging something multiple times with "passed" or "correct" or "complete" or one of many other items indication some kind of finish. Then, they are just checking the count of the tags. Similarly, perhaps they are adding the tag "1" to on each round of run and are using that as a counter for their application. Would this not be valid. Should this be stopped? 

Anyway, below are examples of many use cases and their current outputs with the added code and it covers more than the expected behavior (linked in the issue and referenced below)! 

```
UI - The editor should discard/flag whitespace only tags.
CLI - Metadata service should validate whitespace/empty string tags. 
```

1. expected usage - allowed
elyra-test-expected
There is no effect on a traditional tag without whitespace 

2. whitespace in the middle of the tags - allowed
elyra-test-whitepace-in-the-middle
There is no effect to tags that contain whitespace within the tags themselves

3. example used in issue description - not allowed
elyra-example-from-issue
The example tests with both an empty string tag `''` and an only whitespace tag `'    '`. Neither of these are allowed. And thankfully not when they're together either. We can look at the individual tags submitted below.

4. empty string tag
elyra-empty-tag
The tag stops the code snippet creation on validation due to the tag being too short. Great!

 
5. whitespace only tags - not allowed
elyra-just-space
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

6. whitespace before the tag - not allowed
elyra-test-begin-space
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

7. whitespace after the tag - not allowed
elyra-test-end-space
The tag stops the code snippet creation on validation due to a failure on the second pattern match. Great!

8. whitespace before and after the tag - not allowed
elyra-test-space-both-ends
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

9. duplicated tags - to discuss
elyra-test-dup
The tag stops the code snippet creation on validation due to a failure on the uniqueness! Great! Or is it?

Fixes: #2019

### DIFF
--- a/elyra/metadata/schemas/airflow.json
+++ b/elyra/metadata/schemas/airflow.json
@@ -178,7 +178,15 @@
         "tags": {
           "title": "Tags",
           "description": "Tags for categorizing Apache Airflow",
+          "uniqueItems": true,
           "type": "array",
+          "items": {
+            "allOf": [
+              { "minLength": 1 },
+              { "pattern": "^[^ \t]+" },
+              { "pattern": "[^ \t]+$" }
+            ]
+          },
           "uihints": {
             "field_type": "tags"
           }
--- a/elyra/metadata/schemas/code-snippet.json
+++ b/elyra/metadata/schemas/code-snippet.json
@@ -36,6 +36,14 @@
           "title": "Tags",
           "description": "Tags for categorizing snippets",
           "type": "array",
+          "uniqueItems": true,
+          "items": {
+            "allOf": [
+              { "minLength": 1 },
+              { "pattern": "^[^ \t]+" },
+              { "pattern": "[^ \t]+$" }
+            ]
+          },
           "uihints": {
             "field_type": "tags"
           }
--- a/elyra/metadata/schemas/kfp.json
+++ b/elyra/metadata/schemas/kfp.json
@@ -168,7 +168,15 @@
         "tags": {
           "title": "Tags",
           "description": "Tags for categorizing Kubeflow pipelines",
+          "uniqueItems": true,
           "type": "array",
+          "items": {
+            "allOf": [
+              { "minLength": 1 },
+              { "pattern": "^[^ \t]+" },
+              { "pattern": "[^ \t]+$" }
+            ]
+          },
           "uihints": {
             "field_type": "tags"
           }
--- a/elyra/metadata/schemas/metadata-test.json
+++ b/elyra/metadata/schemas/metadata-test.json
@@ -80,6 +80,16 @@
           "type": "string",
           "pattern": "^[a-z0-9][a-z0-9-.]*[a-z0-9]$"
         },
+        "string_complex_test": {
+          "title": "String Complex Test",
+          "description": "Property used to test strings with multiple restrictions",
+          "type": "string",
+          "allOf": [
+            { "minLength": 1 },
+            { "pattern": "^[^ \t]+" },
+            { "pattern": "[^ \t]+$" }
+          ]
+        },
         "enum_test": {
           "title": "Enum Test",
           "description": "Property used to test properties with enums",
--- a/elyra/metadata/schemas/runtime-image.json
+++ b/elyra/metadata/schemas/runtime-image.json
@@ -35,7 +35,15 @@
         "tags": {
           "title": "Tags",
           "description": "Tags for categorizing runtime images",
+          "uniqueItems": true,
           "type": "array",
+          "items": {
+            "allOf": [
+              { "minLength": 1 },
+              { "pattern": "^[^ \t]+" },
+              { "pattern": "[^ \t]+$" }
+            ]
+          },
           "uihints": {
             "field_type": "tags"
           }
--- a/elyra/tests/metadata/test_metadata.py
+++ b/elyra/tests/metadata/test_metadata.py
@@ -158,6 +158,37 @@ def test_manager_add_display_name(tests_manager, schemaspace_location):
         tests_manager.metadata_store.fetch_instances(metadata_name)
 
 
+@pytest.mark.parametrize("complex_string, valid", [
+    ("", False),
+    (" ", False),
+    (" starting-whitespace", False),
+    ("ending-whitespace ", False),
+    (" whitespace-both-ends ", False),
+    ("whitespace in between", True),
+    ("no-whitespace", True),
+])
+def test_manager_complex_string_schema(tests_manager, schemaspace_location, complex_string, valid):
+    metadata_name = 'valid_metadata_instance'
+    metadata_dict = {**valid_metadata_json}
+    metadata_dict['metadata']['string_complex_test'] = complex_string
+    metadata = Metadata.from_dict(METADATA_TEST_SCHEMASPACE_ID, metadata_dict)
+
+    if not valid:
+        with pytest.raises(ValidationError):
+            tests_manager.create(metadata_name, metadata)
+
+    else:
+        instance = tests_manager.create(metadata_name, metadata)
+        assert instance.metadata.get('string_complex_test') == complex_string
+
+        # And finally, remove it.
+        tests_manager.remove(metadata_name)
+
+        # Verify removal using metadata_store
+        with pytest.raises(MetadataNotFoundError):
+            tests_manager.metadata_store.fetch_instances(metadata_name)
+
+
 def test_manager_get_include_invalid(tests_manager):
     metadata_list = tests_manager.get_all(include_invalid=False)
     assert len(metadata_list) == 2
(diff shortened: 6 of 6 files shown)


JoshuaAlter:
According to the issue referenced above there are problems with `whitespace only` and `empty` tags. In fact, upon looking into this issue there are some other concerns that we should look to address. The proposed solution in the schema will touch upon a couple of items:

1. Whitespace at the beginning of the line. A [regex](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html#regular-expressions) was added following the allowed pattern options in a json schema to stop any number of whitespace after the beginning of a tag - `"^[^ \t]+"` - this can be read as "after the beginning of line, only allow non-whitespace characters for any length." This stop only whitespace as well because a first character whitespace will not be a non-whitespace character.  
2. Whitespace at the end of the line. Very similar to the above, `"[^ \t]+$"` was added to stop whitespace after a tag. This can be read as "Only allow non-whitespace characters of any length before ending the tag". Great! This will stop that extra space after a tag being allowed - something we do not want. 
3. [Minimum length tags](http://json-schema.org/understanding-json-schema/reference/string.html?highlight=minlength#id5) Following the linked item, we added a item parameter to stop 0 length tags. This is aimed to stop the empty tag situation.
4. [Unique items](https://json-schema.org/understanding-json-schema/reference/array.html?highlight=unique#uniqueness) While we're here, we should look think. This seems to make sense logically, why should there be duplicated tags? Well, we stop this with a uniqueness boolean, but it is something to be discussed. I can imagine a situation where somebody is scripting tag creation and they are tagging something multiple times with "passed" or "correct" or "complete" or one of many other items indication some kind of finish. Then, they are just checking the count of the tags. Similarly, perhaps they are adding the tag "1" to on each round of run and are using that as a counter for their application. Would this not be valid. Should this be stopped? 

Anyway, below are examples of many use cases and their current outputs with the added code and it covers more than the expected behavior (linked in the issue and referenced below)! 

```
UI - The editor should discard/flag whitespace only tags.
CLI - Metadata service should validate whitespace/empty string tags. 
```

1. expected usage - allowed
elyra-test-expected
There is no effect on a traditional tag without whitespace 

2. whitespace in the middle of the tags - allowed
elyra-test-whitepace-in-the-middle
There is no affect to tags that contain whitespace within the tags themselves

3. example used in issue description - not allowed
elyra-example-from-issue
The example tests with both an empty string tag `''` and an only whitespace tag `'    '`. Neither of these are allowed. And thankfully not when they're together either. We can look at the individual tags submitted below.

4. empty string tag
elyra-empty-tag
The tag stops the code snippet creation on validation due to the tag being too short. Great!

 
5. whitespace only tags - not allowed
elyra-just-space
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

6. whitespace before the tag - not allowed
elyra-test-begin-space
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

7. whitespace after the tag - not allowed
elyra-test-end-space
The tag stops the code snippet creation on validation due to a failure on the second pattern match. Great!

8. whitespace before and after the tag - not allowed
elyra-test-space-both-ends
The tag stops the code snippet creation on validation due to a failure on the first pattern match. Great!

9. duplicated tags - to discuss
elyra-test-dup
The tag stops the code snippet creation on validation due to a failure on the uniqueness! Great! Or is it?

kevin-bates:
Hi @JoshuaAlter - this is great - thank you!  I agree with your 4 bullet points for attacking whitespace in tags and love the use of `allOf` there.  

Regarding uniqueness, I suppose your use cases make sense in some cases, and this particular attribute (`uniqueItems: true`), I believe, would apply to the schemas _we_ provide.  Since folks can bring their own schemas for their own functionality, then they could choose to omit (or set to `false`) that attribute in their schemas.  There's nothing "built-in" that requires that attribute be present, so I think we're good.  If others believe any of our existing schemas require duplicate tags, then we should discuss, but I think starting with them being unique is the reasonable default behavior.

Here are some other items to address:
- There are several other schemas that reference a `tags` property, so we'll want to update those as well. 
- We should also add this `allOf` stanza under an `allOf_test` attribute to the [metadata_test schema ](https://github.com/elyra-ai/elyra/blob/master/elyra/metadata/schemas/metadata-test.json)and add a test to ensure `allOf` references are properly behaved.  (Note that the metadata tests operate exclusively off the metadata-test schemas and not the application schemas.  I'm not sure we need front-end/UI tests to test tag behaviors in code-snippets, runtimes, etc., although those tests should already have something like that and perhaps they can be extended to introduce illegal whitespace.)

kevin-bates (approved):
This looks great @JoshuaAlter.  I really like (and appreciate) you being able to address this issue strictly via JSON schema, and the use of parameters in your test addition is excellent - thank you!

ptitzler:
I'm not sure items 1 and 2 of the proposed behavior are user friendly and align with the behavior in other parts of the Elyra code. 

Where applicable (e.g. environment variable names, mount points, pvc names in node properties, pipeline names, ...), we consider leading and trailing whitespaces in user-provided text input as irrelevant and strip them off and don't return an error. This way the input is standardized prior to processing and the user doesn't need to manually resolve minor issues that the application can easily handle.

ptitzler:
I'm not sure items 1 and 2 of the proposed behavior are user friendly and align with the behavior in other parts of the Elyra code. 

Where applicable (e.g. environment variable names, mount points, pvc names in node properties, pipeline names, ...), we consider leading and trailing whitespaces in user-provided text input as irrelevant and strip them off and don't return an error. This way the input is standardized prior to processing and the user doesn't need to manually resolve minor issues that the application can easily handle.

kevin-bates:
@ptitzler 
> I'm not sure items 1 and 2 of the proposed behavior are user friendly and align with the behavior in other parts of the Elyra code.

You're correct.  It seems we've only recently had a focus on that aspect of things and the bulk of this PR was done well before then.  That said, I agree that we should have a consistent UX.  The issue here is that this kind of implicit handling for specific properties is not well-suited to a schema-driven model inherent in our metadata-based instances.  Because the front-end and CLI tooling are also schema-driven, I suspect the correct approach would be to introduce a _metadata class_ (via the `metadata_class_name` property in the schema, which applies these kinds of changes.

I think introducing classes specific to each of the "system-owned" schemas would be best (at least those that define `tags` although we need to decide if we want this in general) and each of these classes would either derive from or use a mixin class (multiple inheritance) that is responsible for trimming _explicitly enumerated_ string-valued properties (including lists of strings).  The idea is that this "trim" class defines an empty list attribute (e.g., `trim_properties: List`) that is inherited by the derived class and the derived class initializes that list to _its_ set of properties that should be trimmed.  The trim class would then implement `pre_save()` to trim the values of whichever properties are in the `self.trim_properties` list.

If we feel that **all** string-valued properties (and lists) should be trimmed unconditionally, then we could just build this into the base metadata manager and forgo the additional instance class overrides.  I think that might be a tough ask for our _cyrstal ball_ since there may be some special case somewhere that then breaks this approach.

As far as this PR goes, I suggest we go ahead and merge this as a backend _catch all_ for untrimmed `tags` values.

Thoughts and suggestions are welcome...

akchinSTC (approved):
Im good with this going in as is for now since its scoped to just `tags` as a catch all until we can figure out the bigger picture with whitespace handling in properties

PR #2564, opened 2022-03-14
Enable display of custom component definitions

### LINKED ISSUE #2548: Provide a REST API that allows for retrieval of a component definition
Currently users cannot access component definitions (YAML file for KFP, Python file for Airflow) in the VPE. To support visualization in JupyterLab, the back-end needs to provide a REST API endpoint that the front-end can invoke, when the user invokes the `Open File` action on a node (via the context menu or double-click):

image

The API endpoint expects a component id as input and returns the YAML file content for KFP custom components and the Python source code for Airflow custom components. (The Python source code might contain definitions for multiple components, which should be okay because co-located components typically have a parent-child or sibling relationship.)

### LINKED ISSUE #2549: Enable display of component definitions

Double-click (or `Open File` from the context menu) on a node that is implemented using a custom component opens the component specification in a JupyterLab editor window. 

image

The behavior is identical to the behavior that users observe when they perform the same task on a node that is implemented by a generic component (notebook or script).

To avoid confusion whether or not changes made to the displayed code have an impact on Elyra, a disclaimer comment needs to be injected, indicating that changes are ignored by the VPE. (In most cases Elyra does not have write access to the original location of the source code, which could be, for example,  a read-only catalog.)

image

- Requires https://github.com/elyra-ai/elyra/issues/2548

kiersten-stokes:
Fixes #2548
Fixes #2549 

Dependent upon https://github.com/elyra-ai/pipeline-editor/pull/188

Adds an API to retrieve the component definition given its `id` and runtime type:

```
pipeline/components/<runtime_type>/<component_id>
```

The value of the component `definition` attribute defaults to `None` if there is no definition content, which is currently only the case for the generic components. 

The component APIs have also been modified to allow either the 'old' runtime name (e.g. `kfp`, `airflow`) or the runtime platform type name (e.g. `KUBEFLOW_PIPELINES`, `APACHE_AIRFLOW`).

This PR is in draft pending the addition of a test case for the new endpoint.

- [x] addition of new test cases
- ~[ ] update of the Elyra api YAML (most of these changes can come from #2522)~ I'll open a follow-up issue to cover this situation
- [x] merge and release https://github.com/elyra-ai/pipeline-editor/pull/188 and update package.json

Test case(s) to be added
 
           in the file; or

           it.


elyra/elyra_app.py
```diff
@@ -97,6 +98,8 @@ def initialize_handlers(self):
             (f'/{self.name}/pipeline/components/{processor_regex}', PipelineComponentHandler),
             (f'/{self.name}/pipeline/components/{processor_regex}/{component_regex}/properties',
              PipelineComponentPropertiesHandler),
+            (f'/{self.name}/pipeline/components/{processor_regex}/{component_regex}/definition',
```
kevin-bates, line 101: Since the component definition is a strict subset of the properties API, I wonder if we need the `/definition` in the endpoint?  The thought being that `/pipeline/{processor_regex}/{component_regex}` gets you the definition and `/pipeline/{processor_regex}/{component_regex}/properties` gets you the definition in terms of the canvas properties - like its just a little more specific.

kiersten-stokes: Yeah I was toying with this idea too, especially since the definition handler class is basically a complete repeat of the properties handler class code

ajbozarth:
As noted in this morning's scrum, it would help the front end if the API returned a json dict that included the "content" as a string as well as mimetype for the content.

For reference this is the code mirror function that jupyterlab uses to look up the mimetype based on the file extension: https://codemirror.net/mode/meta.js and this function: https://github.com/jupyterlab/jupyterlab/blob/a2d2c5b26f16b97e605b43e47cfb9a301620921d/packages/codemirror/src/mimetype.ts#L36 wraps it to override two cases, `md` and `ipy`

Based on that the mime types the api would most likely return are `text/x-python` for py files and `text/yaml` for yaml and yml files

ptitzler:
A rather common scenario (in multi-user environments where pipeline files can be shared) is that a referenced component is not found. How would the API communicate this issue?

kevin-bates:
Shouldn't it return a `404` error?

elyra/pipeline/handlers.py
```diff
@@ -174,6 +174,29 @@ async def get(self, processor, component_id):
...
+        self.log.debug(f'Retrieving component definition for component: {component_id}')
+
+        if PipelineProcessorManager.instance().is_supported_runtime(processor) is False:
+            raise web.HTTPError(400, f"Invalid processor name '{processor}'")
+
+        if not component_id:
+            raise web.HTTPError(400, "Missing component ID")
+
+        component: Optional[Component] = \
+            await PipelineProcessorManager.instance().get_component(processor, component_id)
```
kevin-bates, line 191: I think this will return `None` if the `component_id` can't be found, so we should check for that and raise a 404. (We should probably consider raising a NotFound exception from the cache instead and include the id in the message, etc. - for another day.)

The 400 raised if the component_id is missing entirely seems like a good match.

ajbozarth:
Based on our Webex and your last update, the API on this LGTM. I won't speak to the actual python code though.

ajbozarth:
I just pushed an initial working version of the front end code, including a merge with latest master.

It includes a dependency on https://github.com/elyra-ai/pipeline-editor/pull/188/files and currently points to the code sandbox for that PR.

While working on the frontend I found a couple issues with the API:
- The endpoint only works when hitting `.../components/<runtime_type>/<component_id>` (`/definition` must be left off the url)
- The value of `runtime_type` is still using the old `runtime` values not `runtime_type` values (eg `kfp` instead of `KUBEFLOW_PIPELINES`. This means it has to be is derived using a hack on the frontend, we may want to address this now rather than waiting longer (note the value is derived using `__doNotUseInFutureMapTypeToRandomProcessor__`). If this is a API breaking issue we can just leave the hack in for now and make sure to remove it with an API update in 4.0

@kiersten-stokes dont forget to rebase
@ptitzler this should be ready for review of the functionality now
@kevin-bates any input on the `runtime_type` issue noted above

kiersten-stokes:
> - The endpoint only works when hitting `.../components/<runtime_type>/<component_id>` (`/definition` must be left off the url)

Ah yeah, my bad. Kevin and I discussed this change in an earlier round of review, and it looks like I neglected to update the PR description accordingly. I'll do that!

ptitzler:
Initial test observation:

- Add a custom component to the [KFP] catalog 
- Create a pipeline and add the component
- Open the component definition and inspect the [server] console log

```
WARNING: Component 'url-catalog:0e585ee0e744' specifies extensions '['.yaml']' but                                    no entry in the 'parameter_ref' dictionary for 'filehandler' and                                    cannot participate in drag and drop functionality as a result.
```

ptitzler:
For the following error scenario:
 - open a pipeline in the VPE that includes one or more custom components that are not registered in the catalog
 - try to open the component definition
     image

Can we change the behavior to issue a message similar to the one that is displayed when the user tries to open the properties for a non-existent component?

image

ptitzler:
Should this also close https://github.com/elyra-ai/elyra/issues/2549, since the PR also includes the front-end changes?

ptitzler:
For very long component/operator names the code viewer tab displays only part of the name, e.g.

image

Would it be possible to enable a tool tip that renders the complete name, similar to the behavior that can be observed if one hovers over the VPE tab?
image

elyra/pipeline/handlers.py
```diff
@@ -137,15 +139,29 @@ class PipelineComponentHandler(HttpErrorMixin, APIHandler):
...
+        elif processor_manager.is_supported_runtime(runtime_type):
+            # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
+            # 'airflow'). This case and its associated functions should eventually be removed
+            # in favor of using the RuntimeProcessorType name in the request path.
+            self.log.warning(f"Deprecation warning: when calling endpoint '{self.request.path}' "
+                             f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
+                             f"instead of shorthand name (e.g., 'kfp', 'airflow')")
+            components: List[Component] = await processor_manager.get_components(runtime_type)
+        else:
+            raise web.HTTPError(400, f"Invalid runtime type '{runtime_type}'")
```
kevin-bates, line 163: This might be cleaner if we check for the runtime first and get its runtime type, then fall into the cache access block.  Perhaps something like this...
```python
        if processor_manager.is_supported_runtime(runtime_type):
            # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
            # 'airflow'). This case and its associated functions should eventually be removed
            # in favor of using the RuntimeProcessorType name in the request path.
            self.log.warning(f"Deprecation warning: when calling endpoint '{self.request.path}' "
                             f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
                             f"instead of shorthand name (e.g., 'kfp', 'airflow')")
            runtime_processor_type = processor_manager.get_runtime_type(runtime_type)
        elif processor_manager.is_supported_runtime_type(runtime_type):
            runtime_processor_type = RuntimeProcessorType.get_instance_by_name(runtime_type)
        else:
            raise web.HTTPError(400, f"Invalid runtime type '{runtime_type}'")

        # The request path uses the appropriate RuntimeProcessorType name.
        # Include generic components for all runtime types
        components: List[Component] = ComponentCache.get_generic_components()

        # Add additional runtime-type-specific components, if present
        components.extend(ComponentCache.instance().get_all_components(platform=runtime_processor_type))
```

elyra/pipeline/handlers.py
```diff
@@ -160,25 +176,46 @@ def get_mime_type(self, ext: Optional[str]) -> str:
...
+        elif processor_manager.is_supported_runtime(runtime_type):
+            # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
+            # 'airflow'). This case and its associated functions should eventually be removed
+            # in favor of using the RuntimeProcessorType name in the request path.
+            self.log.warning(f"Deprecation warning: when calling endpoint '{self.request.path}' "
+                             f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
+                             f"instead of shorthand name (e.g., 'kfp', 'airflow')")
+            component: Optional[Component] = await processor_manager.get_component(runtime_type, component_id)
+        else:
+            raise web.HTTPError(400, f"Invalid runtime type '{runtime_type}'")
```
kevin-bates, line 218: Same comment as previous, check for runtime (schema) first.

ajbozarth:
@ptitzler I've partially addressed you comments. I've added the tooltip for the tab title and have re-added my fallback code for when a component is not in the palette. The fallback will pass the component `op` value for the node when it can't find the component in the palette to look up it's id (note that our current airflow and kfp examples mirror `op` and `id` but generic components do not). This at least makes the backend error more readable. As for displaying an error message like the one in the properties panel, that is a bit more difficult. It would be straight forward to display a custom error message, but that error message occurs whenever the API returns any error not just a component not found so we would have to then distinguish between the types of errors. At that point I think it would make more sense for the api to just return more detailed error message (`reason`) that the front end service would then display automatically as it does now.

elyra/pipeline/handlers.py
```diff
@@ -154,6 +154,16 @@ async def get(self, processor):
 class PipelineComponentPropertiesHandler(HttpErrorMixin, APIHandler):
     """Handler to expose method calls to retrieve pipeline component_id properties"""
 
+    def get_mime_type(self, ext: Optional[str]) -> str:
+        """
+        Get the MIME type for the component definition content.
+        """
+        if ext == '.yaml':
```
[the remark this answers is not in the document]
kiersten-stokes: Hmm, I'm not sure if GitHub is having some issues, but I do include `.yml` in my commit as of last night. But I'll be updating this anyway according to the above comment

elyra/pipeline/handlers.py
kiersten-stokes: That's much nicer 😄

elyra/pipeline/component_catalog.py
```diff
@@ -236,7 +236,7 @@ def get_all_components(self, platform: RuntimeProcessorType) -> List[Component]:
         for catalog_name, catalog_components in platform_components.items():
             components.extend(list(catalog_components.values()))
 
-        if not components:
+        if not components and platform.name != "LOCAL":
```
kevin-bates, line 239: Since we have a `RuntimeProcessorType` we should probably compare it directly...
```suggestion
        if not components and platform != RuntimeProcessorType.LOCAL:
```

elyra/pipeline/catalog_connector.py
```diff
@@ -83,17 +83,20 @@ class CatalogEntry(object):
...
     categories: List[str]
+    file_extensions: str
 
     def __init__(self,
                  entry_data: EntryData,
                  entry_reference: Any,
                  catalog_instance: Metadata,
-                 hash_keys: List[str]):
+                 hash_keys: List[str],
+                 file_extensions: List[str]):
```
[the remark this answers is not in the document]
kiersten-stokes: The way it's currently set up, no because this is the object that we instantiate (not the one the catalog connector author instantiates). But this gave me an idea that I'm going to try

marthacryan (approved):
Frontend code LGTM but haven't tested locally

ajbozarth:
@ptitzler I've pushed the updates we discussed (ended up doing it today >.<), here's a screenshot of how it looks now:
Screen Shot 2022-03-23 at 7 12 03 PM
Note that the JSON parse doesn't work (the code is there) because the backend is encoding it in python format not json and so it's not parseable/ With some research online I've found that this is a common issue with no nice solution (usually really ugly hacky regex solutions). If this is a longterm issue then we should have the backend start encoding the value as a json string instead of a python string. (unsure if that would require some sort of migration though)

ptitzler:
@ajbozarth can you please make sure the 'component not found'  implementation can accomodate display of a URL that links to a document that outlines what to do? I've opened issue https://github.com/elyra-ai/elyra/issues/2588 to create a documentation topic that contains the necessary information.

ajbozarth:
just pushed a fix to the error dialog now that component_source is json formatted.

> @ajbozarth can you please make sure the 'component not found' implementation can accomodate display of a URL that links to a document that outlines what to do? I've opened issue https://github.com/elyra-ai/elyra/issues/2588 to create a documentation topic that contains the necessary information.

This should be straightforward to do once the url is available

ajbozarth:
new screenshot
Screen Shot 2022-03-24 at 1 01 36 PM

ptitzler:
> just pushed a fix to the error dialog now that component_source is json formatted.
> 
> > @ajbozarth can you please make sure the 'component not found' implementation can accomodate display of a URL that links to a document that outlines what to do? I've opened issue #2588 to create a documentation topic that contains the necessary information.
> 
> This should be straightforward to do once the url is available

Great. See https://github.com/elyra-ai/elyra/pull/2591

ajbozarth:
I've update the pipeline0editor to use the new 1.7 release and I've added the new doc link to the error message (note the docs are not yet showing the update for me)

Screen Shot 2022-03-25 at 11 10 12 AM

ptitzler (approved):
LGTM for AA and KFP. Confirmed that the help link works as expected

ajbozarth:
> We should also update the create-release.sh script to pin the new https://elyra.readthedocs.io/en/latest/user_guide/best-practices-custom-pipeline-components.html#troubleshooting-missing-pipeline-components URL to the "current" version, like we are already doing in other places, e.g. https://github.com/elyra-ai/elyra/blob/master/create-release.py#L175-L177

nice catch, there was actually a doc ref in the same file that was missing from this, iiic they should both get updated with my latest commit

kevin-bates (approved):

PR #2570, opened 2022-03-15
Enable CodeQL checks in CI workflows

akchinSTC:

Enabled CodeQL tests as part of the CI process to check for python and JS issues/vulnerabilities

Ran workflow in personal branch of master, check for results under the security tab of the repository
 
           in the file; or

           it.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,7 +1,11 @@
-name: "My CodeQL config"
+name: "Elyra CodeQL config"
 
 queries:
   - uses: security-and-quality
 
 paths-ignore:
+  - elyra/tests
   - node_modules
+  - packages/services/src/test
+  - packages/pipeline-editor/src/tests
+  - tests
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,7 @@
+name: "My CodeQL config"
+
+queries:
+  - uses: security-and-quality
+
+paths-ignore:
+  - node_modules
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 8 * * 3'
+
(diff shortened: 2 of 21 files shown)


akchinSTC:
by default, scanning will pickup python and JS. Unsure if we need JS activetoo if we have linting and dependabot alerts active.

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,7 @@
+name: "My CodeQL config"
```
kevin-bates, line 1:
```suggestion
name: "Elyra CodeQL config"
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 8 * * 3'
```
kevin-bates, line 15: Could we add a comment, I always have to look this up and I would imagine others do as well...
```suggestion
    # Run every Wednesday at 8 am 
    - cron: '0 8 * * 3'
```
I can't find anything on which timezone this would be but since it's only once a week (besides PRs), it probably doesn't matter much.

kevin-bates (commented):
Thanks Alan.  One of my concerns with this kind of thing is that built remains "broken" for a long time.  We should probably add a scrum item to discuss what approaches we have for correcting the initial issues ASAP.  I suppose we could all party on this PR, or merge this PR with errors present and create a high-priority issue/PR to make a sweep.  That kind of thing would probably be best when there are no large outstanding PRs in progress, so we might want to time the merge of this accordingly.

elyra/pipeline/component.py
```diff
@@ -381,6 +381,10 @@ def determine_type_information(self, parsed_type: str) -> SimpleNamespace:
                 elif data_type in ['list', 'set', 'array', 'arr']:
                     data_type = "list"
                     default_value = []
+                else:
+                    # TODO: Unknown container type, surface error?
+                    data_type = "unknown"
+                    default_value = None
```
kevin-bates, line 387: I think this `else:` block needs to be removed.  The non-container types are handled in the `else` block associated with exhausting the `for` lopp (which is still a hard thing for me to remember with python).  `option` is one of the supported container types by virtue of making into the first `if` statement.

Instead, the `elif` should be an `else` (if that's what the CodeQL complaint was about).
```suggestion
                else:  # data_type is one of ['list', 'set', 'array', 'arr']
                    data_type = "list"
                    default_value = []
```

ajbozarth:
@akchinSTC I pushed a few fixes for frontend warnings, the remaining frontend warnings actually might be causing the race condition bug it describes is possible and need to be followed up on afterwards.

PR #2583, opened 2022-03-22
Update python dependencies to address security issues

### LINKED ISSUE #2579: [Security] Click dependency before version 8 contains vulnerability
**Describe the issue**
Elyra requires since version 3.0.0 requires `click>=7.1.1,<8`: https://github.com/elyra-ai/elyra/blob/master/setup.py#L67
Click version before 8.0 contains the following vulnerability:
> The package does not properly create temporary files and uses tempfile.mktemp(), which allows a local attacker to overwrite arbitrary files and obtain sensitive information via a symlink attack on the temporary file. 'The package does not properly create temporary files and uses tempfile.mktemp(), which allows a local attacker to overwrite arbitrary files and obtain sensitive information via a symlink attack on the temporary file. Creating temporary files using insecure methods exposes the application to race conditions on filenames: a malicious user can try to create a file with a predictable name before the application does. A successful attack can result in other files being accessed, modified, corrupted or deleted.

git issue: https://github.com/pallets/click/issues/1752

**Expected behavior**
Install `click>=8`

**Deployment information**
 - Elyra version: 3.6.0
 - Installation source: PyPi
 - Operating system: -

akchinSTC:

Updating python dependencies to address security fixes. For now the following fixes will result in an error free installation of elyra without any conflicts. 

FYI, `kfp` sdk package is now pinned to python 3.9 or below. 

Few findings, 
- Trying to find out where `asteroid` dependency(transient) and by association `wrapt` is coming from. A search of the dependency tree doesnt reveal anything.
- The caps of both the `jsonschema` and `typing-extension` packages coming from the KFP package are not being honored by the resolver. I suspect(not 100%) this might be due to the use of the legacy resolver. If thats the case, we should look into fixing our overall python build strategy to use the newer resolver sooner than later. 
- `kfp-tekton` 1.2.0 was mainly tested against `kfp` 1.8.11, which is what we've been resolving to, hence the bump.

Fixes #2579 

 Unit and integration tests were run locally with a few basic pipelines for sanity. 

           in the file; or

           it.


setup.py
```diff
@@ -64,12 +64,12 @@
...
         'autopep8>=1.5.0,<1.5.6',
-        'click>=7.1.1,<8',
+        'click>=8',
         'colorama',
         'deprecation',
         'entrypoints>=0.3',
         'jinja2>=2.11',
-        'jsonschema>=3.2.0',
+        'jsonschema>=3.2.0,<4.0',  # Cap from KFP
         'jupyter_core>=4.0,<5.0',
```
kevin-bates: There's no reason to cap the non-existent `jupyter_core` release, could we please remove this?

setup.py
```diff
@@ -90,10 +90,12 @@
         'requests>=2.25.1,<3.0',
```
kevin-bates: If we're feeling it is too risky to uncap `requests` (whose latest release is [`2.27.1`](https://pypi.org/project/requests/)), then let's add a TODO comment to revisit early in the next sprint.

setup.py
```diff
@@ -90,10 +90,12 @@
...
         'tornado>=6.1.0',
+        'typing-extensions>=3.10,<4',  # Cap from KFP
         'traitlets>=4.3.2',
         'urllib3>=1.26.5',
         'watchdog>=2.1.3',
         'websocket-client',
+        'wrapt>=1.11,<1.14',  # Add from asteroid
         'yaspin',
         # KFP runtime dependencies
         'kfp>=1.7.0,<2.0,!=1.7.2',
```
kevin-bates: Can we also add a comment here?  The model is that anything capped must be on our radar to uncap as soon as possible.

setup.py
```diff
@@ -90,10 +90,12 @@
         'requests>=2.25.1,<3.0',
         'rfc3986-validator>=0.1.1',
         'tornado>=6.1.0',
+        'typing-extensions>=3.10,<4',  # Cap from KFP
         'traitlets>=4.3.2',
         'urllib3>=1.26.5',
         'watchdog>=2.1.3',
         'websocket-client',
+        'wrapt>=1.11,<1.14',  # Add from asteroid
```
kevin-bates: I don't see `asteroid` in our dependencies, do we know what of our dependencies is pulling in `asteroid` and, if so, could that be added to the comment?

setup.py
```diff
@@ -43,7 +43,7 @@
 ]
 
 runtime_extras = {
-    'kfp-tekton': ['kfp-tekton~=1.1.1', ],
+    'kfp-tekton': ['kfp-tekton~=1.2.0', ],
```
kevin-bates, line 46: This is effectively a cap of `<1.3` and probably warrants a comment as to why it exists.

kevin-bates (commented):
Looks like some changes were made during my initial review (like the cap on jupyter_core was removed - thank you!).  Generally speaking, I believe we should adopt an approach that all caps have an explanation with the understanding they will be revisited soon.

`kfp` appears to be a serious problem when it comes to capping dependencies.

setup.py
akchinSTC: unpinned

setup.py
akchinSTC: comment added

setup.py
akchinSTC: still trying to track this sucker down....

setup.py
```diff
@@ -63,43 +63,45 @@
...
         'jinja2>=2.11',
-        'jsonschema>=3.2.0',
-        'jupyter_core>=4.0,<5.0',
+        'jsonschema>=3.2.0,<4.0',  # Cap from kfp
+        'jupyter_core>=4.6.0',
         'jupyter_client>=6.1.7',
         'jupyter-packaging>=0.10',
         'jupyter_server>=1.7.0',
         'jupyterlab~=3.3.0',
         'jupyterlab-git~=0.32',
```
kevin-bates, line 78: Since these are essentially capped, could we add comments indicating why we're capping lab to <= 3.3.x and the git extension to <= 0.x?

setup.py
kevin-bates: Just saw the update, thanks.

kevin-bates (approved):
Thanks Alan - this is helpful.

PR #2600, opened 2022-03-28
Relax component YAML validation requirements for certain fields

kiersten-stokes:
Some component YAML files do not contain values for all their keys, e.g.:
```
name: Test Component
description:
inputs:
- {name: test_string_no_default, description: , type: String}
...
```

The simple component YAML validation functionality was listing this as invalid since the 'description' property is `null`/`None`. We want to relax this validation, as these fields should not be explicitly required for our purposes.

This PR relaxes the requirements that certain fields be of a certain data type, if present. These fields can now be either `null`/`None` or their corresponding data type. Note that this does not mean that these fields are required - it only means that if the key is present, the value is still optional. More specifically, these fields are component 'description', input/output property 'description', input/output property 'type', and input/output property 'optional' flag. Note that input/output property 'default value' is not listed in the validation criteria because it is not a required field and its data type can be any type, so there is nothing to validate.

The `kfp_test_operator.yaml` file was edited to have a `null`/`None` value for one of the input property descriptions in order to cover this case (tests are passing anyway indicating that this property having a null value still passes validation).
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/kfp/kfp_component_utils.py
+++ b/elyra/pipeline/kfp/kfp_component_utils.py
@@ -19,7 +19,7 @@ component_yaml_schema = {
             "type": "string"
         },
         "description": {
-            "type": "string"
+            "type": ["string", "null"]
         },
         "inputs": {
             "type": "array",
@@ -30,13 +30,13 @@ component_yaml_schema = {
                         "type": "string"
                     },
                     "description": {
-                        "type": "string"
+                        "type": ["string", "null"]
                     },
                     "type": {
-                        "type": "string"
+                        "type": ["string", "null"]
                     },
                     "optional": {
-                        "type": "boolean"
+                        "type": ["boolean", "null"]
                     }
                 }
             },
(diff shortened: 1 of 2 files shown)


elyra/pipeline/kfp/kfp_component_utils.py
```diff
@@ -19,7 +19,7 @@
             "type": "string"
         },
         "description": {
-            "type": "string"
+            "type": ["string", "null"]
```
lresende, line 22: why we use "null" instead of `null`?

kiersten-stokes: Strangely, that's how JSON schemas handles the 'null' case: https://json-schema.org/understanding-json-schema/reference/null.html

ptitzler (approved):
LGTM!

elyra/pipeline/kfp/kfp_component_utils.py
lresende: Thanks for the clarification, I was researching only the json part of it and not the jsonSchema.

PR #2623, opened 2022-04-01
[HOTFIX] Fix missing f string typo in operator

akchinSTC:
After formatting with black, missed an f-string typo on this line causing the generic operator to fail.

           in the file; or

           it.


kiersten-stokes (approved):
Good catch! Working for me after some tests with generic operators

kevin-bates (approved):
Thanks Alan.

ptitzler (approved):
LGTM as a hotfix. However, we should follow up in the future and harden error detection. Whenever the operator is unable to download the prerequisites, e.g. in https://github.com/elyra-ai/elyra/blob/master/elyra/kfp/operator.py#L178-L179, no error is returned by curl, resulting in downstream errors, such as the one we've just encountered.

ptitzler:
See https://curl.se/docs/manpage.html#--fail-with-body, which should return a non-zero return code and terminate processing early

Issue #2659, opened 2022-04-13
Dockerhub elyra/kf-notebook:latest is 3.6.0 instead of 3.7.0


RobinKa:
**Describe the issue**
`elyra/kf-notebook:latest` on Dockerhub points to the 3.6.0 image instead of the actual latest one (3.7.0).

See https://hub.docker.com/r/elyra/kf-notebook/tags

**Screenshots or log output**
![image](https://user-images.githubusercontent.com/2614101/163167237-251e2231-343e-4765-aec8-b717c0bbd982.png)

kevin-bates:
Hi @RobinKa - thanks for bringing this to our attention!  I've retagged `latest` images for `elyra` and `kf-notebook` and pushed those tagged images to the respective registries (`docker.io` and `quay.io`).

We'll leave this issue open until it's resolved from a _process_ standpoint.

PR #2660, opened 2022-04-13
Publish latest-tagged images when building releases

kevin-bates:
When using the `Makefile` targets to build and publish images for each release, we were not equating the built images with a `latest` tag.  As a result, this needed to be a manual step that, often, fell through the cracks.  This pull request applies the `latest` tag to the corresponding image _when releases are being built_ (i.e., when the `Makefile` variable `TAG` does not equal `"dev"`), and pushes those tagged images to the respective registry.

This was tested by temporarily prefixing the `docker push` commands with `echo` since I didn't want to update any hash values relative to the images already present.  This test ensured that the `if [ "$(TAG)" != "dev" ];` statement was executed correctly.   I also used `make -n publish-container-images` and `make -n TAG=3.7.0 publish-container-images` to visually see the results - which only echos the logic flow although one can't really ensure the `if` statement is properly executed (thus the echo-based test).

Resolves: #2659

 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,9 @@ PYTHON_PIP=$(PYTHON) -m pip
 
 TAG:=dev
 ELYRA_IMAGE=elyra/elyra:$(TAG)
+ELYRA_IMAGE_LATEST=elyra/elyra:latest
 KF_NOTEBOOK_IMAGE=elyra/kf-notebook:$(TAG)
+KF_NOTEBOOK_IMAGE_LATEST=elyra/kf-notebook:latest
 
 # Contains the set of commands required to be used by elyra
 REQUIRED_RUNTIME_IMAGE_COMMANDS?="curl python3"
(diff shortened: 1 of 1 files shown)


ptitzler:
The reason why we didn't do this in the past is that not every `dev` release should be tagged `latest`. 

Problem scenario:
- current version 3.7 (`latest`)
- we need to publish version 3.6.x (or 3.5.x, ...) to fix a major issue

`latest` should always refer to to most current release, not the last recently updated release because it could implicitly cause users to switch between major or minor versions.

akchinSTC:
I misunderstood what others were saying earlier. The release script doesn't publish the container images, that is done manually post-python package release and push to pypi. 
With that said, I don't think we need to make any changes here. This PR will only update the Makefile on master, which will update the tag accordingly since it will always be the 'latest' release. 
Existing maintenance branches, since they will all have their own respective Makefiles without this update, will continue to publish using their `publish` targets without ever updating the `latest` tag, which is what we want.

kevin-bates:
I think we need to incorporate the branch from which the release is being built and assuming that only 'latest" releases are built from the `master` branch.  

> release 3.8.x -> breaks latest

This will occur on a branch other than `master`.  As a result, I'm adding a Makefile variable `LATEST_RELEASE:=0` that, when `1` will trigger the tagging and publication of the `latest` tags.  The release script that modifies the `Makefile` `TAG` variable, will also, conditionally based on `git_branch`, modify the `LATEST_RELEASE` variable.  This form of the `Makefile` will then be tagged with the release and checked into GIT under that tag.  Then, once the release has been pushed to PyPI, the "release operator" will run `make publish-container-images` using this tagged form of the `Makefile` which will appropriately trigger the tagging and publication of the `latest` images.

ptitzler:
I like the idea! If we can, we should avoid the need to patch the `Makefile`, as that might have an unintended side-effect if the "release operator" commits the `Makefile` changes. 

`git rev-parse --abbrev-ref HEAD` will give us the current branch name and could be evaluated in the Makefile itself to determine whether we are on `master` and need to do extra work.

kevin-bates:
> we should avoid the need to patch the Makefile

I agree, but that's what already happens so this fits the current process.  I'd rather not introduce a _process_ change in this PR - especially since I figured it would be a 5-minute effort - which is clearly not the case.

Since you're proposing placing this in the `Makefile`, that seems fine.  I'd rather not change the current "patching" of the Makefile.  I'll go ahead and revert the release script changes, adding `git branch --show-current` to the Makefile publish targets.

ptitzler (approved):
LGTM based on code review.

akchinSTC (approved):

PR #2673, opened 2022-04-20
Update development dockerfiles to use prebuilt extensions

### LINKED ISSUE #2546: reduce size of dev Dockerfiles using prebuilt extensions
Currently our "dev" Dockerfiles use `make install-all` to install the source code into themselves:
- [`etc/docker/elyra/Dockerfile.dev`](https://github.com/elyra-ai/elyra/blob/33c087091ec5612a6d8471a0333719bcfd046c0e/etc/docker/elyra/Dockerfile.dev#L61)
- [`etc/docker/kubeflow/Dockerfile.dev`](https://github.com/elyra-ai/elyra/blob/33c087091ec5612a6d8471a0333719bcfd046c0e/etc/docker/kubeflow/Dockerfile.dev#L58)

This results in all the source code and build-dependencies being left inside the built container image.

After https://github.com/elyra-ai/elyra/pull/2447 you no longer need to run `jupyter lab build` after installing elyra, so our pip wheels are fully self-contained. 

This lets us change our "dev" Dockerfiles into two-stage builds (with two FROM statements) that:
1. Runs `make release` to produce the pip wheels
2. Installs the built pip wheels to a clean container with `pip install ...` (just like the non-dev containers)

akchinSTC:
Adds a new `container-image-dev` target to Makefile
Modifies existing dev dockerfiles to use pre built and packaged extensions.

Fixes #2546 

New make target was run and container images and tags verified. 
existing non dev targets should remain unaffected and should continue to function during release process.
 
           in the file; or

           it.


Makefile
```diff
@@ -29,6 +29,7 @@ SHELL:=/bin/bash
 PYTHON?=python3
 PYTHON_PIP=$(PYTHON) -m pip
 
+ELYRA_VERSION:=$$(python3 -c "import elyra._version; print(elyra._version.__version__)")
```
kevin-bates, line 32: Should the `python3` here reference `$(PYTHON)` instead in case folks don't have a `python3` instance?

This approach captures the version prior to any targets and pulls the version information from the elyra package that is currently _installed_.  This seems susceptible to potentially capturing an older version.  Might a grep-style approach against the local `_version.py` be more definitive?  E.g., something like...
```bash
grep __version__ elyra/_version.py | awk '{print $3}' | sed s/\"//g
```

kevin-bates (commented):
My goodness, this is some serious cleanup - very nice!

I just had one comment about capturing the version.

Makefile
akchinSTC: updated version capture, used a slightly simplified cut with quote delimiter, seems to do the trick

Makefile
```diff
@@ -215,21 +216,38 @@ docs: docs-dependencies ## Build docs
...
+	cp etc/docker/elyra/start-elyra.sh build/dev/docker/start-elyra.sh
+	cp etc/docker/elyra/requirements.txt build/dev/docker/requirements.txt
+	cp dist/elyra-$(ELYRA_VERSION)-py3-none-any.whl build/dev/docker/
+	docker buildx build \
+        --progress=plain \
+        --output=type=docker \
+		--tag docker.io/$(ELYRA_IMAGE) \
+		--tag quay.io/$(ELYRA_IMAGE) \
+		--build-arg TAG=$(ELYRA_VERSION) \
+		build/dev/docker/;
```
kevin-bates, line 231: (This probably isn't a big deal but thought I'd mention it.)  
With these dev targets, if either of the `-image-dev` targets are invoked when the `ELYRA_VERSION` is not a `.devN` version (e.g., `3.8.0`), then the image produced by these targets will overwrite the non-dev images.  Not sure if its worth adding some kind of detection, like confirm `ELYRA_VERSION` contains `".dev"`, prior to proceeding.  Since these targets are referenced by `container-images-dev` and not `container-images` it probably isn't something to be concerned about.  I also haven't looked closely into the differences between the dev and non-dev dockerfiles used by these targets as it could be the case that, when using a released wheel file, the two are essentially equivalent images.

kevin-bates, line 231: (This probably isn't a big deal but thought I'd mention it.)  
With these dev targets, if either of the `-image-dev` targets are invoked when the `ELYRA_VERSION` is not a `.devN` version (e.g., `3.8.0`), then the image produced by these targets will overwrite the non-dev images.  Not sure if its worth adding some kind of detection, like confirm `ELYRA_VERSION` contains `".dev"`, prior to proceeding.  Since these targets are referenced by `container-images-dev` and not `container-images` it probably isn't something to be concerned about.  I also haven't looked closely into the differences between the dev and non-dev dockerfiles used by these targets as it could be the case that, when using a released wheel file, the two are essentially equivalent images.

kevin-bates (approved):

lresende:
Is there a way to use this effort to try to clean up vs adding additional duplication? Now that we are supposedly building proper pre-built images that can be installed, can't we identify the tag name with something like [this](https://github.com/apache/incubator-toree/blob/master/Makefile) and remove the duplicated makefile tasks and dockerfile?

akchinSTC:
> Is there a way to use this effort to try to clean up vs adding additional duplication? Now that we are supposedly building proper pre-built images that can be installed, can't we identify the tag name with something like [this](https://github.com/apache/incubator-toree/blob/master/Makefile) and remove the duplicated makefile tasks and dockerfile?

Id rather keep this separate for the time being and give it time to see if any 

> Is there a way to use this effort to try to clean up vs adding additional duplication? Now that we are supposedly building proper pre-built images that can be installed, can't we identify the tag name with something like [this](https://github.com/apache/incubator-toree/blob/master/Makefile) and remove the duplicated makefile tasks and dockerfile?

@lresende - done

lresende:
Thanks, @akchinSTC, this looks a lot cleaner. I can give it a try on the functionality over the weekend but otherwise looks good.

PR #2679, opened 2022-04-25
Support key-value type list properties in a more generic way

### LINKED ISSUE #2678: Generalize processing of properties consisting of key-value pairs (eg env vars)
Env vars are handled differently than other pipeline properties (e.g. runtime images) where a union of the key/value pairs is taken (preferring the value defined directly on the node property in the case of duplicate keys). As a result, there is special handling for env vars in various parts of the backend code. This will become more of an issue when adding support for PVCs (#2675), which will introduce a new key/value pair type property that requires the same behavior as env vars.

Make processing of properties consisting of key/value pairs more generic. 3 changes likely need to be made:

- Update the pipeline properties template to include some metadata indicating that this property will be one consisting of key/value pairs (which are currently of the format `key=value`)
- Generalize the pipeline defaults propagation step to include the same propagation behavior for all properties of key/value pair type (union with node-level override)
- Generalize the `Operator` object to store or otherwise expose these types of properties in their dict form for access during process/export

kiersten-stokes:
Fixes #2678 

Supports #2675

First, key-value type properties are identified in the canvas JSON with the addition of a key `keyValueEntries` in the `data` stanza. 

The `list` built-in type is extended in this PR in a class called `KeyValueList`. Key-value pair properties will have values of this type. The `KeyValueList` type has various methods that perform conversion of a list of character-separated key-value pairs (e.g. `"key=value"`) to a dictionary and vice-versa. Node and pipeline key-value list properties identified with the `keyValueEntries` metadata mentioned above have their values converted ASAP, during the construction of the `PipelineDefinition` object. A new function `coerce_kv_pair_properties` is added that takes place before default property propagation that makes this conversion from `list` to `KeyValueList` so that later functions dealing with the processing of these values will succeed.

This allows for the removal of some duplicate code and explicit references to env_vars.

A new test case (`test_convert_kv_properties`) was added to ensure all key-value props are converted to `KeyValueList` types.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -447,24 +447,25 @@ class KeyValueList(list):
             kv_dict[key] = value
         return kv_dict
 
-    def dict_to_kv_list(self, kv_dict: Dict) -> "KeyValueList":
+    @classmethod
+    def from_dict(cls, kv_dict: Dict) -> "KeyValueList":
         """
         Convert a set of key-value pairs stored in a dictionary to
         a KeyValueList of strings with the defined separator.
         """
-        str_list = [f"{key}{self._key_value_separator}{value}" for key, value in kv_dict.items()]
+        str_list = [f"{key}{cls._key_value_separator}{value}" for key, value in kv_dict.items()]
         return KeyValueList(str_list)
 
-    def merge_kv_pairs(self, primary_list: "KeyValueList") -> "KeyValueList":
+    @classmethod
+    def merge(cls, primary: "KeyValueList", secondary: "KeyValueList") -> "KeyValueList":
         """
         Merge two key-value pair lists, preferring the values given in the
-        primary_list in the case of a matching key between the two lists.
+        primary parameter in the case of a matching key between the two lists.
         """
-        primary_dict = primary_list.to_dict()
-        secondary_dict = self.to_dict()
+        primary_dict = primary.to_dict()
+        secondary_dict = secondary.to_dict()
 
-        merged_list = self.dict_to_kv_list({**secondary_dict, **primary_dict})
-        return KeyValueList(merged_list)
+        return KeyValueList.from_dict({**secondary_dict, **primary_dict})
 
     @staticmethod
     def log_message(msg: str, logger: Optional[Logger] = None, level: Optional[int] = logging.DEBUG):
--- a/elyra/pipeline/pipeline.py
+++ b/elyra/pipeline/pipeline.py
@@ -21,6 +21,8 @@ from typing import Dict
 from typing import List
 from typing import Optional
 
+from elyra.pipeline.pipeline_constants import KEY_VALUE_SEPARATOR
+
 # TODO: Make pipeline version available more widely
 # as today is only available on the pipeline editor
 PIPELINE_CURRENT_VERSION = 7
@@ -46,7 +48,7 @@ class Operation(object):
     ) -> "Operation":
         """Class method that creates the appropriate instance of Operation based on inputs."""
 
-        if classifier in Operation.generic_node_types:
+        if Operation.is_generic_operation(classifier):
             return GenericOperation(
                 id, type, name, classifier, parent_operation_ids=parent_operation_ids, component_params=component_params
             )
@@ -89,7 +91,7 @@ class Operation(object):
         self._classifier = classifier
         self._name = name
         self._parent_operation_ids = parent_operation_ids or []
-        self._component_params = component_params
+        self._component_params = component_params or {}
         self._doc = None
 
         # Scrub the inputs and outputs lists
@@ -177,20 +179,6 @@ class Operation(object):
         )
 
     @staticmethod
-    def _log_info(msg: str, logger: Optional[Logger] = None):
-        if logger:
-            logger.info(msg)
-        else:
-            print(msg)
-
-    @staticmethod
-    def _log_warning(msg: str, logger: Optional[Logger] = None):
-        if logger:
-            logger.warning(msg)
-        else:
-            print(f"WARNING: {msg}")
-
-    @staticmethod
     def _scrub_list(dirty: Optional[List[Optional[str]]]) -> List[str]:
         """
         Clean an existing list by filtering out None and empty string values
@@ -202,8 +190,8 @@ class Operation(object):
         return [clean for clean in dirty if clean]
 
     @staticmethod
-    def is_generic_operation(operation_type) -> bool:
-        return operation_type in Operation.generic_node_types
+    def is_generic_operation(operation_classifier) -> bool:
+        return operation_classifier in Operation.generic_node_types
 
 
 class GenericOperation(Operation):
@@ -231,7 +219,7 @@ class GenericOperation(Operation):
                                  a non-standard operation instance
 
         Component_params for "generic components" (i.e., those with one of the following classifier values:
-        ["execute-notebook-node", "execute-python-node", "exeucute-r-node"]) can expect to have the following
+        ["execute-notebook-node", "execute-python-node", "execute-r-node"]) can expect to have the following
         entries.
                 filename: The relative path to the source file in the users local environment
                          to be executed e.g. path/to/file.ext
@@ -275,6 +263,11 @@ class GenericOperation(Operation):
         self._component_params["gpu"] = component_params.get("gpu")
         self._component_params["memory"] = component_params.get("memory")
 
+        self.generic_kv_properties_from_template = ...
+        for param_name, value in self._component_params.items():
+            if param_name in self.generic_kv_properties_from_template:
+                self._component_params[param_name] = KeyValueList(value)
+
     @property
     def name(self) -> str:
         if self._name == os.path.basename(self.filename):
@@ -302,8 +295,12 @@ class GenericOperation(Operation):
         return self._component_params.get("include_subdirectories")
 
     @property
-    def env_vars(self) -> Optional[List[str]]:
-        return self._component_params.get("env_vars")
+    def env_vars(self) -> Optional["KeyValueList"]:
+        return KeyValueList(self._component_params.get("env_vars"))
+
+    @property
+    def volume_mounts(self) -> Optional["KeyValueList"]:
+        return KeyValueList(self._component_params.get("mounted_volumes", []))
 
     @property
     def cpu(self) -> Optional[str]:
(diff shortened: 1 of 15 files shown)


elyra/pipeline/pipeline.py
```diff
@@ -440,3 +406,72 @@ def __eq__(self, other: "Pipeline") -> bool:
...
+    def merge_kv_pairs(self, primary_list: "KeyValueList") -> "KeyValueList":
+        """
+        Merge two key-value pair lists, preferring the values given in the
+        primary_list in the case of a matching key between the two lists.
+        """
+        primary_dict = primary_list.to_dict()
+        secondary_dict = self.to_dict()
+
+        merged_list = self.dict_to_kv_list({**secondary_dict, **primary_dict})
+        return KeyValueList(merged_list)
```
kevin-bates, line 467: I find it more intuitive if this action is performed against the instance on which the method is called.  In this case, I think it would be the current instance is the "primary", so an instance is basically being merged with a "secondary".   I'm just not sure how to "adjust" self to the new list.

Perhaps it would be cleaner to also let this be a class factory pattern as well, using another class method:
```suggestion
    @classmethod
    def merge(self, primary: "KeyValueList", secondary: "KeyValueList) -> "KeyValueList":
        """
        Merge two key-value pair lists, preferring the values given in the
        primary parameter in the case of a matching key between the two lists.
        """
        primary_dict = primary.to_dict()
        secondary_dict = secondray.to_dict()
        return KeyValueList.from_dict({**secondary_dict, **primary_dict})
```

elyra/pipeline/pipeline_definition.py
```diff
@@ -326,6 +336,7 @@ def __init__(
         if validate:
             self.validate()
 
+        self.coerce_kv_pair_properties()
```
kevin-bates, line 339: Per your earlier comment in our side conversation, I think it would be better to "protect" this method (with an _ prefix) and call it within `propagate_pipeline_default_properties()` since its used exclusively there.  Looking at the two methods, they do appear to overlap wrt to walking the pipeline properties and nodes and I wonder if the KVL coercion could be performed _as_ the properties are being walked thereby eliminating the need for a separate method (and an additional pass across the properties).  Might be worth exploring.

elyra/pipeline/pipeline_definition.py
```diff
@@ -440,6 +458,44 @@ def validate(self) -> list:
...
+        generic_node_properties = PipelineDefinition.get_canvas_properties_from_template(
+            package_name="templates/components", template_name="generic_properties_template.jinja2"
+        )
+        for node in self.pipeline_nodes:
+            if not Operation.is_generic_operation(node.op):
+                continue
+
+            for property_name, value in generic_node_properties["current_parameters"].items():
+                if property_name.startswith("elyra_"):
+                    property_name = property_name.replace("elyra_", "")
```
kevin-bates, line 488: I understand stripping off the `elyra_` prefix in order to check if this property is a KV pairs property, but that's only because the [`elyra_` prefix is stripped when building the `kv_pair_properties` list](https://github.com/elyra-ai/elyra/pull/2679/files#diff-79810d1ae48b2ef6e0b3b2602ffdf1ca2cf2a452f3695c67168495fbe73e3eafR649-R650) anyway.  Since we don't appear to persist the non-prefixed values, can we forgo both of these "stripping" locations?

I'm not finding any place in the old code where these prefixes are removed so just wondering why we need to do that.

elyra/pipeline/pipeline_definition.py
```diff
@@ -450,38 +506,18 @@ def propagate_pipeline_default_properties(self):
...
+                    continue
+
+                node_value = node.get_component_parameter(pipeline_default_prop)
+                if not node_value:
+                    node.set_component_parameter(pipeline_default_prop, pipeline_default_value)
+                    continue
+
+                if isinstance(pipeline_default_value, KeyValueList):
+                    merged_list = pipeline_default_value.merge_kv_pairs(node_value)
+                    node.set_component_parameter(pipeline_default_prop, merged_list)
```
kevin-bates, line 520: We should probably ensure that `node_value` is also an instance of `KeyValueList`.  If using the class factory approach described previously, this would become:
```suggestion
                if isinstance(pipeline_default_value, KeyValueList):
                    assert(isinstance(node_value, KeyValueList), "'node_value' is not of type KeyValueList!")
                    merged_list = KeyValueList.merge(node_value, pipeline_default_value)
                    node.set_component_parameter(pipeline_default_prop, merged_list)
```

kevin-bates (commented):
This looks good so far. I had some suggestions that might make things cleaner from a class perspective.

elyra/pipeline/pipeline_definition.py
kiersten-stokes: Thanks for the reminder - this slipped my mind!

elyra/pipeline/pipeline_definition.py
kiersten-stokes: Unfortunately I don't think we can. Here and in the function you linked, we're taking property names directly from the templates where the `elyra_` prefix is present. The property name that we get during submit/export (and that we're checking for matches with the key-value pair property names), doesn't have the prefix - the frontend removes it at some point. 

I wonder if we could forego the `elyra_` prefix altogether at some point and just indicate this in the canvas JSON data stanza in some other way.

elyra/pipeline/pipeline.py
kiersten-stokes: That would make more sense. And agreed on the primary vs secondary meaning.

elyra/pipeline/pipeline_definition.py
kevin-bates: I see.  Ok, no worries.
> I wonder if we could forego the elyra_ prefix altogether at some point and just indicate this in the canvas JSON data stanza in some other way.

This sounds like something that might be a helpful update at some point.  
cc: @ajbozarth

elyra/pipeline/pipeline.py
```diff
@@ -440,3 +406,73 @@ def __eq__(self, other: "Pipeline") -> bool:
...
+                    f"separator character: '{self._key_value_separator}'."
+                )
+
+            key, value = kv.split(self._key_value_separator, 1)
+            if not key.strip():
+                KeyValueList.log_message(f"Skipping inclusion of property '{kv}': no key found", logger, logging.WARN)
+                continue
+            if not value:
+                KeyValueList.log_message(
+                    f"Skipping inclusion of property '{key}': no value specified", logger, logging.INFO
```
kevin-bates, line 443: I'm not sure `INFO` is the right severity for this one - thinking more `DEBUG`.  The reason being is that one of these log messages will be produced for every environment variable name that is programmatically parsed from the generic component since none of those results includes a value (and assuming the user didn't follow up on that by manually adding a value).  When `DEBUG` is not enabled, I'm not sure it's useful information since that is essentially default behavior.

elyra/pipeline/pipeline.py
```diff
@@ -440,3 +406,73 @@ def __eq__(self, other: "Pipeline") -> bool:
...
+            if not kv:
+                continue
+
+            if self._key_value_separator not in kv:
+                raise ValueError(
+                    f"Property {kv} does not contain the expected "
+                    f"separator character: '{self._key_value_separator}'."
+                )
+
+            key, value = kv.split(self._key_value_separator, 1)
```
ptitzler, line 437: We should remove leading and trailing whitespaces in key and value given the current use case scenarios.

elyra/pipeline/pipeline.py
kiersten-stokes: Yeah, I agree. I'm actually surprised I never noticed this being logged at INFO before now (when this functionality resided in `Operation`

elyra/pipeline/pipeline_definition.py
```diff
@@ -440,48 +457,70 @@ def validate(self) -> list:
...
+        """
+        Convert pipeline- and node-level list properties that have been identified
+        as sets of key-value pairs from a plain list type to the KeyValueList list
+        type. Properties are identified as key-value pair properties when they set
+        keyValueEntries to true in the pipeline JSON canvas properties definition.
+        """
+        all_pipeline_properties = PipelineDefinition.get_canvas_properties_from_template(
+            package_name="templates/pipeline", template_name="pipeline_properties_template.jinja2"
+        )
+        kv_pair_properties = PipelineDefinition.get_key_value_pair_properties(all_pipeline_properties)
```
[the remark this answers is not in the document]
kiersten-stokes: Did a good bit of re-factoring in `pipeline_definition.py` in the latest commit to address this!

kiersten-stokes:
Moved this into WIP right now because I haven't been able to test the recent changes as thoroughly as I'd like to. I still have to add new test case(s) as well.

elyra/pipeline/pipeline_definition.py
```diff
@@ -457,58 +488,26 @@ def validate(self) -> list:
...
-        for pipeline_default_prop, pipeline_default_value in pipeline_default_properties.items():
-            if pipeline_default_prop in kv_pair_properties:
-                # Coerce plain list to KeyValueList
-                self.set_pipeline_default_property(pipeline_default_prop, KeyValueList(pipeline_default_value))
 
-        generic_node_properties = PipelineDefinition.get_canvas_properties_from_template(
-            package_name="templates/components", template_name="generic_properties_template.jinja2"
-        )
+        # Convert any key/value list pipeline default properties to the KeyValueList type
+        self.primary_pipeline.convert_kv_pair_properties(kv_pair_properties)
```
kevin-bates, line 500: This is a nit, but I think it increases readability/maintainability.  Could we move the initialization of `pipeline_default_properites` directly before the `for` loop since its only used within, then group the other two statements since they're dependent?

```python
        # Convert any key/value list pipeline default properties to the KeyValueList type
        kv_properties = PipelineDefinition.get_kv_properties()
        self.primary_pipeline.convert_kv_properties(kv_properties)

        pipeline_default_properties = self.primary_pipeline.get_property(pipeline_constants.PIPELINE_DEFAULTS, {})
````

File: elyra/pipeline/pipeline_definition.py [hunk 9]
```diff
@@ -638,14 +612,18 @@ def get_canvas_properties_from_template(package_name: str, template_name: str) -
         return json.loads(output)
 
     @staticmethod
-    def get_key_value_pair_properties(canvas_properties) -> List[str]:
+    def get_pipeline_kv_pair_properties() -> List[str]:
```
kevin-bates, line 615: Since this method is on `PipelineDefintion`, I think we can safely assume the kv properties are relative to the pipeline and drop that from the method name as well, leaving...
```python
 def get_kv_properties() -> List[str]:
```

elyra/pipeline/pipeline_definition.py
```diff
@@ -188,6 +192,22 @@ def set_property(self, key: str, value: Any):
 
         self._node["app_data"]["properties"][key] = value
 
+    def convert_kv_pair_properties(self, kv_pair_properties: List[str]):
```
kevin-bates, line 195: Would you mind reducing `kv_pair` to `kv` throughout?  kv-pair seems a bit redundant since key-value is the pair.
```suggestion
    def convert_kv_properties(self, kv_pair_properties: List[str]):
```

elyra/pipeline/pipeline_definition.py
kiersten-stokes: I had the same thought while I was refactoring, not sure why I didn't pull the trigger on that! 😀

kevin-bates (approved):
LGTM - thank you @kiersten-stokes!

ptitzler:
Perhaps I am missing something, but should this processing be still required https://github.com/kiersten-stokes/elyra/blob/generic-kv-props/elyra/pipeline/processor.py#L596-L604 when volume mounts are exposed as kv properties?

PR #2683, opened 2022-04-27
Update print stmts to use f-string

sandhya-nayak:
Update print and debug statements to use f-strings.

Ran make clean, make install, make docs, make test.

 
           in the file; or

           it.

### DIFF
--- a/docs/source/developer_guide/logging.md
+++ b/docs/source/developer_guide/logging.md
@@ -32,15 +32,15 @@ since the Jupyter Notebook server uses Tornado as its web framework. <p>
 Use INFO level for routines(functions), like when handling requests or state changes and DEBUG for troubleshooting and diagnostics.<p>
 ```
   def foo (arg):
-    self.debug.info("doing something with this %s", arg)
+    self.debug.info(f"doing something with this {arg}")
     do something with arg
-    self.log.info("Result of arg is ... %s", result)
+    self.log.info(f"Result of arg is ... {result}")
 ```    
 Use WARNING level when it is important, but not an error, for example, when a user attempts to login with an incorrect password<p>
 ```
   def authenticate(username, password):
     if username and password doesn't work:
-        self.log.warn("Failed login attempt from %s : $s", username, password)
+        self.log.warn(f"Failed login attempt from {username} : {password}")
 ```     
 Use ERROR level when something is actually wrong, like when an exception is thrown, IO operation failure or connectivity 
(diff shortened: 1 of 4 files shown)


ptitzler:
I did a quick scan and it appears that there are two `<string>.format(` instances left in the source code:
- https://github.com/elyra-ai/elyra/blob/90fb4994d19ae4f82884d6446cea92b2468ec02e/elyra/tests/pipeline/resources/components/bash_operator.py#L120
- https://github.com/elyra-ai/elyra/blob/90fb4994d19ae4f82884d6446cea92b2468ec02e/elyra/metadata/metadata_app.py#L770 (@kiersten-stokes, @kevin-bates I remember we discussed this one when we tried to resolve all issues that black reported.  Was there a reason why this one could not be rewritten to use an f-string?)

kevin-bates:
> I remember we discussed this one when we tried to resolve all issues that black reported. Was there a reason why this one could not be rewritten to use an f-string?

I don't think we spent much time looking into it, but it looks like f-strings are captured when "executed" while `format` enables a "functional behavior" that would allow the "string" to be re-constructed. So even using a dummy `schemaspace` placeholder wherever the subclass sets up its description string won't really work.  That said, the subclass may know its `schemaspace` (in this particular scenario) and we could probably forgo the "dynamism" in play here.
```python
$ python
Python 3.10.2 | packaged by conda-forge | (main, Mar  8 2022, 16:02:23) [Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> subcommand_description = f"This is schemaspace '{schemaspace}'."
  File "<stdin>", line 1, in <module>
NameError: name 'schemaspace' is not defined
>>> schemaspace = None
>>> subcommand_description = f"This is schemaspace '{schemaspace}'."
>>> print(subcommand_description)
This is schemaspace 'None'.
>>> schemaspace = 'foo'
>>> print(subcommand_description)
This is schemaspace 'None'.
>>> 
```

The same may be true in the first example.  `format()` essentially enables a "functional string", while `f-string` is a one-time "capture" of the parameters (as evidenced in the REPL example).  But you may want to try converting the first and see.  (Too lazy [and late] to try _that_ in the REPL. :smile:)

sandhya-nayak:
@ptitzler So we're leaving this one in place?
https://github.com/elyra-ai/elyra/blob/710aee782c8a6d2f6f7dbe4e16829ad7932bc343/elyra/metadata/metadata_app.py#L770

I made the other change that you had suggested.

PR #2709, opened 2022-05-05
Develop Elyra against the Jupyterlab source repo

ajbozarth:
Adds documentation and some additions to the build scripts to allow
Elyra developers to develope Elyra against a local build of
Jupyterlab.

Updates include:
- Updated documentation with step by step instructions on how to
step up the devlopment environment
- Updating our `page_config.json` to the new format and location.
We were previously using a deprecated format and location
uncompatable with this feature.
- Added a way to run `clean-jupyterlab.sh` without installing a new
version of jupyterlab
- Added new npm targets and make targets to leverage a yarn link
version of `@jupyterlab/builder`. This was added to the existing
support for `make dev-link`, but the order of make targets had to
be rearranged to make sure lerna was installed before calling it.
- Made a small backwards compatable change to fix a type conflict
found when running against jupyterlab 4.0.0 alpha

These changes are blocked by
- [x] https://github.com/jupyterlab/jupyterlab/pull/12533

without that change you also need to update the version of
`@jupyterlab/builder` in every package.json to prevent an error

           in the file; or

           it.


ajbozarth:
@kevin-bates is there any way to remove this dependency on ipython_genutils? https://github.com/elyra-ai/elyra/blob/720fb3663ef6c30bd420d320b5f1b32325a8aff6/elyra/metadata/metadata.py#L22

It causes issues in completely clean environments since lab no longer depends on it so we're no longer getting the import second hand. Also the readme for that library is literally just a "do not use this" message. I wasn't sure how to remove it since I don't quite understand what it's doing in our code.

kevin-bates:
> @kevin-bates is there any way to remove this dependency on ipython_genutils?

Yes, this dependency should be removed.

> It causes issues in completely clean environments since lab no longer depends on it so we're no longer getting the import second hand. Also the readme for that library is literally just a "do not use this" message. 

This package has been removed from several projects within the ecosystem.  I didn't think to check we'd be using it but, well, we are. 

> I wasn't sure how to remove it since I don't quite understand what it's doing in our code.

The package, in general, provides utility methods mostly geared toward bridging the gap between python 2 and 3 codebases.  However, there are some utilities that are just that, utilities.  In this case, `import_item` essentially _hydrates_ a [class definition from its dotted notation string](https://github.com/elyra-ai/elyra/blob/8d4fd8c5fafcdf1159c692324bc3520c15fd1394/elyra/metadata/metadata.py#L99-L101) and is used by the metadata service to enable [_per-schema instances_](https://github.com/elyra-ai/elyra/blob/8d4fd8c5fafcdf1159c692324bc3520c15fd1394/elyra/metadata/schemas/kfp.json#L8) in which application-specific logic can take place.

I should have a PR for this shortly.

ajbozarth:
@karlaspuldaro I believe I've addressed all your documentation comments if you would like to take another look

ajbozarth:
I also merged master to grab @kevin-bates fix in https://github.com/elyra-ai/elyra/pull/2710 so this will also work on completely fresh conda envs as well now

karlaspuldaro (approved):
Thank you for the updates! The steps to build Elyra against lab src repo work as expected.

Issue #2674, opened 2022-04-20
Confirm that `env_vars` contains no keys before writing pipelines


MichaelTiemannOSC:
When creating piplines, it is important to strip out any access keys or secret from the env_vars data.

```
 "pipelines": [
    {
      "id": "primary",
      "nodes": [
        {
          "id": "...",
          "type": "execution_node",
          "op": "execute-notebook-node",
          "app_data": {
            "component_parameters": {
              "filename": "notebooks/notebook_filename.ipynb",
              "outputs": [],
              "env_vars": [ NO KEYS HERE ]
```

kiersten-stokes:
We've decided to move forward in implementing support for Kubernetes secrets. A single environment variable backed by a secret necessitates entering 3 string values:

1. the environment variable name (as it appears in the notebook or script)
2. the secret name (as shown in a `kubectl get secrets` command on the runtime instance)
3. the secret key (as a single secret can have multiple keys, each of which has its own value)

Based on some offline discussion, there are a few items that we need to finalize before implementing a full solution:

Design:
- what will the new control for a multi-valued property (e.g. a Kubernetes secret property) look like? and will it be considered a 'keyValueEntries' type?
- what will the new pipeline JSON look like for the above control?
- how will copying/pasting such a property between nodes work?
- how will the case be handled where an environment variable and kubernetes secret variable with the same name both have values?
- should a refresh button be available (similar to how it is with vanilla env vars now)?

Admin:
- should we merge the backend code first and followup with a frontend-focused PR (with some backend changes required)?

The backend code has been mostly implemented; draft PR to come soon.

ptitzler:
> should a refresh button be available (similar to how it is with vanilla env vars now)?

What would be the purpose of the button? Elyra does not have any Kubernetes connectivity and wouldn't be able to auto-detect or validate secrets.

kiersten-stokes:
> What would be the purpose of the button? Elyra does not have any Kubernetes connectivity and wouldn't be able to auto-detect or validate secrets.

It would behave just like the env vars one currently does. Basically, parse the notebook or script for certain uses (e.g. `os.getenv`) and populate the **name** of the env var only - the secret name and key still have to be entered manually (like the value does for 'vanilla' env vars now). Right now, these cases will appear in the 'Environment Variables' property area but not the 'Kubernetes Secrets' area, which could be a potential cause for confusion.

kevin-bates:
[...]

PR #2715, opened 2022-05-06
Add backend support for Kubernetes secrets environment variables

kiersten-stokes:
Provides backend support for #2674 

Frontend piece will be delivered when RJSF support is available (#2464)

### What changes were proposed in this pull request?l
These changes add support for Kubernetes-secret-backed environment variables. This PR is currently working off the assumption that secrets will be input as a string in a strict format shown below:

```
ENV_VAR=secret_name:secret_key
```

Some backend changes will be required when the frontend piece is delivered. 

First, a "kubernetes_secret" list (`StringArrayControl`) property is added to both the pipeline and generic node properties. Behavior works similarly to environment variables and mounted volumes: the property is treated as a `KeyValueList` type after pipeline submission, propagating the pipeline default values when applicable.

One additional consideration is necessary for Kubernetes secrets. A pipeline may need to be executed both locally and remotely. In the case of a local submission, any Kubernetes secrets-based variables will never be accessed. Instead, the corresponding variable in the Environment Variable property should have a value assigned. In this case of remote execution, however, any environment variable that has a value in both the Environment Variables property and the Kubernetes Secrets property will defer to using the Kubernetes secret. The corresponding Environment Variable key/value pair will be stripped from the list. This avoids sending any sensitive information for remote execution, but at the same time allows for local execution of the same files if desired (as long as the variable is set appropriately in both locations). This occurs during `propagate_pipeline_default_properties`.

After the relevant properties are converted to `KeyValueList` types and propagated from pipeline-level to node-level, each secret and volume is then further converted to their well-defined types: `KubernetesSecret` and `VolumeMount`. This is done in `convert_data_class_properties` on the `Node` object of a `PipelineDefinition`. No validation is performed in this function; the attributes are simply propagated to their associated types. (This function can perhaps be made less explicit when the pipeline JSON structure changes as a result of the new RJSF editor. E.g. maybe we can add metadata to the properties templates that indicates which data class to use and properties can be fed directly to their constructor without the need to parse). Doing this conversion in the construction of the `PipelineDefinition` object will allow for them to be accessed from the pipeline CLI (specifically the `describe` command) in a follow-up PR.

During processing, attributes of these new data classes are now accessed directly. The `KubernetesSecret` data class object has properties of with `env_var`, `name`, and `key` attributes. These attributes are converted to the appropriate form for each runtime type (in `kfp/operator.py init` for KFP and in the DAG template for Airflow). The `VolumeMount` data class object has properties `path` and `pvc_name` attributes is returned and accessed appropriately during processing.

Finally, validation checks have been added to the validation service that check for valid Kubernetes resource names (for PVC names in volumes and secret names in secrets) and valid Kubernetes secret key format. This has the benefit that we are catching these invalid values early on before any more expensive processing occurs.

Documentation has been updated to reflect the new property.

- [x] Add test to validation service to check volumes and secrets for valid values
- [x] Remove `test_processor_base` anymore?

 
           in the file; or

           it.


elyra/kfp/operator.py
```diff
@@ -106,6 +109,7 @@ def __init__(
           gpu_limit: maximum number of GPUs allowed for the operation
           workflow_engine: Kubeflow workflow engine, defaults to 'argo'
           volume_mounts: data volumes to be mounted
+          kubernetes_secrets: secrets to be made available as environment variables
```
ptitzler, line 112: Kubernetes secrets can also be mounted as volumes. In case we also support that in a future release we should probably pick a more discriminating variable name now that makes it possible to distinguish between secrets that are exposed as an env variable and secrets that are mounted as a volume.

elyra/pipeline/processor.py
```diff
@@ -605,3 +608,31 @@ def _get_volume_mounts(self, operation: Operation) -> Optional[Dict[str, str]]:
...
+        valid_secrets = []
+        secrets = operation.component_params.get(KUBERNETES_SECRETS)
+        if secrets and isinstance(secrets, KeyValueList):
+            for env_var_name, secret in secrets.to_dict().items():
+                secret_tuple = secret.split(":", 1)
+                if len(secret_tuple) != 2:
+                    self.log.warning(f"Ignoring invalid secret for '{env_var_name}': missing secret name and/or key.")
+                    continue
+                secret_name, secret_key = secret_tuple[0].strip(), secret_tuple[1].strip()
+                if not secret_name:
```
ptitzler, line 629: We also need to make sure that the user-provided secret name meets the Kubernetes naming convention. If it doesn't, we have to ignore it because it would cause a failure during pipeline execution. See https://github.com/elyra-ai/elyra/blob/master/elyra/util/kubernetes.py for related code

kevin-bates:
> This PR is currently working off the assumption that secrets will be input as a string in a strict format shown below:
>```
>ENV_VAR_NAME=secret_name:secret_key
>```
>Since this is not a long-term solution, some backend changes will be required when the frontend piece is delivered.

Could we say that this format is the "contract" that when the front-end finalizes its changes, it is responsible for producing a list of strings in this format?  That way, no back-end changes should be required.

elyra/kfp/operator.py
kevin-bates: When "mounted as a volume" are those particular, secret-based, PV definitions differentiated from regular PV definitions? That is, do we need to have an additional "type" of mount or secret?

I agree that a distinguishing name for the env-based field is probably warranted.  Just curious about how distinguishing things are.

elyra/pipeline/pipeline.py
```diff
@@ -487,6 +487,24 @@ def merge(cls, primary: "KeyValueList", secondary: "KeyValueList") -> "KeyValueL
...
         return KeyValueList.from_dict({**secondary_dict, **primary_dict})
 
+    @classmethod
+    def difference(cls, minuend: "KeyValueList", subtrahend: "KeyValueList") -> "KeyValueList":
+        """
+        Given two lists, convert to dictionaries and remove any keys found in the
+        subtrahend from the minuend, if present.
+
+        :param minuend: list to be subtracted from
+        :param subtrahend: list whose keys will be removed from the minuend
```
kevin-bates, line 497: :nerd:

elyra/pipeline/processor.py
kiersten-stokes: I was going to ask you about this - I'll add in some calls to this utility function!

elyra/kfp/operator.py
kiersten-stokes: Yeah totally agree about the distinguishing names - I was thinking that during the discussion yesterday.

> When "mounted as a volume" are those particular, secret-based, PV definitions differentiated from regular PV definitions? That is, do we need to have an additional "type" of mount or secret?

I believe the two different types are consumed/accessed in completely different ways, making an additional type unnecessary. But I may be misunderstanding, so I'd defer to @ptitzler expertise

ptitzler: In the running container this is exposed like other volumes are. afaik the consumer does not have any awareness what the source is, just like the consumer doesn't know wether an env variable value was derived from a constant or a kubernetes resource/secret. I haven't tested what happens if one tries to mount a volume and a secret to the same mount point in the container, but would expect it to fail.

elyra/pipeline/pipeline.py
```diff
@@ -487,6 +487,24 @@ def merge(cls, primary: "KeyValueList", secondary: "KeyValueList") -> "KeyValueL
 
         return KeyValueList.from_dict({**secondary_dict, **primary_dict})
 
+    @classmethod
+    def difference(cls, minuend: "KeyValueList", subtrahend: "KeyValueList") -> "KeyValueList":
+        """
+        Given two lists, convert to dictionaries and remove any keys found in the
+        subtrahend from the minuend, if present.
```
kevin-bates, line 494:
```suggestion
        Given two KeyValueLists, convert to dictionaries and remove any keys found in the
        second (subtrahend) from the first (minuend), if present.
```
By adding a positional notation, it might help folks with the unusual (albeit correct) terminology.

kiersten-stokes:
> Could we say that this format is the "contract" that when the front-end finalizes its changes, it is responsible for producing a list of strings in this format? That way, no back-end changes should be required.

I remember when this came up in discussion that there was some pushback on this, but I can't remember why exactly. One thing that we were worried about is the case where a colon would appear in the secret name or key, but I believe Kubernetes requires that both of those values are completely alphanumeric, so I don't think that would be an issue.

kevin-bates:
Yeah, all I'm saying is that how these are displayed - whether they be 3 separate fields (preferred) or not, is independent of the format in which they are sent to the server and that the server should dictate that format.  If we dictate the format to be an "encoded" string, then we should certainly pick "separators" that can't appear in the sub-strings - which seems to be the case.  I think we're good.

elyra/pipeline/processor.py
```diff
@@ -605,3 +608,31 @@ def _get_volume_mounts(self, operation: Operation) -> Optional[Dict[str, str]]:
...
+    def _get_kubernetes_secrets(self, operation: Operation) -> Optional[List[Tuple]]:
+        """
+        Loops through an Operation's kubernetes secrets to strip whitespace
+        and re-format as a tuple.
+
+        :param operation: the operation to check for secrets
+        :return: tuple of env var name, secret name, and secret key
+        """
+        valid_secrets = []
+        secrets = operation.component_params.get(KUBERNETES_SECRETS)
```
ptitzler, line 621: Ideally the 'client' (in this case the processor, but the CLI will need this too because this PR introduces a new dependency that `elyra-pipeline describe` needs to expose) shouldn't have to deal with decomposing a KV-pair into its parts. Code maintenance would be a lot easier, if (for example) this call would return `list(dict)` or `list(some-class)`. The dict or class would encapsulate only valid definitions because there is nothing client can do with an invalid definition.

elyra/templates/components/generic_properties_template.jinja2
```diff
@@ -189,6 +193,22 @@
...
+          "custom_control_id": "StringArrayControl",
+          "label": {
+            "default": "Kubernetes Secrets"
+          },
+          "description": {
+            "default": "Kubernetes secrets to make available as environment variables to this node. The secret name and key given must be present in the Kubernetes namespace where the node is executed or this node will not run.",
+            "placement": "on_panel"
+          },
+          "data": {
+            "placeholder": "env_var_name=secret_name:secret_key",
```
ptitzler, line 208: To be consistent with the input format for environment variables, we should use `env_var` instead of `env_var_name` now that we won't have a dedicated input widget.

```suggestion
            "placeholder": "env_var=secret_name:secret_key",
```

elyra/templates/pipeline/pipeline_properties_template.jinja2
```diff
@@ -99,6 +103,22 @@
...
+        "custom_control_id": "StringArrayControl",
+        "label": {
+          "default": "Kubernetes Secrets"
+        },
+        "description": {
+          "default": "Kubernetes secrets to make available as environment variables to all nodes.",
+          "placement": "on_panel"
+        },
+        "data": {
+          "placeholder": "env_var_name=secret_name:secret_key",
```
ptitzler, line 118:
```suggestion
          "placeholder": "env_var=secret_name:secret_key",
```

elyra/tests/pipeline/resources/properties.json
```diff
@@ -189,6 +193,22 @@
...
+        "custom_control_id": "StringArrayControl",
+        "label": {
+          "default": "Kubernetes Secrets"
+        },
+        "description": {
+          "default": "Kubernetes secrets to make available as environment variables to this node. The secret name and key given must be present in the Kubernetes namespace where the node is executed or this node will not run.",
+          "placement": "on_panel"
+        },
+        "data": {
+          "placeholder": "env_var_name=secret_name:secret_key",
```
ptitzler, line 208:
```suggestion
          "placeholder": "env_var=secret_name:secret_key",
```

elyra/pipeline/pipeline.py
```diff
@@ -314,6 +326,79 @@ def __eq__(self, other: "GenericOperation") -> bool:
     def _validate_range(self, value: str, min_value: int = 0, max_value: int = sys.maxsize) -> bool:
         return int(value) in range(min_value, max_value)
 
+    @staticmethod
+    def get_valid_volume_mounts(
```
ptitzler, line 330: Should this be an instance method `get_valid_volume_mounts()` to make it easier for consumers of the `Pipeline` object? Right now consumers have to make two calls - one to fetch the list and one to coerce it into the desired data structure. If there was a consumer that did require access to the raw data, it would still be able to gain access via `get_component_parameter(...)`

kiersten-stokes: Well this is a result of the issue with potentially using these methods from the `describe` command, for example. The processors access their parameter values from the `Operation` object, but `describe` accesses parameters from the `PipelineDefinition` object (and each have their own method of providing that value). I haven't fully checked yet to see what the potential implications of using a `Pipeline` object instead of a `PipelineDefinition` object in `describe` would be yet, but I'm not sure it would be as straightforward as it sounds 😄

elyra/pipeline/pipeline.py
```diff
@@ -314,6 +326,79 @@ def __eq__(self, other: "GenericOperation") -> bool:
...
+                    continue
+                secret_name, secret_key = secret_tuple[0].strip(), secret_tuple[1].strip()
+                if not is_valid_kubernetes_resource_name(secret_name):
+                    msg = (
+                        f"Ignoring invalid secret for '{env_var_name}': the secret name "
+                        f"'{secret_name}' is not a valid Kubernetes resource name."
+                    )
+                    Operation.log_message(msg=msg, logger=logger, level=logging.WARN)
+                    continue
+                if not is_valid_kubernetes_resource_name(secret_key):
```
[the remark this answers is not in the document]
kiersten-stokes: Ha, so keys allow underscores and names don't! I somehow missed this when I first read through it.

elyra/pipeline/pipeline.py
kevin-bates: Can we move these to the base `Operation` class in preparation for their general availability?

elyra/util/kubernetes.py
```diff
@@ -37,3 +38,16 @@ def is_valid_kubernetes_resource_name(name: str) -> bool:
...
             return False
     return True
+
+
+def is_valid_kubernetes_key_name(name: str) -> bool:
+    """
+    Returns a truthy value indicating whether name meets the kubernetes
+    naming constraints, as outlined in the link below.
+
+    https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data
```
kevin-bates, line 48: It appears secrets and volumes are "validated" just prior to platform submission.  Should these kinds of resources (names and keys) be validated during pipeline validation and converted to their "usable forms" then?  If so, we should probably ensure length is `< MiB`.

elyra/pipeline/pipeline_definition.py
```diff
@@ -515,13 +580,14 @@ def propagate_pipeline_default_properties(self):
...
+        pipeline_default_properties = self.primary_pipeline.get_property(PIPELINE_DEFAULTS, {})
         for node in self.pipeline_nodes:
             if not Operation.is_generic_operation(node.op):
                 continue
 
-            # Convert any key-value list node properties to the KeyValueList type
-            node.convert_kv_properties(kv_properties)
+            # Convert any key-value list node properties to the KeyValueList type if not done already
+            if node.kv_properties_not_converted(kv_properties):
+                node.convert_kv_properties(kv_properties)
```
kevin-bates, line 590: Seems like `convert_kv_properties()` could have its loops structured like `kv_properties_not_converted()` and just do the conversion or exit if a `KeyValueList` instance is encountered and remove `kv_properties_not_converted()`.  It looks like any KVL instance implies the entire thing has been converted and since only those properties in `kv_properties` are checked, this could simplify things a tad - assuming I'm understanding the overall flow.

elyra/pipeline/pipeline_definition.py
```diff
@@ -346,14 +363,46 @@ def remove_env_vars_with_matching_secrets(self):
...
-        if self.get_component_parameter(ENV_VARIABLES) and self.get_component_parameter(KUBERNETES_SECRETS):
+        env_vars = self.get_component_parameter(ENV_VARIABLES)
+        secrets = self.get_component_parameter(KUBERNETES_SECRETS)
+        if env_vars and isinstance(env_vars, KeyValueList) and secrets and isinstance(secrets, KeyValueList):
             new_list = KeyValueList.difference(
                 minuend=self.get_component_parameter(ENV_VARIABLES),
                 subtrahend=self.get_component_parameter(KUBERNETES_SECRETS),
             )
             if new_list:
                 self.set_component_parameter(ENV_VARIABLES, new_list)
```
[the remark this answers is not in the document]
kevin-bates: Sorry. I just noticed the help-string on the method and didn't realize it was that close to this block, so the comment portion of this suggestion isn't necessary.

docs/source/user_guide/pipelines.md
```diff
@@ -88,9 +88,14 @@ The [tutorials](/getting_started/tutorials.md) provide comprehensive step-by-ste
          - **Environment variables**
            - A list of environment variables to be set in the container that executes the Jupyter notebook or script. Format: `ENV_VAR_NAME=value`. Entries that are empty (`ENV_VAR_NAME=`) or malformed are ignored.
          - **Data volume**
```
ptitzler, line 90: Nit: since it's a list value we should use plural

```suggestion
         - **Data volumes**
```

elyra/pipeline/pipeline_definition.py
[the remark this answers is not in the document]
kiersten-stokes: Agree on the necessity of a comment!

> Hmm, I don't think we want the check against a Falsey result. Since None can't be returned, we should unconditionally reset ENV_VARIABLES to the empty list in the case when there are only secrets present.

I must have had this check in here because [set_component_parameter](https://github.com/elyra-ai/elyra/blob/c31164f5fa1ca354cc1e3ef43e48fd790e790302/elyra/pipeline/pipeline_definition.py#L315-L316) actually throws a `ValueError` if you send it a Falsey value of any type. I'm not sure if there's a reason for this that I can't think of but it seems pretty unnecessary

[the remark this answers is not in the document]
kiersten-stokes: Thoughts on changing that `set_component_parameter` line to `if value is None`?

docs/source/user_guide/pipelines.md
```diff
@@ -88,9 +88,14 @@ The [tutorials](/getting_started/tutorials.md) provide comprehensive step-by-ste
...
            - Data volumes are not mounted when the pipeline is executed locally.
+         - **Kubernetes secrets**
+           - A list of [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to be accessed as environment variables during Jupyter notebook or script execution. Format: `ENV_VAR=secret-name:secret-key`. Entries that are empty (`ENV_VAR=`) or malformed are ignored. Entries with a secret name considered to be an [invalid Kubernetes resource name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) or with [an invalid secret key](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data) will raise a validation error after pipeline submission or export.
+           - The referenced secrets must exist in the Kubernetes namespace where the generic pipeline nodes are executed.
+           - Secrets are ignored when the pipeline is executed locally. For remote execution, in the case where a node has a Kubernetes secret and an environment variable with the same name, the Kubernetes secret will be preferred and the corresponding environment variable will be removed from the list.
```
ptitzler, line 97: Two nits:
 - The wording is a bit ambiguous 'has a Kubernetes secret and an environment variable with the same name': should try to clarify that 'same name' refers to the environment variable name that the secret is associated with
 - The wording 'the corresponding environment variable will be removed from the list' unintentionally seems to target developers not users because it references internal behavior. 

Perhaps change this to something like

For remote execution, if an environment variable was assigned a static value (via the `environment variables` property) and a secret's value, the secret's value is used.

An example might also help.

docs/source/user_guide/pipelines.md
```diff
@@ -136,15 +141,20 @@ The [tutorials](/getting_started/tutorials.md) provide comprehensive step-by-ste
...
    - Example: `/mnt/vol1=data-pvc`
 
-1. Associate each node with a comment to document its purpose.
+   **Kubernetes Secrets**
+   - Optional. A list of [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to be accessed as environment variables during Jupyter notebook or script execution. Format: `ENV_VAR=secret-name:secret-key`. Entries that are empty (`ENV_VAR=`) or malformed are ignored. Entries with a secret name considered to be an [invalid Kubernetes resource name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) or with [an invalid secret key](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data) will raise a validation error after pipeline submission or export. The referenced secrets must exist in the Kubernetes namespace where the generic pipeline nodes are executed.
+   - Secrets are ignored when the pipeline is executed locally. For remote execution, in the case where a node has a Kubernetes secret and an environment variable with the same name, the Kubernetes secret will be preferred and the corresponding environment variable will be removed from the list.
```
ptitzler, line 150: See earlier comment regarding wording ambiguity.

elyra/pipeline/pipeline_definition.py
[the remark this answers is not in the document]
kevin-bates: I see. Of course, I didn't check that prior to my comment.  😅 

I think the check in `set_component_parameter` should be updated to `value is not None` and "empty values" should be allowed assuming they don't cause issues downstream.   It looks like Falseys are tolerated in these structures based on existing calls to `get_component_parameter()` so I suspect this should be okay.

I wonder if _any_ value "kind" should be tolerated (`None`, `""`, `[]`, etc.) so long as the `key` has a legitimate (non-falsey) value.  Then, it's really up to the consumer (where it matters) to make a decision regarding `None` and other falsey conditions.  This may be outside the scope of this PR, but I think, from a general design model, that makes sense (and would eliminate lots of these kinds of checks).

ptitzler:
We should also add a short motivation to https://elyra.readthedocs.io/en/latest/user_guide/best-practices-file-based-nodes.html, indicating that secrets should be used to make sensitive information available to nodes because their values are not exposed in the pipeline file, the pipeline editor, or the runtime environment.

elyra/pipeline/pipeline.py
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import dataclasses
+from dataclasses import dataclass
```
kevin-bates, line 17: Should we update the `@dataclass` decorators to `@dataclasses.dataclass` and remove the second import, or remove the first import and extend the second with `is_dataclass, asdict` and remove the `dataclasses.` prefix from the static methods?

I'm inclined to go with the first approach since it is more descriptive.

kevin-bates: I suppose the first approach is contrary to "import only what is needed", so perhaps the import for `asdict` could be aliased to something like `asdict as dataclasses_asdict`.  `is_dataclass` seems fairly unique and self-describing.

kiersten-stokes: Yeah I had that thought re: the first approach as well. Will do!

kiersten-stokes: Whoops, I only saw your latest comment after I posted mine 😄 . Second approach it is, then

kevin-bates (approved):
Looks good - thanks @kiersten-stokes

ptitzler (approved):
LGTM

Issue #2719, opened 2022-05-11
Elyra pipelines failing with version conflicts even with system owned images


yarnabrina:
**Describe the issue**
`Elyra` is failing with a conflict, likely due to [requirements file](https://raw.githubusercontent.com/elyra-ai/elyra/v3.8.1/etc/generic/requirements-elyra.txt)

**To Reproduce**
1. Create a notebook with a single cell containing `print("hello world")`
2. Create a pipeline with image being `Pandas 1.4.1`
3. Submit a `Kubeflow` pipeline

**Screenshots or log output**
Error Message:

> ERROR: Cannot install nbclient==0.4.1 and nbconvert==6.4.5 because these package versions have conflicting dependencies.

<details>
<summary>Full Logs</summary>

```
Downloading https://raw.githubusercontent.com/elyra-ai/elyra/v3.8.1/elyra/kfp/bootstrapper.py
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 26 28961   26  7748    0     0  16520      0  0:00:01 --:--:--  0:00:01 16520
28961  100 28961    0     0  60335      0 --:--:-- --:--:-- --:--:-- 60209
Downloading https://raw.githubusercontent.com/elyra-ai/elyra/v3.8.1/etc/generic/requirements-elyra.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 1277  100  1277    0     0   2455      0 --:--:-- --:--:-- --:--:--  2460
Collecting packaging
  Downloading packaging-21.3-py3-none-any.whl (40 kB)
Collecting pyparsing!=3.0.5,>=2.0.2
  Downloading pyparsing-3.0.9-py3-none-any.whl (98 kB)
Installing collected packages: pyparsing, packaging
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
Successfully installed packaging-21.3 pyparsing-3.0.9
WARNING: You are using pip version 21.2.4; however, version 22.0.4 is available.
[...]

Issue #2491, opened 2022-02-18
Notebook pipelines are failing on KFP and Airflow


ptitzler:
**Describe the issue**

[Confirmed. Not pipeline specific]

```
load_data.ipynb
[I 18:07:22.295] 'analyze_NOAA_weather_data-0218100646':'load_data' - dependencies processed (0.067 secs)
[I 18:07:22.296] 'analyze_NOAA_weather_data-0218100646':'load_data' - executing notebook using 'papermill load_data.ipynb load_data-output.ipynb' 
[D 18:07:22.994] Found kernel python3 in /usr/local/share/jupyter/kernels
[I 18:07:23.115] Input Notebook:  load_data.ipynb
[I 18:07:23.115] Output Notebook: load_data-output.ipynb
[E 18:07:23.121] Unexpected error: <class 'KeyError'>
[I 18:07:23.360] 'analyze_NOAA_weather_data-0218100646':'load_data' - converting from load_data-output.ipynb to load_data.html 
Traceback (most recent call last):
  File "bootstrapper.py", line 373, in execute
    papermill.execute_notebook(notebook, notebook_output, kernel_name=kernel_name)
  File "/usr/local/lib/python3.8/site-packages/papermill/execute.py", line 90, in execute_notebook
    nb = load_notebook_node(input_path)
  File "/usr/local/lib/python3.8/site-packages/papermill/iorw.py", line 461, in load_notebook_node
    nb_upgraded = nbformat.v4.upgrade(nb)
  File "/usr/local/lib/python3.8/site-packages/nbformat/v4/convert.py", line 39, in upgrade
    _warn_if_invalid(nb, from_version)
  File "/usr/local/lib/python3.8/site-packages/nbformat/v4/convert.py", line 21, in _warn_if_invalid
    validate(nb, version=version)
  File "/usr/local/lib/python3.8/site-packages/nbformat/validator.py", line 274, in validate
    for error in iter_validate(nbdict, ref=ref, version=version,
  File "/usr/local/lib/python3.8/site-packages/nbformat/validator.py", line 298, in iter_validate
    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
  File "/usr/local/lib/python3.8/site-packages/nbformat/validator.py", line 69, in get_validator
    schema_json = _get_schema_json(v, version=version, version_minor=version_minor)
  File "/usr/local/lib/python3.8/site-packages/nbformat/validator.py", line 103, in _get_schema_json
[...]

PR #2721, opened 2022-05-11
Validate runtime image environments in CI testing

kevin-bates:
This pull request adds an additional GitHub action named "Validate Image Environment" which runs for each supported version of Python and validates the [`requirements-elyra.txt`](https://github.com/elyra-ai/elyra/blob/master/etc/generic/requirements-elyra.txt) and requirements-elyra-py37.txt (in this PR) files that consists of the package dependencies used in the generic component containers can be successfully installed relative to the corresponding version of Python.
This also modifies the validate-runtime-image target to validate the elyra sample runtime images against these requirements

Resolves: #2719
 
           in the file; or

           it.


Makefile
```diff
@@ -318,8 +318,14 @@ validate-runtime-images: # Validates delivered runtime-images meet minimum crite
...
 			fi; \
+			if [ $$cmd == "python3" ]; then \
+				IMAGE_PYTHON3_MINOR_VERSION=`docker run --rm $$image $$cmd --version | cut -d' ' -f2 | cut -d'.' -f2` ; \
+				if [[ $$IMAGE_PYTHON3_MINOR_VERSION -lt 8 ]]; then \
+					echo WARNING: Image $$image requires at Python 3.8 or greater for generic component dependency installation; \
+				fi; \
+			fi; \
 		done; \
-		docker run -v $(pwd)/etc/generic:/opt/elyra/ --rm $$image python3 -m pip install -r /opt/elyra/requirements-elyra.txt > /dev/null 2>&1 ; \
+		docker run -v $$GITHUB_WORKSPACE/etc/generic:/opt/elyra/ --rm $$image python3 -m pip install -r /opt/elyra/requirements-elyra.txt > /dev/null ; \
```
kevin-bates, line 328: What is the default value for `GITHUB_WORKSPACE`?  Seems like we should have something like:
```
GITHUB_WORKSPACE?=$(pwd)
```
so this target can still be used outside of GH actions.

akchinSTC: Updated, didnt want to include a one off GITHUB* in the makefile..kinda weird, so I just replaced the env directly

akchinSTC:
Ive added second version of the requirements file for ipython <8 (e.g. py37 support)for generic components, which is essentially the same as the original with after reverting the updates to it. This would mean vulnerabilities that we attempted to close in #2491 -> #2487 would exist for this version of the file.

The original requirements file has been updated with bumps for nbclient for the original issue, and a bump for pyzmq due to failures in testing with python 3.10 env and the conda runtime image due to missing or incompatible native compilation libraries (gcc).

Follow up PR will include operator changes pull in both requirements files, bootstrapper would change to detect which version is running and pull in the appropriate version of the requirements file.

.github/workflows/build.yml
```diff
@@ -171,6 +171,18 @@ jobs:
       - name: Build
         run: make docs
 
+  validate-image-env:
+    name: Validate Image Environment
```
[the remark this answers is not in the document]
akchinSTC: > General comment: Could we start adding descriptive comments to 'targets' to make it clearer what their purpose is? For example, as is looking at this file it's not very clear what the difference is between `validate-image-env` and `validate-image`. With descriptions we can avoid that:
> 
> ```
> validate-image-env:
>     # Validate that Elyra's requirements for generic component processing can be installed in container images, which are included as system-managed runtime image configurations.
>     ...
> validate-images:
>     # Validate that included system-managed runtime images meet Elyra's requirements documented in https://elyra.readthedocs.io/en/latest/user_guide/runtime-image-conf.html#prerequisites
> ```

The targets have been renamed respectively with their descriptions
```
elyra-image-env: ## Creates a conda env consisting of the dependencies used in images
validate-runtime-images: # Validates delivered runtime-images meet minimum criteria

```

akchinSTC:
> Is what is listed address the vulnerabilities comment?

Yes looks like those versions would resolve the vulnerabilities. Updated the requirements file to match. 

Pro-active...should we also do the same for the 3.8+ version and bump the associated requirements?

kevin-bates:
> Pro-active...should we also do the same for the 3.8+ version and bump the associated requirements?

This seems reasonable provided we can re-test things. It would be nice to avoid introducing per-version requirements for 3.8+, so perhaps using the versions found against 3.8 will be sufficient (and until we learn otherwise). 

Also, in visiting each of the capped packages in [`setup.py`](https://github.com/elyra-ai/elyra/blob/master/setup.py), it looks like we'd be violating the caps for `jsonschema`, `pyyaml`, and `typing-extensions`.  These caps are due to KFP and I'm not sure if those same caps should be in play for the image requirements, but wanted to point that out.  Should these also apply to the image?  Since we're not using the KFP API within the container (at least that I'm aware), I'm not sure the caps are necessary.

akchinSTC:
> Since we're not using the KFP API within the container (at least that I'm aware), I'm not sure the caps are necessary.

Pretty sure they dont apply. We dont use any kfp libraries to my knowledge in the generic component execution

Makefile
```diff
@@ -23,16 +23,23 @@
 .PHONY: docs-dependencies docs
 .PHONY: elyra-image publish-elyra-image kf-notebook-image publish-kf-notebook-image
```
ptitzler, line 24:
```suggestion
.PHONY: elyra-image elyra-image-env publish-elyra-image kf-notebook-image publish-kf-notebook-image
```

ptitzler (approved):
LGTM

PR #2722, opened 2022-05-11
Elyra pipeline describe enhancements

### LINKED ISSUE #2685: Extend 'elyra-pipeline describe' to identify PVC and notebook|script dependencies
The `elyra-pipeline describe` command was introduced, among other things, to provide users with the means to understand what kind of dependencies a particular pipeline file has. Consider the following scenario:
 - User 1 creates a pipeline in environment 1. This pipeline has several dependencies, which would prevent it from running successfully in any other environment.
 - User 2 wants to run this pipeline in environment 2 and needs to know what the prerequisites are that need to be in place.

The command currently lists the following pipeline dependencies:
 - files (not including the referenced notebooks or scripts)
 - custom components

 Extend the `describe` command to also list
 - the notebooks/scripts that the pipeline file references if it includes generic components
 - information about mounted volumes/pvcs (supported by Elyra 3.8+)

The ultimate goal is to enable user 2 to confirm (to the extend possible) that environment 2 has all prerequisites in place that the pipeline needs to run.

binayakdutta:
**Enhancing "describe" to display Notebook/Scripts and Mount Volumes when applicable**

Extended pipeline_app.py to display additional details
Extended pipeline_constants.py to add extra constant to represent notebook extension
Extended test_pipeline_app.py to add additional test functions
Added test pipeline file to surface scenario and data for displaying newer details

Test Details:
make test-server works perfectly well expect ~/kfp/test_bootstrapper.py which is expected in local setup
pytest -k "describe" runs successfully including existing and new test functions

FIXES #2685


elyra/pipeline/pipeline_constants.py
```diff
@@ -21,3 +21,4 @@
 PIPELINE_META_PROPERTIES = ["name", "description", "runtime"]
 # optional static prefix to be used when generating an object name for object storage
 COS_OBJECT_PREFIX = "cos_object_prefix"
+NOTEBOOK_EXT = "ipynb"
```
ptitzler, line 24: There is no need to declare this as a constant. Some background information for context that should help clarify why not. Most constants in this file represent property names that various pieces of code use to access information that the [Pipeline class](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline.py#L331)  or [Operation class](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline.py#L219) store. These property names might change over time (e.g. `cos_object_prefix` -> `os_object_prefix`) and we wanted to avoid the need to change the code in several locations. The value that represents a Jupyter notebook file extension (`ipynb`) is not going to change over time, hence it's not a pipeline specific constant.

elyra/cli/pipeline_app.py
```diff
@@ -526,15 +538,29 @@ def describe(json_option, pipeline_path):
...
+    describe_dict["mounted_volumes"] = set()
     for node in primary_pipeline.nodes:
         # collect information about file dependencies
         for dependency in node.get_component_parameter("dependencies", []):
             describe_dict["file_dependencies"].add(f"{dependency}")
         # collect information about component dependencies
         if node.component_source is not None:
             describe_dict["component_dependencies"].add(node.component_source)
+        # collect information of mounted volumes
+        for mounted_volume in node.get_component_parameter("mounted_volumes", []):
```
ptitzler, line 554: Use the constant value for `mounted_volumes` as defined in https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline_constants.py#L20. (See other comment, which includes a clarification what these constants were introduced for)

elyra/cli/pipeline_app.py
```diff
@@ -526,15 +538,29 @@ def describe(json_option, pipeline_path):
...
+            describe_dict["mounted_volumes"].add(f"{mounted_volume}")
+            # mounted_volume_set.add(f"{mounted_volume}")
+        # collect notebook / script name when pipeline is generic
+        if describe_dict["type"] is None:
+            temp_value = node.get_component_parameter("filename", [])
+            if temp_value.split(".")[1].strip() == pipeline_constants.NOTEBOOK_EXT:
```
ptitzler, line 560: `temp_value.split(".")[1].strip()` does not yield the expected value if the filename contains more than one `.`. It's safer to use [pathlib's `Path.suffix`](https://docs.python.org/3/library/pathlib.html), e.g. `Path(temp_value).suffix`.

elyra/cli/pipeline_app.py
```diff
@@ -526,15 +538,29 @@ def describe(json_option, pipeline_path):
...
         # collect information about component dependencies
         if node.component_source is not None:
             describe_dict["component_dependencies"].add(node.component_source)
+        # collect information of mounted volumes
+        for mounted_volume in node.get_component_parameter("mounted_volumes", []):
+            describe_dict["mounted_volumes"].add(f"{mounted_volume}")
+            # mounted_volume_set.add(f"{mounted_volume}")
+        # collect notebook / script name when pipeline is generic
+        if describe_dict["type"] is None:
+            temp_value = node.get_component_parameter("filename", [])
```
ptitzler, line 559: Filename should default to an empty string. There should only be one for each generic operation.

elyra/cli/pipeline_app.py
```diff
@@ -512,8 +513,19 @@ def describe(json_option, pipeline_path):
...
-    iter_keys = {"file_dependencies", "component_dependencies"}
+    pipeline_keys = [
+        "name",
+        "description",
+        "type",
+        "version",
+        "nodes",
+        "file_dependencies",
+        "component_dependencies",
+        "mounted_volumes",
```
ptitzler, line 524: Use the constant value for `mounted_volumes` as defined in https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline_constants.py#L20. (See other comment, which includes a clarification what these constants were introduced for)

ptitzler:
I ran in a couple tests and the describe feature appeared to work as expected! Good start given the limited background information the original issue provided! There are a few things though that we should improve to help us get the output of the command to where it needs to be to properly support automation.

Starting with some background information: the `describe` command was introduced to enable users to identify dependencies that a pipeline might have. There are two output formats that this command supports: human readable (the default) and machine readable (if `--json` is specified). The machine readable output can be used to automate certain tasks. For example, a user might write a script that extracts the name of all PVCs and runs the appropriate `kubectl get pvc ..` command to verify that these PVCs are present in the Kubernetes cluster where the pipeline will be executed.

There are a few obstacles that would prevent the user from doing this:
  - [This behavior was not introduced by this PR.]The machine readable output includes a header, which shouldn't be printed because it renders the JSON output invalid:
     ```
     ──────────...
     Elyra Pipeline details
    ───────────...
     ```
    The header should not be printed if machine-readable output is produced.
  - The machine-readable output for mounted volumes might look as follows:
     ```
    "mounted_volumes": [
        "/mnt/vol2=my-pvc2"
    ]
     ```
    If a user wanted to use this output (see earlier example) to validate the existence of the specified PVC/s (here `my-pvc2`), it would be necessary to extract the name from `/mnt/vol2=my-pvc2`. We should avoid the need to do so by only returning the PVC name. @kiersten-stokes we had a similar  discussion in https://github.com/elyra-ai/elyra/pull/2715. Should 2722 piggy back on the changes you are making for that one?

ptitzler:
For reference a few examples how a user could extract information from the machine-readable output using [`jq`](https://stedolan.github.io/jq/):

```
$ elyra-pipeline describe my-pipeline.pipeline | jq .scripts  

[
  "/Users/.../workspace/a-script.py",
  "/Users/.../workspace/a-script.r"
]

$ elyra-pipeline describe my-pipeline.pipeline | jq .mounted_volumes[]

"my-pvc2"

```

kiersten-stokes:
>  @kiersten-stokes we had a similar  discussion in [Add support for Kubernetes secrets environment variables #2715](https://github.com/elyra-ai/elyra/pull/2715). Should 2722 piggy back on the changes you are making for that one?

Good question. I think we're safe to move forward with this PR whenever its ready and then perhaps any required changes as a result of #2715 can be addressed in a follow-up. Especially because 2715 may take a bit longer since I'll be out a few days coming up.

ptitzler:
@binayakdutta even though this hasn't been part of the original requirements in https://github.com/elyra-ai/elyra/issues/2685, would you be open to also adding support to include  runtime images for Jupyter notebooks/scripts? Code wise this should be very similar to what you've already added for files (there's a 1-to-1 relationship between generic nodes and runtime images, which are used to execute the notebook/script in).

An example use case scenario that would be enabled by this:
 - User created several pipelines that utilize a particular container image as runtime, say `docker.io/my-org/my-image:v1.0.0`.
 - ... time passes ...
 - User creates a new version of this container image, `docker.io/my-org/my-image:v2.0.0`, which includes newer versions of the installed packages.
 - User wants to update the pipelines that use the old image to use the newer version and needs to find out what those pipelines are.
 - User creates a script that locates pipeline files in the filesystem, runs the `elyra-pipeline describe --json ...` command and extracts the runtime image information

binayakdutta:
> @binayakdutta even though this hasn't been part of the original requirements in #2685, would you be open to also adding support to include runtime images for Jupyter notebooks/scripts? Code wise this should be very similar to what you've already added for files (there's a 1-to-1 relationship between generic nodes and runtime images, which are used to execute the notebook/script in).
> 
> An example use case scenario that would be enabled by this:
> 
>     * User created several pipelines that utilize a particular container image as runtime, say `docker.io/my-org/my-image:v1.0.0`.
> 
>     * ... time passes ...
> 
>     * User creates a new version of this container image, `docker.io/my-org/my-image:v2.0.0`, which includes newer versions of the installed packages.
> 
>     * User wants to update the pipelines that use the old image to use the newer version and needs to find out what those pipelines are.
> 
>     * User creates a script that locates pipeline files in the filesystem, runs the `elyra-pipeline describe --json ...` command and extracts the runtime image information

yes definitely. Will be happy to incorporate.

elyra/pipeline/pipeline_constants.py
binayakdutta: > I ran in a couple tests and the describe feature appeared to work as expected! Good start given the limited background information the original issue provided! There are a few things though that we should improve to help us get the output of the command to where it needs to be to properly support automation.
> 
> Starting with some background information: the `describe` command was introduced to enable users to identify dependencies that a pipeline might have. There are two output formats that this command supports: human readable (the default) and machine readable (if `--json` is specified). The machine readable output can be used to automate certain tasks. For example, a user might write a script that extracts the name of all PVCs and runs the appropriate `kubectl get pvc ..` command to verify that these PVCs are present in the Kubernetes cluster where the pipeline will be executed.
> 
> There are a few obstacles that would prevent the user from doing this:
> 
>     * [This behavior was not introduced by this PR.]The machine readable output includes a header, which shouldn't be printed because it renders the JSON output invalid:
>       ```
>       ──────────...
>       Elyra Pipeline details
>       ───────────...
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       The header should not be printed if machine-readable output is produced.
> 
>     * The machine-readable output for mounted volumes might look as follows:
>       ```
>       "mounted_volumes": [
>          "/mnt/vol2=my-pvc2"
>       ]
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       If a user wanted to use this output (see earlier example) to validate the existence of the specified PVC/s (here `my-pvc2`), it would be necessary to extract the name from `/mnt/vol2=my-pvc2`. We should avoid the need to do so by only returning the PVC name. @kiersten-stokes we had a similar  discussion in [Add support for Kubernetes secrets environment variables #2715](https://github.com/elyra-ai/elyra/pull/2715). Should 2722 piggy back on the changes you are making for that one?

Done.
1. Header Banner is conditional now, applicable when --json is not chosen
2. For mounted_volumes, I split by "=" and took second element
installed and ran jq with the output. Looks good
image

binayakdutta: > There is no need to declare this as a constant. Some background information for context that should help clarify why not. Most constants in this file represent property names that various pieces of code use to access information that the [Pipeline class](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline.py#L331) or [Operation class](https://github.com/elyra-ai/elyra/blob/master/elyra/pipeline/pipeline.py#L219) store. These property names might change over time (e.g. `cos_object_prefix` -> `os_object_prefix`) and we wanted to avoid the need to change the code in several locations. The value that represents a Jupyter notebook file extension (`ipynb`) is not going to change over time, hence it's not a pipeline specific constant.

Understood. Done

elyra/cli/pipeline_app.py
binayakdutta: @ptitzler , just one doubt. 
For referencing pipeline dictionary keys, I am replacing the key "mounted_volumes" with the corresponding constant. This is clear.
But for list, set initialization also, should we be doing this ? Rechecking my understanding since the list is local and technically may have any names.
Kindly let me know. Thnx

ptitzler: Good question! The `describe_dict` keys (e.g. `describe_dict["component_dependencies"]`) should remain as is for now because their names are used to produce the output (`Component Dependencies: ...`) and should remain consistent over time. For example, let's say we decide in the future to rename the value of `MOUNTED_VOLUMES` in the pipeline constants file from `component_dependencies` to `custom_component_dependencies`. We don't necessarily want this internal name change to  surface to the user.  If we were to use the constant value, we always would and lose a layer of abstraction that is useful.

There's definitely room for improvement in the describe method code to implement this more cleanly -- e.g. each dict entry gets assigned a value for a human-readable display name (`Component dependencies`) and a machine-readable property name (`custom_components`) that is used when the output is JSON formatted).

binayakdutta: Fixed as discussed. Removed usage of iter_keys and pipeline_keys and used describe_dict straight away for producing JSON and non-JSON output

elyra/cli/pipeline_app.py
```diff
@@ -526,20 +521,41 @@ def describe(json_option, pipeline_path):
...
+                describe_dict["notebooks"].add(temp_value)
+            else:
+                describe_dict["scripts"].add(temp_value)
 
     if not json_option:
-        for key in pipeline_keys:
+        click.echo()
+
+        print_banner("Elyra Pipeline details")
+        for key in list(describe_dict.keys()):
```
ptitzler, line 556: Nit:
```suggestion
        for key in describe_dict:
```

elyra/cli/pipeline_app.py
```diff
@@ -526,20 +521,41 @@ def describe(json_option, pipeline_path):
...
             readable_key = " ".join(key.title().split("_"))
-            if key in iter_keys:
+            if type(describe_dict[key]) is set:
```
ptitzler, line 558: Nit - slightly better: 
```suggestion
            if isinstance(describe_dict[key], set):
```

elyra/cli/pipeline_app.py
```diff
@@ -526,20 +521,41 @@ def describe(json_option, pipeline_path):
...
                 click.echo(f"{readable_key}:")
                 if describe_dict.get(key, set()) == set():
```
ptitzler, line 560: Nit: You could exploit that an empty set evaluates to False

```suggestion
                if not describe_dict.get(key):
```

elyra/cli/pipeline_app.py
```diff
@@ -549,11 +565,12 @@ def describe(json_option, pipeline_path):
...
-        for key in iter_keys:
-            describe_dict[key] = list(describe_dict[key])
-        for key in pipeline_keys:
+        for key in list(describe_dict.keys()):
+            if type(describe_dict[key]) is set:
+                describe_dict[key] = list(describe_dict[key])
+
             value = describe_dict.get(key)
-            if value is None or (key in iter_keys and len(value) == 0):
+            if value is None or (key in list(describe_dict.keys()) and len(str(value)) == 0):
```
ptitzler, line 573: Not: this should yield the same result for `None` and '':

```suggestion
            if not value:
```

elyra/tests/cli/test_pipeline_app.py
```diff
@@ -253,6 +253,40 @@ def test_describe_with_missing_kfp_component():
         assert result.exit_code == 0
 
 
+def test_describe_wth_mount_volume():
```
ptitzler, line 256: Typo:
```suggestion
def test_describe_with_mount_volume():
```

ptitzler:
To close the remaining gaps in test coverage, can you please update the tests to 
- also verify that `--json` produces the expected results
- also use a pipeline that that only includes [custom components](https://elyra.readthedocs.io/en/stable/user_guide/pipeline-components.html#custom-components) (or at least contains one custom component) - for these components your describe command enhancements should not report any dependencies. https://github.com/elyra-ai/elyra/blob/master/elyra/tests/pipeline/resources/validation_pipelines/kf_inputpath_parameter.pipeline would be a suitable candidate

elyra/cli/pipeline_app.py
binayakdutta: Changed to 
for key in list(describe_dict)
to avoid dict mutation error  (as discussed)

elyra/cli/pipeline_app.py
binayakdutta: Done

elyra/cli/pipeline_app.py
binayakdutta: Done

elyra/cli/pipeline_app.py
binayakdutta: Done

elyra/tests/cli/test_pipeline_app.py
binayakdutta: Done

elyra/cli/pipeline_app.py
```diff
@@ -526,34 +521,64 @@ def describe(json_option, pipeline_path):
...
-            describe_dict[key] = list(describe_dict[key])
-        for key in pipeline_keys:
+        # for key in list(describe_dict.keys()):
+        for key in list(describe_dict):
+            if isinstance(describe_dict[key], set):
+                describe_dict[key] = list(describe_dict[key])
+
             value = describe_dict.get(key)
-            if value is None or (key in iter_keys and len(value) == 0):
+            # if value is None or (key in list(describe_dict.keys()) and len(str(value)) == 0):
```
kiersten-stokes, line 580: I'm sure you're already tracking this, but just in case, this line can probably be removed now, as well as 574 and 565 🙂

kiersten-stokes (approved):
I know we're still waiting on the version update for the new test pipelines, but functionality looks great!

elyra/cli/pipeline_app.py
binayakdutta: Resolved

kiersten-stokes (approved):
Looks great! Thanks for tackling this

Issue #1199, opened 2021-01-15
Save notebook cell as new code snippet


karlaspuldaro:
Currently the user can select a test from a notebook cell and save as a new code snippet.
As suggested in [this comment](https://github.com/elyra-ai/elyra/pull/1186#issuecomment-760318118), users could also have the option to save an entire notebook cell (or cells) and as a new code snippet, without having to select the text.

xlegs:
I would like to contribute to this issue.

PR #2726, opened 2022-05-12
Create code snippet from whole cells (#1199)

xlegs:
The user can now create a code snippet without explicitly selecting text.

- This works when one or more cells are selected.
- If more than one cell is selected, their contents are combined, separated by an empty line, and added to a single code snippet.
- If text is selected, the the text selection will take priority, even if multiple cells are selected.

Please review. 

If there any more requirements, such as documentation or testing, please let me know.

### DIFF
--- a/create-release.py
+++ b/create-release.py
@@ -143,6 +143,13 @@ def update_version_to_release() -> None:
             rf"https://elyra.readthedocs.io/en/v{new_version}/",
         )
 
+        # Update documentation version for elyra-metadata cli help
+        sed(
+            _source("elyra/metadata/metadata_app_utils.py"),
+            r"https://elyra.readthedocs.io/en/latest/",
+            rf"https://elyra.readthedocs.io/en/v{new_version}/",
+        )
+
         sed(
             _source("packages/pipeline-editor/src/EmptyPipelineContent.tsx"),
             r"https://elyra.readthedocs.io/en/latest/user_guide/",
@@ -268,6 +275,13 @@ def update_version_to_dev() -> None:
             _source("elyra/cli/pipeline_app.py"),
             rf"https://elyra.readthedocs.io/en/v{new_version}/",
             rf"https://elyra.readthedocs.io/en/latest/",
+        )
+
+        # Update documentation version for elyra-metadata cli help
+        sed(
+            _source("elyra/metadata/metadata_app_utils.py"),
+            rf"https://elyra.readthedocs.io/en/latest/",
+            rf"https://elyra.readthedocs.io/en/v{new_version}/",
         )
 
         sed(
--- a/elyra/metadata/metadata_app_utils.py
+++ b/elyra/metadata/metadata_app_utils.py
@@ -640,10 +640,10 @@ class AppBase(object):
         print("-----------")
         print("Subcommands are launched as `elyra-metadata cmd [args]`. For information on")
         print("using subcommand 'cmd', run: `elyra-metadata cmd -h` or `elyra-metadata cmd --help`.")
+        print("\nFind more information at https://elyra.readthedocs.io/en/latest/")
         print()
         for subcommand, desc in self.subcommands.items():
-            print(subcommand)
-            print(f"    {desc[1]}")
+            print(f"{subcommand:<10}{desc[1]:>10}")
 
     @staticmethod
     def exit(status: int):


marthacryan (approved):
These changes look good -

ptitzler (approved):
LGTM. Tested functionality and reviewed doc updates. Nicely done!

ptitzler:
I'm adding this new feature to the [release highlights for version 3.9](https://github.com/elyra-ai/elyra/releases).

marthacryan (changes requested):
It'd actually be good if you did add those tests referenced above. A couple scenarios to test would be to just right click a cell and check that it opens the code snippet editor with the correct contents. Another scenario would be to check for the disabled menu option if there is no text or cell selected. The file that Karla referenced (https://github.com/elyra-ai/elyra/blob/master/tests/integration/codesnippet.ts) has examples of using selectors to click on an item, and she also referenced the integration test documentation for more information on where to find the API etc. Let me know if you have any further questions!

kiersten-stokes:
Hey @xlegs, just FYI we had a few conflicts between two recently merged PRs that are causing the current server tests failures. We fixed those yesterday in #2740. If you merge with master again now, we should be good to go! Sorry for the inconvenient timing there 🙂

xlegs:
I have added tests to this PR. They test:

- that the code snippet menu entry is disabled when the selected cell is empty
- that creating a snippet from a single cell works correctly
- that creating a snippet from multiple cells works correctly

I wanted to create a fourth to verify text selection, but cypress does not support this. There are workarounds, but Jupyter uses JS instead of native HTML form elements for cells, which complicates matters. See: https://github.com/cypress-io/cypress/issues/6200

Please review.

create-release.py
```diff
@@ -143,13 +143,6 @@ def update_version_to_release() -> None:
             rf"https://elyra.readthedocs.io/en/v{new_version}/",
         )
 
-        # Update documentation version for elyra-metadata cli help
```
karlaspuldaro, line 146: Is this file changed by accident during a merge?

elyra/metadata/metadata_app_utils.py
```diff
@@ -640,10 +640,10 @@ def print_subcommands(self):
         print("-----------")
         print("Subcommands are launched as `elyra-metadata cmd [args]`. For information on")
         print("using subcommand 'cmd', run: `elyra-metadata cmd -h` or `elyra-metadata cmd --help`.")
-        print("\nFind more information at https://elyra.readthedocs.io/en/latest/")
```
karlaspuldaro, line 643: Same for this file

tests/integration/codesnippetfromselectedcells.ts
```diff
@@ -0,0 +1,135 @@
+/*
```
karlaspuldaro, line 1: Since this is testing the code snippet extension, I suggest the new tests are located in `tests/integration/codesnippet.ts` instead of a separate test file.

tests/integration/codesnippetfromselectedcells.ts
```diff
@@ -0,0 +1,135 @@
...
+    cy.get(
+      'li.lm-Menu-item[data-command="codesnippet:save-as-snippet"]'
+    ).click();
+
+    cy.wait(2000);
+
+    cy.get('span[role="presentation"]:visible').should(
+      'have.text',
+      'print("test")'
+    );
```
karlaspuldaro, line 71: I'm guessing at this point we're checking the code entry field from the metadata editor form? To make it more clear maybe you could add a comment above or make this selector more specific as you have done in line 107.

karlaspuldaro (commented):
Hi @xlegs ! Thx so much for writing integration tests for this enhancement.
For some reason they didn't pass locally for me, it could be my environment, but I trust it would fail CI as well if something went wrong. I just have a few questions/comments.

elyra/metadata/metadata_app_utils.py
xlegs: I did not explicitly change this file. Feel free to make any changes to it.

tests/integration/codesnippetfromselectedcells.ts
xlegs: `tests/integration/codesnippet.ts` contains tests for the dialog that allows the user to create and edit snippets. 

These new tests are primarily for the notebook editor. I felt that they were different enough to have a separate file.

karlaspuldaro: That's a good point, although we have been writing integration tests on a single test file per extension, I'm thinking we would need to start breaking them apart in the future (ie. pipeline editor test file is getting huge). We could start a code-snippet directory inside `tests/integration/` to organize them better at least. 
@marthacryan @ajbozarth what do you think?

ajbozarth: I would agree we should split up our tests into different files and sort them into directories. I'm ok leaving this as is in this PR an creating a followup issue for next sprint to split up and restructure our current tests

karlaspuldaro (approved):
LGTM! Thanks a lot for the contribution :)

PR #2756, opened 2022-06-01
Fix save as code snippet

marthacryan:
Fixes a bug introduced by #2464 that breaks the "Save as a code snippet" functionality. 

 
           in the file; or

           it.

### DIFF
--- a/packages/metadata-common/src/MetadataEditorWidget.tsx
+++ b/packages/metadata-common/src/MetadataEditorWidget.tsx
@@ -87,6 +87,8 @@ export interface IMetadataEditorProps {
   themeManager?: IThemeManager;
 
   titleContext?: string;
+
+  code?: string[];
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/metadata-common/src/MetadataEditorWidget.tsx
```diff
@@ -86,7 +86,15 @@ export interface IMetadataEditorProps {
    */
   themeManager?: IThemeManager;
 
+  /**
+   * String used to make the title of the editor more readable
+   */
   titleContext?: string;
+
+  /**
+   * Optional field used for the "Save as code snippet" feature
```
ajbozarth, line 95: I would make the description more generic, don't reference the exact use case but a potential one, such as "a default value for the code properties" Also it's worth noting that this would insert this for all code properties if it exists, not sure if that's a feature we want, or if we should try to "aim" where this goes better?

marthacryan: Updated the comment - in terms of the behavior, that's worth working on / discussing but this is just getting it back to the state it was in before. As far as our functionality is concerned that wouldn't come up because we'd only be giving the code prop for empty metadata.

ptitzler (approved):
LGTM. Thank you! No longer seeing the two issues (code is not inserted and multiple code snippet editor window open)

karlaspuldaro (approved):

Issue #2759, opened 2022-06-02
allow setting component inputs from files (KFP & Airflow)


thesuperzapper:
Currently, we only provide two methods to specify component inputs:

1. Plain Strings (no newlines)
2. From the output of a parent node

We can provide a third option that selects a local file as the input. During pipeline run/export we would pass the content of this file as a string to the input. 

For example, if a component takes an SQL query as an input, it would be amazing to just reference a local `query.sql` file, as right now newlines can't even be input:

![Screen Shot 2022-06-02 at 16 05 47](https://user-images.githubusercontent.com/5735406/171563804-d56b1eb3-c144-4c6d-a0b8-3b6422429545.png)

---

Considerations:
- KFP generates Argo-Workflow YAML, and there are limits on how big of a YAML Kubenrentes will accept, so we might want to  fail if the input file would cause the YAML to be too big (but I am not sure on the exact max value).

ptitzler:
This would be similar to the 'local file dependencies' we currently support for generic components, right? For now users could use notebooks/scripts to accomplish this, instead of custom components. (Not arguing against it, just wanted to bring this up for others who come across this request and are looking for a solution that is available today.)

thesuperzapper:
> This would be similar to the 'local file dependencies' we currently support for generic components, right? For now, users could use notebooks/scripts to accomplish this, instead of custom components. (Not arguing against it, just wanted to bring this up for others who come across this request and are looking for a solution that is available today.)

@ptitzler the main difference to the "local file dependencies" is that this approach would not use the COS bucket, as the content of the input file would be read at pipeline "run" and "export" time and then input as a string to the Kubeflow/Airflow component.

ptitzler:
Since pipelines cannot be exported or run without these input files being present, the `elyra-pipeline describe` command also needs to be extended to list them as dependencies.

PR #2767, opened 2022-06-03
Support file contents as component property input value

kiersten-stokes:
#2759
May depend on #2780 

When complete, this PR will support the ability to use file contents as the input value for component properties. Right now, this is an extremely rough draft with changes to the properties template temporarily hardcoded in order to figure out the correct JSON structure required for correct frontend rendering.

For now, the following questions need to be answered:

- [ ] How to best structure component properties JSON
  - Depending on how this ends up, we may need to test that the removal of `json.loads()` for component properties does not cause any unexpected issues; `json.loads()` removes duplicate keys within the same sub-stanza, which we may need to allow in the case of 2 different types of `StringControl`'s (one for text input and one for file input)
- [ ] How to best structure pipeline JSON (small changes may be required since both plain strings and file types are `StringControl` types, but their processing will be different)
- [ ] How to enforce the size limit Kubernetes sets on YAMLs

- [ ] Add tests
- [ ] Fix existing tests

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_catalog.py
+++ b/elyra/pipeline/component_catalog.py
@@ -673,7 +673,7 @@ class ComponentCache(SingletonConfigurable):
             template = ComponentCache.load_jinja_template("canvas_properties_template.jinja2")
 
         canvas_properties = template.render(component=component)
-        return json.loads(canvas_properties)
+        return canvas_properties
 
 
 class ManifestFileChangeHandler(FileSystemEventHandler):
--- a/elyra/templates/components/canvas_properties_template.jinja2
+++ b/elyra/templates/components/canvas_properties_template.jinja2
@@ -6,7 +6,7 @@
         "elyra_{{ property.ref }}":
     {% if property.control_id == "OneOfControl" %}
         {
-        "activeControl": "{{ property.default_control_type }}",
+        "activeControl": "OneOfControl",
         "{{ property.default_control_type }}":
     {% endif %}
     {% set property_data_type = property.data_type %}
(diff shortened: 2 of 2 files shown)


kiersten-stokes:
@ajbozarth, as promised here is the open PR. The code won't be very helpful (and it's super rough anyway since changes will definitely be required), so instead, I'll paste the `uihints/parameter_info` stanza of the properties for a single component and a screenshot of what is rendered below. I'm not really sure whether the issue is the way I have the things structured or if there's a frontend piece missing. If it's the former, let me know how I should be structuring it!

```json
{
    "parameter_ref": "elyra_curl_options",
    "control": "custom",
    "custom_control_id": "OneOfControl",
    "label": {
        "default": "curl options"
    },
    "description": {
        "default": "Additional options given to the curl program. See https://curl.haxx.se/docs/manpage.html (type: string)",
        "placement": "on_panel"
    },
    "data": {
        "controls": {
            "OneOfControl": {
                "label": "Please select an input type :",
                "data": {
                    "controls": {
                        "StringControl": {
                            "label": "Please enter a string value :",
                            "format": "string"
                        },
                        "StringControl": {
                            "label": "Please choose a file :",
                            "format": "file"
                        }
                    }
                }
            },
            "NestedEnumControl": {
                "label": "Please select an output from a parent :",
                "format": "inputpath",
                "allownooptions": false
            }
        },
        "required": true
    }
},
```

kiersten-stokes:
This functionality will be delivered in #2780

PR #2769, opened 2022-06-03
Add errors to tag editor

marthacryan:
This adds displaying errors to the tag editor. Uses the same css as all the other errors. 
image

 
           in the file; or

           it.


akchinSTC:
hrmm is this a must have for 3.9? 
wondering if we could somehow display a more helpful message for someone not familiar with regex.

akchinSTC:
if this is just to surface the fact that there is an error though, im ok with it

marthacryan:
@akchinSTC Yeah this isn't super readable but I'm just surfacing the error that rjsf created. It shouldn't actually be surfaced now that we're trimming it with #2768 - are there other cases that could break this pattern?

ptitzler:
> @akchinSTC Yeah this isn't super readable but I'm just surfacing the error that rjsf created. It shouldn't actually be surfaced now that we're trimming it with #2768 - are there other cases that could break this pattern?

It doesn't look like it's working as intended. If a user enters a string with leading/trailing spaces the error message  is surfaced:

image

ptitzler (approved):
LGTM!

PR #2728, opened 2022-05-12
Prepend baseUrl to icon source

ajbozarth:
Currently we only prepend `/` to the icon path when using icons
hosted on the server. This causes the icons to not be found when
the server has base_url set in in it's config.

To address this I've prepended baseUrl to the icon path instead as
we also do when making API calls to the server.

Note: base_url is used in the server config json, but baseUrl is used when
accessing the python and js objects

base_url can be configured by including the following when running `jupyter lab`:

` --ServerApp.base_url="/foo"`

Fixes #2723

Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #2775, opened 2022-06-08
Do not persist the jupyter baseUrl in pipelines

### LINKED ISSUE #2773: pipeline file contains reference to potentially invalid custom component image
**Describe the issue**
For custom components the .pipeline file contains references to a deployment specific image URL (note the port, which is not the Jupyter default): 

```
           "ui_data": {
              "label": "Split File",
              "image": "http://localhost:8890/static/elyra/kubeflow.svg",
              "x_pos": 219,
              "y_pos": 283.5,
              "description": "Split file into two"
            }
```

While the image appears to be displayed properly in the VPE irrespective of whether the specified port is the port where JupyterLab was started, the metadata should probably not include protocol, host and port information.

**To Reproduce**
1. Launch JupyterLab specifying a non-default port (`--port xxxx`)
2. Create a KFP/AA pipeline and add a custom component
3. Save the pipeline and review the .pipeline file

**Deployment information**
 - Elyra version: 3.9.0

ajbozarth:
The changes in #2728 cause the pipeline-editor to persist the full
url for a node's image. This could cause issues with surfacing PI,
as such I've updated that fix to instead use the full URL path,
which will behave the same as before the fix, but including the
URL path from the baseUrl if one exists.

Note that if a baseUrl path conatains PI that would still be
persisted, but putting PI in a URL path is considered inappropriate
handling of PI by the server owner, not Elyra.

Fixes #2773

           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/pipeline-hooks.ts
+++ b/packages/pipeline-editor/src/pipeline-hooks.ts
@@ -213,10 +213,9 @@ export const componentFetcher = async (type: string): Promise<any> => {
       category.node_types?.[0]?.runtime_type ?? 'LOCAL';
 
     const type = types.find((t: any) => t.id === category_runtime_type);
-    const defaultIcon = URLExt.join(
-      ServerConnection.makeSettings().baseUrl,
-      type?.icon || ''
-    );
+    const defaultIcon = URLExt.parse(
+      URLExt.join(ServerConnection.makeSettings().baseUrl, type?.icon || '')
+    ).pathname;
 
     category.image = defaultIcon;
 


kevin-bates (approved):
LGTM.  Confirmed entries change:
`ui_data` entry goes from ...
```
            "ui_data": {
              "label": "Calculate data hash",
              "image": "http://localhost:8888/static/elyra/kubeflow.svg",
              "x_pos": 37,
              "y_pos": 364.5
            }
```
to
```
            "ui_data": {
              "label": "Calculate data hash",
              "image": "/static/elyra/kubeflow.svg",
              "x_pos": 37,
              "y_pos": 364.5
            }
```

ajbozarth:
Also as a note due to how this field is handled this will also update any pipeline that is opened/edited, so there no worry about pipelines that already persisted the full url with a domain or ip

ptitzler:
> Note that if a baseUrl path conatains PI that would still be
persisted, but putting PI in a URL path is considered inappropriate
handling of PI by the server owner, not Elyra.

A user might not necessarily have control over this. For example, Elyra deployments running as Kubeflow Notebook Server the URL is  
`http://<host-domain>:<port>/notebook/<kubernetes-namespace>/<user-provided-name>/static/elyra/kubeflow.svg`:

image

The namespace is something an administrator configures, not necessarily the person who deploys Elyra as a notebook server.

kevin-bates:
@ptitzler - that's a good point.  Couldn't we persist the "suffix portion" and strip anything prior to `static/elyra`? Then, these values are identical across pipelines and independent of `base_url`.

ajbozarth:
> Couldn't we persist the "suffix portion" and strip anything prior to static/elyra

That was the behavior prior to my fix that caused this, so that would just revert the fix and bring back the issue it solved.

Overall this is a limitation of canvas not elyra, and if including any of the baseUrl is a problem then we will just need to return to using base64 encoded strings for the images like we did before moving them to the server

lresende:
Note that, storing environment-specific details, such as host+port also breaks the ability to share the pipelines with others.

ptitzler:
> Note that, storing environment-specific details, such as host+port also breaks the ability to share the pipelines with others.

Right, that concern was why I've created the issue in the first place! In this specific case though it turns out it doesn't impact portability of pipeline files between environments. We definitely need to continue to watch out for these things.

Issue #2778, opened 2022-06-09
jupyter-server 1.17.1 breaks server tests


ptitzler:
**Describe the issue**
Elyra server tests 'fail' since 6/7, as seen here https://github.com/elyra-ai/elyra/actions/runs/2458875061, even though no Elyra changes were introduced. @kevin-bates' initial analysis identified  `jupyter-server` 1.17.1 as the likely source. The problem cannot be reproduced with version 1.17.0.

**To Reproduce**
1. Run `pytest -v elyra/tests/pipeline/test_handlers.py`

**Screenshots or log output**
If applicable, add screenshots or log output to help explain your problem.
<details><summary>Log Output</summary>
<pre>
Paste the log output here.
</pre>
</details>

**Expected behavior**

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.10/main

ptitzler:
Closing since #2781 prev ents the issue

PR #2783, opened 2022-06-10
CI: fetch packages info if server tests fail

### LINKED ISSUE #2777: CI: capture additional information if the server tests fail
Occasionally server tests (`make pytest`) fail because a new version of a package dependency was released. We should consider adding an additional CI step to https://github.com/elyra-ai/elyra/blob/main/.github/workflows/build.yml#L76-L95, which is only executed if the previous step https://github.com/elyra-ai/elyra/blob/main/.github/workflows/build.yml#L94 failed. This step could collect additional information that would enable us to determine more quickly if a failure might be the result of a dependency change that the build picked up. 

- append a step to https://github.com/elyra-ai/elyra/blob/main/.github/workflows/build.yml#L76 that is executed only if https://github.com/elyra-ai/elyra/blob/main/.github/workflows/build.yml#L94 fails
- the step parses the output of `pip freeze`, retrieves for each package its metadata by sending an HTTP `GET` request to `https://pypi.org/pypi/{PKG_NAME}/json`, scans the release history and issues a message if a new package version was released within the past X days

**Additional context**

tal66:
Hi, this closes #2777.
I used `pip list --outdated`. So basically, this script is only necessary because of the 'past X days' requirement.

see my (other) experimental branch for demo run that fails on purpose: https://github.com/tal66/elyra/actions/runs/2477060776

### DIFF
--- a/.github/workflows/scripts/fetch_packages_metadata.py
+++ b/.github/workflows/scripts/fetch_packages_metadata.py
@@ -1,7 +1,7 @@
 from requests import get
 from datetime import datetime, timedelta
 from concurrent.futures import ThreadPoolExecutor
-from itertools import islice
+from itertools import islice, chain
 import sys
 import subprocess
 
@@ -23,6 +23,7 @@ def get_outdated_pkg_list():
 
 
 def check_release_date(pkg_list):
+    pkgs_output = []
     for row in pkg_list:
         try:
             pkg, v = row.split("==")
@@ -44,10 +45,10 @@ def check_release_date(pkg_list):
         if today - release_date > timedelta(days=max_days):
             continue
 
-        print(f"Newer version for <{pkg}>:")
-        print(f" - released {latest}  {latest_upload_time[:10]}")
-        print(f" - current  {v}")
-        print("")
+        pkgs_output.append(
+            f"Newer version for <{pkg}>:\n - released {latest}  {latest_upload_time[:10]}\n - current  {v}\n"
+        )
+    return pkgs_output
 
 
 if __name__ == "__main__":
@@ -65,7 +66,10 @@ if __name__ == "__main__":
     parts[-1] = l
 
     with ThreadPoolExecutor(max_workers=workers) as executor:
-        executor.map(
+        results = executor.map(
             check_release_date,
             [islice(pkg_list, parts[i], parts[i + 1]) for i in range(len(parts) - 1)],
         )
+
+    for item in chain(*results):
+        print(item)
--- a/.github/workflows/scripts/fetch_packages_metadata.py
+++ b/.github/workflows/scripts/fetch_packages_metadata.py
@@ -27,7 +27,8 @@ def check_release_date(pkg_list):
         try:
             pkg, v = row.split("==")
         except Exception:
-            print(f"skipping {row}")
+            if len(row) > 1:
+                print(f"skipping {row}")
             continue
 
         res = get_info_from_pypi(pkg)
@@ -55,6 +56,7 @@ if __name__ == "__main__":
         max_days = int(sys.argv[1])
 
     print(f"packages that were updated in the last {max_days} days:")
+    print()
 
     pkg_list = get_outdated_pkg_list()
     l = len(pkg_list)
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/fetch_packages_metadata.py
```diff
@@ -0,0 +1,75 @@
+from requests import get
```
ptitzler, line 1: Due to it's location this Python file wasn't processed by the linter/black. Can you update https://github.com/elyra-ai/elyra/blob/main/Makefile#L111 to also process files in .github/workflows/scripts? This will surface a few import order issues in your code.

.github/workflows/scripts/fetch_packages_metadata.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        res = get(url).json()
+        return res
+    except Exception as e:
+        print(f"request to pypi failed for <{pkg}>: {e}")
+        return None
+
+
+def get_outdated_pkg_list():
+    out = subprocess.check_output(["python", "-m", "pip", "list", "--outdated", "--format", "freeze"])
```
ptitzler, line 20: If we use `--outdated` then we won't process any packages that are up-to-date, which is a problem. The [issue that motivated this enhancement request](https://github.com/elyra-ai/elyra/issues/2778) is a good example for why we cannot ignore these packages. What happened was that in test build XXX package version x.y.z was installed and everything worked. The following day (test build XXX+1) package version x.y.z+1 was installed and the test failed. We try to avoid capping versions and it's therefore common that two test builds use different dependency versions, even though nothing has changed in the Elyra code.

.github/workflows/scripts/fetch_packages_metadata.py
tal66: added .github in this line

.github/workflows/scripts/fetch_packages_metadata.py
tal66: changed to pip freeze

ptitzler (approved):
LGTM! Very nice work, thank you for your contribution and awesome turnaround time!

ptitzler:
General question: currently the output is sorted by package name, which makes perfect sense if one wants to quickly check if one of the 'usual suspects' (packages that have broken Elyra in the past) has changed:

```
Newer version for <google-api-python-client>:
 - released 2.50.0  2022-06-07	(>)
 - current  1.12.11

Newer version for <google-auth>:
 - released 2.7.0  2022-06-07	(>)
 - current  1.35.0

Newer version for <google-cloud-core>:
 - released 2.3.1  2022-06-09	(=)
 - current  2.3.1

Newer version for <google-cloud-storage>:
 - released 2.4.0  2022-06-08	(>)
 - current  1.44.0

Newer version for <ipykernel>:
 - released 6.13.1  2022-06-06	(=)
 - current  6.13.1

Newer version for <jupyter-client>:
 - released 7.3.4  2022-06-08	(=)
 - current  7.3.4

Newer version for <jupyter-server>:
 - released 1.17.1  2022-06-07	(=)
 - current  1.17.1

Newer version for <jupyterlab>:
 - released 3.4.3  2022-06-07	(=)
 - current  3.4.3

Newer version for <kfp-server-api>:
 - released 1.8.2  2022-06-10	(=)
 - current  1.8.2

Newer version for <networkx>:
 - released 2.8.3  2022-06-04	(>)
 - current  2.6.3

Newer version for <notebook>:
 - released 6.4.12  2022-06-07	(=)
 - current  6.4.12

Newer version for <pipenv>:
 - released 2022.6.7  2022-06-07	(>)
 - current  2020.11.15

Newer version for <pylint>:
 - released 2.14.1  2022-06-06	(=)
 - current  2.14.1

Newer version for <requests>:
 - released 2.28.0  2022-06-09	(=)
 - current  2.28.0

Newer version for <watchdog>:
 - released 2.1.9  2022-06-10	(=)
 - current  2.1.9
```

However, since we are also running automated daily tests (e.g. https://github.com/elyra-ai/elyra/actions/runs/2478594976) where probably sorting the output by the most recent change date (in descending order) might be a bit more useful. Would be interested in hearing what other think.

ptitzler:
We've discussed this in todays's dev meeting and decided to leave the report output as is for now.

PR #2785, opened 2022-06-17
Support volumes for custom components

### LINKED ISSUE #2760: mount "data volumes" from "generic nodes defaults" on KFP components
### Background 

Currently, we support mounting data volume on "generic nodes" while in a Kubeflow Pipeline.
I think we should also mount these volumes to the "KFP component nodes".

This will allow data to be exchanged between "generic" and "KFP" components through the mounted volumes, (at least partially solving https://github.com/elyra-ai/elyra/issues/1765), and would also facilitate data exchange between "KFP" components themselves.

### Implementation

For "generic nodes" we currently pass [`volume_mounts` to the `ExecuteFileOp` constructor](https://github.com/elyra-ai/elyra/blob/eb71bd2ad00b868b426f0340db9d1fc2b6e58c96/elyra/pipeline/kfp/processor_kfp.py#L549), which are [looped-over and added with `self.add_volume()`](https://github.com/elyra-ai/elyra/blob/eb71bd2ad00b868b426f0340db9d1fc2b6e58c96/elyra/kfp/operator.py#L238-L253). 

_(NOTE: `self.add_volume()` works on `ExecuteFileOp` because it extends from `kfp.dsl.ContainerOp`)_

Similarly, we can add these volumes to the [`ContainerOps` generated from KFP component definitions](https://github.com/elyra-ai/elyra/blob/eb71bd2ad00b868b426f0340db9d1fc2b6e58c96/elyra/pipeline/kfp/processor_kfp.py#L653) by looping over the list with `add_volume()`.

### Considerations

- We should rename the "Generic nodes defaults" to "Node Defaults" as they would no longer only be for generic nodes
     - _(We may want to say in the description that "Data Volumes" are not applied to Airflow components)_
- We should add node-specific configs for "Data Volume" mounts on KFP components (like we already do for generic ones)
- If we want, we can also apply the global "Environment Variables" using [`container.add_env_variable()`](https://github.com/elyra-ai/elyra/blob/eb71bd2ad00b868b426f0340db9d1fc2b6e58c96/elyra/kfp/operator.py#L259)

kiersten-stokes:
Fixes #2760 

Will also require documentation updates, either here or in a follow-up PR.

I'll flush this description out further after making a few changes.

The `mounted_volumes` property is added to the canvas properties for custom components. It will exist as a sibling-stanza to the `component_parameters` stanza of the pipeline JSON, which is different from how this property is handled for generic components (where it appears within the `component_parameters` stanza). As a result of this difference, some of the property propagation has been updated in `pipeline_definition.py`. Two functions -- `get_property()` and `set_property()` -- are added that deal with this difference. The logic for `get_property()` is below (`set_property` is similar). I'm not sure I like how this is handled so I may fix that later (suggestions welcome).

```python
def get_property(self, key: str, default_value=None) -> Any:
   if self.is_generic_node():
        value = self.get_component_parameter(key, default_value)
   else:
         value = self.get(key, default_value)
   return None if value == "None" else value
```

A new `mounted_volumes` property is also added to both `GenericOperation` and `Operation` (also accessed differently depending on the type). In adding this property, I found that simplifying the `Operation` object construction made things more straightforward, so some of the logic previously done in `parser.py` has been moved to `pipeline.py` in the `Operation` constructors and some utility classes (`KeyValueList` and the dataclasses) have been moved to a new file, `pipeline_utilities.py` to prevent circular reference issues. 

Finally, logic was added to both `processor_kfp.py` and `processor_airflow.py` to handle the addition of volumes for custom components. I also refactored how the Airflow processor handles volumes and secrets in order to simplify the DAG template. Now, all volumes and secrets for the whole pipeline are collected _before_ being template to render. 

Tests will have to be updated and added

 
           in the file; or

           it.


elyra/pipeline/pipeline.py
```diff
@@ -308,6 +355,14 @@ def memory(self) -> Optional[str]:
     def gpu(self) -> Optional[str]:
         return self._component_params.get("gpu")
 
+    @property
+    def mounted_volumes(self) -> List[VolumeMount]:
+        return self._component_params.get(MOUNTED_VOLUMES)
```
kevin-bates, line 360: What is preventing `GenericOperations` from using the same location for `mounted_volumes` and `kubernetes_secrets` as `Operations` (i.e., in `app_data` rather than `app_data.component_parameters`) or vice versa?  This difference is referenced in the opening description and it seems like it would be worth a possible migration to place these pieces of information in the same location.

Using a property to handle this difference is nice, but it would be good if we could have as few differences within the file structure as we can.

kiersten-stokes: Hmm yeah not sure why I hadn't thought of that as an option. Thinking through this, it looks like the main thing that would have to be worked through is that this would necessitate different logic among those properties that need to be propagated. For example, we need to propagate both `env_vars` and `mounted_volumes` (and presumably `env_vars` should stay in the `app_data.component_parameters` stanza). That being said, once we got past that part, this approach would probably result in fewer differences between `Operation`s and `GenericOperation`s. I'm going to give it a try and hope that nothing weird comes up 🙂

kiersten-stokes:
While working through this PR, an issue was encountered that complicated things to a degree that made this approach a lot more complicated than it was worth. Long story short, the `elyra_` prefix that we use to denote which properties are 'component parameters' or 'pipeline default' properties was affecting both how the component properties appeared in the GUI and how they were processed. Will close this and open a new PR in which the `mounted_volumes` property is added as a child to the 'component_parameters' stanza.

PR #2790, opened 2022-06-22
Install jupyterlab during normal builds

ajbozarth:
In #2709 I removed jupyterlab from the build_requires.txt so that
the local versaion of jupyterlab could be used if installed. This
has caused frustration in users, so I've added the dependency back
and update the make dev-install target to not use the dependency
file at all, instead using a new make target to install the
required dependency.

           in the file; or

           it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,6 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
       - name: Build
         run: |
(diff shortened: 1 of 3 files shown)


ajbozarth:
A note on review, we should make sure the following cases all work as expected:

- CI
- Using `make clean install` in a old env
- Using `make clean install` in a fresh env
- Following the steps to use a local instance of jupyterlab as in the docs

ptitzler (approved):
LGTM! Thanks Alex!

Issue #2789, opened 2022-06-22
Empty Output Files Directory Should Be Valid With Wildcard


cupdike:
Validation fails on the Output Files directory when a directory with a wildcard for files is specified (e.g. `data/*`).  Since output files may be generated by the pipeline node, this should be valid.

**To Reproduce**
Create a node with Output Files set to  `data/*` where that directory is empty.  Try to run the pipeline triggering validation.
Note that if you drop an empty file into the directory, it will pass validation.

**Screenshots or log output**
{
  "severity": 1,
  "source": "Elyra Pipeline Validation Service",
  "type": "invalidFilePath",
  "message": "Property(wildcard) has an invalid path to a file/dir or the file/dir does not exist.",
  "data": {
    "nodeID": "9eccdc64-edbe-4dad-8ae5-0d90405a5c13",
    "nodeName": "elyra.trainer.ipynb",
    "propertyName": "dependencies",
    "value": "/home/jovyan/MNIST/data/*"
  }
}

**Expected behavior**
The pipeline should pass validation in this condition.

**Deployment information**
Describe what you've deployed and how:
 - Elyra 3.9.1

**Pipeline runtime environment**
If the issue is related to pipeline execution, identify the environment where the pipeline is executed
 - Kubeflow Pipelines 1.0.4

cupdike:
Closing this.  It's because I had defined File Dependencies in the second node on the mistaken notion that they were treated like inputs from the prior node.  It makes sense that files should exist at a location specified by File Dependencies.  Apologies for the noise...

ptitzler:
+1. We've updated the `elyra-pipeline describe` CLI command to list more dependencies, as shown [in section `Pipeline CLI: identify pipeline dependencies` in the release notes](https://github.com/elyra-ai/elyra/releases/tag/v3.9.0). Unfortunately there is currently no equivalent functionality in the pipeline editor. In your specific scenario the command would have listed  `data/*` as a dependency and might have made it easier to identify the root cause.

cupdike:
> Pointed out by the team, I dont think we currently support wildcards in output file declarations, see #995

Just so you are aware, the docs suggest they are supported: https://elyra.readthedocs.io/en/v3.9.0/user_guide/pipelines.html
> Output Files
> Optional. A list of files generated by the notebook inside the image to be passed as inputs to the next step of the pipeline. Specify one file, directory, or expression per line. Supported patterns are * and ?.
Example: data/*.csv

[...]

Issue #995, opened 2020-10-23
bootstrapper cannot handle input with wildcards


ptitzler:
version 1.3

- create a notebook1 that produces several output files, e.g. `data/file1.csv, data/file2.csv, ...`
- create a notebook2 (it doesn't matter what it does because it won't run)
- create a pipeline notebook1 -> notebook2, configuring the output files for notebook1 as `data/*`
- run the pipeline on kubeflow pipelines

Result:
 - processing of notebook1 succeeds and the output files are properly uploaded to the COS bucket
 - processing of notebook2 fails:

```
[I 23:08:27.715] 'test_load_viz-1023160715':'Data_Viz' - downloaded Data_Viz-2289c970-b214-418b-bf93-68d880326eb0.tar.gz from bucket: pipeline-artifacts, object: test_load_viz-1023160715/Data_Viz-2289c970-b214-418b-bf93-68d880326eb0.tar.gz (0.042 secs)
Traceback (most recent call last):
  File "bootstrapper.py", line 402, in <module>
    main()
  File "bootstrapper.py", line 393, in main
    file_op.process_dependencies()
  File "bootstrapper.py", line 97, in process_dependencies
    self.get_file_from_object_storage(file.strip())
  File "bootstrapper.py", line 137, in get_file_from_object_storage
    self.cos_client.fget_object(bucket_name=self.cos_bucket,
  File "/usr/local/lib/python3.8/site-packages/minio/api.py", line 719, in fget_object
    stat = self.stat_object(bucket_name, object_name, sse)
  File "/usr/local/lib/python3.8/site-packages/minio/api.py", line 1138, in stat_object
    response = self._url_open('HEAD', bucket_name=bucket_name,
  File "/usr/local/lib/python3.8/site-packages/minio/api.py", line 2017, in _url_open
    raise ResponseError(response,
minio.error.NoSuchKey: NoSuchKey: message: The specified key does not exist.
```
 
The corresponding lines of code are
```
        if inputs:
            input_list = inputs.split(INOUT_SEPARATOR)
            for file in input_list:
                self.get_file_from_object_storage(file.strip()).     <------ FAIL
```

To figure out which input file couldn't be processed I had to export the pipeline and inspect the generated bootstrapper script:
```
- name: data-viz
    container:
      args: ['mkdir -p ./jupyter-work-dir/ && cd ./jupyter-work-dir/ && curl -H "Cache-Control:
[...]

Issue #2794, opened 2022-06-23
Output Files That Exists Beget A FileNotFoundError


cupdike:
The training node tries to save a model file to Minio but reports FileNotFound when the file does exist at the intended path.

To Reproduce
Configure the pipeline to mount the PVC of the "authoring" Jupyter Notebook at `/home/jovyan` in the node images.  Create a node with Output Files set to `model/model.pt` and write a file to that location (specifically, `/home/jovyan/model/model.pt` where the pipeline and node notebooks are in `/home/jovyan`). 

![image](https://user-images.githubusercontent.com/133214/175380557-e54259ba-9ab3-48d2-b46a-43f9710eabdd.png)

NOTE: I observe that the current directory in the node at runtime is NOT `/home/jovyan` as I would expect:
#### Incoming CWD: /workspace/jupyter-work-dir\

However, I am forcing it to that so things run:
#### Adjusted CWD: /home/jovyan

If the upload to minio is still happening relative to that unexpected current directory, that might explain why the upload attempt gets a `FileNotFound` (and then the question is why is the current directory not `/home/jovyan`.

Error from "training" node: 
```
Traceback (most recent call last):
  File "bootstrapper.py", line 697, in <module>
    main()
  File "bootstrapper.py", line 687, in main
    file_op.execute()
  File "bootstrapper.py", line 367, in execute
    raise ex
  File "bootstrapper.py", line 359, in execute
    self.process_outputs()
  File "bootstrapper.py", line 148, in process_outputs
    self.process_output_file(file.strip())
  File "bootstrapper.py", line 331, in process_output_file
    self.put_file_to_object_storage(matched_file)
  File "bootstrapper.py", line 309, in put_file_to_object_storage
    self.cos_client.fput_object(bucket_name=self.cos_bucket, object_name=object_to_upload, file_path=file_to_upload)
  File "/home/jovyan/.local/lib/python3.8/site-packages/minio/api.py", line 981, in fput_object
    file_size = os.stat(file_path).st_size
FileNotFoundError: [Errno 2] No such file or directory: 'model/model.pt'
```
Logs from the trainer image indicate that the file does exist at the intended location:
> model_path /home/jovyan/model
model_file_path = model_path / "model.pt"
torch.save(model.state_dict(), model_file_path)
print("#### Does model file exist?", model_file_path.exists())
> '#### Does model file exist? True\n'

Expected behavior
The file should get saved to Minio.

Deployment information
Elyra 3.9.1
[...]

PR #2798, opened 2022-06-24
Update docs to reflect proper generic op output file usage

akchinSTC:

Updated the docs and placeholder text in response to 

https://github.com/elyra-ai/elyra/issues/2789 and 
https://github.com/elyra-ai/elyra/issues/995#issuecomment-715633815

           in the file; or

           it.

### DIFF
--- a/docs/source/user_guide/best-practices-file-based-nodes.md
+++ b/docs/source/user_guide/best-practices-file-based-nodes.md
@@ -63,7 +63,8 @@ Invalid directory layout examples:
 
 #### File output
 
-All changes to the file system (e.g. new files or modified files) are discarded after processing of the notebook/script has completed. To retain these files you must store those files on cloud storage, or declare those files as output files in the notebook/script node properties. 
+All changes to the file system (e.g. new files or modified files) are discarded after processing of the notebook/script has completed. To retain these files you must store those files on cloud storage, or declare those files as output files in the notebook/script node properties.  
+NOTE: If you need to output a large number of files, instead of using a wildcard expression e.g. `directory/*`, compress them into an archive (e.g. tar, zip) as a step in your notebook/workflow and then add the archive here.
 
 ![Define output files](../images/user_guide/best-practices-file-based-nodes/vpe-node-output-files.png)
 


ptitzler (approved):
LGTM! Should we perhaps also reference the data volume topic as it provides an alternative for large (in terms of disk size or number of files) outputs ?

akchinSTC:
> LGTM! Should we perhaps also reference the data volume topic as it provides an alternative for large (in terms of disk size or number of files) outputs ?

Im probably going to follow that up in a separate PR since Id like to address the issues brought up by cupdike in https://github.com/elyra-ai/elyra/issues/2794. Maybe updating the best practices doc and explain the diff between the intermediate volume workspace and the output file parameter

cupdike:
Apologies if I am missing context here, but doesn't this suggest that wildcards actually are supported?
https://github.com/elyra-ai/elyra/blob/main/elyra/airflow/bootstrapper.py#L197

akchinSTC:
thank you @cupdike.
@ptitzler, im probably going modify this PR a bit more while fixing #995, since the current documentation is correct (albeit with the bug). Listing files in outputs with wildcards does function correctly, but subsequent child nodes that we pass the wildcard path to cannot properly expand it and grab from s3. 
wondering if we should disable wildcarding for now, so that subsequent nodes are able to proceed and provide guidance on how to properly push/pull artifacts to/from s3 (via compressed archives), or perhaps add the compression steps when wildcards are present to the bootstrapper and make the process transparent ?

ptitzler:
> Listing files in outputs with wildcards does function correctly, but subsequent child nodes that we pass the wildcard path to cannot properly expand it and grab from s3. wondering if we should disable wildcarding for now, so that subsequent nodes are able to proceed 

In a way doing so would result in a breaking change, as there _might_ be users that use this feature to make artifacts available in the last (or only) pipeline node.  
> and provide guidance on how to properly push/pull artifacts to/from s3 (via compressed archives), or perhaps add the compression steps when wildcards are present to the bootstrapper and make the process transparent ?

Should we perhaps we update the best practices document for now and fix the implementation (based on your suggestion) in a future release? We are at the tail end of the 3.10 release and I'm not sure the latter can be handled in time.

PR #2799, opened 2022-06-24
Support data exchange between custom & generic components via volumes

### LINKED ISSUE #2760: mount "data volumes" from "generic nodes defaults" on KFP components

kiersten-stokes:
Fixes #2760

While working through #2785, an issue was encountered that complicated things to a degree that made that approach almost useless. Long story short, the `elyra_` prefix that we use to denote which properties are 'component parameters' or 'pipeline default' properties was affecting both how the component properties appeared in the GUI and how they were processed. In this PR, a different approach is taken, as described below.

The `mounted_volumes` property is added to the canvas properties for custom components. It will exist as a child within the `component_parameters` stanza of the pipeline JSON, which is the same as how this property is handled for generic components. Processing proceeds the same for both generic and custom components in the case where the custom component does not define a property of the same id (`mounted_volumes`) in its component definition.

A few additional safeguards are required to handle the case where a property id collision does occur (e.g. the component defines its own property called `mounted_volumes`). First, a new keyword arg is passed to the properties template for custom components that tells the template not to render any Elyra-owned properties that have the same id as any component-defined properties that are parsed from the definition. Similar logic is added to `pipeline_definition.py` that checks for any property id collisions and does not perform certain logic (property propagation from pipeline defaults, conversion to relevant dataclass objects, etc.) if the property is determined to be component-defined. Any value entered for a component-defined property of the same id should not be touched, but rather passed directly to the component constructor (via the DAG or the `ContainerOp` object).

Much of the above paragraph could be simplified or removed should we decide that we want to do a pipeline migration, in which case we could change the id for the built-in `mounted_volumes` property to something that we could be nearly sure would not result in a collision of property ids (though I suppose this could never be guaranteed so perhaps we wouldn't want to remove the safeguards?).

A new `mounted_volumes` property is also added to the `Operation` class. A `kubernetes_secret` property is added to the `GenericOperation` class as well just to make processing more straightforward.

Finally, logic was added to both `processor_kfp.py` and `processor_airflow.py` to handle the addition of volumes for custom components. I also refactored how the Airflow processor handles volumes and secrets in order to simplify the DAG template. Now, all volumes and secrets are rendered by functions defined in `processor_airflow.py` and passed to the template env. (I think we could also further re-factor this in a future PR to make the DAG template even more straightforward).

Documentation has been updated to reflect the availability of volumes on custom components.

Existing server tests have been updated. Tests have been added to cover the following scenarios:
 1. Ensure that a component that defines its own property called `mounted_volumes` results in the appropriate properties JSON (i.e., skips the system-defined `mounted_volumes` property and only defines its own)
 2. Ensure that a pipeline with custom components and mounted volumes defined has its properties propagated appropriately (propagation works the same as with generic nodes)
 3. Ensure that a custom component that defines its own property called `mounted_volumes` does not go through the processing steps that the system-defined property of the same name does (propagation, conversion to dataclass, etc.).

 
           in the file; or

           it.


kiersten-stokes:
> I view things like mounted_volumes, kubernetes_secrets, env_vars, etc., as system-owned properties. If we were to reserve the elyra_ prefix as an indicator of system-owned properties and not only apply the prefix in a temporal manner (as we currently do) and make it permanent, would that be sufficient to move forward, albeit at the cost of a migration?

Just to clarify, do you mean adding the `elyra_` prefix to all `component_parameters` (including system owned) and then another `elyra_` prefix to those system-owned properties? E.g., for the purposes of the canvas template, `mounted_volumes` would become `elyra_elyra_mounted_volumes` and then for the purposes of pipeline processing,  the `MOUNTED_VOLUMES`constant would become `elyra_mounted_volumes`?

I think that would be all that's needed! And I would imagine it would be fairly easy migration code too (I may even be able to do it myself 😄 ). 


> (I'm assuming our system-owned properties have corresponding display names so the elyra_ wouldn't appear in various UI dialogs.)

Correct! The property id (called the ref in the canvas properties) is never displayed.

kevin-bates:
> Just to clarify, do you mean adding the elyra_ prefix to all component_parameters (including system owned) and then another elyra_ prefix to those system-owned properties? 

No, only prefix system-owned properties.  This way, `elyra_` or whatever we choose as the prefix _is_ the indicator that we control that property.

elyra/pipeline/pipeline_definition.py
```diff
@@ -15,23 +15,26 @@
 #
 import json
 import os
-from typing import Any
+from typing import Any, Set
```
kiersten-stokes, line 18: Hmm not sure why this is passing lint? (and same with line 23 below). I'll fix it but might be worth looking into further?

kiersten-stokes: Per discussion, it appears that the linting rules that we use will allow duplicates per line for the `typing` module. I've already separated them in this PR, but something to keep in mind for the future.

docs/source/user_guide/pipelines.md
```diff
@@ -140,15 +141,17 @@ The [tutorials](/getting_started/tutorials.md) provide comprehensive step-by-ste
...
    - Optional. A list of [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to be accessed as environment variables during Jupyter notebook or script execution. Format: `ENV_VAR=secret-name:secret-key`. Entries that are empty (`ENV_VAR=`) or malformed are ignored. Entries with a secret name considered to be an [invalid Kubernetes resource name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) or with [an invalid secret key](https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data) will raise a validation error after pipeline submission or export. The referenced secrets must exist in the Kubernetes namespace where the generic pipeline nodes are executed.
    - Secrets are ignored when the pipeline is executed locally. For remote execution, if an environment variable was assigned both a static value (via the 'Environment Variables' property) and a Kubernetes secret value, the secret's value is used.
    - Example: `ENV_VAR=secret-name:secret-key`
+   
+   Both generic and [custom components](pipeline-components.html#custom-components) include the following system-owned property:
```
ptitzler, line 149: Remove internal 'lingo'

```suggestion
   Both generic and [custom components](pipeline-components.html#custom-components) include the support the following property:
```

ptitzler:
- Documentation is looking good
- Kubeflow Pipelines test completed (including CLI validation)
- Apache Airflow test in progress ...

akchinSTC (approved):
LGTM, kfp portion. thanks!

PR #2815, opened 2022-06-29
Fix Airflow operation processing for number data types

kiersten-stokes:
In testing #2799, I came across this issue where number-type component parameter values are being rendered by the DAG as follows because these types were being skipped during processing:

```
int_default_non_zero= {
  "activeControl": "StringControl",
  "StringControl": "['value1', 'value2']",
},
```

After these changes, the same property appears as it should:

```
int_default_non_zero=2,
```

This is not an issue for KFP.

A simple 2-line change takes care of handling number-type parameter values (which was erroneously being skipped prior to this). The bulk of the changes here are test-related (see below).

Previously, no test existed that checked for how number-type values were being processed by Airflow. A case for that is added here to an existing test (`test_create_file_custom_components`). To support this test, I've added a new catalog instance (`AIRFLOW_TEST_OPERATOR_CATALOG`) to `conftest.py` that loads the `TestOperator` defined in `tests/pipeline/resources/components/airflow_test_operator.py`, which includes a component parameter for each supported data type. I've also updated the sample pipeline used during this test to include the `TestOperator` instead of the `BashOperator`. 

Related, I used the same strategy to update `test_same_name_operator_in_pipeline` (also in `test_airflow_processor.py`). This test now uses the `AIRFLOW_TEST_OPERATOR_CATALOG` that includes the `TestOperator` and the test pipeline resource is similarly updated to use `TestOperator` parameters. While not strictly necessary, I think with only a few more similar changes, we can completely remove the `BashOperator` from our test resources.

           in the file; or

           it.


ptitzler (approved):
LGTM!

Issue #2826, opened 2022-07-11
Investigate possible memory leak issue using jsonschema.validate()


kevin-bates:
In last week's [Jupyter Server meeting](https://github.com/jupyter-server/team-compass/issues/15#issuecomment-1180451886) it was brought to our attention that the `jsonschema` package can exhibit a memory leak when the `validate` method is repeatedly called to validate an instance of a given schema.  The leak isn't necessarily in the validation of the instance itself, but rather in the acquisition of the _validator_ used to perform the validation and the `validate` method's documentation even states:

_"If you know you have a valid schema already, especially if you intend to validate multiple instances with the same schema, you likely would prefer using the Validator.validate method directly on a specific validator (e.g. Draft7Validator.validate)."_

Since Elyra validates all instances on retrieval (and save) and is intended to remain running for long periods of time, we should look into how this impacts Elyra vs. the cost to address this issue.

PR #2829, opened 2022-07-11
Add support for schema validators

kevin-bates:
This pull request adds support for schema validators after learning that the default `validate()` method in `jsonschema` can leak memory when called repeatedly.  (See #2826 for details)

Aside from using the `Draft7Validator` by default, we also expose the ability for `SchemasProvider` implementors to bring their own validators as part of their provider.

Rather than create a validator instance whenever validation is required (as was done in the results for #2826), the `SchemaManager` captures the validator's instance upon loading the `SchemasProvider`.  By doing this, the overall test time went from 16 seconds to 2 seconds, thereby gaining an overall 50X performance improvement over using the default `validate()` method done originally. 
 
Resolves: #2826

           in the file; or

           it.


kiersten-stokes (approved):
Good idea to allow custom validators - everything is working well for me!

Issue #2782, opened 2022-06-10
Review/address 'Elyra Tests: Test Server' warnings


ptitzler:
**Describe the issue**
The most recent CI 'Elyra Tests: Test Server' tests produce 700+ warnings. 

**To Reproduce**
1. Open https://github.com/elyra-ai/elyra/actions
2. Choose one of the 'Elyra Tests'
3. Review the 'Test Server' log

Many warnings are deprecation messages that will result in an error in the future.

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.10/main

ptitzler:
Earmarking 3.11 for investigation

ptitzler:
Warning related to KFP/generic components:
> FutureWarning: Please create reusable components instead of constructing ContainerOp instances directly. Reusable components are shareable, portable and have compatibility and support guarantees.

To resolve, the [implementation of generic components](https://github.com/elyra-ai/elyra/blob/main/elyra/kfp/operator.py#L70) would have to be redone as a custom component. Sizing estimate: ~med

kevin-bates:
Following the merge of #2833 and #2829, there are still 1688 warnings.  However, 1632 (all but 56) of these can be attributed to the recent release of tornado 6.2 vs. 6.1.

Each of the following entries produce 544 warning each in our test - adding to 1632:
```
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/ioloop.py:350: DeprecationWarning: make_current is deprecated; start the event loop first
    self.make_current()

/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/pytest_tornasync/plugin.py:65: DeprecationWarning: make_current is deprecated; start the event loop first
    loop.make_current()

/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/pytest_tornasync/plugin.py:67: DeprecationWarning: clear_current is deprecated
    loop.clear_current()
```

Of the remaining 56 warnings...

15 can be attributed to (KFP SDK?):
```
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/kfp/dsl/_container_op.py:75: PendingDeprecationWarning: `dsl.ContainerOp.image` will be removed in future releases. Use `dsl.ContainerOp.container.image` instead.
    warnings.warn(
```

14 to (Papermill):
```
[...]

PR #2833, opened 2022-07-13
Address test warnings

kevin-bates:
This pull request is a "first pass" at addressing the warnings produced following the server CI tests.  Prior to this PR `3791` warnings were produced.  This PR takes the warnings down to `474` (87%)  and when merged with #2829, the warnings will be around `55` (98%).  (Note, these values were produced on my local dev env.   In running CI on my fork, and without #2829, I still see `2108` warnings, but `474` on my local system.  I suspect the additional warnings are related to recent deprecations in `tornado` regarding event loop management and I'm running with an older version.)

The primary issue here was in handling class initialization parameters within the `Configurable` class hierarchy.  The PR currently has two commits, each taking a different approach to resolving the issue.  
- The first was to add a positional argument (for the most part `root_dir`) so that it is not included in the `**kwargs` that _flow_ to the superclasses.  This approach probably consisted of fewer changes than the second because that was closer to what was happening before.  
- The second approach was to use `**kwargs` but _pop_ parameters like `root_dir` out, so that it doesn't interfere in the superclasses.  This approach was chosen because I found it was used _within_ the `traitlets` package, lending credence to that approach.

I also refactored the `PipelineProcessorResponse` class, introducing a `RuntimePipelineProcessorResponse` class.  The KFP and Airflow response objects derive from the latter while the Local response object derives directly from `PipelineProcessorResponse` since it doesn't need the location information in the response and was just passing empty string placeholders for those values.  In addition, this class hierarchy now mirrors that of the `PipelineProcessor` class hierarchy.

In some cases, you'll see the `__init__()` method completely removed. This is intentional as it is not required if the traitlet (e.g., `root_dir`) is the only item needed from the `**kwargs` and is functionality the `traitlets` package provides automatically.

I'm going ahead and marking #2782 as resolved by this PR since most all of the remaining errors can be attributed to upstream components (Tornado, KFP, and Papermill) and should probably be addressed with separate issues for each.

Resolves: #2782
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -43,8 +43,8 @@ from elyra.pipeline.pipeline import Operation
 from elyra.pipeline.pipeline import Pipeline
 from elyra.pipeline.pipeline_constants import COS_OBJECT_PREFIX
 from elyra.pipeline.processor import PipelineProcessor
-from elyra.pipeline.processor import PipelineProcessorResponse
 from elyra.pipeline.processor import RuntimePipelineProcessor
+from elyra.pipeline.processor import RuntimePipelineProcessorResponse
 from elyra.pipeline.runtime_type import RuntimeProcessorType
 from elyra.util.cos import join_paths
 from elyra.util.github import GithubClient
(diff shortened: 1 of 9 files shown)


ptitzler (approved):
LGTM! Without those warnings, it's now much easier to identify test issues. Much appreciated!

kiersten-stokes (approved):
> In some cases, you'll see the __init__() method completely removed. This is intentional as it is not required if the traitlet (e.g., root_dir) is the only item needed from the **kwargs and is functionality the traitlets package provides automatically.

Cool! LGTM

PR #2862, opened 2022-07-28
Remove deprecation warning for local runtime type

### LINKED ISSUE #2855: Deprecation warnings in pipeline endpoints should be addressed
I just noticed the following deprecation warnings are produced when starting Elyra.  Calls to these endpoints should be updated to use the appropriate runtime type names.
```
[W 2022-07-26 09:54:44.221 ServerApp] Deprecation warning: when calling endpoint '/elyra/pipeline/components/local' use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') instead of shorthand name (e.g., 'kfp', 'airflow')
[W 2022-07-26 09:54:44.229 ServerApp] Deprecation warning: when calling endpoint '/elyra/pipeline/local/properties' use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') instead of shorthand name (e.g., 'kfp', 'airflow')
[W 2022-07-26 09:54:44.439 ServerApp] Deprecation warning: when calling endpoint '/elyra/pipeline/components/local/notebook/properties' use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') instead of shorthand name (e.g., 'kfp', 'airflow')
[W 2022-07-26 09:54:44.443 ServerApp] Deprecation warning: when calling endpoint '/elyra/pipeline/components/local/python-script/properties' use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') instead of shorthand name (e.g., 'kfp', 'airflow')
[W 2022-07-26 09:54:44.445 ServerApp] Deprecation warning: when calling endpoint '/elyra/pipeline/components/local/r-script/properties' use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') instead of shorthand name (e.g., 'kfp', 'airflow')
```
To most easily reproduce the issue, start Elyra w/o debug logging enabled.  These will be produced on startup and subsequent browser refreshes.

kiersten-stokes:
Fixes #2855 

Based on some discussion related to #2855, we decided to consider lowercase `local` equivalent to uppercase `LOCAL` when checking against runtime types. This PR suppresses the warning message about using (eventually) deprecated runtime type names (`kfp`, `airflow`) for the local runtime only. This is the 'path of least resistance' for this issue. If we ever fully remove the ability to query with the lowercase `local`, we'll probably want the frontend to adjust its API call to use the `LOCAL` `RuntimeProcessorType` Enum member.

Queried the `elyra/pipeline/components/{runtime_type}` endpoint with `local` as the `runtime_type`, and no warning message is shown in the logs. A new test is added to cover this case as well.
 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/handlers.py
+++ b/elyra/pipeline/handlers.py
@@ -57,14 +57,15 @@ def get_runtime_processor_type(runtime_type: str, log: Logger, request_path: str
     """
     processor_manager = PipelineProcessorManager.instance()
     if processor_manager.is_supported_runtime(runtime_type):
-        # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
-        # 'airflow'). This case and its associated functions should eventually be removed
-        # in favor of using the RuntimeProcessorType name in the request path.
-        log.warning(
-            f"Deprecation warning: when calling endpoint '{request_path}' "
-            f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
-            f"instead of shorthand name (e.g., 'kfp', 'airflow')"
-        )
+        if runtime_type != "local":
+            # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
+            # 'airflow'). This case and its associated functions should eventually be removed
+            # in favor of using the RuntimeProcessorType name in the request path.
+            log.warning(
+                f"Deprecation warning: when calling endpoint '{request_path}' "
+                f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
+                f"instead of shorthand name (e.g., 'kfp', 'airflow')"
+            )
         return processor_manager.get_runtime_type(runtime_type)
     elif processor_manager.is_supported_runtime_type(runtime_type):
         # The request path uses the appropriate RuntimeProcessorType name. Use this
--- a/elyra/pipeline/handlers.py
+++ b/elyra/pipeline/handlers.py
@@ -56,21 +56,20 @@ def get_runtime_processor_type(runtime_type: str, log: Logger, request_path: str
     :returns: the RuntimeProcessorType for the given runtime_type, or None
     """
     processor_manager = PipelineProcessorManager.instance()
-    if processor_manager.is_supported_runtime(runtime_type):
-        if runtime_type != "local":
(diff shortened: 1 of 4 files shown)


elyra/pipeline/handlers.py
```diff
@@ -57,14 +57,15 @@ def get_runtime_processor_type(runtime_type: str, log: Logger, request_path: str
...
+            # The endpoint path contains the shorthand version of a runtime (e.g., 'kfp',
+            # 'airflow'). This case and its associated functions should eventually be removed
+            # in favor of using the RuntimeProcessorType name in the request path.
+            log.warning(
+                f"Deprecation warning: when calling endpoint '{request_path}' "
+                f"use runtime type name (e.g. 'KUBEFLOW_PIPELINES', 'APACHE_AIRFLOW') "
+                f"instead of shorthand name (e.g., 'kfp', 'airflow')"
+            )
         return processor_manager.get_runtime_type(runtime_type)
     elif processor_manager.is_supported_runtime_type(runtime_type):
```
kevin-bates, line 70: If we switch the order of these if statements, checking for runtime-type first, and loosen the case-sensitivity in https://github.com/elyra-ai/elyra/blob/ef4cba3892999028ed616b24fd797fcc5d57c4ca/elyra/pipeline/processor.py#L90 and
https://github.com/elyra-ai/elyra/blob/ef4cba3892999028ed616b24fd797fcc5d57c4ca/elyra/pipeline/runtime_type.py#L48
(upcasing each item for the comparison and lookup) then I think this extra condition could be avoided and similar scenarios from the CLI (that don't go through the handlers) could also be addressed.

kiersten-stokes: This is how I originally handled it, but I was worried I was over-complicating 😄 but it is the way I like slightly better since it would allow removal of the hard-coded `local`

kevin-bates: Trust your instincts! :smile:  (Of course, watch something subtle about all of this bite us!)

kevin-bates (commented):
Sorry, but would it be possible to add a handlers test that used `local` (and reproduces the problem prior to this fix)?  If that turns out to be a hassle, I'm okay moving forward, but I think it would be good to have in case we make changes later that wind up side-affecting this PR.

kevin-bates (approved):
Thanks for adding the test - this LGTM.

PR #2868, opened 2022-08-05
Add support for Kubernetes annotations

### LINKED ISSUE #2849: Allow to specify annotation for steps execution pods
We are using KFP integrated with DAPR, where DAPR interactions is all driven by annotations. As a result, we do need to enable
annotating pods.

We would like to be able to specify annotations on the specific pods, similar to how environment variables are currently supported. In a most simplistic approach, a drop-down on the pod allowing for specifying a set of annotations for a given step

modifying the resulting YAML file, which does not seem to be practical

Here is an example of how we are doing this now:

````
    ray_op(
              times = times, 
              ray_address = ray_address,
              exp = exp,
              runid = runid)\
        .add_env_variable(
            k8s_client.V1EnvVar(name="PUBSUB_NAME", value="pubsub"))\
        .add_env_variable(
            k8s_client.V1EnvVar(name="TOPIC_NAME", value="postevents"))\
        **.add_pod_annotation(
            name="dapr.io/enabled", value="true")\
        .add_pod_annotation(
            name="dapr.io/app-id", value="ray-processor")**
````

### PR DESCRIPTION
**Pull request by ptitzler**
This PR adds support for [Kubernetes annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) for Kubeflow Pipelines and Apache Airflow runtime environments.

Annotations are supported for generic components and custom components and can be defined as pipeline defaults or for individual nodes.

Input needs to be entered as a string using the `annotation_key=annotation_value` format until proper UI support (https://github.com/elyra-ai/elyra/pull/2780) is released. 

Annotation keys are validated according to [this spec](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set).

Closes #2849

**Pipeline default settings in VPE**
  Displayed in section "Node defaults" (Settings in this section apply to all nodes):

   image

**Generic node setting in VPE**
 
   image

**Custom node setting in VPE**
Displayed in section "Additional properties"
  
   image

**Applied annotations (KFP)**

Example `describe pod` output (KFP)
 ```
 $ kubectl describe pod lambda-... -n kubeflow-user-example-com
Name:         lambda-6sj47-3256227022
Namespace:    kubeflow-user-example-com
Priority:     0
Node:         cloning1.fyre.ibm.com/9.30.189.14
Start Time:   Sun, 07 Aug 2022 22:57:16 -0700
Labels:       pipeline/runid=e0b80cae-9c4b-4e89-a594-0b7c7eb55c16
              pipelines.kubeflow.org/cache_enabled=true
              pipelines.kubeflow.org/cache_id=
              pipelines.kubeflow.org/enable_caching=true
              pipelines.kubeflow.org/kfp_sdk_version=1.8.13
              pipelines.kubeflow.org/metadata_context_id=1
              pipelines.kubeflow.org/metadata_execution_id=1
              pipelines.kubeflow.org/metadata_written=true
              pipelines.kubeflow.org/pipeline-sdk-type=kfp
              workflows.argoproj.io/completed=true
              workflows.argoproj.io/workflow=lambda-6sj47
Annotations:  a-to-Z: soup to nuts
              annotationA: annotation default  A
              pipelines.kubeflow.org/arguments.parameters: {"URL": "http://www.bing.com"}
              pipelines.kubeflow.org/component_ref: {"digest": "8e4384f422a088e4814024df7955e952c1488bd091fa0d4873d5f611d741ceb4"}

 ```

**Applied annotations (Airflow)**

The applied annotations can be inspected using the `kubectl describe pod` command or in the _DAG task instance details_ page in the Airflow UI:

 image

Notes:
  - Status (if checked, functionality is implemented and was verified):
    - [x] Pipeline editor: validation  
    - [x] KFP generic node property
    - [x] KFP custom node property
    - [x] KFP pipeline default
    - [x] Airflow generic node property
    - [x] Airflow custom node property
    - [x] Airflow pipeline default

- Added new optional 'Kubernetes pod annotations' pipeline default property
- Added new optional 'Kubernetes pod annotations' node property
- Updated the `user_guide/pipelines.html` documentation 

- Added new validation tests
- Manual testing (see scope and status in previous section)
- Reviewed output of `make docs`

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -334,6 +334,7 @@ be fully qualified (i.e., prefixed with their package names).
                     "doc": operation.doc,
                     "volumes": operation.mounted_volumes,
                     "secrets": operation.kubernetes_secrets,
+                    "kubernetes_pod_annotations": operation.kubernetes_pod_annotations,
                 }
 
                 if runtime_image_pull_secret is not None:
@@ -447,6 +448,7 @@ be fully qualified (i.e., prefixed with their package names).
                     "is_generic_operator": operation.is_generic,
                     "doc": operation.doc,
                     "volumes": operation.mounted_volumes,
+                    "kubernetes_pod_annotations": operation.kubernetes_pod_annotations,
                 }
 
                 target_ops.append(target_op)
@@ -499,6 +501,7 @@ be fully qualified (i.e., prefixed with their package names).
                 "render_executor_config_for_custom_op": AirflowPipelineProcessor.render_executor_config_for_custom_op,
                 "render_secrets_for_generic_op": AirflowPipelineProcessor.render_secrets_for_generic_op,
                 "render_secrets_for_cos": AirflowPipelineProcessor.render_secrets_for_cos,
+                "render_executor_config_for_generic_op": AirflowPipelineProcessor.render_executor_config_for_generic_op,
             }
             template.globals.update(rendering_functions)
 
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: elyra/pipeline/airflow/processor_airflow.py [hunk 1]
```diff
@@ -636,18 +639,59 @@ def render_executor_config_for_custom_op(op: Dict) -> Dict[str, Dict[str, List]]
...
+                    {
+                        "key": annotation.key,
+                        "value": annotation.value,
+                    }
+                )
+
+        return executor_config
+
+    @staticmethod
+    def render_executor_config_for_generic_op(op: Dict) -> Dict[str, Dict[str, List]]:
```
kevin-bates, line 676: Should this also take an `executor_config: Dict` parameter and then we could just call it in `render_executor_config_for_custom_op()`?  What is unique about a generic operation here?

However, I think what would be ideal is to extend the `@dataclass` objects for these "add-ons" to be richer (and siblings of the same base class) where we could also hang all of the validation logic.   There are multiple patterns forming here that I think could be eliminated (or condensed) - validation, "rendering", etc.

If items are specific to a platform, the processor_xxx for platform xxx could use subclasses specific to xxx and apply those methods there - so that platform specifics don't need to be included in the "base" classes.

kiersten-stokes: > However, I think what would be ideal is to extend the @dataclass objects for these "add-ons" to be richer (and siblings of the same base class) where we could also hang all of the validation logic. There are multiple patterns forming here that I think could be eliminated (or condensed) - validation, "rendering", etc.

You've read my mind - I've got such changes in the works 🙂

ptitzler: > Should this also take an executor_config: Dict parameter and then we could just call it in render_executor_config_for_custom_op()? What is unique about a generic operation here?

The `KubernetesPodOperator` supports certain parameters natively that other operators don't. Therefore the `executor_config` isn't necessarily the same for generic and custom components.

kevin-bates: I understand and agree that the overall methods should remain separate.  What I'm getting at is that L661-L666 == L679-L684 and, at that granularity, we shouldn't duplicate things - which led to the comment about converting the `@dataclass` instances to actual class instances.

ptitzler: I guess a general overhaul of the Airflow code generation in an upcoming release might be the best path forward as your feedback also applies to https://github.com/elyra-ai/elyra/pull/2848. As is, this PR and 2848 will result in a ton of merge conflicts that still need to be sorted out and we have only a few days left before the planned release date. 

Also it appears that the current code generation for volumes unnecessarily differentiates between generic and custom ops. Both can utilize the `executor_config` to define mounts/volumes, resulting in the removal of a non-trivial amount of code.

wdyt?

kevin-bates: Sounds good.  We can circle back on this when time permits.

kevin-bates (approved):
Thanks @ptitzler - this will be extremely useful!

kiersten-stokes (approved):
All my scenarios are testing as expected! 

(Also, 20 files changed, wow! Hoping to reduce that for future params with some upcoming re-factoring 🤞 )

PR #2888, opened 2022-08-16
Remove .ONESHELL Makefile directive so multi-step targets properly fail

### LINKED ISSUE #2886: CI does not properly handle lint errors
**Describe the issue**
The `lint-server` github action https://github.com/elyra-ai/elyra/blob/main/.github/workflows/build.yml#L46-L52 does not terminate as expected if `flake8` raises an error, as seen [here](https://github.com/elyra-ai/elyra/runs/7811501848?check_suite_focus=true) and all subsequent CI runs.

Note that `flake8` returns a non-zero status code, yet the next commands (echo + black) in the `Makefile` are executed: 
image
Since the last command (`python -m black ...`) returns a zero return code, the action's outcome is interpreted as success.

`make lint-server` behaves as expected in a local build environment:

```
$ make lint-server
python3 -m flake8 elyra .github
elyra/tests/pipeline/test_handlers.py:33:1: I100 Import statements are in the wrong order. 'from elyra.pipeline.processor import PipelineProcessorManager' should be before 'from elyra.tests.util.handlers_utils import expected_http_error'
elyra/tests/pipeline/test_handlers.py:33:1: I202 Additional newline in a group of imports. 'from elyra.pipeline.processor import PipelineProcessorManager' is identified as Application and 'from elyra.tests.util.handlers_utils import expected_http_error' is identified as Application.
elyra/tests/pipeline/test_handlers.py:36:1: I100 Import statements are in the wrong order. 'from elyra.pipeline.parser import PipelineParser' should be before 'from elyra.pipeline.validation import ValidationResponse'
make: *** [lint-server] Error 1

```

kevin-bates:
#2886 documents an issue where the `lint-server` Makefile target does not stop when the `flake8` portion of the multi-step target encounters a non-zero exit code.  Based on initial investigation, there appears to have been a change in behavior between GNU `make` 3.81 (which I, and probably others, use on our desktops) and 4.2.1 (which is used in CI, with 4.3 being the latest).  In versions < 4, the multi-step targets will stop on non-zero exit status, but with versions < 4, that is not the case.

It turns out that the `.ONESHELL` directive was causing this issue - which makes sense - because only the last command of the target is the status that is checked.  Not sure why 3.x versions work - perhaps they don't properly support `.ONESHELL`.

At any rate, `.ONESHELL` was added in #2721 and there's no mention during the review as to why.  I have confirmed appropriate behavior without the directive so perhaps the reason for it is no longer present or it was thought to be an improvement - especially since desktop builds were still behaving okay.  Since this is a fairly serious issue, we should move forward without `.ONESHELL` and tackle any related issues of that when they come up.

Resolves #2886 
 
           in the file; or

           it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,6 @@
 .PHONY: elyra-image elyra-image-env publish-elyra-image kf-notebook-image publish-kf-notebook-image
 .PHONY: container-images publish-container-images validate-runtime-image validate-runtime-images
 
-.ONESHELL:
-
 SHELL:=/bin/bash
 
 # Container execs


akchinSTC (approved):
confirmed that once one shell was removed linting and exiting was running as it should 
image

kevin-bates:
Thanks @akchinSTC.  I'm curious what version of `make` you use.

akchinSTC:
> Thanks @akchinSTC. I'm curious what version of `make` you use.

image

Issue #2256, opened 2021-10-26
Add a control for dictionary types in the node properties panel


kiersten-stokes:
**Is your feature request related to a problem? Please describe.**
With the support for BYO catalog types, what is currently the `Component Source` property in the node properties panel must change from a generic string indicating a url or path to a dictionary of key-value pairs that indicate where that component specification file can be found.

e.g., the new version of `Component Source` should be able to support the following format (as received in the properties JSON for a component)
```
{
  "hostname": "http://localhost",
  "port": "8000",
  "component_id": "filter-text"
}
```

**Describe the solution you'd like**
The new control should be able to display a dictionary with string-valued keys and string values. This also needs to be read-only; a user should not be able to edit anything in this property.

PR #2927, opened 2022-09-14
Make multi-valued input fields more user-friendly

### LINKED ISSUE #2738: Syntax for mounting PVC Volumes is confusing
Currently, we use a very esoteric syntax for mounting PVCs to pipelines `/mnt/path=existing-pvc-name`.

The current UI asks the user to specify a list of strings, without even explaining what the syntax is in the UI.

__Screenshot of current state in the "node properties" section:__

Screen Shot 2022-05-25 at 12 28 01

We can make this MUCH easier to use by making it a list with TWO fields, one for "mount path" and the other for "PVC name". 
(OR: At very least, we can update the text above the list to explain the `MOUNT_PATH=PVC_NAME` syntax)

### LINKED ISSUE #2750: Simplify input of Kubernetes secrets in the pipeline editor
Version 3.9.0 introduced [support for Kubernetes secrets](https://elyra.readthedocs.io/en/latest/user_guide/best-practices-file-based-nodes.html#handling-sensitive-information), enabling users to make sensitive information available to notebooks/scripts via environment variables, without exposing sensitive information in the pipeline editor^, the pipeline file^^, or the runtime environment^^^. The current input format `ENV_VAR=secret-name:secret-key` (see [instructions](https://elyra.readthedocs.io/en/latest/user_guide/pipelines.html#adding-nodes)) is not user friendly.

^ the secret's value is never displayed in the GUI, nor is it ever accessed or logged by Elyra
^^ only the secret name and the secret's key is stored in the .pipeline file and generated DAGs, but not its value
^^^ assuming that access to the Kubernetes secret is protected and its value not exposed by the notebooks/scripts during execution (e.g. via logging)
  
Replace the current string based input with a widget that's intuitive to use.

See related issue https://github.com/elyra-ai/elyra/issues/2738, which raises the same requirement for data volume input.

### LINKED ISSUE #2906: Add support for three-valued BooleanControl

The `BooleanControl` widget enables users to specify a boolean property value by checking/unchecking a box. This implementation does not support scenarios where users should be able to choose from three values: true, false, and null, with null representing 'not set'.

**Example**
 Enable users to prevent use of cached node outputs
 - user explicitly declines use of cached node outputs
 - user explicitly accepts use of cached node outputs
 - user defers the decision to the runtime environment (Elyra does not have  any insights into how the environment is configured) 

Perhaps a select box with three options. The label text for each option should be customizable, and the returned values be one of true/false/null. 

It might probably be possible to implement a one-off solution using `EnumControl` but it wouldn't be re-usable.

### LINKED ISSUE #2923: Airflow pipeline editor generates unnecessary code for custom components
**Describe the issue**

The pipeline editor always generates the following code snippet if in the Airflow runtime configuration the COS authentication type is set to `KUBERNETES_SECRET`.

```
# Ensure that the secret 'airflow-cos-secret' is defined in the Kubernetes namespace where the pipeline is run
env_var_secret_id = Secret(
    deploy_type="env",
    deploy_target="AWS_ACCESS_KEY_ID",
    secret="airflow-cos-secret",
    key="AWS_ACCESS_KEY_ID",
)
env_var_secret_key = Secret(
    deploy_type="env",
    deploy_target="AWS_SECRET_ACCESS_KEY",
    secret="airflow-cos-secret",
    key="AWS_SECRET_ACCESS_KEY",
)
```

If a pipeline only includes custom operators these variables are never used because they are only relevant to generic operators, which use them to access cloud storage.

**To Reproduce**
1. Create an airflow RTC, which is configured to use `KUBERNETES_SECRET` for object storage authentication
2. Create an airflow pipeline and add any custom operator
3. Export the pipeline and review the generated DSL

**Expected behavior**

**Deployment information**
 - Elyra version: source/main

kiersten-stokes:
Fixes #2750
Fixes #2738
Fixes #2923 
Fixes #2906

Corresponds to elyra-ai/pipeline-editor#201.

Excuse all the extraneous commits - I was building and merging with the `pipeline-rjsf` branch so often and for so long that I didn't want to mess with the history by cherry-picking. The diff, however, is accurate.

Note that the pipeline editor may go blank when trying to edit multi-valued fields at this point during development. Front end changes required before things are fully functional. 

This PR re-factors a significant portion of backend node and pipeline properties handling in order to centralize and simplify validation and processing of Elyra-owned properties (which includes properties such as `mounted_volumes`, `kubernetes_secrets`, etc. that are not properties considered a part of the operator/component definition itself).

TODOs:
- [x] **frontend**: pipeline editor goes blank
- [x] **frontend**: tooltip for airflow nodes doesn't render properties properly
- [x] **frontend**: update CSS/styling of new input format
- [x] **frontend**: update default property propagation for new format
- [x] **frontend**/**backend**: integrate UI placeholders (or decide on a title) for each input field
- [x] **frontend**: update property display on node mouseover
- [x] **frontend**: remove/do not persist properties with all empty values (may be tackled in followup since backend validation will catch)
- [x] **frontend**: fix env var parsing and refresh for new format
- [x] ~**frontend**: frontend validation (may require schema updates on backend)~ backend validation will handle most cases; frontend validation can be added in a follow-up if required
- [x] ~**frontend**/**backend**: investigate improving display of Component Source property (#2256) (may be better tackled in a follow-up PR)~ non-trivial, will have to tackle in a follow-up
- [x] **frontend**: migration
- [x] **documentation** Update [node properties reference topic in pipelines topic](https://elyra.readthedocs.io/en/latest/user_guide/pipelines.html#node-properties-reference) (for many properties we currently discuss the string input format)
- [x] pipeline editor release + merge

- [x] Fix backend tests
- [x] Fix integration tests
- [x] Reviewed output of `make docs` (section "Node properties reference" in the pipelines documentation topic)

Manual test checklist. If checked, end-to-end test was successful:
 - node properties are rendered correctly (this includes validation for missing and invalid property input)
 - node tooltip on canvas renders properties correctly
 - properties are persisted as expected  in `.pipeline` file
 - properties are persisted as expected in exported DAG
 - properties are properly applied to pod 

pipeline properties:
  - [x] data volumes
  - [x] annotations
  - [x] environment variables
  - [x] tolerations
  - [x] secrets
  - [x] disable node caching
 
node properties (KFP)
  - [x] data volumes - generic components
  - [x] data volumes - custom components
  - [x] annotations - generic components
  - [x] annotations - custom components
  - [x] environment variables - generic components
  - [x] tolerations - generic components
  - [x] tolerations - custom components
  - [x] secrets - generic components
  - [x] disable node caching - custom components
  
node properties (Airflow)
  - [x] data volumes - generic components
  - [x] data volumes - custom components
  - [x] annotations - generic components
  - [x] annotations - custom components
  - [x] environment variables - generic components
  - [x] tolerations - generic components
  - [x] tolerations - custom components
  - [x] secrets - generic components

 
           in the file; or

           it.


ptitzler:
image


If one chooses (or leaves) the default selection, this is persisted in the `.pipeline` file:
```
"disallow_cached_output": "Use runtime environment default",
```

Since `Use runtime environment default` is a UI label that could change over time, it'd be safer to persist `null` instead to indicate that the user did not choose an explicit property value.

ptitzler:
image

Console log:
```
3935.4159b022aa6d82e44127.js?v=4159b022aa6d82e44127:2 TypeError: Cannot set properties of undefined (setting 'ui:placeholder')
    at index.js:1:37929
    at Proxy.forEach (<anonymous>)
    at p (index.js:1:37718)
    at index.js:1:38167
    at Immer.produce (immer.cjs.development.js:792:1)
    at t.propagatePipelineDefaultProperties (index.js:1:37339)
    at t.getAllPaletteNodes (index.js:1:37213)
    at index.js:1:67826
    at Nu (3935.4159b022aa6d82e44127.js?v=4159b022aa6d82e44127:2:61501)
    at uo (3935.4159b022aa6d82e44127.js?v=4159b022aa6d82e44127:2:69686)
```

elyra/pipeline/component_parameter.py
```diff
@@ -0,0 +1,818 @@
...
+
+    property_id = MOUNTED_VOLUMES
+    generic = True
+    custom = True
+    _display_name = "Data Volumes"
+    _json_data_type = "array"
+    _keys = ["path"]
+    _ui_details_map = {
+        "path": {"display_name": "Mount Path", "placeholder": "/mount/path"},
+        "pvc_name": {"display_name": "Volume Claim Name", "placeholder": "pvc-name"},
```
ptitzler, line 400:
```suggestion
        "pvc_name": {"display_name": "Persistent Volume Claim Name", "placeholder": "pvc-name"},
```

ptitzler:
Generic nodes: the `runtime image` property is now rendered as the last property in the list, even though it is the most important one:

Pipeline properties:
image


Node properties:
image

ptitzler:
Validation doesn't detect missing annotation value

image
As a result the invalid key/value combination is persisted in the `.pipeline` file and exported, producing an invalid annotation.

ptitzler:
Environment variables widget:

- If no value is specified, a cryptic validation error is raised:

  image

  ```
  Property has an improperly formatted env variable key value pair.
  ```

  In previous releases an empty string was considered valid. I believe we have to retain that behavior because otherwise the environment variable detection code ("refresh" button) would produce invalid list entries if the analyzed source code does not set a default value.

 - "Refresh" button doesn't work - nothing is displayed in the env variable list
 - spaces in environment variable names should not be allowed. while it might be possible to process such variables programmatically, it's generally a major pain in shell environments/scripts

ptitzler:
`data volumes` widget:

- Validation error metadata is inconsistent across widgets. One would expect the same validation error to be raised if a constraint violation is detected. For example, assume widget X requires input of a key/value pair and widget Y also requires input of a key/value pair. If the user doesn't provide a value, validation should return messages that are consistent. For data volumes the validation error metadata is 
   
image

Note the use of `"<missing required value>"`

elyra/pipeline/processor.py
```diff
@@ -352,15 +366,56 @@ def _sort_operation_dependencies(operations_by_id: dict, ordered_operations: lis
                     )
             ordered_operations.append(operation)
 
+    def add_runtime_image(self, instance, execution_object: Any, **kwargs) -> None:
```
kevin-bates, line 369: Should these be on `RuntimePipelineProcessor` instead since they don't apply to local execution?

Could we use a forward reference to annotate the type on `instance`?  Would a name like `runtime_processor` be more clear?
```suggestion
    def add_runtime_image(self, runtime_processor: "RuntimePipelineProcessor", execution_object: Any, **kwargs) -> None:
```

kiersten-stokes:
@ptitzler thanks again for the review! I believe I've addressed everything that I could from the above comments. Details below


> Since `Use runtime environment default` is a UI label that could change over time, it'd be safer to persist null instead to indicate that the user did not choose an explicit property value.

I've changed this such that only `True` and `False` are listed. The empty string is persisted, which is handled during property conversion in the `DisallowCachedOutput` class (though full disclosure, I haven't been able to thoroughly test this due to the pipeline editor white-outs).

> Generic nodes: the runtime image property is now rendered as the last property in the list, even though it is the most important one

I've removed the `RuntimeImage` `ElyraProperty` class and added the property manually in the generic and pipeline properties templates (as it was before), restoring it to it's rightful place in the property panel 😄 

> Validation doesn't detect missing annotation value

Addressed and added tiny test case to the existing test for invalid annotations.

Screen Shot 2022-09-20 at 3 06 31 PM

> [environment variables] Need to add placeholder values (ENV_VAR, value)that are consistent with other widgets

I haven't quite been able to figure this one out. The placeholders exist in the properties JSON the same way as for the others but aren't showing up. @marthacryan (or @ajbozarth) I think I might need your help on this one. [Here's](https://github.com/elyra-ai/elyra/pull/2927#issuecomment-1252038890) the originating comment.

> if no key/value pair is enters a fatal error occurs during validation

Fixed for all properties!

> [environment variables] If no value is specified, a cryptic validation error is raised:

Now when no value is specified, instances are silently removed from the list (same behavior as current `main`).

> "Refresh" button doesn't work - nothing is displayed in the env variable list

Yeah this has been on my radar, and I believe after looking into it that it is a frontend task - tagging @ajbozarth since I think he implemented the functionality. [Here's](https://github.com/elyra-ai/elyra/pull/2927#issuecomment-1252038890) the originating comment.

> spaces in environment variable names should not be allowed. 

Fixed with below message and added test case to existing test

Screen Shot 2022-09-20 at 3 16 05 PM

> no validation error is raised if only a pvc name is specified, but not a mount path

Fixed with below message and added test case to existing test

Screen Shot 2022-09-20 at 3 23 27 PM

> validation - replace <missing value> with what the actual entered value is

Done! As seen in the above screenshots

> [kubernetes tolerations] key is not a required property. the previously implemented validation code should already have the complete logic built that verifies the constraints that need to be enforced

Empty key, value, and effect are now all allowed in the correct circumstances

> [kubernetes tolerations] the initial selected value for operator should be Equal to properly model the Kubernetes behavior
> [kubernetes tolerations] operator property is not properly recognized (note Exists -> missing required value)

I've updated this on the backend such that "Equal" is set as the default in the schema. I believe the second part, however, will require a frontend change. Or at least feedback as to what I might need to change in the schema. Basically, the value is only persisted once it's changed for the first time. @marthacryan let me know if there's something I could change to fix this! Originating comment [here](https://github.com/elyra-ai/elyra/pull/2927#issuecomment-1252358337)

> effect property should be an select list enum ["", "NoExecute", "NoSchedule", "PreferNoSchedule"]

Done

Screen Shot 2022-09-20 at 3 31 05 PM

> the property order is different (pipeline defaults (lhs) vs node property (rhs))

Fixed! As seen in screenshot immediately above

elyra/pipeline/processor.py
kiersten-stokes: In this case the instance refers to the `ElyraProperty` (since each individual property calls `runtime_processor.add_{property_name}(self, execution_object`). But I've gone ahead and type-hinted those anyway since it's probably for the best 😄

ptitzler:
We should tweak the CSS. The current label padding for select lists (see highlighted areas) on top is much smaller than on the bottom. As a result the label is rendered closer to the previous element than the one that the label applies to. image

ptitzler:
Manual test checklist. If checked, end-to-end test was successful:
 - proper rendering
 - proper labels
 - persistence in `.pipeline` file
 - persistence in exported DAG
 - pod properties verified

pipeline properties:
  - [ ] data volumes
  - [ ] annotations
  - [ ] environment variables
  - [ ] tolerations
  - [ ] secrets
 
node properties (KFP)
  - [ ] data volumes
  - [ ] annotations
  - [ ] environment variables
  - [ ] tolerations
  - [ ] secrets
  
node properties (Airflow)
  - [ ] data volumes
  - [ ] annotations
  - [ ] environment variables
  - [ ] tolerations
  - [ ] secrets

ptitzler:
Tolerations widget (node properties): missing "" in enum. the widget should render the same no matter where it is displayed (node properties vs pipeline defaults)
image

ptitzler:
>> Since Use runtime environment default is a UI label that could change over time, it'd be safer to persist null instead to indicate that the user did not choose an explicit property value.

> I've changed this such that only True and False are listed. The empty string is persisted, which is handled during property conversion in the DisallowCachedOutput class (though full disclosure, I haven't been able to thoroughly test this due to the pipeline editor white-outs).

Observed behavior is:
 - label not shown in UI (but should be shown for clarity)
 - empty string is not persisted (which is what should make the most sense, or null)
 - property value is stored as string (`"True"`, `"False"`) but should be JSON boolean (`true`, `false`)

ptitzler:
Volume mount widget:
 - if no value is provided for the mount path the validation message metadata lists `/` instead of ""
   ```
   "data": {
     "nodeID": "cf9a1a8e-6d75-4c28-a9fb-41ac0ad9d6e5",
     "nodeName": "Untitled.ipynb",
     "propertyName": "mounted_volumes",
     "value": {
       "path": "/",
       "pvc_name": ""
     }
   }
   ```

ptitzler:
Environment variables widget :

Environment variables that are detected in the notebook/script

image

are not properly populated

image

ptitzler:
Added bug fix for https://github.com/elyra-ai/elyra/pull/2927#issuecomment-1253461505 in https://github.com/elyra-ai/elyra/pull/2927/commits/d96cb11255c4f2f33ced53c39b73d0b0dcd00d0b and addressed two other issues not listed above

ptitzler:
Added https://github.com/elyra-ai/elyra/pull/2927/commits/1d2875ef0f8f9691b8391e01201de977487be7b4, which updates section "Node properties reference" in the pipeline documenation.

ptitzler:
https://github.com/elyra-ai/elyra/pull/2927 renames all occurrences and variations of "disallow cached output" to "disable node caching" in the source code (variables, methods, etc)

elyra/pipeline/local/processor_local.py
```diff
@@ -114,6 +115,18 @@ def process(self, pipeline):
...
+    def supported_properties(self) -> List[str]:
+        """A list of Elyra-owned properties supported by this runtime processor."""
+        return [
+            pipeline_constants.ENV_VARIABLES,
+            pipeline_constants.KUBERNETES_SECRETS,
+            pipeline_constants.MOUNTED_VOLUMES,
+            pipeline_constants.KUBERNETES_POD_ANNOTATIONS,
+            pipeline_constants.KUBERNETES_TOLERATIONS,
+            pipeline_constants.DISABLE_NODE_CACHING,
+        ]
```
kevin-bates, line 128: I think only env variables are support by the local processor.  I'm guessing this is another example of not promoting local _mode_ to be its own _processor type_, and what is really meant by this list are the properties that can be associated with Notebook, Python, and R nodes.  The `LocalPipelineProcessor` (today) only knows how to deal with envs.

kiersten-stokes: Yes, that's exactly it. I had this in the PipelineProcessor before, perhaps I should move it back there for clarity?

kevin-bates: I see, thanks.  I think it makes sense to leave here, but only contain `env_vars`, since others don't apply to `LocalPipelineProcessor`.

I suppose you could move to `PipelineProcessor` but only define env_vars there (since, I'm guessing, every runtime will support env variables).  Then you wouldn't need the attribute check.  If we found that some new processor didn't support envs, then we could have the base class return the empty array and move the envs_vars to `LocalPipelineProcessor` at that time - which still removes the need for the attribute check.

kiersten-stokes: Well the issue is that the call to grab the "local" properties has to display all the properties available for generic nodes in any runtime type since they can be run anywhere. That's actually why, to determine the properties to render for each runtime, I'm using this "supported_properties" method instead of the class attribute check for an "add_{property}" method. Since "local" needs to pull in all these properties for which an "add" function will never be relevant, this list seemed a slightly better alternative. And, as we chatted about, we can't use the parent class to do the "add_{property}" attribute check since then there will never be a difference in properties available between runtimes.

ajbozarth:
I just pushed the update that add migration. If you created a pipeline on this branch previously the migration will error, but you can just update the version in the json to 8 and it should be valid

elyra/pipeline/local/processor_local.py
kiersten-stokes: I think I could probably do a thing where, when "local" properties are being queried, take the set of properties supported by all subclasses of 'PipelineProcessor' (or 'RuntimePipelineProcessor'). For reasons too long to explain here, this wouldn't require an explicit reference to "local" in the code

kevin-bates: Thank you for the explanations.  This isn't a big enough deal to spend time on at the moment.  It's helpful to know that this is a side-effect of _not_ making "local mode" its own runtime and something we can address should we decide to promote it to its own runtime down the road.  Thanks for spending time on this.  I know things like this can be extremely frustrating.

elyra/pipeline/component_parameter.py
```diff
@@ -0,0 +1,804 @@
...
+        """
+        validation_errors = []
+
+        # Ensure the operator is valid
+        if self.operator not in ["Exists", "Equal"]:
+            validation_errors.append(
+                f"'{self.operator}' is not a valid operator: the value must be one of 'Exists' or 'Equal'."
+            )
+
+        if self.operator == "Equal" and (self.key is None or len(self.key) == 0):
```
kevin-bates, line 493: This can be cleaned up a bit using a Falsey...
```suggestion
        if self.operator == "Equal" and not self.key:
```

elyra/pipeline/component_parameter.py
```diff
@@ -0,0 +1,804 @@
...
+            self.effect is not None
+            and len(self.effect) > 0
+            and self.effect not in ["NoExecute", "NoSchedule", "PreferNoSchedule"]
+        ):
+            validation_errors.append(
+                f"'{self.effect}' is not a valid effect: effect must be one "
+                f"of 'NoExecute', 'NoSchedule', or 'PreferNoSchedule'."
+            )
+
+        if self.operator == "Exists" and self.value is not None and len(self.value) > 0:
```
kevin-bates, line 508: Ditto here...
```suggestion
        if self.operator == "Exists" and self.value:
```

ajbozarth:
just pushed updates to address @ptitzler issues discussed in scrum today

ptitzler:
> just pushed updates to address @ptitzler issues discussed in scrum today

Thank you! I confirmed 

-  the width is now consistent
  
  image


- label padding for select lists is now more balanced

image

ptitzler:
Currently properties for custom nodes are visually grouped into categories:
 - Inputs
 - Outputs
 - Additional properties
 - Component source
 
We should do the same for generic nodes (either in this PR or a follow-up). Doing so would provide for a more uniform user experience and prepare for the migration of the current one-off implementation to one that utilize a genuine custom component to run noteboooks and scripts.

ptitzler (approved):
LGTM! Very nice work. Gone are the days of esoteric input!

ajbozarth:
just pushed to update to point at the RC release of pipeline editor now that https://github.com/elyra-ai/pipeline-editor/pull/201 is merged and released

PR #2957, opened 2022-10-06
Prepare to make all Elyra-owned properties object-valued

kiersten-stokes:
Supports #2942. This PR to be merged before merge of the shared memory PR.

This PR introduces some implementation details that simplify the addition of new properties by making all properties object-valued. The bulk of the change is the renaming of the `ui_details_map` to `attribute_details`, and also includes a major change to the data structure. Whereas before this structure was a dictionary, it is now a list of `PropertyAttribute` objects, each of which contains the information needed to render and process a single property attribute (e.g. "env_var" and "value" are the `PropertyAttribute`s for the `EnvironmentVariable` class). Additionally, the creation of the properties schema now assumes that all properties are either `object`s or `array`s (with the exception of `DisableNodeCaching`, explained below). All the above allows us to generalize, and removes the need for a `create_instance` method on each class.

Note that the `DisableNodeCaching` property is still single-valued for now. The reason being that fully converting this property to an object would require a pipeline migration, and we would like to avoid a migration so soon after the 3.12 migration, if possible.

No changes to the frontend are included or necessary, unless we decide that we want to migrate `DisableNodeCaching`.

Finally, a distinction is made between Elyra-owned properties (those with a corresponding `ElyraProperty` class) and properties that should be propagated. There is currently only one case where the set of propagated properties != the set of Elyra-owned properties, which is the propagation of `runtime_image` for generic components. This is handled in a new property on the `Node` object called `propagated_properties`, which is accessed in the node propagation loop.

Manually tested several scenarios. Changes are self-contained and only one minor test update was required.

 
           in the file; or

           it.


elyra/pipeline/component_parameter.py
```diff
@@ -169,129 +262,132 @@ def add_to_execution_object(self, runtime_processor: RuntimePipelineProcessor, e
...
-        """Returns a boolean indicating whether this instance is considered a no-op."""
-        return False
-
 
 class DisableNodeCaching(ElyraProperty):
-    """Disable caching to force node re-execution in the target runtime environment."""
+    """An ElyraProperty representing node cache preference"""
+
+    applies_to_generic = False
+    applies_to_custom = True
```
kevin-bates, line 270: Where are these used?  I only see them being set.

elyra/pipeline/component_parameter.py
```diff
@@ -169,129 +262,132 @@ def add_to_execution_object(self, runtime_processor: RuntimePipelineProcessor, e
...
+    def get_all_validation_errors(self) -> List[str]:
+        return []
+
     def add_to_execution_object(self, runtime_processor: RuntimePipelineProcessor, execution_object: Any, **kwargs):
-        """Add DisableNodeCaching info to the execution object for the given runtime processor"""
         runtime_processor.add_disable_node_caching(instance=self, execution_object=execution_object, **kwargs)
 
 
-class ElyraPropertyListItem(ElyraProperty):
+class ElyraPropertyListItem(ElyraProperty, ABC):
```
kevin-bates, line 302: nit: Because this class no longer has abstract methods, its inheritance from `ABC` is unnecessary.  Unfortunately, you can't prevent a subclass of `ABC` from being instantiated unless it also defines at least one abstract method in Python.

elyra/pipeline/pipeline_definition.py
```diff
@@ -625,9 +616,14 @@ def propagate_pipeline_default_properties(self):
                     merged_list = ElyraPropertyList.merge(node_value, pipeline_value)
                     node.set_component_parameter(property_name, merged_list)
 
+            # Runtime image is a special case that must be handled separately
```
kevin-bates, line 619: Why is this a special case?  Can we extend this comment to clarify for future maintainers?

I noticed above that it doesn't map to an `ElyraProperty` class hierarchy and wondering if that would simplify its handling in these kinds of places.

setup.py
```diff
@@ -77,7 +77,7 @@
         "jupyter_server>=1.7.0",
         "jupyterlab>=3.4.6",  # comment out to use local jupyterlab
         "jupyterlab-lsp>=3.8.0",  # comment out to use local jupyterlab
-        "jupyterlab-git~=0.32",  # Avoid breaking 1.x changes
+        "jupyterlab-git~=0.32",  # Avoid baking 1.x changes
```
kevin-bates, line 80: This seems like a typo.

kevin-bates (commented):
This refactoring is great!  Just had a few minor comments and nits.

setup.py
kiersten-stokes: 😄  That's very weird... I definitely didn't consciously change this but I'm not sure why my IDE would have done so for me either

elyra/pipeline/component_parameter.py
kiersten-stokes: These are used in `get_classes_for_component_type` in the same file (in line 188 right now). If a property has `applies_to_{component_type} == True`, then it is added to the list of properties to render for that component. I had these attributes just named `generic` and `custom` up until a few commits ago, but thought that that name maybe wasn't descriptive enough.

setup.py
ptitzler: 🍪

kevin-bates: > I definitely didn't consciously change this but I'm not sure why my IDE would have done so for me either

I've found that pycharm will "touch" unexpected files during refactoring and depending on the selected options - which I find infuriating.  Perhaps you did a "refactor->rename" that dealt with "re"?  No worries though - just mentioning.

elyra/pipeline/component_parameter.py
kevin-bates: Ah - thanks.  I had searched for the complete variable name and didn't see its templated use.  Thank you.

elyra/pipeline/pipeline_definition.py
kiersten-stokes: Yeah runtime image is sort of a tricky case. Like you say, it's the only pipeline default property (right now) that doesn't have a corresponding `ElyraProperty`. I played around with making it an `ElyraProperty` subclass, but there's so much special processing of runtime image in the processors and validation that it became pretty messy. 

I think what I might do instead of this special handling block is to define a new property on `Node` that corresponds to the set of properties that _should_ be propagated. Then the docstring for that property can explain why in some cases `elyra_owned_properties != <the set of properties to propagate>`.

setup.py
kiersten-stokes: I bet it was something like that!

elyra/pipeline/component_parameter.py
```diff
@@ -44,21 +49,85 @@
...
+    ):
+        """
+        :param use_in_key: whether this attribute should be used when constructing a
+            key for an instance that will be used to de-duplicate list items
+        """
+        super().__init__(attribute_id, display_name, placeholder, default_value, input_type, enum, hidden, required)
+        self.use_in_key = use_in_key
+
+
+class ElyraProperty(ABC):
```
ptitzler, line 117: Should this be also extending `EnforceOverrides` to raise errors as documented at https://pypi.org/project/overrides/
> Use `EnforceOverrides` to require subclass methods that shadow superclass methods to be decorated with @overrides.

```suggestion
class ElyraProperty(ABC, EnforceOverrides):
```

kevin-bates (approved):
Good stuff - thank you!

elyra/pipeline/component_parameter.py
kiersten-stokes: per offline discussion, the commit to add `overrides` will be reverted and a new issue opened to investigate the addition later on

Issue #2964, opened 2022-10-10
Runtime configuration named 'local' is treated as local runtime


ptitzler:
**Describe the issue**

**To Reproduce**
1. Create a KFP RTC and name it local
   ```
   $ elyra-metadata list runtimes
   Available metadata instances for runtimes (includes invalid):

   Schema    Instance                         Resource                                                                             
   ------    --------                         --------                                                                             
   kfp       local                            /Users/patti/Library/Jupyter/metadata/runtimes/local.json   
   ```
3. Create a valid KFP pipeline that runs a notebook
4. Run the pipeline using the 'local' RTC
5. Observe the output. It indicates that the pipeline was executed locally. this can be confirmed by inspecting the notebook's cell outputs.

![2022-10-10_16-18-09 (1)](https://user-images.githubusercontent.com/13068832/194966724-68faa333-3a04-49c9-8cee-2989c7054b73.gif)

**Expected behavior**

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.12

kevin-bates:
This is occurring in the server where it tries to determine the runtime.  In this case, `runtime_config` has a value of "local". 

https://github.com/elyra-ai/elyra/blob/797a17b3d41fed15499c1e158ba4e7bfb639e852/elyra/pipeline/validation.py#L198-L199

However, the front end will also submit generic pipelines for local processing by setting `runtime_config` to the string "local" - as does the CLI utility.

https://github.com/elyra-ai/elyra/blob/71a1ffdc2e8ea93c04ab1ee038489a2d6782ab44/elyra/cli/pipeline_app.py#L504

As a result, I think we should explore letting the `None` value for `runtime_config` indicate that local processing takes place (for now).   Ideally, it seems like we should determine the runtime from `runtime_config`'s schema name - but that can only happen once the Local runtime processor supports runtime configuration metadata.

PR #2968, opened 2022-10-12
Tolerate runtime configuration instances named 'local'

kevin-bates:
When submitting pipelines to Kubeflow or Airflow runtimes, Elyra runs the pipeline locally if the corresponding runtime configuration is named "local" as described in #2964.  This is because the server looks at the runtime_config field in the pipeline for either no value (`None`) or the name `"local"`.  This pull request implements changes such that only invocations for local processing will submit pipelines with no value for `runtime_config`.

Should we decide to _promote_ the LOCAL runtime to a full-fledged runtime (where there's an associated schema and support for runtime configuration instances), most of these changes should still work since there _will_ be a value for `runtime_config` in these cases, which will then determine the target pipeline processor from the referenced schema name within the runtime configuration instance.  We should only then need to make changes where we set the `runtime_config` to `None` during submission (and in both the UI and CLI applications).

Still to do:
- [x] Set `runtime_config` to `null` (or `None`) in the UI when submitting pipelines for local execution

Resolves: #2964

           in the file; or

           it.

### DIFF
--- a/elyra/cli/pipeline_app.py
+++ b/elyra/cli/pipeline_app.py
@@ -31,7 +31,7 @@ from kfp import Client as ArgoClient
 
 from elyra._version import __version__
 from elyra.metadata.manager import MetadataManager
-from elyra.metadata.schema import SchemaManager
+from elyra.metadata.metadata import Metadata
 from elyra.metadata.schemaspaces import Runtimes
 from elyra.pipeline import pipeline_constants
 from elyra.pipeline.component_catalog import ComponentCache
@@ -46,7 +46,6 @@ from elyra.pipeline.processor import PipelineProcessorManager
 from elyra.pipeline.processor import PipelineProcessorResponse
 from elyra.pipeline.runtime_type import RuntimeProcessorType
 from elyra.pipeline.runtime_type import RuntimeTypeResources
-from elyra.pipeline.runtimes_metadata import RuntimesMetadata
 from elyra.pipeline.validation import PipelineValidationManager
 from elyra.pipeline.validation import ValidationSeverity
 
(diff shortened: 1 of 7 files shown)


lresende:
@kevin-bates could you add to the description how the info will be submitted (or expected) when local and non-local? (e.g. runtime-config and runtime-schema)

kevin-bates:
> @kevin-bates could you add to the description how the info will be submitted (or expected) when local and non-local? (e.g. runtime-config and runtime-schema)

From the description...
> This pull request implements changes such that only invocations for local processing will submit pipelines with no value for runtime_config.

The schema for the runtime (and thus which processor to use) comes from the runtime configuration's instance referenced by `runtime_config`.

karlaspuldaro:
@kevin-bates I believe this is where you would set the `runtime_config` value 
https://github.com/elyra-ai/elyra/blob/main/packages/pipeline-editor/src/PipelineEditorWidget.tsx#L762

kevin-bates:
Thanks @karlaspuldaro - that's the very location I had identified as well, so it's great reassurance!

karlaspuldaro (approved):
Code looks good to me :)

ptitzler (approved):
LGTM based on spot testing

lresende:
All the confusion described in #2964 seems to be related to the UI only propagating the runtime_config which is the key to finding the runtime configuration profile, and when an existing runtime tries to use it as local it becomes problematic. 

How about, instead of using workarounds, we don't actually also propagate the runtime type, which is how the ui finds all the runtime configuration available for each runtime type thus we just need to make sure the backend uses both runtime type and runtime config to find the proper processor to execute the pipeline?

NOT A CONTRIBUTION

kevin-bates:
Thanks for your review @lresende.

Adding `runtime_type` to the API will certainly work, but I disagree that it's any better.  Aside from an API change, this will only add to the confusion since there will be _two_ sources for determining the type of runtime.  What's the point of having `runtime_type` as part of the runtime schema in the first place?

The "workaround" is due to the fact that we use a "magic config name" named 'local' to indicate that the pipeline is run locally.  The source code was already using a "None" or "local" approach when checking `runtime_config` and this pull request simplifies that further to only allow `None` for that specification.  This nicely positions us to introduce true runtime configuration for local processing - which is the previous decision (to not do that) we're constantly hacking our way around.  If local runtimes are ever introduced, this PR will require zero changes since, then, the runtime name will be passed and the code will simply bypass the checks against `None` and use whatever runtime is specified in the `runtime_config` schema - **just like all non-local runtimes work today**.  In addition, non-local runtimes can choose any name they want, including "local".

PR #2971, opened 2022-10-17
Add integration tests for Script debugger

### LINKED ISSUE #2897: Add basic debugger tests
Add basic unit tests + integration tests to the script debugger feature.

karlaspuldaro:
Closes #2897

Integration test cases added in `scriptdebugger.ts` :
- test for debugger button to be enabled for default Python kernel
- test for debugger button state persistence on page reload
- test for debugger button state persistence on reopening editor tab
- test for debugger button to be disabled for kernel without debug support
 
 ~~All flaky on CI environment~~ 
**Follow up note**: new test file name has number on purpose (for now) so it is the first to run on CI as a workaround for  failures, as suggested in [this comment](https://github.com/elyra-ai/elyra/pull/2971#issuecomment-1295245416) . 
Follow up issue #2995 to track investigating other test(s) file(s) that may be affecting this new one.

           in the file; or

           it.


karlaspuldaro:
@kevin-bates you were right, looks like the execution order matters!
There are still 2 tests commented out for being more inconsistent, I'll introduce them back and check how things go.

kevin-bates:
Thanks for checking that @karlaspuldaro.  This implies that _another_ test (or more) is side-affecting the button's location (or something like that if I'm reading the test failures correctly).

When I've dealt with this kind of thing before, it usually means a "binary search" exercise of placing the affected test in different locations until you find the test which your test immediately fails after, but works when moved one test prior.   Then, the side-effecting test can be more closely examined if the solution hasn't been discovered in the course of the search exercise.  It can be a long road, but also interesting once the solution is found.  Good luck. 🍀

akchinSTC (approved):

kevin-bates:
@karlaspuldaro - since it appears we still don't know _why_ the test location matters, could you please open an issue to revisit this as this is the kind of thing that can bite us down the road?  I suspect it's one of the two "editor" tests that are causing the issue.  This could be determined by moving this test immediately before those and ensuring it still passes.  If there are failures, then this particular hypothesis is incorrect.

Issue #2936, opened 2022-09-23
Feature (parameter) parity "run as pipeline" vs "node properties"


romeokienzler:
**Is your feature request related to a problem? Please describe.**
The "run as pipeline" dialog doesn't have the same parameter set as the properties page on a node in the pipeline editor. For example I can't set env. variables in the dialog

**Describe the solution you'd like**
Feature (parameter) parity "run as pipeline" vs "node properties"

**Describe alternatives you've considered**
Hard - code parameters before submitting...

**Additional context**
![image](https://user-images.githubusercontent.com/5694071/192051923-05f6c3a7-5394-434d-ac6e-c4628918999f.png)

ozen:
Also, I get "must be greater than zero" error when I enter 0 for GPU in "run as pipeline" dialog. Not an issue in "node properties".

PR #2780, opened 2022-06-09
Add initial support for rjsf in pipeline properties

marthacryan:
Corresponding to elyra-ai/pipeline-editor#197. 

Fixes #2759 

Todo:
- [x] basic pipeline properties support
- [x] basic generic node properties support
- [x] basic "canvas properties" support
- [x] finish airflow support
- [x] link in custom components
- [x] link in validation
- [x] add in required fields
- [x] add inputpath as oneof
- [x] frontend: pipeline migration
  - [x] backend: double check that relevant test pipelines are migrated properly (once migration is finalized)
- [ ] go through code to make sure all utilities and edits etc. are all corresponding to the updated schema

Testing Follow Up TODOs:
- [x] section headers: description 
- [x] backend validation requires updates (pending some frontend changes)
- [x] pipeline JSON occasionally gets corrupted (e.g. one node's properties/values find their way into another node's `component_parameters` stanza) when connected 2 nodes (no longer seeing; uncheck if not the case)
- [x] frontend: properties don't always save if you navigate away or sometimes disappear after connecting 2 nodes - possibly related to above task
- [x] frontend: schema shows up in pipeline JSON
- [x] frontend: headers appear in pipeline properties `app_data.properties.pipeline_defaults` stanza
- [x] frontend: headers appear in node properties `component_parameters` stanzas
  - [x] backend: code can be updated (`validation.py:493-500`) after this change is made
- [x] frontend: properties not listed as required in the schema show as required in properties panel
- [x] frontend: change look of some items (see [comment](https://github.com/elyra-ai/elyra/pull/2780#issuecomment-1226803610))
- [x] frontend: pipeline editor canvas goes blank occasionally when editing array values (haven't been able to reproduce recently)
- [x] `inputpath` structure in pipeline JSON needs fixed
- [x] drag and drop of notebooks/scripts
- [x] pipeline properties - placeholders for list-valued properties
- ~[ ] user-friendly key-value inputs (env variables etc all)~ - pushing to next release
- [x] pipeline default property propagation (including display in node properties)
- [x] pipeline defaults: section headers (node defaults, generic node defaults)
- [x] custom components: component description missing
- [x] generic components: [node] label description missing
- [x] certain string property validation (frontend) erroneously require objects after connecting to another node
- [x] properties with empty strings as defaults erroneously pass frontend validation when first dragging component into pipeline
- [x] `component_source` parameter should be moved out of `component_parameters` stanza of pipeline JSON (some frontend work required in order to still display the property in properties panel)
[...]

PR #2972, opened 2022-10-17
Fix handling of `cos_object_prefix` pipeline property

kiersten-stokes:
Prior to this PR, pipelines with a `cos_object_prefix` defined will fail to run on `main` and the 3.12 release.

This PR does a handful of related things. First, the `pipeline_parameters` property on the `pipeline_definition.py` `Pipeline` object is removed - it is no longer necessary now that `cos_object_prefix` is stored in the `pipeline_defaults` stanza of the pipeline JSON. (Prior to the changes in #2780, `cos_object_prefix` existed as a sibling to the `pipeline_defaults` stanza). The `PIPELINE_META_PROPERTIES` constant is also no longer needed and is removed. A handful of helper methods are added to `component_parameter.py` and, as a result, the `convert_elyra_owned_properties` methods are simplified.

The key-value pairs in the `pipeline_defaults` stanza are passed to the `pipeline.py` `Pipeline` object and accessed by the runtime processors in the same way as before, but now this property is named `pipeline_properties` rather than `pipeline_parameters` (in order to prepare for an implementation of #2936).

Of note here is that no pipeline migration is necessary right now. In order to avoid a migration so soon in 3.13 (3.12 just required migration), a workaround is added to the backend that supports `cos_object_prefix` both as a child of the `pipeline_defaults` and as a sibling. See `pipeline_definition:191:197`.

- [x] Airflow - a case was added to an existing test to cover `cos_object_prefix` functionality; existing case that uses the same sample pipeline was also updated accordingly
- [x] KFP -added a new test (`test_create_yaml_complex_pipeline`) to cover `cos_object_prefix` and other functionality; mimics the `create_test_file` test in `test_airflow_processor.py`
- [x] Manual test with `cos_object_prefix` defined in the `pipeline_defaults` stanza of the pipeline json
- [x] Manual test with `cos_object_prefix` defined in the `app_data` stanza of the pipeline json

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/airflow/processor_airflow.py
+++ b/elyra/pipeline/airflow/processor_airflow.py
@@ -187,7 +187,7 @@ be fully qualified (i.e., prefixed with their package names).
             if pipeline.contains_generic_operations():
                 object_storage_url = f"{cos_endpoint}"
                 os_path = join_paths(
-                    pipeline.pipeline_parameters.get(pipeline_constants.COS_OBJECT_PREFIX), pipeline_instance_id
+                    pipeline.pipeline_properties.get(pipeline_constants.COS_OBJECT_PREFIX), pipeline_instance_id
                 )
                 object_storage_path = f"/{cos_bucket}/{os_path}"
             else:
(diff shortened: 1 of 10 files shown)


ptitzler (approved):
LGTM!

akchinSTC (approved):
FYI, I think using `|` to represent the typehint Unions only work for python >=3.10. 

anywho not a biggie, LGTM

kevin-bates:
> FYI, I think using | to represent the typehint Unions only work for python >=3.10.

Older python versions can use that syntax with `from __future__ import annotations`.  That said, I've found that the `overrides` package (which we're exploring) has heartburn equating, for example, signatures of `type | None` to `Optional[type]` (or `Union[type, None]`) and when either signature is not in the local package, you're forced to choose the approach of the foreign signature (speaking from experience).  So I guess my point is that we should be consistent. :smile:

> anywho not a biggie

Agreed.

PR #2977, opened 2022-10-24
Fix debugger button state on kernel selection change

### LINKED ISSUE #2916: Debugger should be disabled when selecting kernel not supporting debugging
**Describe the issue**
The script debugger reads kernel selections and either enables/disables the debug button accordingly.
With the spec example below, when the user selects that kernel, the debug button should be disabled, although that currently doesn't update the UI :
```
 "display_name": "Python 3 - fakecopy",
 "language": "python",
 "metadata": {
  "debugger": false
 }
```
It only disables the button properly on first load - eg. if there is a kernel installed for R or Scala those editors show the debug button disabled.

This happens because [the callback function to update the debug state returns when it finds the selected kernel has no debug support](https://github.com/elyra-ai/elyra/blob/main/packages/script-debugger/src/index.ts#L68), when it should still follow the regular method flow and let the debug service (in lab) take care of reading the kernel state and updating the UI - requiring a few extra listeners on the elyra side since it has its custom debug handler.

Ideally, this issue could also be fixed by implementing the [SessionContext](https://jupyterlab.readthedocs.io/en/stable/api/classes/apputils.sessioncontext-1.html) into the Script Editor, replacing the current kernel toolbar with this class and all the kernel services it includes (as used by notebook and console).

**To Reproduce**
1. Install a python kernel without debug support (it can be by copying the ipykernel spec folder and edit the `kernel.json` as above)
2. Open a python editor - ipykernel is selected by default and the debug button is enabled
3. Select the non debug kernel from the toolbar dropdown - the debug button is still enabled

**Expected behavior**
The debug button should be disabled for the use case above

**Context information**
 - Elyra version: v3.12.0-dev
 - Installation source: source
 - Operating system: macOS

karlaspuldaro:
Fixes #2916
Changing kernel selection now updates the debug button state accordingly.
debug-button

Updated  the `updateDebugger`  method logic to handle a few areas where UI update was not being triggered for the expected workflow.

Manually tested by mocking a python kernel without debug support as below:
1. jupyter kernelspec list
2. copy default python3 kernel directory and content to the same kernels/ directory
3. Change kernel.json content to:
```
 "display_name": "Python 3 - test",
 "language": "python",
 "metadata": {
  "debugger": false
 }
```
When changing kernels verify the debug button state (enabled/disabled) gets updated accordingly (sometimes it takes a few ms to update).
 
           in the file; or

           it.


akchinSTC (approved):
works as advertised. LGTM!

PR #2999, opened 2022-11-04
Allow for custom export file name

### LINKED ISSUE #1417: Pipeline export: allow for custom export file name

Currently export produces a file with the pipeline name, e.g. `my-pipeline.py` or `my-pipeline.yaml` for Kubeflow Pipelines environments. With the newly added support for Apache Airflow as a runtime environment in version 2.1, export also produces a `my-pipeline.py` file when AA is selected as target environment. (The content of `my-pipeline.py` (KFP) and `my-pipeline.py` (AA) is very different.)

To make it easier for users to determine what the purpose of a generated file is, Elyra should
- generate a default file name that identifies the export target, e.g. `my-pipeline-kfp-dsl.py`
- allow the user to choose a custom name prior to processing the export request

## Update 
Since each runtime type now only has a single export format -- `.py` for Airflow and `.yaml` for KFP) -- it's probably sufficient to allow the user to customize the export file name and have the default file name pre-populated as the pipeline file name. This behavior is similar to that which already exists for pipeline run/submission.

salonee13:
Fixes #1417 

Added functionality that allows the user to change the export file name to their own custom name if needed, if not, keeps it as the export path name

This shows the default export name: 
export_pt1

This shows that it allows you to customize the export name to whatever you want:  
export_pt2

This shows that the file exports and saves successfully with the custom export name:  
export_pt3

Tested manually, see screenshots shown above. After the integration tests run, we will add a test to make sure this functionality works. 

 
           in the file; or

           it.

### DIFF
--- a/packages/pipeline-editor/src/PipelineExportDialog.tsx
+++ b/packages/pipeline-editor/src/PipelineExportDialog.tsx
@@ -60,15 +60,6 @@ export const PipelineExportDialog: React.FC<IProps> = ({
 }) => {
   return (
     <form className="elyra-dialog-form">
-      <label htmlFor="export_name">Export Name:</label>
-      <br />
-      <input
-        type="text"
-        id="export_name"
-        name="export_name"
-        defaultValue={exportName}
-        data-form-required
-      />
       <RuntimeConfigSelect
         runtimeData={runtimeData}
         pipelineType={pipelineType}
@@ -78,6 +69,17 @@ export const PipelineExportDialog: React.FC<IProps> = ({
           return <FileTypeSelect fileTypes={info?.export_file_types ?? []} />;
         }}
       </RuntimeConfigSelect>
+      <label htmlFor="export_name">Export Filename:</label>
+      <br />
+      <input
+        type="text"
+        id="export_name"
+        name="export_name"
+        defaultValue={exportName}
+        data-form-required
+      />
+      <br />
+      <br />
       <input
         type="checkbox"
         className="elyra-Dialog-checkbox"
--- a/packages/pipeline-editor/src/PipelineExportDialog.tsx
+++ b/packages/pipeline-editor/src/PipelineExportDialog.tsx
@@ -49,15 +49,26 @@ interface IProps {
   runtimeData: IRuntimeData;
   runtimeTypeInfo: IRuntimeType[];
   pipelineType?: string;
+  exportName: string;
 }
 
 export const PipelineExportDialog: React.FC<IProps> = ({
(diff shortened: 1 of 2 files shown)


kiersten-stokes (commented):
This is looking really great!

Could we add some space beneath the export name field to bring it in like with how it looks in the submission dialog (shown below)?

Screen Shot 2022-11-07 at 2 26 42 PM

packages/pipeline-editor/src/PipelineExportDialog.tsx
```diff
@@ -49,15 +49,26 @@ interface IProps {
...
     <form className="elyra-dialog-form">
+      <label htmlFor="export_name">Export Name:</label>
+      <br />
+      <input
+        type="text"
+        id="export_name"
+        name="export_name"
+        defaultValue={exportName}
+        data-form-required
+      />
```
karlaspuldaro, line 71: Can you add one or two more break lines after this input element, so it gives some spacing between that and the next element below?

karlaspuldaro:
Thank you @salonee13 for implementing this enhancement, it looks great and works as expected :)
I only had one comment regarding the styling but everything else looks good to me!

For the integration test, you can add a new file (maybe call it `pipelineexport.ts`) to the [integration tests folder](https://github.com/elyra-ai/elyra/tree/main/tests/integration).
One test case should be sufficient, and the workflow/commands would look something like:
1. create a runtime config
2. open a pipeline test file (`helloworld.pipeline`)
3. click export button
4. edit export name to something else
5. select the expected runtime config
6. press ok
7. check export name is in the `succeeded` dialog
8. press ok o the `succeeded` dialog
9. check the new file is listed in the file browser

ptitzler:
Could you also change the order in which information is rendered and rename the label to `Export Filename`?

image

By moving the filename input to the bottom we group related information: the first inputs are specific to the runtime configuration that will be applied and the last three are specific to the exported file that will be produced:
 - choose a file format
 - choose a filename
 - choose whether to overwrite the file if it already exists

packages/pipeline-editor/src/PipelineExportDialog.tsx
salonee13: Made the changes, and added two break lines.

kiersten-stokes (approved):
I'm approving as-is because we decided offline to add the above-mentioned integration test a follow-up issue/PR

Thank you so much, Salonee! I'll be getting a lot of use out of this feature

karlaspuldaro (approved):

PR #3012, opened 2022-11-14
Fix bug with propagation of node caching property

### LINKED ISSUE #3010: Pipeline defaults are not properly propagated
**Describe the issue**
There appears to be a problem with the propagation of pipeline defaults.

**To Reproduce**
1. Create a KFP pipeline
2. Add one custom node to the pipeline (any node will do)
3. Set the `Disable node caching` _pipeline default property_ to `True`
4. Export the pipeline (any format)
5. In the generated output search for `P0D` (that's a zero) - there is no match
6. Set the _node's_ `Disable node caching` property to `True`
7. Export the pipeline (any format)
8. In the generated output search for `P0D` - there is a match

Debug output indicated that [here](https://github.com/elyra-ai/elyra/blob/main/elyra/pipeline/kfp/processor_kfp.py#L1122) the value of `selection` is `False` when `True` was specified as pipeline default property. Other properties might be impacted as well.

**Deployment information**
 - Elyra version: main/3.13

kiersten-stokes:
Fixes #3010

`DisableNodeCaching` currently requires special handling, as it is our only Elyra-owned property that is not stored and processed as an object. As a result, values weren't being stored in their instances properly and therefore weren't propagating properly. This PR adds a `should_discard` method to `DisableNodeCaching` in order to: 1.) fix the propagation issue now, and 2.) prepare for when this property is finally converted into an object-valued property in the pipeline JSON.

I added a case to an existing test because this has unfortunately come up before. We could potentially remove this case if desired once we switch the handling/storage of `DisableNodeCaching` to an object-valued properly. Instead, however, we should probably have a test case for each Elyra-owned property.

 
           in the file; or

           it.

### DIFF
--- a/elyra/pipeline/component_parameter.py
+++ b/elyra/pipeline/component_parameter.py
@@ -294,7 +294,7 @@ class DisableNodeCaching(ElyraProperty):
         self.selection = None
         if selection in ["True", "true", True]:
             self.selection = True
-        elif selection == ["False", "false", False]:
+        elif selection in ["False", "false", False]:
             self.selection = False
 
     @classmethod
--- a/elyra/pipeline/component_parameter.py
+++ b/elyra/pipeline/component_parameter.py
@@ -291,13 +291,18 @@ class DisableNodeCaching(ElyraProperty):
     _json_data_type = "string"
 
     def __init__(self, selection, **kwargs):
-        self.selection = selection == "True"
+        self.selection = None
+        if selection == "True":
+            self.selection = True
+        elif selection == "False":
+            self.selection = False
 
     @classmethod
     def get_single_instance(cls, value: Optional[Any] = None) -> ElyraProperty | None:
         if isinstance(value, ElyraProperty):
             return value  # value is already a single instance, no further action required
-        return DisableNodeCaching(selection=value)
+        instance = DisableNodeCaching(selection=value)
+        return None if instance.should_discard() else instance
 
     @classmethod
     def get_schema(cls) -> Dict[str, Any]:
@@ -317,6 +322,9 @@ class DisableNodeCaching(ElyraProperty):
         """Add DisableNodeCaching info to the execution object for the given runtime processor"""
         runtime_processor.add_disable_node_caching(instance=self, execution_object=execution_object, **kwargs)
 
+    def should_discard(self) -> bool:
+        """Ignore this DisableNodeCaching instance if neither True nor False was selected."""
+        return self.selection is None
 
 class CustomSharedMemorySize(ElyraProperty):
     """An ElyraProperty representing shared memory size for a node."""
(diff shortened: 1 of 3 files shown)


elyra/pipeline/component_parameter.py
```diff
@@ -290,14 +290,19 @@ class DisableNodeCaching(ElyraProperty):
     property_description = "Disable caching to force node re-execution in the target runtime environment."
     _json_data_type = "string"
 
-    def __init__(self, selection, **kwargs):
-        self.selection = selection == "True"
+    def __init__(self, selection: Union[str, bool], **kwargs):
+        self.selection = None
+        if selection in ["True", "true", True]:
+            self.selection = True
+        elif selection == ["False", "false", False]:
```
kevin-bates, line 297: Typo:
```suggestion
        elif selection in ["False", "false", False]:
```

elyra/pipeline/component_parameter.py
```diff
@@ -290,14 +290,19 @@ class DisableNodeCaching(ElyraProperty):
     property_description = "Disable caching to force node re-execution in the target runtime environment."
     _json_data_type = "string"
 
-    def __init__(self, selection, **kwargs):
-        self.selection = selection == "True"
+    def __init__(self, selection: Union[str, bool], **kwargs):
+        self.selection = None
+        if selection in ["True", "true", True]:
```
kevin-bates, line 295: Is there a reason this is not just a pure boolean type?  Does the front end only persist string values into the pipeline JSON for this field?

elyra/tests/pipeline/resources/sample_pipelines/pipeline_valid_with_pipeline_default.json
```diff
@@ -99,7 +100,8 @@
             ],
             "mounted_volumes": [
               { "path": "/mnt/vol2", "pvc_name": "pvc-claim-2" }
-            ]
+            ],
+            "disable_node_caching": "True"
```
kevin-bates, line 104: If the front end can persist boolean types (per the previous question), it might be good to have one value that is derived from a boolean.

kevin-bates (commented):
This looks good. My only comment is regarding the dual types that need to be handled (which is kind of a hassle) and just wondering if this is historical (and avoids a migration).

elyra/pipeline/component_parameter.py
kiersten-stokes: You're exactly right that right now it avoids migration 😄 . Currently the frontend persists strings only, but we intend to try to change this in 3.14. I only say try because we have some other potential limitations, e.g. the placeholder value is a string. Checking for boolean types is probably unnecessary right now, but I'll probably open a PR later this week that *would* require this change, so just 'preparing' for that

ptitzler (approved):
LGTM!

kevin-bates (commented):
LGTM

PR #3029, opened 2022-11-23
Support config GPU vendor

typhoonzero:
This is a following up work of https://github.com/elyra-ai/elyra/pull/2973.

Removed "vgpu memory" boxes.

截屏2022-11-23 16 35 52

The K8s cluster used for GPU training may have other GPU resource types other than `nvidia.com/gpu`, there are other vendors like AMD or opensource vGPU implementations, like https://github.com/4paradigm/k8s-vgpu-scheduler, https://github.com/tkestack/gpu-manager etc. I propose to add support for setup user inputed resource types.

Tests added at `test_pipeline_constructor.py`


ptitzler (dismissed):
If the user enters an invalid `gpu_vendor` value (e.g. only spaces '`   `' or a string that violates Kubernetes naming constraints) pod execution will fail:

Screenshot 2022-11-23 at 10 04 51

We therefore also need to add a few lines to the existing validation code to prevent these failures.

I checked [our utility library](https://github.com/typhoonzero/elyra/blob/gpu_vendor_support/elyra/util/kubernetes.py) and it appears that function [`is_valid_annotation_key`](https://github.com/typhoonzero/elyra/blob/gpu_vendor_support/elyra/util/kubernetes.py#L70) performs the kind of checking we need. We could therefore create a wrapper function  [like this one](https://github.com/typhoonzero/elyra/blob/gpu_vendor_support/elyra/util/kubernetes.py#L119) and invoke it [after this code]( https://github.com/typhoonzero/elyra/blob/gpu_vendor_support/elyra/pipeline/validation.py#L421-L430).

typhoonzero:
@ptitzler Thanks for the detailed comments, the PR is updated. By the way, I'm looking for a way to put some UI components under an "advanced options" section, which can be toggled by the user. I've noticed that there are such components used at the left bar for metadata UI, is there such component in the right bar and how to apply it?

ptitzler:
> @ptitzler Thanks for the detailed comments, the PR is updated. By the way, I'm looking for a way to put some UI components under an "advanced options" section, which can be toggled by the user. I've noticed that there are such components used at the left bar for metadata UI, is there such component in the right bar and how to apply it?

Unfortunately not yet. We've discussed this in our dev meeting a while ago and have identified a potential solution that would enable the Pipeline Editor to distinguish between core properties (that are always visible, e.g. because they are required) and properties that are only of relevance to advanced users. This solution should be flexible enough to enable users to switch between basic display mode and advanced mode. This should help improve the up-and-running experience, as there is currently a fairly large number of [Kubernetes] properties that can be intimidating at first.

elyra/pipeline/validation.py
```diff
@@ -428,6 +429,20 @@ async def _validate_generic_node_properties(self, node: Node, response: Validati
                         resource_value=resource_value,
                         response=response,
                     )
+            for resource_vendor in ["gpu_vendor"]:
+                vendor = node.get_component_parameter(resource_vendor)
+                if not is_valid_kubernetes_device_plugin_name(vendor):
```
ptitzler, line 434: Function `is_valid_kubernetes_device_plugin_name` is returning `False` if `vendor` is not set, which causes two tests in elyra/tests/pipeline/test_validation.py to fail:
 - `test_valid_node_property_pipeline_filepath`
 - `test_invalid_node_property_structure`

```
'message': 'Property is not a valid resource vendor name.', 'severity': 1, 'source': 'Elyra Pipeline Validation Service', ...}, {'data': {'nodeID': '1e7ffe03-54b8-412b-8ce0-2707181fd5e8', 'nodeName': 'notebookD.ipynb', 'propertyName': 'gpu_vendor', 'value': None},
```

Since the behavior of `is_valid_kubernetes_device_plugin_name` is correct (an empty vendor name is not valid), we need to only call the function if a value was provided. (Validation is performed before the processor is invoked and therefore the 'default' value `nvidia.com/gpu` hasn't been assigned yet.

ptitzler:
The changes are looking good!! Before we merge, can you please update the documentation topic at  https://github.com/elyra-ai/elyra/blob/main/docs/source/user_guide/pipelines.md#resources-cpu-gpu-and-ram to mention the new property and perhaps link to https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/ (or another URL that might be useful to the user.) Thank you!

ptitzler (approved):
LGTM! Thank you for your contribution! The functionality will be included in Elyra 3.14.

(I tweaked the documentation a tiny bit. For example, I reversed the proposed header change because it broke existing references to the topic.)

PR #3032, opened 2022-11-29
Address warnings produced by tests

kevin-bates:
This pull request takes a look at removing/ignoring a majority of the warnings that are produced by the server tests (via `make test-server`).  Many of these warnings are ignored as they stem from upstream modules.  That said, there are a number that have been resolved via code changes (primarily in tests) where things like file descriptors were not closed, etc.  

We should strive to get to a place where all warnings (except those explicitly handled) are treated as errors.  This will enable their capture in a timely manner so things like `DeprecationWarning` can be handled and adjusted (if necessary) or added to the list of ignored warnings (if appropriate).

This PR adds a number of warnings that are now ignored where, like caps on dependencies, a comment is provided indicating why the warning is ignored and/or additional context about the warning.  At this point, the `error` filter is disabled.  We could also consider ignoring these warnings using [decorators](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#pytest-mark-filterwarnings) in the test files themselves.

It looks like the number of warnings varies based on which version of Python is in use, but I'm seeing the warnings decrease from thousands of warnings to these 15:
```
================================================================================= warnings summary =================================================================================
elyra/tests/pipeline/test_pipeline_definition.py: 2 warnings
elyra/tests/pipeline/test_pipeline_parser.py: 1 warning
elyra/tests/pipeline/test_validation.py: 1 warning
elyra/tests/pipeline/airflow/test_component_parser_airflow.py: 9 warnings
elyra/tests/pipeline/airflow/test_processor_airflow.py: 2 warnings
  <unknown>:26: DeprecationWarning: invalid escape sequence '\_'

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================ slowest durations =================================================================================

(1714 durations < 60s hidden.  Use -vv to show these durations.)
============================================================= 570 passed, 2 skipped, 15 warnings in 347.53s (0:05:47) ==============================================================
```
 
           in the file; or

           it.

### DIFF
--- a/elyra/tests/pipeline/kfp/test_processor_kfp.py
+++ b/elyra/tests/pipeline/kfp/test_processor_kfp.py
@@ -896,7 +896,7 @@ def test_generate_pipeline_dsl_compile_pipeline_dsl_workflow_engine_test(
     monkeypatch.setattr(processor, "_upload_dependencies_to_object_store", lambda w, x, y, prefix: True)
     monkeypatch.setattr(processor, "_verify_cos_connectivity", lambda x: True)
 
-    compiled_output_file = Path(tmpdir) / os.path.basename(test_pipeline_file.with_suffix(".yaml"))
+    compiled_output_file = Path(tmpdir) / test_pipeline_file.with_suffix(".yaml").name
     compiled_output_file_name = str(compiled_output_file.absolute())
 
     # generate Python DSL for the specified workflow engine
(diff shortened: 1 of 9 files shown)


elyra/tests/pipeline/kfp/test_processor_kfp.py
```diff
@@ -992,7 +992,7 @@ def test_generate_pipeline_dsl_compile_pipeline_dsl_one_generic_node_pipeline_te
     monkeypatch.setattr(processor, "_upload_dependencies_to_object_store", lambda w, x, y, prefix: True)
     monkeypatch.setattr(processor, "_verify_cos_connectivity", lambda x: True)
 
-    compiled_argo_output_file = Path(tmpdir) / test_pipeline_file.with_suffix(".yaml")
+    compiled_argo_output_file = Path(tmpdir) / os.path.basename(test_pipeline_file.with_suffix(".yaml"))
```
ptitzler, line 995: Nice catch! Would you mind switching to the pathlib-based approach? For the past couple of months I've been replacing `os.path` to their more modern equivalent. There should be seven `test_pipeline_file.with_suffix` in this file. (Next time I need to touch the parser tests I will also replace the `tmpdir` fixture in all tests with the `tmp_path` fixture, which yields a path object instead of a string.)

```suggestion
    compiled_argo_output_file = Path(tmpdir) / test_pipeline_file.with_suffix(".yaml").name
```

kevin-bates: Sounds good - thanks.  Btw, I'm only finding five of these instances.

ptitzler (approved):
LGTM!

elyra/tests/pipeline/kfp/test_processor_kfp.py
ptitzler: oops, I can't count

PR #3148, opened 2023-05-06
Update pytorch runtime to fix validate images check

typhoonzero:

1. Since Python 3.7 runtime image support is removed, this PR updates default pytorch cuda runtime image version to fix the "validate images" check in github actions.
2. Python 3.10.11 reported `ImportError: cannot import name 'appengine' from 'urllib3.contrib' ` at current main branch CI jobs, adding `appengine-python-standard` can fix this, see: https://stackoverflow.com/questions/76175487/sudden-importerror-cannot-import-name-appengine-from-requests-packages-urlli


lresende (approved):

lresende:
Validated main issues while running `make validate-runtime-images` and that they are gone after applying these changes.

Issue #3149, opened 2023-05-08
Pipeline editor doesn't connect to insecured pipeline server


harshad16:
**Describe the issue**

The pipeline editor looks for a secured pipeline server for connection and doesn't allow execution of the pipeline with insecure pipeline server.
When tried to connect pipeline editor with insecured pipeline server, the editor throws following error:

```
HTTPServerRequest(protocol='http', host='jupyter-nb-kube-3aadmin-rhods-notebooks.apps.hnalla.dev.datahub.redhat.com', method='POST', uri='/notebook/rhods-notebooks/jupyter-nb-kube-3aadmin/elyra/pipeline/schedule?1683579505300', version='HTTP/1.1', remote_ip='127.0.0.1')
Traceback (most recent call last):
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/connectionpool.py", line 703, in urlopen
httplib_response = self._make_request(
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/connectionpool.py", line 386, in _make_request
self._validate_conn(conn)
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/connectionpool.py", line 1042, in _validate_conn
conn.connect()
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/connection.py", line 419, in connect
self.sock = ssl_wrap_socket(
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/util/ssl_.py", line 449, in ssl_wrap_socket
ssl_sock = _ssl_wrap_socket_impl(
File "/opt/app-root/lib64/python3.8/site-packages/urllib3/util/ssl_.py", line 493, in _ssl_wrap_socket_impl
return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
File "/usr/lib64/python3.8/ssl.py", line 500, in wrap_socket
return self.sslsocket_class._create(
File "/usr/lib64/python3.8/ssl.py", line 1040, in _create
self.do_handshake()
File "/usr/lib64/python3.8/ssl.py", line 1309, in do_handshake
self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1131)
```

**To Reproduce**
1. Setup ODH or kubeflow pipeline on an insecure cluster
2. Execute a notebook with elyra installed
3. Try to execute a pipeline example 

[...]

PR #3150, opened 2023-05-08
Allow users to include ssl_sa_certs via env vars for kfp_tekton client

harshad16:

As the pipeline editor fails to connect to an unsecured pipeline server due to a certs issue, this PR includes an environment variable that can be pointed to the ssl_cert, which would be used for setting up the connection.

`KF_PIPELINES_SSL_SA_CERTS` can be set with cert path.

and `ssl_ca_cert` var would be set with the value and passed to [kfp SDK](https://github.com/kubeflow/kfp-tekton/blob/2d7cc156d131c66c9274a6c62a4235e3d575a88b/sdk/python/kfp_tekton/_client.py#L103).

Fixes: #3149 

           in the file; or

           it.

### DIFF
--- a/docs/source/user_guide/runtime-conf.md
+++ b/docs/source/user_guide/runtime-conf.md
@@ -221,6 +221,9 @@ The KubeFlow Pipelines API endpoint you want to utilize. This setting is require
 
 Example: `https://kubernetes-service.domain.com/pipeline`
 
+*Note*: In setups where the Kubeflow Pipeline server is secured with a custom certificate, users can specify this certificate path by setting the environment variable `KF_PIPELINES_SSL_SA_CERT` in the jupyter environment. This variable directs the Kubeflow Pipelines SDK client to use the provided certificate for accessing the Kubeflow Pipeline server. For more information, see the [ssl_ca_cert](https://kubeflow-pipelines.readthedocs.io/en/stable/source/client.html#kfp.client.Client.__init__.ssl_ca_cert) attribute in the `kfp` SDK documentation.  
+Ex: `KF_PIPELINES_SSL_SA_CERT:/path/to/certs`
+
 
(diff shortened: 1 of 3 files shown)


lresende:
This only allows for one global cert path, should we actually add a config on the runtime metadata to allow passing this on the UI per server?

kevin-bates:
> This only allows for one global cert path, should we actually add a config on the runtime metadata to allow passing this on the UI per server?

Is the idea that a site could target multiple KFP clusters (e.g., dev, test, and production) where the "valid" certs vary from target to target?

This makes sense, although I don't recall what the overhead is to extend runtime configurations with an optional property.  Does this require UI work or is that portion dynamic?

harshad16:
> > This only allows for one global cert path, should we actually add a config on the runtime metadata to allow passing this on the UI per server?
> 
> Is the idea that a site could target multiple KFP clusters (e.g., dev, test, and production) where the "valid" certs vary from target to target?
> 
> This makes sense, although I don't recall what the overhead is to extend runtime configurations with an optional property. Does this require UI work or is that portion dynamic?


@lresende  @kevin-bates thanks for the comments
If we want to provide this option via runtime configurations then it would require UI change.

With this my idea, was the simply provide a way for user with unsecured cluster to utilize the feature of elyra pipeline , with a simple env var set.
as i understand from the notes, the insecure cluster support for overall elyra feature is not in avenue of development for now.

shalberd:
Nice work, I like this current proposed no-frills way

lresende:
LGTM, could you also provide a little update to the docs mentioning the var so that other users can leverage it.

harshad16:
> could you also provide a little update to the docs mentioning the var so that other users can leverage it.

@lresende , sorry totally missed this PR.
As you instructed, I updated the docs with details.
PTAL, at your convenience

lresende (approved):

PR #3167, opened 2023-06-21
Make generic pipeline work with airflow2x

shalberd:
fixes #3166 

@ianonavy @lresende NOT intended to work with Airflow 1.x

### What changes were proposed in this pull request?

changes to airflow processor and airflow pipeline template with regards to Airflow 2.8.2 or higher support
also added pendulum instead of days_ago since deprecated

https://github.com/apache/airflow/pull/21653/files

Conceptual overlap with a fork, came to the same code in parallel

change to the Kubernetes client SDK in generic pipeline part of template since the Airflow abstractions
were all deprecated and removed except for Secret.

Finally, Airflow 2 adds logic that makes config_file mutually exclusive
with in_cluster, so we need to ensure that None is passed as None and
not string "None".

See also

https://github.com/kflow-ai/elyra/commit/f9d132954e008d30145f18794aa543d97f121a5f#diff-dc6c3f666aad9271fa5e9b8c31e3f0582cd39a7d2516cbc2240731fe4456e641

### How was this pull request tested?
In contrast to kubeflow pipelines, even for Airflow 1.x and the different pipeline editors, there do not seem to be any tests.
I'd like to test the built wheel file in a docker image in conjunction with ODH.
Mostly seeing whether the generated DAG code works with Airflow 2.8.2 and higher.
 
Developer's Certificate of Origin 1.1

       By making a contribution to this project, I certify that:

       (a) The contribution was created in whole or in part by me and I
           have the right to submit it under the Apache License 2.0; or

       (b) The contribution is based upon previous work that, to the best
           of my knowledge, is covered under an appropriate open source
           license and I have the right under that license to submit that
           work with modifications, whether created in whole or in part
           by me, under the same open source license (unless I am
           permitted to submit under a different license), as indicated
           in the file; or

       (c) The contribution was provided directly to me by some other
           person who certified (a), (b) or (c) and I have not modified
           it.

       (d) I understand and agree that this project and the contribution
           are public and that a record of the contribution (including all
           personal information I submit with it, including my sign-off) is
           maintained indefinitely and may be redistributed consistent with
           this project or the open source license(s) involved.

PR #3202, opened 2023-12-20
Adding resource limits that can be selected from the UI

### LINKED ISSUE #3168: add limits for container resources like cpu. memory, GPU in addition to requests
@kevin-bates Currently, in the Elyra GUI, i.e. for generic pipelines and a component therein, you can set values for CPU, Memory, and GPU and GPU Vendor.

Bildschirmfoto 2023-06-27 um 16 35 51

https://elyra.readthedocs.io/en/stable/user_guide/pipelines.html#resources-cpu-gpu-and-ram

Those values are set as container request sizes, except [GPU, which is set as GPU limit](https://github.com/typhoonzero/elyra/blob/main/elyra/pipeline/kfp/processor_kfp.py#L562) and [used as GPU limit](https://github.com/elyra-ai/elyra/blob/main/elyra/pipeline/kfp/processor_kfp.py#L841), as far as I see. That goes against [the doc](https://github.com/elyra-ai/elyra/blob/main/docs/source/user_guide/pipelines.md) that speaks of all as "resources required".

Resources: CPU, GPU, and RAM

    - Resources that the notebook or script requires. RAM takes units of gigabytes (109 bytes).
    - Specify a custom Kubernetes GPU vendor, if desired. The default vendor is nvidia.com/gpu. See [this topic in the Kubernetes documentation](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/) for more information.
    - The values are ignored when the pipeline is executed locally.
    - Example: amd.com/gpu

The problem is that not setting any limits, and not giving the user any way to set limits, leads to cluster instability potentially on Kubernetes and Openshift.

Add fields and properties for CPU, GPU, Memory limits in GUI, to be used later in processors and templates.
Possibly also add default factor of x1, overwritable of course, for the limits field values, based on requests fields.

Would have to add those new fields / properties to the [generic properties template](https://github.com/elyra-ai/elyra/blob/main/elyra/templates/components/generic_properties_template.jinja2) as well.

No real way around it if using the graphical pipeline editors with setting for runtime containers.

**Additional context**
Came across this when discussing Kubernetes / Openshift Resource Limits setting in an Airflow2 compatible way. Discussion quickly came to, wait a sec, no limits set ...

https://github.com/elyra-ai/elyra/pull/3167#issuecomment-1603036641

paloma-rebuelta:
I have changed the backend to replace cpu and memory for cpu_request and memory_request, as well as adding cpu_limit and memory_limit. This has also been added to the UI.

These changes should fix the issue where the pipeline breaks when there are pre-defined argo resource limits that are lower than the requests, fixes #3168 

milestone 4.0 

I have tested this by adding two unit tests and building the code to run in a kubeflow server. I tested a pipeline with all the resource requests/limits combinations (including leaving it blank). I cannot run airflow pipelines in my setup but have also edited the code for these. 

Screenshot 2023-12-20 at 19 44 43


elyra/pipeline/pipeline.py
```diff
@@ -260,20 +262,32 @@ def __init__(
             raise ValueError("Invalid pipeline operation: Missing field 'operation filename'.")
         if not component_props.get("runtime_image"):
             raise ValueError("Invalid pipeline operation: Missing field 'operation runtime image'.")
-        if component_props.get("cpu") and not self._validate_range(component_props.get("cpu"), min_value=1):
-            raise ValueError("Invalid pipeline operation: CPU must be a positive value or None")
+        if component_props.get("cpu_request") and not self._validate_range(
+            component_props.get("cpu_request"), min_value=1
+        ):
+            raise ValueError("Invalid pipeline operation: CPU request must be a positive value or None")
+        if component_props.get("cpu_limit") and not self._validate_range(component_props.get("cpu_limit"), min_value=1):
```
kevin-bates, line 269: I think the limit values should be dependent upon the existence of the corresponding request value.  That is, limits do not come into play unless the corresponding request has a value.  If we make that statement, then the `min_value` on the limits should be the request value and we have our check to ensure limits are >= requests (which I don't see in this PR).

This rule would also prevent only limits from getting set, which I think we should avoid as well.

kevin-bates (commented):
@paloma-rebuelta - thank you for acting on this so quickly - its much appreciated!

I had a general comment regarding the relationship between limits and requests and how to enforce conditions we'd like to impose.

elyra/pipeline/pipeline.py
paloma-rebuelta: We have some users that leave default values for requests but want to make sure they control the maximum amount of resources that a pod uses (in case it gets stuck or is not behaving as expected), so they set a limit explicitly. I think have both optional is good, but I agree that there should be a check that the limit is equal or greater than the request. What do you think?
I can also see my tests don't work as they would only fail at the argo workflow level (which I tested upon build), so need to change them to be effective

shalberd:
@romeokienzler @lresende maybe we should merge in Airflow 2.x related changes to processor airflow and airflow jinja2 template first before the changes here. @paloma-rebuelta in Airflow 2, a lot of changes happened and resource requests and limits are now specified a little differently, see my [Airflow 2.x PR 3167](https://github.com/elyra-ai/elyra/pull/3167/files#diff-dc6c3f666aad9271fa5e9b8c31e3f0582cd39a7d2516cbc2240731fe4456e641R50)

docs/source/getting_started/changelog.md
```diff
@@ -2,6 +2,10 @@
 
 A summary of new feature highlights is located on the [GitHub release page](https://github.com/elyra-ai/elyra/releases).
 
+## Release 3.15.1 - 03/28/2023
```
shalberd, line 5: milestone or release target when looking at elyra-ai:main branch should be 4.0, should it not? @lresende what do you think?

elyra/pipeline/pipeline.py
paloma-rebuelta: Okay fixed this which solves the test issue too 👍🏽

docs/source/getting_started/changelog.md
paloma-rebuelta: Should I change it or leave it blank so it can be filled when the next release is created?

elyra/templates/components/generic_properties_template.jinja2
```diff
@@ -35,16 +35,28 @@
             "items": []
           }
         },
-        "cpu": {
+        "cpu_request": {
           "type": "integer",
-          "title": "CPU",
-          "description": "For CPU-intensive workloads, you can choose more than 1 CPU (e.g. 1.5).",
+          "title": "CPU request",
+          "description": "For CPU-intensive workloads, you can request more than 1 CPU (e.g. 1.5, this is optional).",
```
kevin-bates, line 41: In thinking about introducing millicores, I was surprised to see the example in the comment regarding 1.5 CPUs.  Since this is an integer field, do fractional entries even work?   Seems like this field should be a `number`.

elyra/templates/components/generic_properties_template.jinja2
```diff
@@ -35,16 +35,28 @@
             "items": []
           }
         },
-        "cpu": {
+        "cpu_request": {
```
kevin-bates, line 38: I don't think we can change the names of _existing_ pipeline fields w/o introducing a migration.  As a result, I would recommend retaining the names `cpu` and `memory` and only introducing `cpu_limit` and `memory_limit`.

(See the test pipeline files.  Any changes to field name will require a migration.)

elyra/tests/pipeline/resources/test_pipelines/kfp/kfp-multi-node-generic.pipeline
```diff
@@ -156,9 +156,9 @@
               "mounted_volumes": [],
               "filename": "producer.ipynb",
               "runtime_image": "tensorflow/tensorflow:2.8.0",
-              "cpu": 1,
+              "cpu_request": 1,
```
kevin-bates, line 159: See "migration comment".

kevin-bates (commented):
I think this introduces a migration issue that either needs to be handled (or avoided).

Also, can fractional values for CPUs be entered, despite the schema type being `integer`?

kevin-bates:
Regarding the screen layout (thank you for including that), I find it more intuitive to have CPU request and limit on the same line, rather than CPU and Memory request, then CPU and Memory  limit.  I think keeping the "subject" on its own line would be better, and is also consistent relative to the CPU line (as its fields are on the same line as well).  Thanks.

elyra/templates/components/generic_properties_template.jinja2
paloma-rebuelta: I left the example as it was but I just tested it in my environment and the pipeline was triggered fine, it just automatically changes it from 1.5 to 1500m
Screenshot 2023-12-21 at 19 02 39

elyra/templates/components/generic_properties_template.jinja2
paloma-rebuelta: I have changed all the test pipelines files too I believe, what would the migration involve?

paloma-rebuelta:
> Regarding the screen layout (thank you for including that), I find it more intuitive to have CPU request and limit on the same line, rather than CPU and Memory request, then CPU and Memory limit. I think keeping the "subject" on its own line would be better, and is also consistent relative to the CPU line (as its fields are on the same line as well). Thanks.

done :)
Screenshot 2023-12-21 at 20 09 43

elyra/templates/components/generic_properties_template.jinja2
kevin-bates: This doesn't address all the existing pipeline files that still use `cpu`.  Those pipelines will be loaded but the cpu (and memory) request value(s) will not be found.  I really don't think we should introduce a migration for this but instead update the `description` entries to emphasize these are _request_ values.  The UI will still reflect proper naming, only the pipeline file field entries would remain `cpu` and `memory`.

> what would the migration involve?

We'd need to increment the pipeline version and when an older version is detected, read the `cpu` and `memory` fields into `cpu_request` and `memory_request` and ensure the updated pipeline is persisted.  There is a framework for this in place already although I have never worked with it.

kevin-bates:
> @romeokienzler @lresende maybe we should merge in Airflow 2.x related changes to processor airflow and airflow jinja2 template first before the changes here. @paloma-rebuelta in Airflow 2, a lot of changes happened and resource requests and limits are now specified a little differently, see my [Airflow 2.x PR 3167](https://github.com/elyra-ai/elyra/pull/3167/files#diff-dc6c3f666aad9271fa5e9b8c31e3f0582cd39a7d2516cbc2240731fe4456e641R50)

@shalberd - what is the ETA of the Airflow 2 PR - its been open for some time now.  That PR speaks of really _wanting_ limit settings and the changes to add CPU/Memory requests are trivial.  In addition, this PR still supports Airflow 1x while [I can't determine if your PR _replaces_ that support or not (although I suspect it does)](https://github.com/elyra-ai/elyra/pull/3167#issuecomment-1867858962).

As a result, I think this PR should take priority ([assuming we don't require a migration](https://github.com/elyra-ai/elyra/pull/3202#discussion_r1435198142)), set the limit handling in place against _existing_ user installations (preferrably with the release), then release support for Airflow 2x.  If 2x support is a replacement, that's another issue worth discussing (if it hasn't been discussed already).

elyra/templates/components/generic_properties_template.jinja2
paloma-rebuelta: I might not have access to those older pipelines, so I have reverted the request names to cpu and memory to avoid any issues

kevin-bates: Pipeline files (`.pipeline`) are tantamount to notebook files (`.ipynb`) and will exist wherever elyra has been installed and used to create a pipeline.  This is why any changes to the "file format" need to be well-intentioned.  I'm not saying this PR doesn't warrant an update to file format, but that update introduces more moving parts (and points for error) than is probably worth the effort.  

Thank you for your patience on this pull request.

kevin-bates (approved):
Thank you for this great addition @paloma-rebuelta!  

We can address millicores (and values < 1 core) in another PR, which will likely require a pipeline migration, at which time we can add proper adjustments to the request-related resources if we choose.

kevin-bates:
@lresende, since this change adds a helpful feature (and is low risk IMO), I wonder if we should include it in 3.x and use Airflow 2 support as the trigger for 4.x.  Are there other merged PRs that are only slated for 4.x?  If we do decide to include this in 3.x we should probably adjust the [changelog updates](https://github.com/elyra-ai/elyra/pull/3202/files#diff-7434fe02e77d7b869d9e543249324a5febef5e7918fafec6734acbe1053f30c6).

shalberd:
@kevin-bates @lresende Happy New Year. I agree that this PR here should take precedence, as adding limits is something that is extremely useful no matter what target runtime and is backward compatible.
As for the Airflow PR: yes, after we have merged in this limits PR here, I will make the few modifications to the assembly of Airflow 2.x code, i.e. the limits fields in Airflow 2.x compatible syntax. My Airflow PR is intended **not** for backward compatibility, but for replacing Airflow 1.x support with Airflow 2.x support in version 4 of Elyra. I agree that we can make this limits PR here part of 3.x Elyra. Thanks again to @paloma-rebuelta for the great work on this here.

paloma-rebuelta:
> @kevin-bates @lresende Happy New Year. I agree that this PR here should take precedence and is low risk, as adding optional limits is something that is extremely useful no matter what target runtime and is backward compatible. As for the Airflow PR: yes, after we have merged in this limits PR here, I will make the few modifications to the assembly of Airflow 2.x code, i.e. the limits fields in Airflow 2.x compatible syntax. My Airflow PR is intended **not** for backward compatibility, but for replacing Airflow 1.x support with Airflow 2.x support in version 4 of Elyra. I agree that we can make this limits PR here part of 3.x Elyra. Thanks again to @paloma-rebuelta for the great work on this here.

Happy new year! I am happy to edit the milestone, is there a specific version of 3.x I should put? 3.15.1 or 3.16.0 maybe?

shalberd:
@paloma-rebuelta we will make this part of release 4.0, looks good.

elyra/templates/components/generic_properties_template.jinja2
shalberd: yeah, in future, we might think about leaving out the minimum 1 validation though, as there is no reason to request at least 1 CPU or 1000 millicores. Same for validating at least 1 Gig of memory for a task. 
But that is a story for later. I wonder, too, why 1.5 or any decimal value is allowed there, given it is an integer field. Stuff for later.

packages/pipeline-editor/style/index.css
```diff
@@ -864,20 +864,26 @@ input.elyra-Dialog-checkbox.jp-mod-styled {
...
+input#root_component_parameters_memory_limit,
 input#root_component_parameters_gpu,
 input#root_component_parameters_gpu_vendor {
   width: 100%;
 }
 
 #cpu,
+#cpu_limit,
 #gpu,
+#cpu_limit,
```
shalberd, line 886: @paloma-rebuelta in line 886, cpu_limit or memory_limit? I am not a CSS expert, especially also regarding the following more distinct settings for memory.

shalberd:
alos wondering about the env var minus GUI approach vs. env var and GUI approach discussed here before with @kevin-bates https://github.com/elyra-ai/elyra/issues/3168#issuecomment-1864855870

kevin-bates:
@shalberd - regarding:
> we will make this part of release 4.0 (confirmed by @lresende) , looks good. I'll have a look at the code one more time, too, and test it. Having @kevin-bates approved it and your extensive discussions already make this seem mergeable.

I'd still prefer we include this in a 3.x release since its independent of AF 2 support.  I guess I'd prefer to see @lresende comment on [my comment above](https://github.com/elyra-ai/elyra/pull/3202#issuecomment-1868051115) regarding where this should be merged before concluding its only in 4.0.

> alos wondering about the env var minus GUI approach vs. env var and GUI approach discussed here before with @kevin-bates https://github.com/elyra-ai/elyra/issues/3168#issuecomment-1864855870

I'm a big fan of envs as _default_ values, but also feel that (for this application in particular) having a GUI presence is preferred.  I believe the previous discussion assumed that the GUI work was not able to be done, so we'd start with envs, then, when a GUI is in place, use the previous env-based effort to represent defaults/preferences.  Since @paloma-rebuelta will be receiving "extra credit" on this by actually introducing the GUI changes (thank you!), the env portion of things is a bit muted IMHO.

shalberd:
> I believe the previous discussion assumed that the GUI work was not able to be done, so we'd start with envs ... the env portion of things is a bit muted IMHO.

Agreed, just wanted to hear your perspective on this.

shalberd:
@paloma-rebuelta regardless of how wide of size the node properties part, I don't get limits on the same line, but rather on the next line.

Bildschirmfoto 2024-01-10 um 17 29 09

Browser: MS Edge
Will try again tomorrow, I am based off your latest code.

lresende:
> @shalberd - regarding:
> 
> > we will make this part of release 4.0 (confirmed by @lresende) , looks good. I'll have a look at the code one more time, too, and test it. Having @kevin-bates approved it and your extensive discussions already make this seem mergeable.
> 
> I'd still prefer we include this in a 3.x release since its independent of AF 2 support. I guess I'd prefer to see @lresende comment on [my comment above](https://github.com/elyra-ai/elyra/pull/3202#issuecomment-1868051115) regarding where this should be merged before concluding its only in 4.0.
> 
> >

I was suggesting we merge to the main branch and that I would push for the next release to be 4.0 with support for JupyterLab... if others are willing to maintain a 3.x branch I am open for it. At the moment I won't be able to manage 3.x and 4.x releases in parallel.

Feel free to merge this to main once you are happy with it @kevin-bates

shalberd:
@kevin-bates let's hold off for a bit until we clear up the questions related to node editor limits field placement and cascading style sheets, as well as the question as to why for a single file "Submit as pipeline" does not show the new fields.
I am trying out a little, too, but @paloma-rebuelta might be quicker to answer.

shalberd (approved):
great job, working nicely, minor cosmetic GUI comment in my last comment under conversations, no big deal. Thank you so very much, this is so very valuable. Love using it already.

@kevin-bates please disregard my last comment, PR is good to merge

kevin-bates:
> if others are willing to maintain a 3.x branch I am open for it.

No can do either.  So this will first appear in 4.0.  Thanks for the response @lresende

paloma-rebuelta:
Thanks @kevin-bates @shalberd and @lresende for getting this across 🥳

shalberd:
you're welcome; this has been a long-standing wish of many people in the community, so your effort on this is extremely appreciated by not just us, I am sure.

Issue #3087, opened 2023-01-24
UI export/submit handling of INT pipeline parameter with default value of 0


ptitzler:
**Describe the issue**

- Create a KFP pipeline
- Create a required pipeline parameter of type INT as follows:

   ![image](https://user-images.githubusercontent.com/13068832/214431647-69f38384-3c38-4f7a-93d6-176e1fc248ca.png)

- Assign the parameter as a node property value
- Export/submit the pipeline.  Do not enter a value for the parameter. The OK button is disabled, even though a required value is present.
  
  ![image](https://user-images.githubusercontent.com/13068832/214432011-591ad06d-78f5-4272-9584-af5d2415a85e.png)

**Expected behavior**
 - A 0 value for int and 0.0... for float are properly recognized as valid input

**Deployment information**
Describe what you've deployed and how:
 - Elyra version: 3.14.2.rc0

ptitzler:
The problem does not surface for FLOAT-typed parameters

PR #3417, opened 2026-09-05
Coerce pipeline parameter values to their declared numeric type

### LINKED ISSUE #3088: UI export/submit handling of FLOAT pipeline parameter with default value of 0.0...
**Describe the issue**

- Create a KFP pipeline
- Create a required pipeline parameter of type FLOAT as follows:
  
  image

- Assign the parameter as a node property value
- Export/submit the pipeline.  Note that the displayed default value is 0 instead of 0.0...
  
  image

Also note that this behavior also triggers the problem reported in https://github.com/elyra-ai/elyra/issues/3087

**Expected behavior**
 - The exact representation of the user-provided value.

**Deployment information**
 - Elyra version: 3.14.2.rc0

karthikchundi-commits:
## Fixes #3088

**What's happening:** a `Float`-typed pipeline parameter with a whole-number value — the exact-zero default `0.0` from the issue, or any value that arrives as the plain string `"0"` — is stored and later exported/submitted as a Python `int` rather than a `float`. Python's `json` serializer then renders it without a decimal point (`0` instead of `0.0`), even though the user explicitly declared the parameter type as Float. (The issue notes this also feeds #3087.)

**Root cause:** `PipelineParameter.__init__` stores `default_value` and `value` exactly as received, never consulting `selected_type` (which already holds `"Float"`/`"Integer"`). The frontend can't help here — JS `number` has no int/float distinction (`0.0 === 0`), so the backend has to enforce the declared type itself.

**Fix:** `__init__` now coerces both `default_value` and `value` through the numeric Python type implied by `selected_type` — `"Float"` → `float`, `"Integer"` → `int`. Non-numeric types (`String`, `Bool`) are left untouched; a conversion failure leaves the value unchanged rather than raising. This is the backend counterpart to #3412 (which constrained the Integer *input* on the frontend side).

**Tests:** added `test_pipelineparameter_float_type_value_coercion` covering int/string/float inputs for both Float and Integer types (including the exact-zero case), plus the untouched String/Bool paths. The existing `test_pipelineparameter_constructor` never exercised a non-empty falsy default like `0`, which is why this was never caught. `pytest elyra/tests/pipeline/test_properties.py` — 2/2 pass; `ruff check` clean.

## AI writing disclosure

AI-assisted: implemented by Claude Code under my direction.


karthikchundi-commits:
Re-triggered CI (the previous run from Sept 5 failed on \`Run Integration Tests (pipeline)\` at the \`@elyra/ui-components:build:prod\` step with a \`json.decoder.JSONDecodeError\` inside JupyterLab's build tooling — unrelated to this PR, which only touches \`elyra/pipeline/properties.py\`; every other check passed cleanly). Pushed an empty-diff commit to get a fresh run; it's currently sitting in \`action_required\`, awaiting a maintainer approval to actually execute.

